PR #24, opened 2020-06-04
Try to fix wasmparser benchmarks

Robbepop:
This PR fixes the search paths for the benchmarks as well as the `wasmparser` `Cargo.toml` file to respect the `criterion` provided benchmark suite.
Also removes some code dupe in benchmarks and generally polish the implementation.

The error we still see is the following:
```
Benchmarking it works benchmark: Warming up for 3.0000 sthread 'main' panicked at 'unexpected error while reading Wasm Bad magic number (at offset 0)', crates/wasmparser/benches/benchmark.rs:31:42
```

Please tell me how I can fix it or what is actually wrong. :S I assume it is some benchmark `.wast` files that are broken?

### DIFF
--- a/crates/wasmparser/Cargo.toml
+++ b/crates/wasmparser/Cargo.toml
@@ -19,6 +19,10 @@ wat = { path = "../wat" }
 wast = { path = "../wast" }
 rayon = "1.3"
 
+[[bench]]
+name = "benchmark"
+harness = false
+
 [features]
 # The "deterministic" feature supports only Wasm code with "deterministic" execution
 # across any hardware. This feature is very critical for many Blockchain infrastructures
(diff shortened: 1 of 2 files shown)


alexcrichton:
Oops sorry about that, but thanks!

Robbepop:
Thanks @alexcrichton for merging, however as described above there is still the `Bad magic number` error when trying to execute the benchmarks since I wasn't able to identify its cause. Maybe you know more and can tell me where I can look to fix it in a follow-up PR? :)

alexcrichton:
Oh sorry, is that perhaps because the text files in the test directory are being parsed instead of the binary files now?

Robbepop:
> Oh sorry, is that perhaps because the text files in the test directory are being parsed instead of the binary files now?

Ah I remember that there have been some changes to convert all the `.wasm` test files into `.wat` which generally is a good idea but in this case we might have the need to pre-convert them back into `.wasm` before benchmarking so you are probably right about this.

PR #28, opened 2020-06-18
Drop ref.is_null type annotation

eqrion:
This updates the binary format to match WebAssembly/reference-types#99.


eqrion:
Thanks!

PR #30, opened 2020-06-24
wasmpaser: Validate nested modules

alexcrichton:
This commit implements recursive validation of nested modules in
wasmparser. Previously nested modules were simply skipped but now
they're recursed into and actually checked. This commit also comes with
a more complete implementation of handling `alias` directives.

Internally this required quite a bit of refactoring. The validator now
retains a stack of modules which are being validated and remembers
parent-relationships between them. Indices into this stack are then used
for recording type definitions. The type of a
function/instance/module/etc can be defined the current module or any
previous module on the stack. New helper functiosn were added to help
resolve this new `Def` type to ensure it's handled correctly.

### DIFF
--- a/crates/wasmparser/src/module_resources.rs
+++ b/crates/wasmparser/src/module_resources.rs
@@ -302,7 +302,7 @@ pub trait WasmModuleResources {
     /// Returns the global variable at given index.
     fn global_at(&self, at: u32) -> Option<&Self::GlobalType>;
     /// Returns the function signature ID at given index.
-    fn func_type_id_at(&self, at: u32) -> Option<u32>;
+    fn func_type_at(&self, at: u32) -> Option<&<Self::TypeDef as WasmTypeDef>::FuncType>;
     /// Returns the element type at the given index.
     fn element_type_at(&self, at: u32) -> Option<crate::Type>;
 
@@ -336,8 +336,8 @@ where
     fn global_at(&self, at: u32) -> Option<&Self::GlobalType> {
         T::global_at(self, at)
     }
-    fn func_type_id_at(&self, at: u32) -> Option<u32> {
-        T::func_type_id_at(self, at)
+    fn func_type_at(&self, at: u32) -> Option<&<T::TypeDef as WasmTypeDef>::FuncType> {
+        T::func_type_at(self, at)
     }
     fn element_type_at(&self, at: u32) -> Option<crate::Type> {
         T::element_type_at(self, at)
--- a/crates/wasmparser/src/operators_validator.rs
+++ b/crates/wasmparser/src/operators_validator.rs
@@ -523,7 +523,7 @@ impl OperatorValidator {
         function_index: u32,
         resources: impl WasmModuleResources,
     ) -> OperatorValidatorResult<()> {
-        let type_index = match resources.func_type_id_at(function_index) {
+        let ty = match resources.func_type_at(function_index) {
             Some(i) => i,
             None => {
                 bail_op_err!(
@@ -532,7 +532,6 @@ impl OperatorValidator {
                 );
             }
         };
-        let ty = func_type_at(&resources, type_index)?;
         self.check_operands(wasm_func_type_inputs(ty).map(WasmType::to_parser_type))?;
         self.func_state.change_frame_with_types(
             ty.len_inputs(),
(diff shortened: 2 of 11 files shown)


alexcrichton:
FWIW I've been wrestling a lot with the current design of the validator. I find it really difficult to work with the way it's set up, namely how there's one massive state enum and we just transition over that as we would a state machine. This is quite unwieldy and difficult to work with what would otherwise be pretty simple to work with I think.

One major drawback of this implementation is that there's really no avenue for parallel module validation. That's a major design consideration with the section orderings, however, and is something that I think we should enable.

I've also found, however, that there's a number of other drawbacks with the current validator and/or parser such as they can't accept incremental input (e.g. downloaded over the network). Additionally there's a lot of state being learned as part of nested modules which is going to have to be figured out again by consumers after validation. Ideally this would also be structured so consumers could persist information such as "what's the type of this function?" in a way that they don't have to figure out all the parent module relationships again.

What I'm thinking, however, is basically a rewrite of the validator. While this isn't necessarily a huge undertaking it's large enough that I don't want to entangle it here or anything like that. I'm hoping that as this come up in `wasmtime` we can talk about if and how validation/parsing should be rewritten.

alexcrichton:
Ok [this ran on CI](https://github.com/bytecodealliance/wasm-tools/pull/30/checks?check_run_id=818660937#step:3:187), but I'm not sure how to read the output myself..

alexcrichton:
Does that look as expected to you @yurydelendik ?

yurydelendik:
> Does that look as expected to you

The results are good, but it looks like it did not checkout "main" (see ` error: pathspec 'main' did not match any file(s) known to git`). So it is false positive. I'll run locally.

yurydelendik:
I see, so not much effect from this PR

```
group                           after                                  before
-----                           -----                                  ------
it works benchmark              1.00      0.4±0.03ns        ? B/sec    0.99      0.4±0.01ns        ? B/sec
validate benchmark              1.00      1.4±0.03ns        ? B/sec    1.00      1.4±0.07ns        ? B/sec
validator no fails benchmark    1.00      0.6±0.02ns        ? B/sec    0.76      0.4±0.02ns        ? B/sec
```

P.S. we need to fix ./compare-with-main.sh so it is getting valid "main"

alexcrichton:
What do these numbers mean? Is that something executing in 0.4ns?

alexcrichton:
Yeah 1ns runtimes typically means [tests aren't running](https://github.com/bytecodealliance/wasm-tools/pull/35)!

alexcrichton:
Ok now it is saying 

```
group                           after                                  before
-----                           -----                                  ------
it works benchmark              1.03   215.4±13.99µs        ? B/sec    1.00    208.8±6.97µs        ? B/sec
validate benchmark              1.19   570.5±10.68µs        ? B/sec    1.00   478.6±32.82µs        ? B/sec
validator no fails benchmark    1.28   610.2±21.13µs        ? B/sec    1.00   475.4±30.67µs        ? B/sec
```

I [fixed another issue](https://github.com/bytecodealliance/wasm-tools/commit/9a6b428df6c57ccbb6c68804f59fe13f5baee185) after this measurement was taken, though, which ensures that it actually checks out all submodules so we can include wabt/spec tests in the measurement.

I don't doubt that this iteration is slower (bounds checking) but the benchmarking strategy is also somewhat flawed because the `main` branch has a different set of tests. I don't expect the differences in tests here to account for a 30% slowdown, however.

FWIW I'm not sure that we can do all that much better without `unsafe` compared to the previous iteration. I personally feel that if we really want to eek out the max performance we should refactor validation/parsing along the lines of what I was thinking above.

yurydelendik:
> we should refactor validation/parsing along the lines of what I was thinking above.

I was planning to fade Parser and ValidatingParser away. Hopefully you are not thinking about refactoring this API. Shall we just deprecate them, and use readers + something instead?

alexcrichton:
I don't have any changes in the works, but I've been thinking recently about how things might be refactored. I'm planning on waiting until https://github.com/bytecodealliance/rfcs/pull/1 is settled though and going through that process for making a proposal (and going through a more formal proposal before doing code changes).

To clarify, though, @yurydelendik do you feel it's ok to merge this as-is or would you prefer I look into the perf difference further?

yurydelendik:
> To clarify, though, @yurydelendik do you feel it's ok to merge this as-is or would you prefer I look into the perf difference further?

It is good to go. I'm mostly monitoring huge "it works benchmark" regression.

PR #31, opened 2020-06-25
[wat] Update reftype parsing to match latest proposals

eqrion:
The upstream function-references/gc proposals have had some churn recently around the reftype model. These commits update wat to match.

 * Implements reftype := (ref null? <heaptype>) and associated sugars
 * Removes anyref as an alias of externref (GC proposal still has anyref, we might need to add it as a proper type later)
 * Switches tables to actually use RefType instead of TableElemTyp
 * Fixes ref.null to take <heaptype> now that it's a concept in wat

### DIFF
--- a/crates/wast/src/ast/expr.rs
+++ b/crates/wast/src/ast/expr.rs
@@ -1,4 +1,4 @@
-use crate::ast::{self, kw, RefType};
+use crate::ast::{self, kw, HeapType};
 use crate::parser::{Parse, Parser, Result};
 use std::mem;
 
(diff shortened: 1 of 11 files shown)


crates/wast/src/ast/expr.rs
```diff
@@ -411,7 +411,7 @@ instructions! {
         TableSize(TableArg<'a>) : [0xfc, 0x10] : "table.size",
         TableGrow(TableArg<'a>) : [0xfc, 0x0f] : "table.grow",
 
-        RefNull(RefType<'a>) : [0xd0] : "ref.null",
+        RefNull(HeapType<'a>) : [0xd0] : "ref.null",
```
alexcrichton, line 414: Should this be removed to account for https://github.com/WebAssembly/reference-types/pull/100?

eqrion: Ah no, that just applies to ref.is_null which was #28. This change is from function-references [1].

[1] https://github.com/WebAssembly/function-references/blob/master/proposals/function-references/Overview.md#optional-references

alexcrichton: Wow apparently I need to learn how to read...

PR #33, opened 2020-06-26
Add back 'anyref'

eqrion:
The GC proposal now brings back the 'anyref' type, so we should be able
to parse it.


alexcrichton:
👍

PR #36, opened 2020-06-29
Bump wasmparser to 0.58.0

abrown:
I believe this is necessary to get access to new instructions such as `bitmask`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -80,7 +80,7 @@ jobs:
     - uses: actions/checkout@v2
       with:
         submodules: true
-    - run: git fetch origin main && git branch main origin/main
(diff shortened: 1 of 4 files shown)


abrown:
cc: @yurydelendik

abrown:
Hm... branch names? https://github.com/bytecodealliance/wasm-tools/pull/36/checks?check_run_id=819643898#step:3:39

alexcrichton:
Ok I pushed a few commits which I think fixes the benchmarking stuff. Turned out in the past few hours `cargo critcmp` was broken by criterion updates. What bad timing! I'll leave merging this up to @yurydelendik though.

yurydelendik:
Thank you for the patch!

PR #41, opened 2020-07-12
wast/wat: Make the crates useable in no_std environments

athei:
This PR adds the default enabled feature `std` to the `wast` and `wat` crate. Disabling it makes the crate perform in a `no_std` mode.

## Motivation
We need to compile wat -> wasm in a no_std environment and there is no real reason why this crate cannot work in such an environment given that an allocator is provided.

## Changes made

These are the changes I needed to make to have a working no_std build. Most of them are non-controversial but some could have been done in a different way. I tried to not clutter the codebase with conditional compilation.

### Switch `std` to `core` import statements

I switched all imports that have a representation in `core` to just use those instead. I am not sure if this is a recent addition to the compiler but this also happened to just work in std builds.

### Add `alloc` import statements

Mostly collections but also other stuff (`format!`) from the std prelude needs to be added explicitly in no_std build. This also happens to not break the std build.

### Remove usage of `Path` and `PathBuf`

We breaks the API of `wast` here and request the user to supply the path as `String` to the `set_path` function as no_std does not have these types. I figured that the advantage of storing a `PathBuf` instead of a `String` is too small to justify conditional compilation here.

### Conditionally deactivate the `parse_file` API of `wat`

When `wat` is compiled without the `std` feature we remove this API for obvious reasons.

### Replace `leb128` dependency with `nano-leb128`

The currently used `leb128` needs std as it relies on the `std::io::{Read,Write}` traits. The replacement has a no_std mode and works just fine. It has fewer users but does not use unsafe und is very small. When `leb128` adds a no_std mode we could switch back.

### Replace `HashMap` by `BTreeMap` when no_std

`HashMap` is not available in alloc. We use `BTreeMap` as drop in replacement when `std` is deactivated.

### Replace `thread_local!` by `spin::Mutex` when no_std

This seemed to be the easiest solution to implement the gensym as-is. AFAIK there is no way to only depend on `spin? when `std` is **not** set. Or is there one?

### DIFF
--- a/crates/wast/Cargo.toml
+++ b/crates/wast/Cargo.toml
@@ -13,7 +13,9 @@ Customizable Rust parsers for the WebAssembly Text formats WAT and WAST
 """
 
 [dependencies]
-leb128 = "0.2"
+nano-leb128 = { version = "0.1.0", default-features = false }
+# only used when std is not set to emulate thread_local
+spin = "0.5.2"
 
 [dev-dependencies]
 anyhow = "1.0"
@@ -22,7 +24,7 @@ wasmparser = { path = "../wasmparser" }
 wat = { path = "../wat" }
 
 [features]
-default = ['wasm-module']
+default = ['wasm-module', 'std']
 
 # Includes default parsing support for `*.wat` and `*.wast` files (wasm
 # modules). This isn't always needed though if you're parsing just an
@@ -31,6 +33,8 @@ default = ['wasm-module']
 #
 # This feature is turned on by default.
 wasm-module = []
+
+std = []
 
 [[test]]
 name = "parse-fail"
--- a/crates/wast/README.md
+++ b/crates/wast/README.md
@@ -38,24 +38,33 @@ interface if all you'd like to do is convert `*.wat` to `*.wasm`.
 
 ## Cargo features
 
-By default this crate enables and exports support necessary to parse `*.wat` and
+* std (enabled by default)
+
+Enables usage of the Rust standatd library. This makes the `Error` type implement
+`std::error::Error`. In addition some performance improvements are gained with this
+feature enabled as `thread_local` and `HashMap` are used as opposed to `spin` and
+`BTreeMap`.
+
+* wasm-module (enabled by default)
+
+This enables and exports support necessary to parse `*.wat` and
 `*.wast` files, or in other words entire wasm modules. If you're using this
 crate, however, to parse simply an s-expression wasm-related format (like
-`*.witx` or `*.wit` perhaps) then you can disable the default set of features to
+`*.witx` or `*.wit` perhaps) then you can disable this feature to
 only include a lexer, the parsing framework, and a few basic token-related
 parsers.
 
 ```toml
 [dependencies]
-wast = { version = "17.0", default-features = false }
+wast = { version = "21.0", default-features = false, features = ['std'] }
 ```
 
-# License
+## License
 
 This project is licensed under the Apache 2.0 license with the LLVM exception.
 See [LICENSE](LICENSE) for more details.
 
-### Contribution
+## Contribution
 
 Unless you explicitly state otherwise, any contribution intentionally submitted
 for inclusion in this project by you, as defined in the Apache-2.0 license,
--- a/crates/wast/src/ast/custom.rs
+++ b/crates/wast/src/ast/custom.rs
@@ -1,5 +1,6 @@
 use crate::ast::{self, annotation, kw};
 use crate::parser::{Parse, Parser, Result};
+use alloc::vec::Vec;
 
 /// A wasm custom section within a module.
 #[derive(Debug)]
--- a/crates/wast/src/ast/export.rs
+++ b/crates/wast/src/ast/export.rs
@@ -1,5 +1,6 @@
 use crate::ast::{self, kw};
 use crate::parser::{Cursor, Parse, Parser, Peek, Result};
+use alloc::vec::Vec;
 
 /// A entry in a WebAssembly module's export section.
 #[derive(Debug)]
--- a/crates/wast/src/ast/expr.rs
+++ b/crates/wast/src/ast/expr.rs
@@ -1,6 +1,7 @@
 use crate::ast::{self, kw, HeapType};
 use crate::parser::{Parse, Parser, Result};
-use std::mem;
+use core::mem;
+use alloc::vec::Vec;
 
 /// An expression, or a list of instructions, in the WebAssembly text format.
 ///
--- a/crates/wast/src/ast/func.rs
+++ b/crates/wast/src/ast/func.rs
@@ -1,5 +1,6 @@
 use crate::ast::{self, kw};
 use crate::parser::{Parse, Parser, Result};
+use alloc::vec::Vec;
 
 /// A WebAssembly function to be inserted into a module.
 ///
--- a/crates/wast/src/ast/instance.rs
+++ b/crates/wast/src/ast/instance.rs
@@ -1,5 +1,6 @@
 use crate::ast::{self, kw};
 use crate::parser::{Parse, Parser, Result};
+use alloc::vec::Vec;
 
 /// A nested WebAssembly instance to be created as part of a module.
 #[derive(Debug)]
--- a/crates/wast/src/ast/memory.rs
+++ b/crates/wast/src/ast/memory.rs
@@ -1,5 +1,6 @@
 use crate::ast::{self, kw};
 use crate::parser::{Parse, Parser, Result};
+use alloc::vec::Vec;
 
 /// A defined WebAssembly memory instance inside of a module.
 #[derive(Debug)]
--- a/crates/wast/src/ast/module.rs
+++ b/crates/wast/src/ast/module.rs
@@ -1,5 +1,6 @@
 use crate::ast::{self, annotation, kw};
 use crate::parser::{Parse, Parser, Result};
+use alloc::vec::Vec;
 
 pub use crate::resolve::Names;
 
@@ -80,7 +81,7 @@ impl<'a> Module<'a> {
     ///
     /// If an error happens during resolution, such a name resolution error or
     /// items are found in the wrong order, then an error is returned.
-    pub fn resolve(&mut self) -> std::result::Result<Names<'a>, crate::Error> {
+    pub fn resolve(&mut self) -> core::result::Result<Names<'a>, crate::Error> {
         crate::resolve::resolve(self)
     }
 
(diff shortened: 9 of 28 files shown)


alexcrichton:
Thanks for the PR! Before I dive too much into the details though, can you detail a bit more about your use case? What targets are you interested in compiling for?

The downside of PRs like this is that it makes it very difficult to support this feature over time. The idioms are vastly different which makes code difficult to both read and write, and primarily it also makes it difficult for new contributors because of how different the code is. I personally vastly prefer to simply get `std` compiling for the desired target since it makes everything else easier where possible. That may not fix all the issues you've found here, but it seems like it would fix the lion's share.

athei:
The only target I am interested in here is `wasm32-unknown-unknown`. The bottom line is that we have some benchmarks that must run with this target because that is what we are benchmarking. As part of these benchmarks we need some hand written wasm blobs to facilitate those benchmarks.

We are **not** benchmarking the compilation itself. The compilation is part of the test fixture. I considered the follwing alternatives:

1. Check in compiled wasm blobs and use those in the benchmarks: Dismissed for obvious reasons.
2. Create a `build.rs` and do the compilation there. This might be possible but is of course undesirable because it further complicates the build process. 

> The idioms are vastly different which makes code difficult to both read and write, and primarily it also makes it difficult for new contributors because of how different the code is.

I do not understand what you mean. Which idioms?

alexcrichton:
Hm when you say the target you're intersted in is wasm32-unknown-unknown, isn't that already supported by this crate? I thought `wat` already compiled and worked for wasm?

For idioms I largely mean everything in this PR. Everything here is pretty non-standard or even wrong (`String` is not an infallilble substite for `Path`, `spin` is almost always wrong unless you're in a context where you can disable interrupts, etc)

fitzgen:
For example, this compiles fine on `wasm32-unknown-unknown`:

```toml
# Cargo.toml

[package]
name = "wat2wasm"
version = "0.1.0"
authors = ["Nick Fitzgerald"]
edition = "2018"

[lib]
crate-type = ["cdylib"]

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
wat = "1"
```

```rust
// src/lib.rs

#[no_mangle]
pub unsafe extern "C" fn wat2wasm(text: *const std::os::raw::c_char, out: *mut *mut u8) -> usize {
    let text = std::ffi::CStr::from_ptr(text);
    let text = text.to_str().unwrap();
    let mut data = wat::parse_str(text).unwrap();
    std::ptr::write(out, data.as_mut_ptr());
    let len = data.len();
    std::mem::forget(data);
    len
}
```

athei:
Yes but I need to support a `!#[no_std]` target making use of `wasm32-unknown-unknown`. I see no path for it to move to `std` nor do I think this would be the right decision: I believe using `std` on an `unknown` target is fundamentally wrong. It moves compile time errors to runtime.

Ultimately it is your decision whether you want to support `#![no_std]` targets as you have to maintain this support and I respect that. If you reject the idea of having this support feel free to close the issue.

Otherwise I am happy to iterate on this PR until you are happy with it.

alexcrichton:
Oh ok so I'm just trying to understand more about the target you're compiling the `wast`/`wat` crates for. Can you explain more about it?

athei:
Sure. We develop a [framework for creating blockchains](https://github.com/paritytech/substrate). The logic of the blockchain is compiled to wasm and stored in the chain itself. It is of utmost importance that this wasm blob is small as possible and never panics. For that reason `#![no_std]` was chosen as it was thought of the best way to have more control. This decision might even predate std support for wasm (I am not sure, though). Maybe today it would make sense to use `std` for that blob but this is a huge code base and this transition would be a controversial topic amongst our developers (I imagine). The point is that there are legitimate users of `wasm32-unknown-unknown` that do not use std.

It is that runtime that also executes benchmark of individual functions in order to attach costs to the calling of this functions which is needed to protect the chain from DoS attacks. It is this benchmark code where we need to compile `wat2wasm` (or rather it would be very helpful). The benchmarking code is locked away behind a cargo feature in order to not have this in the actual production wasm blob. 

That said, I am currently exploring the idea of allowing this "benchmarking wasm" to use std crates. I am doing this by not compiling in any handlers and language items that are already provided by std when the benchmarking feature is set.

This already compiles and might be a way forward without needing `wat` to be `no_std`. Despite the fact that `wast`compiles under `wasm32-unknown-unknown`, it still won't work because of its `thread_local!` usage which is just a panic under that target. So at least we need to replace this (we could use  an atomic counter rather than spin).

alexcrichton:
@athei sorry I think I'm still a bit confused. Are you compiling this crate, `wat`, to WebAssembly? Or are you compiling this crate, `wat`, to a different architecture which does not have the standard library?

Sorry I'm still just trying to understand the motivations here, and I'm having a tough time getting past that :(

athei:
> Are you compiling this crate, `wat`, to WebAssembly? Or are you compiling this crate, `wat`, to a different architecture which does not have the standard library?

The first: Compiling to WebAssembly.

alexcrichton:
Ah ok, thanks for the clarification! So in light of that I'm a bit confused by this PR. The WebAssembly target has the standard library so I'm not sure why we'd need to use `#![no_std]` instead. In terms of the final binary any optimizations here are just as applicable when using the standard library vs no, so I don't think there's any reason to use `#![no_std]`. A good example is that if nothing ends up actually being used from `std` then it won't even be linked.

Some things you've mentioned though which I can address for wasm are:

* Code size - definitely agreed that code size matters a lot with wasm! You mentioned panics as bloating the binary a bit, and that can definitely happen on wasm. We've got [online guides for shrinking wasm size](https://rustwasm.github.io/book/reference/code-size.html), so assuming all the compile options are taken care of what's left is code changes. The impact of panicking with and without std is pretty similar, and the benefit of removing panics will affect this crate regardless of whether `#![no_std]` uses. In that sense switching to `#![no_std]` won't automatically make code size smaller, but instead changing idioms is what will make code size smaller.

* Thread locals - this works natively on wasm targets and since there are no threads it compiles them to globals. I'm not sure what you've been seeing, though. Do thread locals appear to not work for you?

* Panics - similar to code size, switching to `#![no_std]` doesn't automatically remove panics. It's definitely possible to change idioms to remove panics, but there's no magic bullet for "remove all panics from this program".

Overall it sounds like you've got some pain points compiling `wat` to WebAssembly itself which I'd love to help fix, but I don't think that fixing these issues requires switching to `#![no_std]` (nor does `#![no_std]` I think automatically fix any of these issues).

athei:
> The WebAssembly target has the standard library so I'm not sure why we'd need to use #![no_std] instead.

Mainly historical and philosophical reasons I guess. And is not easily changeable right now for reasons I mentioned in my previous post. My line of thinking was that providing `no_std` support is best practice when there are no real reasons why not and therefore a PR providing this would be welcome.

That maybe we should move to `std` in the future is a separate topic.

> Thread locals - this works natively on wasm targets and since there are no threads it compiles them to globals. I'm not sure what you've been seeing, though. Do thread locals appear to not work for you?

As a matter of fact, yes they work for me. I wrote that before trying and was thinking they aren't because of this:
https://github.com/rust-lang/rust/blob/e55d3f9c5213fe1a25366450127bdff67ad1eca2/src/libstd/sys/wasm/thread_local.rs#L1-L26

alexcrichton:
Ok so in any case I'm happy to merge updates to this crate which improve code size or reduce panics. I don't feel, though, that this motivation is strong enough to switch to `#![no_std]` and have such an impact on the development of this crate.

Are you interested in investigating what changes here are improving code size or reducing panics and trimming this PR down to that?

athei:
I think you miss-understood me (or I miss-understood you). I was never meaning to say that this crate specifially was causing panics or increase code size.

I just used that as a general argument for `#![no_std]` as there are obviously things that panic in the wasm std like the time APIs. And that just removing `#![no_std]` from our wasm binary increases its size was an observation made a long time ago and should be properly evaluated against current Rust.

In my current experiment the code size actually increases by a good chunk (200k) when using the upstream version of `wast` vs. using my PR branch. I did not evaluate why this is, though. I can imagine a good chunk is the std allocator which is for more complex (I imagine) than what we are using (delegating this completely to the embedding application).

alexcrichton:
Oh sure yeah but as you've mentioned these changes make the crate have a smaller code size, so from that angle *something* here is improving things for y'all. I don't think it has anything to do with `#![no_std]`, however, and probably has more to do with the code changes (e.g. changes in maps or leb128 implementations). The allocator may matter a bit but you can swap that out with and without `#![no_std]`.

FWIW simply marking a crate as `#![no_std]` is unlikely to really affect that much in terms of binary costs. If something drastically changes there's either something else behind the scenes or that's a bug in libstd itself.

athei:
> Oh sure yeah but as you've mentioned these changes make the crate have a smaller code size, so from that angle _something_ here is improving things for y'all.

If I said that, this came across completely wrong then. Where did I say that? Honestly asking.

alexcrichton:
Oh I was just referring to [when you said](https://github.com/bytecodealliance/wasm-tools/pull/41#issuecomment-658826990) "In my current experiment the code size actually increases by a good chunk (200k) when using the upstream version of `wat` vs. using my PR branch"

athei:
> Oh I was just referring to [when you said](https://github.com/bytecodealliance/wasm-tools/pull/41#issuecomment-658826990) "In my current experiment the code size actually increases by a good chunk (200k) when using the upstream version of `wat` vs. using my PR branch"

Yes you are right. I just assumed that this would be due to `std` and not the changes to the `wast` crate. You obviously have another opinion. And it turns out that you are right. I compared the two optimized wasm outputs and just assumed that the smaller one must be the `no_std` version because of my bias. But in reality the output is smaller when just using the upstream version of `wat` in our `#![no_std]` binary:
```
-rw-r--r--  1 alexander  staff   3.5M Jul 15 17:36 no_std.wasm
-rw-r--r--  1 alexander  staff   3.3M Jul 15 17:31 std.wasm
```

I cannot explain why, though.

So this whole PR is based off some of misconceptions I had when opening it:
* `wasm32-unknown-unknown` cannot use `std` (wrong: this was added at some point in the past)
* I cannot depend on a std crate from a no-std crate (I can when not specifying my own panic hander)

Thank you for your patience in clearing that up for me. Naturally, I just lost my immediate interest in getting this merged.

alexcrichton:
No worries!

It is worth pointing out though that I've made basically zero effort trying to optimize the code size of this library. I'm sure there's opportunities to improve it, but in general for handling wasm code size I'd recommend looking at https://rustwasm.github.io/book/reference/code-size.html for profiling, compile flags, etc.

alexcrichton:
Ok I'm going to go ahead and close this as it sounds like some confusion was cleared up and the motivation for this PR has changed. If this comes up again though please feel free to resubmit!

PR #45, opened 2020-07-13
Enable reference types spec tests

alexcrichton:
Update the encoding of `ref.is_null` to run the reference types spec test suite, and tweak a few things here and there to get everything working.


fitzgen:
Do you plan on updating wasmtime for these changes too?

alexcrichton:
Indeed!

alexcrichton:
Ok I'm gonna ignore the benchmark build failure. That's actually fixed in this PR but the script switching to master doesn't handle updating submodules so it'll fail until we get a fix in to master anyway.

PR #53, opened 2020-07-15
wasmprinter: Limit the maximum number of locals

### LINKED ISSUE #52: [crash/fuzzing] Resources exhaustion (CPU/MEM) using wasmprinter::print_bytes()
During fuzzing of `wasmprinter`, I trigger a resources exhaustion bug leading to full CPU usage when parsing a crafted wasm module. The file is only 650 bytes in size.

Download: [huge_cpu_usage_wasmprinter.zip](https://github.com/bytecodealliance/wasm-tools/files/4926087/huge_cpu_usage_wasmprinter.zip)

Repo:
``` rust
use std::env;
use std::fs::File;
use std::io;
use std::io::Read;

/// Read the contents from file path
fn read_contents_from_path(path_str: &String) -> Result<Vec<u8>, io::Error> {
    let mut buffer: Vec<u8> = Vec::new();
    let file_path = std::path::PathBuf::from(path_str);

    println!("file_to_process: {:?}", file_path);

    let mut file = File::open(file_path)?;
    file.read_to_end(&mut buffer)?;
    drop(file);
    Ok(buffer)
}

fn main() {
    println!("Start debugging of wasmprinter_parser");
    let args: Vec<String> = env::args().collect();

    // verify file_to_process is provided
    if args.len() != 2 {
        println!("Usage: wasmprinter_parser <file_to_process>\n");
        return;
    }

    // read data from provided file
    let data = read_contents_from_path(&args[1]).expect("cannot read file content");

    // call the fuzzing target
    wasmprinter::print_bytes(&data).is_ok();

    println!("No crash, everything is OK\n");
}

```

alexcrichton:
Don't inflate a small file to gigabytes of `(local` declarations!

Closes #52

### DIFF
--- a/crates/wasmprinter/src/lib.rs
+++ b/crates/wasmprinter/src/lib.rs
@@ -15,6 +15,8 @@ use std::mem;
 use std::path::Path;
 use wasmparser::*;
 
+const MAX_LOCALS: u32 = 50000;
+
 /// Reads a WebAssembly `file` from the filesystem and then prints it into an
 /// in-memory `String`.
 pub fn print_file(file: impl AsRef<Path>) -> Result<String> {
@@ -562,6 +564,13 @@ impl Printer {
             let mut locals = NamedLocalPrinter::new("local");
             for local in body.get_locals_reader()? {
                 let (cnt, ty) = local?;
+                if MAX_LOCALS
+                    .checked_sub(local_idx)
+                    .and_then(|s| s.checked_sub(cnt))
+                    .is_none()
+                {
+                    bail!("function exceeds the maximum number of locals that can be printed");
+                }
                 for _ in 0..cnt {
                     if first {
                         self.result.push_str("\n   ");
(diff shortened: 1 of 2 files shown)


crates/wasmprinter/src/lib.rs
```diff
@@ -15,6 +15,8 @@ use std::mem;
 use std::path::Path;
 use wasmparser::*;
 
+const MAX_LOCALS: u32 = 50000;
```
tschneidereit, line 18: Are you sure this is enough? While it might seem like it'd obviously be, I vaguely remember some preposterously large functions generated by Emscripten back in the asm.js days. Perhaps toolchains don't do those kinds of things anymore?

fitzgen: I would copy whatever cranelift / wasmparser already have as their limit.

fitzgen: Which is also `50000` :laughing:

tschneidereit: hah, fair enough 😁 I do wish the spec imposed limits here, but oh well.

PR #56, opened 2020-07-16
Implement function-references proposal and update GC proposal

eqrion:
Take 2 :)

### DIFF
--- a/crates/wast/src/resolve/names.rs
+++ b/crates/wast/src/resolve/names.rs
@@ -1833,7 +1833,7 @@ impl<'a, 'b> ExprResolver<'a, 'b> {
                 ));
             }
 
-            Br(i) | BrIf(i) | BrOnCast(i) | BrOnNull(i) => {
+            Br(i) | BrIf(i) | BrOnNull(i) => {
                 self.resolve_label(i)?;
             }
 
@@ -1851,6 +1851,11 @@ impl<'a, 'b> ExprResolver<'a, 'b> {
                 self.resolve_label(&mut b.label)?;
                 self.module.resolve(&mut b.exn, Ns::Event)?;
             }
+            BrOnCast(b) => {
+                self.resolve_label(&mut b.label)?;
+                self.module.resolve_heaptype(&mut b.val)?;
+                self.module.resolve_heaptype(&mut b.rtt)?;
+            }
 
             Select(s) => {
                 if let Some(list) = &mut s.tys {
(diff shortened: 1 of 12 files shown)


crates/wast/src/resolve/names.rs
```diff
@@ -1730,7 +1809,14 @@ impl<'a, 'b> ExprResolver<'a, 'b> {
             }
 
             LocalSet(i) | LocalGet(i) | LocalTee(i) => {
-                self.locals.resolve(i, "local")?;
+                for (depth, scope) in self.scopes.iter().enumerate().rev() {
+                    if let Err(e) = scope.resolve(i, "local") {
+                        if depth == 0 {
+                            return Err(e);
+                        }
+                    }
```
alexcrichton, line 1817: Should this `break` once the local is resolved?

crates/wast/src/resolve/names.rs
```diff
@@ -1328,32 +1326,35 @@ impl<'a> Module<'a> {
...
                     }
 
                     // .. followed by locals themselves
-                    for (id, _, _) in locals {
-                        resolver.locals.register(*id, "local")?;
+                    for local in locals {
+                        scope.register(local.id, "local")?;
                     }
 
+                    resolver.scopes.push(scope);
```
alexcrichton, line 1356: Could you add a debug assert that after `resolve` below the length of `scopes` is 1?

crates/wast/src/resolve/names.rs
```diff
@@ -1675,18 +1688,26 @@ impl<'a, T> Default for Namespace<'a, T> {
...
+    label: Option<Id<'a>>,
+    // Whether this block pushed a new scope for resolving locals
+    pushed_scope: bool,
+}
+
 struct ExprResolver<'a, 'b> {
     module: &'b Module<'a>,
-    locals: Namespace<'a, ()>,
-    labels: Vec<Option<Id<'a>>>,
+    scopes: Vec<Namespace<'a, ()>>,
```
alexcrichton, line 1701: Mind adding a comment to this saying that the scopes here are for locals and that this is dynamically managed as we exit/enter `let` blocks?

crates/wast/src/resolve/names.rs
```diff
@@ -1742,87 +1828,62 @@ impl<'a, 'b> ExprResolver<'a, 'b> {
...
                 }
 
-                // If the inline annotation persists to this point then resolve
-                // all of its inline value types.
-                if let Some(inline) = &mut bt.ty.inline {
-                    inline.resolve(self.module)?;
+                // Register all locals defined in this let
+                let mut scope = Namespace::default();
+                for local in &t.locals {
+                    scope.register(local.id, "local")?;
```
alexcrichton, line 1844: The semantics of this implementation I think means that you can redefine a local name multiple times, it'll just shadow within a scope. Do you know if that's the intention of the `let` proposal? Or is it intended that shadowing is not allowed and redefining a name should be an error?

crates/wast/src/resolve/names.rs
```diff
@@ -1730,7 +1809,14 @@ impl<'a, 'b> ExprResolver<'a, 'b> {
             }
 
             LocalSet(i) | LocalGet(i) | LocalTee(i) => {
-                self.locals.resolve(i, "local")?;
+                for (depth, scope) in self.scopes.iter().enumerate().rev() {
```
alexcrichton, line 1812: Mind throwing an asert above this that `self.scopes.len() > 0`?

crates/wast/src/resolve/names.rs
eqrion: Yup, nice catch. Impressed that it worked without that.

crates/wast/src/resolve/names.rs
eqrion: Yes, that appears to be the intent of the proposal. I originally had it the other way around, and it failed a couple tests in `let.wast` which checked that shadowing was the behavior.

eqrion:
@alexcrichton This should be ready to go now.

alexcrichton:
Looks great to me, thanks again @eqrion!

Would y'all like a publish with this or did you want to reorganize the enum first?

alexcrichton:
Also if you'd like, I'd be happy to add you as an owner of the crate on crates.io so you can push too (and push access here)

PR #61, opened 2020-07-20
Unify parsing of `BrTable`

### LINKED ISSUE #58: validate accepts invalid wasm
The attached wasm file (g
[x.wasm.gz](https://github.com/bytecodealliance/wasm-tools/files/4940433/x.wasm.gz)
zipped due to github file type detection) triggers no errors with the wasmparser validator, but it has an invalidly large label index (0x8180808021) in the br_table default label.

Wabt detects the error:
```
$ wabt/build/wasm2wat x.wasm
000001f: error: unable to read u32 leb128: br_table default target depth
```

alexcrichton:
This commit unifies parsing methods of `BrTable` to make sure there's
only one way to parse it, and then it ensures that the one way to parse
it handles everything necessary to parse it, namely parsing N items as
specified by the count of the table instruction.

Closes #58


crates/wasmparser/src/binary_reader.rs
```diff
@@ -1776,89 +1777,34 @@ impl<'a> BrTable<'a> {
...
-            table.push(reader.read_var_u32()?);
-        }
-        let default_target = table.pop().ok_or_else(|| {
-            BinaryReaderError::new(
-                "br_table missing default target",
-                reader.original_position(),
-            )
-        })?;
-        Ok((table.into_boxed_slice(), default_target))
+    pub fn targets<'b>(&'b self) -> impl Iterator<Item = Result<(u32, bool)>> + 'b {
```
Robbepop, line 1780: I'd love to preserve the little documentation we have in this crate.

crates/wasmparser/src/operators_validator.rs
```diff
@@ -910,7 +910,11 @@ impl OperatorValidator {
             Operator::BrTable { ref table } => {
                 self.check_operands_1(Type::I32)?;
                 let mut depth0: Option<u32> = None;
-                for relative_depth in table {
+                for element in table.targets() {
+                    let (relative_depth, _default) = element.map_err(|mut e| {
```
Robbepop, line 914:
```suggestion
                    let (relative_depth, _is_default) = element.map_err(|mut e| {
```

Robbepop: `is_default` makes it clearer that this is a `bool`. at first I thought you'd always iterate over all targets and the default target but and had to look at the signature to clear this confusion.

crates/wasmprinter/src/lib.rs
```diff
@@ -670,9 +670,9 @@ impl Printer {
                 )?;
             }
             BrTable { table } => {
-                let (table, default) = table.read_table()?;
                 self.result.push_str("br_table");
-                for item in table.iter().cloned().chain(Some(default)) {
+                for item in table.targets() {
+                    let (item, _default) = item?;
```
Robbepop, line 675:
```suggestion
                    let (item, _is_default) = item?;
```

crates/wasmparser/src/binary_reader.rs
```diff
@@ -1808,6 +1807,22 @@ impl<'a> BrTable<'a> {
     }
 }
 
+impl fmt::Debug for BrTable<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let mut f = f.debug_struct("BrTable");
+        f.field("count", &self.cnt);
+        match self.read_table() {
+            Ok((targets, default)) => {
```
Robbepop:
```suggestion
            Ok((targets, is_default)) => {
```

crates/wasmparser/src/binary_reader.rs
```diff
@@ -1808,6 +1807,22 @@ impl<'a> BrTable<'a> {
...
+                f.field("targets", &targets).field("default", &default);
+            }
```
Robbepop:
```suggestion
            Ok((targets, is_default)) => {
                f.field("targets", &targets).field("is_default", &is_default);
            }
```

alexcrichton:
Oops sorry was a bit too hasty in dropping the docs!

PR #62, opened 2020-07-21
Updates for integrating validation in Wasmtime

alexcrichton:
This commit is a batch of updates I've made and found when integrating
wasmparser's validation more intrusively into Wasmtime. The changes made
here are:

* `WasmFeatures` is now exposed as a standalone type so it can be shared
  with other crates. This way only one crate, wasmparser, declares the
  "bag of features" and everyone else can just reference that one bag.

* Parsing locals is deferred in `code_section_entry` to actual
  validation. Helper methods have been added to advance a `BinaryReader`
  by reading the locals, as well as methods added to define locals
  yourself. The intention here is to avoid parsing the locals section
  twice, once in `wasmparser` and once by the application. This also
  removed a few `read_local*` methods from `BinaryReader` to defer
  validation to `Validator`.

* `WasmTypeDef` and `type_at` have been removed from
  `WasmModuleResources`. These weren't ever really that useful and
  callers only ever cared about function types anyway. The previous
  `func_type_at` method has been repurposed for looking up a type index
  and a new `type_of_function` looks up the type of a function by index.

* Parsing for element segments was slightly updated to ensure that
  `externref` segments don't have `ref.func` in them.

### DIFF
--- a/crates/wasmparser/src/binary_reader.rs
+++ b/crates/wasmparser/src/binary_reader.rs
@@ -73,7 +73,7 @@ impl Range {
 }
 
 /// A binary reader of the WebAssembly structures and types.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Hash)]
 pub struct BinaryReader<'a> {
     pub(crate) buffer: &'a [u8],
     pub(crate) position: usize,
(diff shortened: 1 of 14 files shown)


alexcrichton:
Hm actually I forgot there's one more large-ish feature, so I'll be adding that on here.

alexcrichton:
Ok I've added the API intended, which was `FuncValidator::new` to create an arbitrary function validator with a `WasmModuleReferences` instance.

alexcrichton:
Ok, and another commit for `WasmFuncType::{inputs,outputs}` to get at the inputs/outputs a bit more ergonomically.

crates/wasmparser/src/binary_reader.rs
```diff
@@ -73,7 +73,7 @@ impl Range {
 }
 
 /// A binary reader of the WebAssembly structures and types.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Hash)]
```
yurydelendik, line 76: I wonder how/where `Hash` will be used -- `buffer` can be long.

alexcrichton: FWIW this is [used here](https://github.com/bytecodealliance/wasmtime/pull/2059/files#diff-70306a13ab1567aab84c018246072737R523) where we has the contents of functions for the internal cache.

PR #68, opened 2020-07-28
Simplify the `WasmModuleResources` trait

alexcrichton:
* Don't have separate traits for tables/memories/globals when the
  wasmparser `Copy` types by value work for now. This may need to be
  extended later but we can presumably cross that bridge when we get
  there.

* Return bare `T` instead of `&T` where possible, reducing the need for
  long-lasting borrows or self-references where possible.

After this only `WasmFuncType` and `WasmModuleResources` need to be
implemented, and functions are largely just to handle the vector of
arguments/return values in an abstract way (iterators, not slices).

### DIFF
--- a/crates/wasmparser/src/module_resources.rs
+++ b/crates/wasmparser/src/module_resources.rs
@@ -13,26 +13,11 @@
  * limitations under the License.
  */
 
+use crate::{FuncType, GlobalType, MemoryType, TableType, Type};
 use std::ops::Range;
-
-/// Types that qualify as Wasm types for validation purposes.
-///
-/// Must be comparable with `wasmparser` given Wasm types and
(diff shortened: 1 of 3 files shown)


yurydelendik:
Use case for using the deleted interfaces is described at https://github.com/bytecodealliance/wasmparser/pull/172

@Robbepop is it still the case?

alexcrichton:
I'm gonna go ahead and merge this since I think it will still cover existing use cases, but if it ends up not sufficing we can always have further PRs!

Robbepop:
Sorry for not answering. I am no longer really into the codebase well enough to give a quick answer. I guess we should let @alexcrichton be a machine and as they said whenever there is the need for adjustments we can simply do them.

PR #70, opened 2020-07-31
Add parsing for folded `try` statements.

takikawa:
Currently folded `try` statements for the exception handling proposal like `(try (do ...) (catch ...)` are not supported in the parser. This pull request adds support for those.

I added success tests in a `try.wat` and error tests in a `try.wast` as putting the success tests in `.wast` seemed to also trigger validation testing, which isn't supported yet for exceptions AFAICT. (though I did start working on commits for validation as well, not in this PR)

### DIFF
--- a/crates/wast/src/ast/expr.rs
+++ b/crates/wast/src/ast/expr.rs
@@ -57,6 +57,13 @@ enum Level<'a> {
     /// which don't correspond to terminating instructions, we're just in a
     /// nested block.
     IfArm,
+
+    /// Similar to `If` but for `Try` statements, which has simpler parsing
+    /// state to track.
+    Try(Try<'a>),
+
+    /// Similar to `IfArm` but for `(do ...)` and `(catch ...)` blocks.
+    TryArm,
 }
 
 /// Possible states of "what should be parsed next?" in an `if` expression.
@@ -69,6 +76,17 @@ enum If<'a> {
     /// Next thing to parse is the `else` block
     Else,
     /// This `if` statement has finished parsing and if anything remains it's a
+    /// syntax error.
+    End,
+}
+
+/// Possible state of "what should be parsed next?" in a `try` expression.
+enum Try<'a> {
+    /// Next thing to parse is the `do` block.
+    Do(Instruction<'a>),
+    /// Next thing to parse is the `catch` block.
+    Catch,
+    /// This `try` statement has finished parsing and if anything remains it's a
     /// syntax error.
     End,
 }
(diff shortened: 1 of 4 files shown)


alexcrichton:
Awesome, thanks! Are the upstream spec tests using this syntax as well, and if so can the tests be enabled here?

takikawa:
Unfortunately the upstream spec doesn't yet have tests (but would likely use this syntax, as the WIP reference interpreter does). Wabt has this syntax for various tests (e.g., https://github.com/WebAssembly/wabt/blob/master/test/roundtrip/fold-try.txt).

crates/wast/src/ast/expr.rs
```diff
@@ -255,6 +299,51 @@ impl<'a> ExpressionParser<'a> {
...
+        }
+
+        if let Try::Catch = i {
+            self.instrs.push(Instruction::Catch);
+            if parser.parse::<Option<kw::catch>>()?.is_some() {
+                *i = Try::End;
+                self.stack.push(Level::TryArm);
+                return Ok(true);
+            }
+            return Ok(false);
```
alexcrichton, line 342: When this happens (and in `Try::Do` above), the internal state isn't advanced, but I think that might lead to bugs? It's been awhile since I wrote this but in the `If` parsing above each state always advances to the next state unconditionally, even on failed parses I think?

takikawa: Right, I see that `If` does advance the state after each step, but I thought that was because `If` is a bit more permissive than the `Try` case. In particular, if a `do` or `catch` in `try` are missing the intent is to error while for `If` I think it allows the `then` clause to not have the keyword. (or should `try` be more permissive like `if` too?)

My understanding is that if the state isn't advanced, eventually a `Paren::Right` will come up and the `Level::Try(...)` cases there should signal the failing parse. But I might not be understanding the code correctly. :) Happy to try any additional test cases that might clarify this.

alexcrichton: Ah yeah it's ok to not be as flexible as `if`, although could you expand a bit on the purpose of these then? I'm having a bit of a difficult time wrapping my head around what this is enabling by not returning an error or not updating the state when the group doesn't start with what we'd expect

takikawa: Ah, right. So I think this code could return an error immediately here if the keyword parse fails where it currently returns `Ok(false)`. But I think it would still also need the other check in the `Paren::Right` case, making it a bit redundant (the same error would be returned in both spots, I think).

Specifically, for a case like `(try (do))` with a missing `catch` clause. The state will be advanced to `Try::Catch` after parsing `(do)` but won't go into `handle_try_lparen` again as the expression just ends, so the error should be returned by the right paren parsing.

If you'd prefer I could add the error returns in here too so the control flow is clearer. Or maybe I can add a comment clarifying that it doesn't error yet because the right paren case should catch it anyway?

alexcrichton: Aha that makes sense, thanks for clarifying! Yeah if you wouldn't mind throwing a comment here that'd be great, and otherwise this looks good to me!

takikawa: Just pushed a new version of the commit with more comments. :)

alexcrichton:
Thanks!

PR #71, opened 2020-08-03
Implement the multi-memory proposal

alexcrichton:
This commit implements the current state of the [multi memory
proposal][repo] where the general idea is that all instructions which
touch memory now take an index of some form to indicate which memory is
being touched. Most of this was pretty simple, it's just plumbing around
the various places that indexes appear which need to be resolved,
validated, and printed.

[repo]: https://github.com/webassembly/multi-memory

### DIFF
--- a/crates/wast/src/ast/expr.rs
+++ b/crates/wast/src/ast/expr.rs
@@ -117,7 +117,9 @@ impl<'a> ExpressionParser<'a> {
                         // If block/loop show up then we just need to be sure to
                         // push an `end` instruction whenever the `)` token is
                         // seen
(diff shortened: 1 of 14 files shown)


crates/wast/src/ast/expr.rs
```diff
@@ -1109,18 +1120,62 @@ impl<'a> Parse<'a> for TableArg<'a> {
     }
 }
 
+/// Extra data associated with unary memory instructions.
+#[derive(Debug)]
+pub struct MemoryArg<'a> {
+    /// The index of the table argument.
```
fitzgen, line 1126:
```suggestion
    /// The index of the memory space.
```

PR #72, opened 2020-08-04
Initial implementation of memory64 proposal

alexcrichton:
This is an initial implementation of the memory64 proposal for our
tools. The main change here is validation where the index type is now
dynamic depending on the memory's type, but otherwise the text/binary
changes are pretty small. I filed a few issues for clarification along
the way and tried to "do something reasonable" in the meantime.

### DIFF
--- a/crates/wasmparser/src/binary_reader.rs
+++ b/crates/wasmparser/src/binary_reader.rs
@@ -22,7 +22,8 @@ use crate::limits::*;
 use crate::primitives::{
     BinaryReaderError, BrTable, CustomSectionKind, ExternalKind, FuncType, GlobalType, Ieee32,
     Ieee64, LinkingType, MemoryImmediate, MemoryType, NameType, Operator, RelocType,
-    ResizableLimits, Result, SIMDLaneIndex, SectionCode, TableType, Type, TypeOrFuncType, V128,
+    ResizableLimits, ResizableLimits64, Result, SIMDLaneIndex, SectionCode, TableType, Type,
+    TypeOrFuncType, V128,
 };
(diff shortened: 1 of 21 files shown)


crates/wasmparser/src/binary_reader.rs
```diff
@@ -564,6 +587,38 @@ impl<'a> BinaryReader<'a> {
         Ok(result)
     }
 
+    /// Advances the `BinaryReader` up to four bytes to parse a variable
+    /// length integer as a `u64`.
+    /// # Errors
+    /// If `BinaryReader` has less than one or up to four bytes remaining, or
```
fitzgen, line 593:
```suggestion
    /// length integer as a `u64`.
    ///
    /// # Errors
    ///
    /// If `BinaryReader` has less than one or up to four bytes remaining, or
```

crates/wasmparser/src/primitives.rs
```diff
@@ -195,16 +195,36 @@ pub struct ResizableLimits {
...
     pub element_type: Type,
     pub limits: ResizableLimits,
 }
 
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
-pub struct MemoryType {
-    pub limits: ResizableLimits,
-    pub shared: bool,
+pub enum MemoryType {
+    B32 {
```
fitzgen, line 212: What is "B" here? Not that it matters much, but I guess I would expect "Memory" or simply "M".

alexcrichton: I thought "bits", but I like your thinking of "Memory" more!

PR #77, opened 2020-08-17
Update GC proposal slightly

eqrion:
1. Add back the 'struct.new' instruction as a temporary compat measure for SM
2. Update implementation of heaptype syntax to match latest proposal


eqrion:
Looks like the testsuite needs to be updated to include the latest function-references changes.

PR #80, opened 2020-08-19
Fix br_if validation of dead code

### LINKED ISSUE #79: Valid Wasm module fails validation
[test.wasm.gz](https://github.com/bytecodealliance/wasm-tools/files/5096780/test.wasm.gz)

```wat
(module
  (type (;0;) (func))
  (type (;1;) (func))
  (import "" "" (global (;0;) (mut i32)))
  (func (;0;) (type 1)
    (local f64)
    loop  ;; label = @1
      loop  ;; label = @2
        unreachable
        global.get 0
        global.get 0
        br_if 0 (;@2;)
        global.set 0
        unreachable
      end
      unreachable
    end
    unreachable))
```

I need to double check the spec and how unreachable code is validated, but all of `wasm2wat`, `wasm-validate`, and `wasm-dis` accept this module as valid, while `wasmparser`'s validator returns an error.

yurydelendik:
Fixes #79

### DIFF
--- a/crates/wasmparser/src/operators_validator.rs
+++ b/crates/wasmparser/src/operators_validator.rs
@@ -846,11 +846,10 @@ impl OperatorValidator {
             Operator::BrIf { relative_depth } => {
                 self.check_operands_1(Type::I32)?;
                 self.check_jump_from_block(relative_depth, 1)?;
+                self.func_state.change_frame(1)?;
                 if self.func_state.last_block().is_stack_polymorphic() {
                     self.func_state
                         .change_frame_to_exact_types_from(relative_depth as usize)?;
-                } else {
-                    self.func_state.change_frame(1)?;
                 }
             }
             Operator::BrTable { ref table } => {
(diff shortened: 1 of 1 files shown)


fitzgen:
Looks like this one is failing too:

https://github.com/bytecodealliance/wasm-tools/pull/80/checks?check_run_id=1003443236#step:10:188

```
failed test: tests/wabt/test/roundtrip/generate-label-names.txt

Caused by:
    0: failed testing the binary output of `wat`
    1: wasm isn't valid
    2: type mismatch: stack size does not match block type (at offset 44)
```

crates/wasmparser/src/operators_validator.rs
```diff
@@ -229,9 +229,20 @@ impl FuncState {
...
         self.stack_types.truncate(keep);
         self.stack_types.extend_from_slice(&types);
-        last_block.polymorphic_values = None;
+        // Keep polymorphic stack.
+        let polymorphic_values = last_block.polymorphic_values.as_mut().unwrap();
+        *polymorphic_values = if *polymorphic_values < polymorphic_values_used {
+            0
+        } else {
+            *polymorphic_values - polymorphic_values_used
+        };
```
fitzgen, line 245:
```suggestion
        *polymorphic_values = polymorphic_values.saturating_sub(polymorphic_values_used);
```

yurydelendik:
> Ah nice so the i32 wasn't being popped for polymorphic stacks?

Correct.

The main issue was that br_if suppose to infer types from the target block (for dead code), though it truncated stack instead of just adding unknown types.

PR #81, opened 2020-08-19
WIP fix for #79

fitzgen:
This makes it so that the test case in #79 validates, and all the `wasmparser` tests pass, but introduces failures in this whole repo's round trip tests:

<details>

```
failed test: tests/testsuite/proposals/reference-types/unreached-invalid.wast

Caused by:
    2 test failures in tests/testsuite/proposals/reference-types/unreached-invalid.wast:
    
    	--------------------------------
    
    	failed directive on tests/testsuite/proposals/reference-types/unreached-invalid.wast:690:2
    	
    	Caused by:
    	    no wasm validation error found
    
    	--------------------------------
    
    	failed directive on tests/testsuite/proposals/reference-types/unreached-invalid.wast:701:2
    	
    	Caused by:
    	    no wasm validation error found
failed test: tests/testsuite/unreached-invalid.wast

Caused by:
    2 test failures in tests/testsuite/unreached-invalid.wast:
    
    	--------------------------------
    
    	failed directive on tests/testsuite/unreached-invalid.wast:689:2
    	
    	Caused by:
    	    no wasm validation error found
    
    	--------------------------------
    
    	failed directive on tests/testsuite/unreached-invalid.wast:700:2
    	
    	Caused by:
    	    no wasm validation error found
```

</details>

I'm not sure I totally understand how `wasmparser`'s stack polymorphism handling works, but the bug seems to orbit that code. It looks like `wasmparser` is trying to avoid memory overheads of representing the operand stack as `Vec<Option<Type>>` where `None` is a stack polymorphic value, and is instead keeping track on the side of how many stack polymorphic values we have at what index in the stack? I'm not 100% sure it is relevant to this bug, but what happens when there are two places in the operand stack where we have stack polymorphic values? I guess that once you start popping from a stack polymorphic section, you should never be able to get past it again to earlier values (other than exiting the control frame) right?

Anyways, since you're more familiar with `wasmparser`'s validator, I was hoping you could help investigate what's going on here, @yurydelendik. Thanks!

### DIFF
--- a/crates/wasmparser/examples/simple.rs
+++ b/crates/wasmparser/examples/simple.rs
@@ -27,8 +27,8 @@ fn main() -> Result<()> {
                     println!("  Import {}::{}", import.module, import.field.unwrap());
                 }
             }
-            _other => {
-                // println!("found payload {:?}", _other);
+            other => {
+                println!("{:?}", other);
             }
         }
     }
--- a/crates/wasmparser/examples/validate.rs
+++ b/crates/wasmparser/examples/validate.rs
@@ -1,0 +1,12 @@
+fn main() {
+    let mut validator = wasmparser::Validator::new();
+    validator.wasm_features(wasmparser::WasmFeatures {
+        multi_value: true,
+        ..wasmparser::WasmFeatures::default()
+    });
+
+    let bytes = std::fs::read(std::env::args().nth(1).unwrap()).unwrap();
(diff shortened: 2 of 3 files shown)


yurydelendik:
> wasmparser: add a validation example program

there is `cargo run --bin wasm-validate-rs ...`

The complete fix can be found at #80. Closing this one.

PR #82, opened 2020-08-19
Read extended operator code as LEB

### LINKED ISSUE #76: Update encodings for wasm extensions
See also https://github.com/WebAssembly/spec/issues/1228, and in particular https://github.com/WebAssembly/spec/issues/1228#issuecomment-669337584: prefix is a single byte, and what follows is a LEB128, while wasmparser currently accepts a single byte (except for SIMD which accepts a LEB128 already).

yurydelendik:
... for consistency

Fixes #76


yurydelendik:
> Would it be possible to add tests for this too?

The test will be included (?) with https://github.com/WebAssembly/spec/pull/1235/files#diff-59c366fbc32de0cb02b330bf7871317f -- I hope it will be uplifted into testsuite soon

PR #95, opened 2020-08-25
Expose some internals of wast::Error

### LINKED ISSUE #94: Expose internals of wast::Error
I am interested in using `wast` in one of my projects for reporting parse errors. I need to be able to extract useful information about the errors such as the file name, span, error message, etc. But it seems I can only really get access to the displayed string with `wast` because all of the internals of [wast::Error](https://docs.rs/wast/23.0.0/wast/struct.Error.html) are private.

Is there any particular reason for this? Can it be changed?

In theory I could reparse the displayed string to collect this info but that's messy and error prone and I'd rather avoid going that route.

darinmorrison:
This PR fixes #94. /cc @alexcrichton

### DIFF
--- a/crates/wast/src/lib.rs
+++ b/crates/wast/src/lib.rs
@@ -140,6 +140,11 @@ impl Error {
         }
     }
 
+    /// Return the `Span` for this error.
+    pub fn span(&self) -> Span {
+        self.inner.span
+    }
+
     /// To provide a more useful error this function can be used to extract
     /// relevant textual information about this error into the error itself.
     ///
@@ -170,6 +175,14 @@ impl Error {
         match &self.inner.kind {
             ErrorKind::Lex(e) => Some(e),
             _ => None,
+        }
+    }
+
+    /// Returns the underlying message, if any, that describes this error.
+    pub fn message(&self) -> String {
+        match &self.inner.kind {
+            ErrorKind::Lex(e) => e.to_string(),
+            ErrorKind::Custom(e) => e.clone(),
         }
     }
 }


crates/wast/src/lib.rs
```diff
@@ -140,6 +140,11 @@ impl Error {
         }
     }
 
+    /// Return the `Span` for this error.
+    pub fn span(&self) -> &Span {
+        &self.inner.span
```
darinmorrison, line 145: I chose to return a reference to `Span` here since I figured `Span` may not always be `Copy` in the future. But I can change that if you'd prefer.

crates/wast/src/lib.rs
```diff
@@ -172,6 +177,14 @@ impl Error {
             _ => None,
         }
     }
+
+    /// Returns the underlying custom `&str`, if any, that describes this error.
+    pub fn custom_error(&self) -> Option<&str> {
```
alexcrichton, line 182: Could this be renamed to perhaps `message`? I think that this should be implementable for lex errors as well. The term "custom" is mostly specific to the internals of this file and doesn't necessarily make a ton of sense to consumers.

crates/wast/src/lib.rs
alexcrichton: Ah yeah returning a bare `Span` without a reference here should be fine.

crates/wast/src/lib.rs
```diff
@@ -172,6 +177,14 @@ impl Error {
             _ => None,
         }
     }
+
+    /// Returns the underlying message, if any, that describes this error.
+    pub fn message(&self) -> Option<String> {
+        match &self.inner.kind {
+            ErrorKind::Lex(e) => Some(e.to_string()),
+            ErrorKind::Custom(e) => Some(e.clone()),
```
darinmorrison, line 185: We could return `Cow<str>` here to avoid the `.clone()` if you would prefer.

alexcrichton: I think lex errors should always have some sort of a static description, could that be exposed? Additionally now that the return value is always `Some` I think the return value here doesn't need `Option`.

darinmorrison: @alexcrichton Good point about the unnecessary `Option` return. I can remove that.

I don't quite understand what you mean about a static description for `LexError` though. I could factor part of the string description out of the `Display` impl for `LexError`, but since some of the `LexError` cases have arguments (which contribute to the description), I don't see how it can be fully static (since we need to use something like `format!`).

Can you elaborate on what you have in mind for that?

alexcrichton: Hm I was thinking of returning a static string yeah for all lex errors, but I actually think this makes more sense thinking about it. (you bring up a good point too!)

Seems fine by me in that case!

PR #98, opened 2020-09-02
Start reducing memory overhead of the parser

alexcrichton:
We've got a fuzz bug on Wasmtime right now where the wasm module
generated, when converted to text and then parsed, blows the memory
limit on the fuzz infrastructure. The memory usage and performance of
`wast` hasn't ever really been too too much of a concern, so this picks
off a low-hanging-fruit which is the size of the `Source` enum.

Before this commit `Source` was 136 bytes. The way the parser works
right now is that it lexes the entire source and creates a list of
tokens to parse, storing them into a `Vec`. This commit shrinks the size
of this element from 136 bytes to 24 (and flattens `Source` into one
`Token` enum). This drastically reduces the amount of memory allocated
to store the tokens of what we're parsing.

Unfortunately the parser is still hanging on to too much memory, so I'll
be adding a few follow-up PRs as well to handle further memory
improvements.

### DIFF
--- a/crates/wast/src/lexer.rs
+++ b/crates/wast/src/lexer.rs
@@ -48,31 +48,24 @@ pub struct Lexer<'a> {
 /// whitespace. For most cases you'll probably ignore these and simply look at
 /// tokens.
 #[derive(Debug, PartialEq)]
-pub enum Source<'a> {
-    /// A fragment of source that is a comment, either a line or a block
-    /// comment.
-    Comment(Comment<'a>),
+pub enum Token<'a> {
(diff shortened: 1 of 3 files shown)


alexcrichton:
The main trick here was to reduce the size of the integer, float, and string variants of `Token` by hiding them behind a `Box`. This is a pretty "low hanging fruit" option for now without involving too much invasive refactoring. There's lots of other possibilities (such as not hanging onto the entire list of tokens), but at this time that's pretty hard to do.

PR #101, opened 2020-09-06
Rewrite handling of blocks in the operator validator

### LINKED ISSUE #100: Unreachable in block breaks (?) some validation
This wasm file:

```wasm
(module
  (func
    unreachable
    i32.const 0
    select
    block (param i32)
      unreachable
    end))
```

validates with wabt but not with wasmparser:

```
$ cargo run -q --bin wat2wasm-rs test.wat -o test.wasm
$ wasm-validate test.wasm
$ cargo run -q --bin wasm-validate-rs test.wasm --enable-multi-value
Error: type mismatch: stack size does not match block type (at offset 35)
```

alexcrichton:
This commit rewrites how blocks are handled in the `OperatorValidator`
type, primarily motivated by the various bugs we've had recently around
the handling of `unreachable` instructions (#100 being the most recent).

The goal of this refactoring is to update the validator to match, almost
exactly, the [online appendix for the validation algorithm][appendix],
namely around how blocks and `unreachable` are handled. This also should
hopefully make the code a bit easier to read since the trickiness is
handled in a relatively subtle way where the "naive interpretation" ends
up being correct.

I've done some quick benchmarking and this doesn't appear to either
regress or improve performance (yay?), but this does...

Closes #100

[appendix]: https://webassembly.github.io/spec/core/appendix/algorithm.html

### DIFF
--- a/crates/wasmparser/src/operators_validator.rs
+++ b/crates/wasmparser/src/operators_validator.rs
@@ -13,277 +13,18 @@
  * limitations under the License.
  */
 
+// The basic validation algorithm here is copied from the "Validation
+// Algorithm" section of the WebAssembly specification -
+// https://webassembly.github.io/spec/core/appendix/algorithm.html.
+//
+// That algorithm is followed pretty closely here, namely `push_operand`,
+// `pop_operand`, `push_ctrl`, and `pop_ctrl`. If anything here is a bit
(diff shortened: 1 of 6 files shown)


yurydelendik:
Thank you for the patch.

PR #105, opened 2020-09-24
WIP: Update SIMD instructions

darinmorrison:
This PR updates SIMD support to reflect recent syntax changes (and new instructions) in the proposal repo.

This PR allows `wast` to successfully parse all of the test files in the upstream [WebAssembly/simd](https://github.com/WebAssembly/simd) repo (tested with the testsuite for the webassembly language server, which uses `wast`) but unfortunately it breaks a lot of the tests in `tests/local`, `tests/testsuite`, and `tests/wabt` which still use the old syntax.

I tried fixing some of the tests in `tests/local` but wasn't sure about some of the errors. I also tried regenerating the upstream [WebAssembly/testsuite](https://github.com/WebAssembly/testsuite) but the `update-testsuite.sh` script fails on several of the repos with unresolvable merge conflicts and I wasn't sure how to work around that without messing up the script workflow.

So for the meantime I will just leave this here.


alexcrichton:
Thanks for this! Would you be ok linking some of the PRs on the simd repo here for posterity too where the renames happened?

Additionally I think this is fine to land and not block on webassembly/testsuite and/or webassembly/wabt being updated, there's various locations throughout roundtrip.rs where test are ignored and you can add some TODO/FIXME items to get removed in the future.

darinmorrison:
Here are links to the respective PRs where the recent renaming and additions to the SIMD proposal were made:

https://github.com/WebAssembly/simd/pull/321
https://github.com/WebAssembly/simd/pull/322
https://github.com/WebAssembly/simd/pull/341
https://github.com/WebAssembly/simd/pull/344

darinmorrison:
I've updated the PR, temporarily pointing my local branches of `testsuite` and `wabt` until the upstream PRs are merged. All of the tests should pass now.

Note that I had to modify `Integer` parsing in https://github.com/bytecodealliance/wasm-tools/pull/105/commits/98c83b1a8ce6bddc62ab4cbeeb95699234932a36. This was needed in order to be able to fail (correctly) when parsing lane indices with sign prefixes. See [here](https://github.com/WebAssembly/simd/blob/9579d435305e72c7401655324a2e50b6dc260443/test/core/simd/simd_lane.wast#L875-L883).

alexcrichton:
Looks great to me, thanks again for all your help here on this!

To bikeshed a bit, though, the name "signededness" is a bit overloaded with integers themselves and the token in front, and I was a bit perplexed at `Signededness::Unsigned` meaning "no sign token" when I first saw it. That and I always have a hard time spelling and typing "signededness" (even now I'm not sure that it's right...)

Perhaps a name like `SignToken::{Plus, Minus, None}`? Or `Option<SignToken::{Plus, Minus}>`?

alexcrichton:
I've got a commit at https://github.com/alexcrichton/wasm-tools/commit/954df9d643a9d9dca3fdae91116ae318e2b2df78 which updates to the upstream testsuite module now with some other assorted fixes that cropped up, feel free to cherry-pick that here or take what you like from it! With the submodules set back to upstream repos I think this should be good to merge.

darinmorrison:
@alexcrichton I think this should be good to go now.

alexcrichton:
Perfect, thanks again!

PR #110, opened 2020-10-02
Bump wast version to 25.0.0

darinmorrison:
This PR bumps the `wast` version to `25.0.0`. Hoping a new release can be made so that I can update the dependency for the language server (for recent SIMD changes and other testsuite fixes).


alexcrichton:
Thanks!

PR #112, opened 2020-10-07
Correct opcode for i32x4.dot_i8x16_s

lars-t-hansen:
The opcode in the spec PR is 0xfd 0xba, and this is also what LLVM, V8, and SpiderMonkey use.

### DIFF
--- a/crates/wast/src/ast/expr.rs
+++ b/crates/wast/src/ast/expr.rs
@@ -940,12 +940,12 @@ instructions! {
         I32x4ShrU : [0xfd, 0xad] : "i32x4.shr_u",
         I32x4Add : [0xfd, 0xae] : "i32x4.add",
         I32x4Sub : [0xfd, 0xb1] : "i32x4.sub",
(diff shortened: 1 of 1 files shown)


alexcrichton:
Oops, thanks! Want a publish as well?

lars-t-hansen:
> Oops, thanks! Want a publish as well?

That would be most helpful.  Thanks!

alexcrichton:
Sure thing, done now!

lars-t-hansen:
@alexcrichton, could you also bump wat to 1.0.27 and make it require wast 25.0.1?  The reason I ask is because we pull in wat, not wast directly.

alexcrichton:
Oh that shouldn't be necessary since wat 1.0.26 requires wast 25.0.0 which is satisfied with 25.0.1, so you should be able to do `cargo update -p wast` and it'll pull in the newer version of the `wast` crate.

lars-t-hansen:
TIL.  Thanks!

PR #117, opened 2020-10-16
Expose the height of the value stack from the validator

fitzgen:


alexcrichton:
Out of curiosity, what's the use case for this? Should this also return the height of the stack of the whole function or just the current block? Either case seems fine, there's all sorts of info we could expose we don't just yet anyway.

R=me with a test

fitzgen:
> Out of curiosity, what's the use case for this?

I was using this in my template JIT to compute offsets for locals within the current frame.

> Should this also return the height of the stack of the whole function or just the current block?

The function's whole stack.

PR #119, opened 2020-10-19
Add v128.load32_zero and v128.load64_zero SIMD instructions

lars-t-hansen:
These opcodes were promoted to standards track last Friday: https://github.com/WebAssembly/simd/pull/237#issuecomment-710675437


lars-t-hansen:
Oh, and a dot release would be great once these are approved & incorporated.

alexcrichton:
Published!

I also added a [small fixup](https://github.com/bytecodealliance/wasm-tools/commit/4e2589ba86c71358af46d541bf6a11210227887b) since the text parser didn't seem to encode the memory immediate

lars-t-hansen:
Thanks!  I apparently don't know my way around this code very well yet...

lars-t-hansen:
@alexcrichton, actually looking at that fixup: it looks like the argument to MemArg used there is the log of the byte size of the datum, but all other uses of MemArg in that file carry the byte size of the datum.  Care to comment?

alexcrichton:
Oh dear looks like I've also forgotten my way around the code! I'll fix that and publish a release

alexcrichton:
Opened as https://github.com/bytecodealliance/wasm-tools/pull/120

PR #122, opened 2020-10-21
Merge `wasm-smith` into the `wasm-tools` repo

fitzgen:
It makes sense for `wasm-smith` to live here, since it is a Wasm tool :-p

More seriously, for the same reason it is nice to update all the rest of the tools in lockstep when adding new Wasm features and doing releases together, it makes sense to do this with `wasm-smith` as well. For example, see the annoying developer overhead caused by reporting bugs in `wasmparser`'s validator that were fixed in `main` but we just didn't have the versions updated together yet.

This PR was made by

* creating a new, orphan branch for `wasm-smith`
* pulling `fitzgen/wasm-smith#main`
* moving `wasm-smith` in that branch into `crates/wasm-smith`
* and then merging `bytecodealliance/wasm-tools#main`

This way we should retain `wasm-smith`'s full history.


crates/wasm-smith/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+[dependencies]
+arbitrary = { version = "0.4.6", features = ["derive"] }
+leb128 = "0.2.4"
+structopt = { version = "0.3.16", optional = true }
+
+[dev-dependencies]
+criterion = "0.3.3"
+libfuzzer-sys = "0.3.4"
+rand = { version = "0.7.3", features = ["small_rng"] }
+wasmparser = { version = "0.64.0", path = "../wasmparser" }
```
alexcrichton, line 31: You can drop the `version` here to make version bumps easier

alexcrichton (approved):

PR #124, opened 2020-10-23
wasm-smith: Implement reference types support

alexcrichton:
This commit implements the reference types proposal for the wasm-smith
crate, generating modules which use reference types. It covers the new
instructions, multiple-tables, and new kinds of element segments.
Support is gated through `Config::max_tables` and
`Config::reference_types_enabled`.

### DIFF
--- a/crates/wasm-smith/src/code_builder.rs
+++ b/crates/wasm-smith/src/code_builder.rs
@@ -1,6 +1,8 @@
-use super::{BlockType, Config, ConfiguredModule, FuncType, Import, Instruction, MemArg, ValType};
+use super::{
+    BlockType, Config, ConfiguredModule, Elements, FuncType, Import, Instruction, MemArg, ValType,
+};
 use arbitrary::{Result, Unstructured};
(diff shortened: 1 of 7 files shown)


alexcrichton:
Note that this isn't quite ready for merging yet. Running the fuzzers locally it discovered that this module doesn't round-trip:

```wasm
(module (elem declare funcref))
```

This is due to [this block](https://github.com/bytecodealliance/wasm-tools/blob/e8fc421a9f666130d03ec7fd15773803c2e7ef1c/crates/wast/src/binary.rs#L652-L667) which currently serves as compatibility with wabt. I'll file an issue with wabt tomorrow to see if we can sort that out on their side and be able to round trip the above module.

alexcrichton:
Er actually that's easy enough to back out, I've reverted the changes to `roundtrip-valid-module` using `SwarmConfig`, now it's back to only testing MVP modules which means this should be good to go.

fitzgen (approved):
Very nice!

PR #130, opened 2020-10-28
wast: Add a cache for copying types from other modules

### LINKED ISSUE #129: Module linking text-to-binary excessively slow with ~400 modules
This was found with a fuzzer so it's not quite reminiscient of what might be reasonable, but [this module](https://gist.github.com/alexcrichton/fe9e9b6baef0f9e6387244b9bae0926f) takes multiple seconds (~6 locally) to convert to a binary. Conversion from binary to text takes ~40ms, so this is a huge discrepancy which is a bug in the resolver.

Some profiling shows that most of the time is going to "expanding" the module which is elaborating types and such. I didn't exactly write it to be fast, but it looks like it's way slower than I thought it was.

alexcrichton:
This should fix an exponential blowup in runtime for highly-nested
module-linking modules.

Closes #129

### DIFF
--- a/crates/wast/src/resolve/names.rs
+++ b/crates/wast/src/resolve/names.rs
@@ -64,6 +64,13 @@ pub struct Module<'a> {
     /// currently-being-processed field. This should always be empty after
     /// processing is complete.
     to_prepend: Vec<ModuleField<'a>>,
+
+    /// Cache for copying over types from other modules, used for module-linking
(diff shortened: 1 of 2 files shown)


fitzgen (approved):
friendship with "accidentally quadratic" ended :x: my new best friend is "accidentally exponential" :muscle:

PR #141, opened 2020-11-17
Exception handling support

yurydelendik:
Implementation of https://github.com/WebAssembly/exception-handling proposal

### DIFF
--- a/crates/wasmparser/src/operators_validator.rs
+++ b/crates/wasmparser/src/operators_validator.rs
@@ -110,6 +110,8 @@ enum FrameKind {
     If,
     Else,
     Loop,
+    Try,
+    Catch,
 }
 
 impl OperatorValidator {
@@ -276,6 +278,18 @@ impl OperatorValidator {
         Ok(())
     }
 
+    fn push_catch(&mut self, ty: TypeOrFuncType) -> OperatorValidatorResult<()> {
+        // Push a new frame which has a snapshot of the height of the current
+        // operand stack.
+        self.control.push(Frame {
+            kind: FrameKind::Catch,
+            block_type: ty,
+            height: self.operands.len(),
+            unreachable: false,
+        });
+        self.push_operand(Type::ExnRef)
+    }
+
     /// Pops a frame from the control stack.
     ///
     /// This function is used when exiting a block and leaves a block scope.
@@ -390,6 +404,15 @@ impl OperatorValidator {
     fn check_simd_enabled(&self) -> OperatorValidatorResult<()> {
         if !self.features.simd {
             return Err(OperatorValidatorError::new("SIMD support is not enabled"));
+        }
+        Ok(())
+    }
+
+    fn check_exceptions_enabled(&self) -> OperatorValidatorResult<()> {
+        if !self.features.exceptions {
+            return Err(OperatorValidatorError::new(
+                "Exceptions support is not enabled",
+            ));
         }
         Ok(())
     }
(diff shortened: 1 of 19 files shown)


alexcrichton:
Looks pretty good to me! I think it'd be good to run the upstream spec tests as well which I believe can be executed by uncommenting https://github.com/bytecodealliance/wasm-tools/blob/05a2bb8436c72319a05c905ad8f7d5f9f22887fb/tests/roundtrip.rs#L143-L147

yurydelendik:
The one thing I cannot figure out which name is to use in regards of exception section or type: `exception` or `event`?

crates/wasmparser/src/operators_validator.rs
```diff
@@ -561,17 +584,71 @@ impl OperatorValidator {
                 }
                 self.push_ctrl(FrameKind::Else, frame.block_type, resources)?;
             }
+            Operator::Try { ty } => {
+                self.check_exceptions_enabled()?;
+                self.check_block_type(ty, resources)?;
+                for ty in params(ty, resources)?.rev() {
+                    self.pop_operand(Some(ty))?;
+                }
+                self.push_ctrl(FrameKind::Try, ty, resources)?;
```
alexcrichton, line 593: More a note for later, but this, `Loop, `If`, and `Block` are all pretty similar so we may want to look into trying to unify those at some point. Not really need now, just a possibility!

crates/wasmparser/src/operators_validator.rs
```diff
@@ -276,6 +278,18 @@ impl OperatorValidator {
         Ok(())
     }
 
+    fn push_catch(&mut self, ty: TypeOrFuncType) -> OperatorValidatorResult<()> {
```
alexcrichton, line 281: Since this is only called once could this be inlined into where it's called?

crates/wasmparser/src/validator.rs
```diff
@@ -703,6 +717,17 @@ impl Validator {
         Ok(())
     }
 
+    fn event_type(&self, ty: &EventType) -> Result<()> {
+        if ty.attribute != 0 {
```
alexcrichton, line 721: This is validated currently at both the binary layer and this layer, but should it only be validated at the binary layer?

crates/wasmprinter/src/lib.rs
```diff
@@ -694,6 +724,29 @@ impl Printer {
...
             }
             Else => self.result.push_str("else"),
+            Try { ty } => {
+                self.result.push_str("try");
+                self.print_blockty(ty)?;
+                write!(self.result, "  ;; label = @{}", cur_label)?;
+            }
+            Catch => self.result.push_str("catch"),
+            Throw { index } => {
+                write!(self.result, "throw {}", index,)?;
```
alexcrichton, line 734: trailing comma

alexcrichton (commented):
I don't have much of an idea on naming, I'd just follow whatever the upstream proposal is doing which I believe is events?

Can you confirm the spec tests are running though? (or that there's any spec tests for exceptions?) In `wasmparser_validator_for` I don't think `exceptions` is ever set to `true` for the `testsuite` directory, so for running the `*.wast` tests in `tests/testsuite` I think exceptions are always disabled? (or something about that function is buggy in how it's recognizing the test suite)

crates/wasmparser/src/validator.rs
yurydelendik: attribute is specified as 0 for this proposal -- I'll remove the `attributes` field for now

alexcrichton (approved):

alexcrichton:
Mind pushing up the tags too when publishing?

PR #144, opened 2020-11-19
Introduce the `wasm-encoder` crate

fitzgen:
A Wasm encoder crate that is at roughly the same level of abstraction as `wasmparser`, i.e it just lets you write raw Wasm and doesn't try and provide a high-level interface that lets you forget about index spaces or anything like that.

Doesn't try to do fancy things like use the type system to make it impossible to generate invalid Wasm modules; focuses on simplicity instead.

Allows you to encode arbitrary data as a particular section, so that if you are making little tweaks to a Wasm module, you don't need to fully parse and fully encode sections that you aren't modifying.

Light on tests for now. I expect that most testing for this crate will be done via porting other crates in this repo over to using this crate for encoding Wasm.


crates/wasm-encoder/src/code.rs
```diff
@@ -0,0 +1,831 @@
...
+}
+
+impl Section for CodeSection {
+    fn id(&self) -> u8 {
+        SectionId::Code as u8
+    }
+
+    fn encode<S>(&self, sink: &mut S)
+    where
+        S: Extend<u8>,
```
alexcrichton, line 63: Out of curiosity does this optimize well? I'd imagine that while this is super general it'll rarely get used with much other than `Vec`

crates/wasm-encoder/src/code.rs
```diff
@@ -0,0 +1,831 @@
...
+    /// The `n`th function type.
+    FunctionType(u32),
+}
+
+impl BlockType {
+    fn encode(&self, bytes: &mut Vec<u8>) {
+        match *self {
+            BlockType::Empty => bytes.push(0x40),
+            BlockType::Result(ty) => bytes.push(ty as u8),
+            BlockType::FunctionType(f) => bytes.extend(encoders::s33(f as i64)),
```
alexcrichton, line 174: In an effort to remove `as` casts, could this be `i64::from` or `.into()`?

crates/wasm-encoder/src/code.rs
```diff
@@ -0,0 +1,831 @@
...
+    RefIsNull,
+    RefFunc(u32),
+    TableInit { segment: u32, table: u32 },
+    ElemDrop { segment: u32 },
+    TableFill { table: u32 },
+    TableSet { table: u32 },
+    TableGet { table: u32 },
+    TableGrow { table: u32 },
+    TableSize { table: u32 },
+    TableCopy { src: u32, dst: u32 },
```
alexcrichton, line 393: You know at some point we should figure out how to not have a list of instructions in like 30 different places in this repository...

crates/wasm-encoder/src/encoders.rs
```diff
@@ -0,0 +1,123 @@
...
+    });
+    let mut buf = [0; 5];
+    let n = leb128::write::signed(&mut &mut buf[..], x).unwrap();
+    Buf5Iter { buf, n, i: 0 }
+}
+
+/// Encode an `i64` as a SLEB128.
+pub fn s64(x: i64) -> impl ExactSizeIterator<Item = u8> {
+    let mut buf = [0; 9];
+    let n = leb128::write::signed(&mut &mut buf[..], x as i64).unwrap();
```
alexcrichton, line 56: I think this cast can be dropped?

crates/wasm-encoder/src/encoders.rs
```diff
@@ -0,0 +1,123 @@
...
+pub fn u32(n: u32) -> impl ExactSizeIterator<Item = u8> {
+    let mut buf = [0; 5];
+    let n = leb128::write::unsigned(&mut &mut buf[..], n as u64).unwrap();
+    Buf5Iter { buf, n, i: 0 }
+}
+
+/// Encode an `i32` as a SLEB128.
+pub fn s32(x: i32) -> impl ExactSizeIterator<Item = u8> {
+    let mut buf = [0; 5];
+    let n = leb128::write::signed(&mut &mut buf[..], x as i64).unwrap();
```
alexcrichton, line 18: I think this and the above cast can be `.into()`?

crates/wasm-encoder/src/encoders.rs
```diff
@@ -0,0 +1,123 @@
...
+}
+
+impl ExactSizeIterator for Buf5Iter {}
+
+// `[u8; 9]` doesn't have `into_iter()` so we can't do
+// `[..].into_iter().take(n)` :(
+struct Buf9Iter {
+    buf: [u8; 9],
+    n: usize,
+    i: usize,
```
alexcrichton, line 100: FWIW I've generally found a more useful way to store this is to have a `Range<usize>` and that way you can delegate to the range for `next_back`, `len`, `size_hint`, etc. The `next` method in that case is also just `Some(self.buf[self.range.next()?])`

alexcrichton (commented):
Nice!

I'd love to see this hooked up to the `wast` crate as well, but do you think that should hold off?

In some hypothetical ideal world we'd have one shared `wasm-types` crate, but I don't think that everything is similar enough for us to do that just yet.

crates/wasm-encoder/src/code.rs
fitzgen: I just did it because it gives less capabilities to implementations. I originally had a `&mut Vec<u8>` and a comment like "only push data, don't pop anything" and then I realized I could just make it an `Extend<u8>` to enforce that at the type level.

crates/wasm-encoder/src/encoders.rs
fitzgen: oo nice.

fitzgen:
> I'd love to see this hooked up to the `wast` crate as well, but do you think that should hold off?

I just didn't want to spend the time. I don't think there is any reason not to start on that work, but I also don't want to block landing this PR on it. I personally would probably start on `wasm-smith` before `wast` simply because I'm more familiar with it.

> In some hypothetical ideal world we'd have one shared `wasm-types` crate, but I don't think that everything is similar enough for us to do that just yet.

Yeah part of the problem is that this crate doesn't need ownership but other crates (like wasm-smith) do, so we would have to make everything `Cow`-y or something.

fitzgen:
@alexcrichton do you have any objection to landing this now (I just pushed a commit to clean up the iterators and to remove the `as` casts) and then port the other crates in follow up PRs as we find time?

alexcrichton:
I'm a bit hesitant to accumulate too many work-in-progress things in-tree, so could wasm-smith get hooked up before landing?

alexcrichton (approved):
👍

PR #151, opened 2020-11-30
wasm-smith: add some additional configuration options.

cfallin:
This PR adds some options to more tightly control the arbitrary Wasm
modules generated by `wasm-smith`. In particular, it adds minimums
alongside most of the maximums, so that we can require certain elements
to be present (e.g., a memory or at least one function); it allows the
user to omit start functions; and it allows for a limit to be placed on
maximum Wasm memory sizes, which is useful during fuzzing (to reduce
memory overhead).

### DIFF
--- a/crates/wasm-smith/src/config.rs
+++ b/crates/wasm-smith/src/config.rs
@@ -16,9 +16,31 @@ use arbitrary::{Arbitrary, Result, Unstructured};
 /// need to override the methods for things you want to change away from the
 /// default.
 pub trait Config: Arbitrary + Default {
+    /// The minimum number of types to generate. Defaults to 0.
+    fn min_types(&self) -> usize {
+        0
+    }
+
     /// The maximum number of types to generate. Defaults to 100.
     fn max_types(&self) -> usize {
         100
+    }
+
+    /// The minimum number of imports to generate. Defaults to 0.
+    ///
+    /// Note that if the sum of the maximum function[^1], table, global and
+    /// memory counts is less than the minimum number of imports, then it will
+    /// not be possible to satisfy all constraints (because imports count
+    /// against the limits for those element kinds). In that case, we strictly
+    /// follow the max-constraints, and can fail to satisfy this minimum number.
+    ///
+    /// [^1]: the maximum number of functions is also limited by the number of
+    ///       function types arbitrarily chosen; strictly speaking, then, the
+    ///       maximum number of imports that can be created due to
+    ///       max-constraints is `sum(min(num_func_types, max_funcs), max_tables,
+    ///       max_globals, max_memories)`.
+    fn min_imports(&self) -> usize {
+        0
     }
 
     /// The maximum number of imports to generate. Defaults to 100.
@@ -26,14 +48,33 @@ pub trait Config: Arbitrary + Default {
         100
     }
 
-    /// The maximum number of functions to generate. Defaults to 100.
+    /// The minimum number of functions to generate. Defaults to 0.  This
+    /// includes imported functions.
+    fn min_funcs(&self) -> usize {
+        0
+    }
+
+    /// The maximum number of functions to generate. Defaults to 100.  This
+    /// includes imported functions.
     fn max_funcs(&self) -> usize {
         100
     }
 
-    /// The maximum number of globals to generate. Defaults to 100.
+    /// The minimum number of globals to generate. Defaults to 0.  This includes
+    /// imported globals.
+    fn min_globals(&self) -> usize {
+        0
+    }
+
+    /// The maximum number of globals to generate. Defaults to 100.  This
+    /// includes imported globals.
     fn max_globals(&self) -> usize {
         100
+    }
+
+    /// The minimum number of exports to generate. Defaults to 0.
+    fn min_exports(&self) -> usize {
+        0
     }
 
     /// The maximum number of exports to generate. Defaults to 100.
@@ -41,15 +82,31 @@ pub trait Config: Arbitrary + Default {
         100
     }
 
+    /// The minimum number of element segments to generate. Defaults to 0.
+    fn min_element_segments(&self) -> usize {
+        0
+    }
+
     /// The maximum number of element segments to generate. Defaults to 100.
     fn max_element_segments(&self) -> usize {
         100
+    }
+
+    /// The minimum number of elements within a segment to generate. Defaults to
+    /// 0.
+    fn min_elements(&self) -> usize {
+        0
     }
 
     /// The maximum number of elements within a segment to generate. Defaults to
     /// 100.
     fn max_elements(&self) -> usize {
         100
+    }
+
+    /// The minimum number of data segments to generate. Defaults to 0.
+    fn min_data_segments(&self) -> usize {
+        0
     }
 
     /// The maximum number of data segments to generate. Defaults to 100.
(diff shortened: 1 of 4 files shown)


crates/wasm-smith/src/lib.rs
```diff
@@ -1021,54 +1068,61 @@ where
...
 }
 
 pub(crate) fn arbitrary_loop(
     u: &mut Unstructured,
+    min: usize,
     max: usize,
     mut f: impl FnMut(&mut Unstructured) -> Result<()>,
 ) -> Result<()> {
-    for _ in 0..max {
+    for _ in min..max {
```
alexcrichton, line 1125: This may not work as intended? I think perhaps for `0..min` this would unconditionally call `f` and then from `0..(max - min)` would conditionally call it?

(or `keep_going = i < min || ...`)

crates/wasm-smith/src/lib.rs
```diff
@@ -584,29 +577,34 @@ where
```
fitzgen, line 577: This is an existing problem that I just noticed, but these choices should probably not be loop invariant anymore, since we want to respect the config's max functions and globals.

crates/wasm-smith/src/lib.rs
```diff
@@ -584,29 +577,34 @@ where
...
-            }
-            if self.table_imports() < self.config.max_tables() {
-                choices.push(|u, m| Ok(Import::Table(m.arbitrary_table_type(u)?)));
-            }
+        arbitrary_loop(
+            u,
+            self.config.min_imports(),
+            self.config.max_imports(),
+            |u| {
+                choices.truncate(num_stable_choices);
```
fitzgen, line 585: What if `config.min_imports() >= 1` but all of `config.max_{functions,memories,globals,tables}() == 0`? That is, with the introduction of minimums, we can now have contradicting constraints.

I think that is fine so long as we assert that `min <= max` in `arbitrary_loop` and document these requirements for the min/max methods in the `Config` trait (and also document the special case of imports where `min_imports <= sum(max_{functions,memories,globals,tables})`).

crates/wasm-smith/src/lib.rs
```diff
@@ -935,39 +967,54 @@ where
...
-                        if ty == ValType::ExternRef || func_max == 0 || u.arbitrary()? {
-                            None
-                        } else {
-                            Some(u.int_in_range(0..=func_max - 1)?)
+        arbitrary_loop(
+            u,
+            self.config.min_element_segments(),
+            self.config.max_element_segments(),
+            |u| {
+                // Choose an a
```
fitzgen, line 975: While you're here:
```suggestion
                // Choose whether to generate a segment whose elements are
                // initialized via expressions, or one whose elements are initialized
                // via function indices.
```

crates/wasm-smith/src/lib.rs
fitzgen, line 1125: Let's also `assert!(min <= max)` here.

fitzgen: Ah yes, also what @alexcrichton said :)

crates/wasm-smith/src/config.rs
```diff
@@ -82,6 +133,18 @@ pub trait Config: Arbitrary + Default {
         1
     }
 
+    /// The maximum, in 64k Wasm pages, of any memory's initial or maximum size.
+    /// Defaults to 2^16 = 65536 (the maximum possible for 32-bit Wasm).
+    fn max_memory_pages(&self) -> u32 {
```
fitzgen, line 138: Can you hook this up to `SwarmConfig` too?

fitzgen:
Also can you leave a comment somewhere on `SwarmConfig` that we intentionally do not set arbitrary minimums for the swarm to avoid picking conflicting constraints? Thanks!

crates/wasm-smith/src/lib.rs
cfallin: Interesting -- this is assuming that the imported functions and globals count against the limits, yes? My assumption had been that the limits should denote the number of functions or globals defined in the module itself; this simplifies the import generation and also removes the particular conflict between the minimum import count and the maximum allowed element (func/global/mem/table) counts below.

crates/wasm-smith/src/lib.rs
cfallin: So I think the simplest resolution here is (as noted above) just count imports as a separate thing with a separate limit; i.e. an imported function does not count against `max_funcs`. If there is some requirement by the user that the total function index space is limited to some range, then that's a different thing, but it seems to me like "locally defined function bodies" is a thing whose range we want to control separately in most cases. I'm not too tied to this, though.

crates/wasm-smith/src/config.rs
cfallin: Done!

crates/wasm-smith/src/lib.rs
cfallin: Done.

crates/wasm-smith/src/lib.rs
cfallin: Done.

cfallin: D'oh; this is what I get for leaving my brain at home today. Thanks!

cfallin:
Thanks! Updated.

crates/wasm-smith/src/lib.rs
fitzgen: This would make things easier however it doesn’t allow us to do things like set max memories to one to avoid the multi-memory proposal

cfallin: Hmm, that's a good point. It appears that the existing implementation wasn't consistent with this (`arbitrary_imports()` is invoked to create imports before `arbitrary_{funcs,tables,memories,globals}()`, and the latter do not check how many imports of each kind have been created); so I'll fix that as well, if these are the semantics that we want.

cfallin: (Actually, it seems only the function case was not checking the number of imports; I was confused by that case. Will fix nevertheless.)

crates/wasm-smith/src/lib.rs
```diff
@@ -1004,77 +1067,90 @@ where
...
     u: &mut Unstructured,
+    min: usize,
     max: usize,
-    mut f: impl FnMut(&mut Unstructured) -> Result<()>,
+    mut f: impl FnMut(&mut Unstructured) -> Result<bool>,
 ) -> Result<()> {
-    for _ in 0..max {
+    assert!(max >= min);
+    for _ in 0..min {
+        f(u)?;
```
fitzgen, line 1143: Should this also be checking the boolean result? I could go either way here.

fitzgen (approved):
r=me with or without the comment below

fitzgen:
Thanks @cfallin!

crates/wasm-smith/src/lib.rs
cfallin: Hmm, yeah, I suppose that would be more consistent -- done.

cfallin:
CI failed with "internal error", kicking with amended commit...

cfallin:
Green now, @fitzgen would you mind merging if all's well (I'm not a maintainer on this repo)?

fitzgen:
> (I'm not a maintainer on this repo)?

I don't seem to have the authorization to add you to the github team for this repo. +cc @tschneidereit

alexcrichton:
I added @cfallin  to the team that maintains tooling repos

PR #152, opened 2020-12-01
Start to add support for new version of exception handling proposal

takikawa:
Recently the Wasm CG agreed on changing the exception proposal to a different design (due to wanting to accommodate 2-phase exception handling in the future) that doesn't use `exnref` and has tag arguments for `catch` instructions. There are some slides explaining this in [this spec issue](https://github.com/WebAssembly/exception-handling/issues/125) and a new draft [spec explainer](https://github.com/WebAssembly/exception-handling/pull/137).

This pull request starts to add support for this in parsing and in validation. Specifically, it adds `catch_all` and `unwind`, removes `br_on_exn`, and revises `catch` (new tag argument) and `rethrow` (new index argument) to match the new semantics.

I'm submitting this PR as I'm also working on adding [exceptions support](https://bugzilla.mozilla.org/show_bug.cgi?id=1335652) in SpiderMonkey and updated parsing support from the `wast` crate will be helpful for making the SM tests work.

This doesn't yet add the new proposed `delegate` instruction (because no opcode has been proposed yet), and also doesn't remove `exnref` yet (can be done later in a separate patch).

Since other tools like wabt haven't been updated for the new semantics yet, this PR also disables some of those tests (but adds local ones).

Also the new spec isn't final too, so maybe some instructions will keep changing. Happy to follow up with further PRs to keep current.


yurydelendik:
Is it possible to update tests (instead of disabling them)? Do you know what is a timeline to change LLVM and other tooling to match a new spec?

alexcrichton:
I think there's some CI failures as well, but otherwise looks great, thanks!

takikawa:
I believe I've fixed the CI issues with a push earlier today, but it appears that the Github CI infra is [having some trouble](https://www.githubstatus.com/incidents/sxthxvdlljk4) right now so it might take a bit for this to show up.

@yurydelendik Since the tests are in wabt, I think to update them I'd have to go update the wabt implementation as well (which would be nice to do but wasn't on my critical path at the moment). Re: other tooling, I'm not sure. So far I don't think wabt, binaryen, or LLVM have switched over to the new proposal yet.

takikawa:
I've rebased and pushed again so it merges cleanly on top of #154, and CI should be good too.

crates/wasmparser/src/operators_validator.rs
```diff
@@ -578,10 +580,23 @@ impl OperatorValidator {
             }
             Operator::Else => {
                 let frame = self.pop_ctrl(resources)?;
-                if frame.kind != FrameKind::If {
-                    bail_op_err!("else found outside of an `if` block");
+                // The `catch_all` instruction shares an opcode with `else`,
+                // so we check the frame to see how it's interpreted.
+                if self.features.exceptions
+                    && (frame.kind == FrameKind::Try || frame.kind == FrameKind::Catch)
```
alexcrichton, line 586: Perhaps this could be:

```rust
match frame.kind {
    FrameKind::If => ...,
    FrameKind::Try | FrameKind::Catch => ..., // can assume `self.features.exceptions` is true
    _ => bail_op_err!(...),
}
```

To make it a bit more tidy?

I think we can skip the recheck of `self.features.exceptions` because that should be checked pushing the frame on the initial instruction.

alexcrichton (commented):
Looks great to me, thanks! Just one small stylistic nit

alexcrichton (approved):
@yurydelendik do you want to review as well?

crates/wasmparser/src/operators_validator.rs
takikawa: Thanks for the suggestion!

takikawa:
Hi just wanted to give a friendly ping to see if there's anything else I should address on this.

BTW, should I also add a commit to bump crate versions as this adds new AST types?

yurydelendik (approved):
Looks good on wasmparser side too. Thank you for the patch.

alexcrichton:
Oops sorry didn't see @yurydelendik's approval from earlier!

Issue #153, opened 2020-12-01
Issue in looking up event indices in `throw` instruction validation


takikawa:
I think there is a small bug in validating `throw` in cases like the following:

```
(assert_invalid
  (module
    (type (func))     ;; func type at index 0
    (func throw 0))   ;; event at index 0, which is missing
  "throw index out of bounds")
```

I believe this should be invalid, because the `throw` argument index should index into the events vector of the module, but it's looking up in the function types vector instead. Currently it doesn't throw an error.

(I can submit a PR to fix this, though maybe I should wait to see if there are thoughts on PR #152 first as it affects `catch` in that PR too)

yurydelendik:
> I believe this should be invalid, because the throw argument index should index into the events vector of the module

Yes, we need to fail here.

alexcrichton:
Fixed in https://github.com/bytecodealliance/wasm-tools/pull/154

PR #154, opened 2020-12-02
Fix event lookup for validating throw/br_on_exn

takikawa:
This PR addresses issue #153, about event index lookup for `throw` and `br_on_exn`.

I was originally thinking of pushing this after PR #152, but this PR is a simpler patch so maybe better if this is merged before and then I can fix up #152 to merge cleanly with this change.


alexcrichton:
👍 Thanks!

PR #156, opened 2020-12-02
Check memory index in attomics operations

yurydelendik:
Removes hardcoded `0` for memory index in the `check_shared_memarg_wo_align`


yurydelendik:
old version of wabt is used?

alexcrichton (approved):
r=me with tests passing

PR #159, opened 2020-12-02
wasmparser version 0.69.0

yurydelendik:
API did not change much: just validation logic and linking stuff.

### DIFF
--- a/crates/wast/Cargo.toml
+++ b/crates/wast/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wast"
-version = "27.0.0"
+version = "28.0.0"
 authors = ["Alex Crichton <alex@alexcrichton.com>"]
 edition = "2018"
 license = "Apache-2.0 WITH LLVM-exception"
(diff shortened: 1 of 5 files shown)


alexcrichton:
Hm unfortunately though the `WasmModuleResources` trait changed, which is a public API, so I think this needs a major bump. Also can `wat`/`wast` get bumped to account for the module linking encoding update?

crates/wat/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "wat"
-version = "1.0.28"
+version = "1.1.0"
```
alexcrichton, line 3: Oh FWIW this only changes in minor ways so I've been keeping it at 1.0.x instead of 1.1.0

crates/wast/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "wast"
-version = "27.0.0"
+version = "27.1.0"
```
alexcrichton, line 3: Ah unfortunately the changes in this crate are almost always public-facing, so this'll need to go to 28.0.0

alexcrichton:
Sorry at some point we should figure out a better solution for version management in this repo, it's a lot to track right now unfortunately...

PR #160, opened 2020-12-02
Fix aliasing parent items before they're defined 

alexcrichton:
This commit fixes an issue in validating module linking nested modules
using `parent` alias annotations. The specification indicates that only
the items defined before a module are available for aliasing, but we
were accidentally validating with the entire parent's context available.
This adds a snapshot of the length of the type/module vectors whenever a
module is defined, and then we use that snapshot when validating parent
alias annotations to ensure they're in-bounds.

This also updates the encoding of `*.wat` modules where there's a hack
to sort all types first in a module, but we avoid doing that if we don't
actually need to elaborate any module types. This will preserve our
round-trip-ability and also makes it possible to write a textual test
for this issue.

### DIFF
--- a/crates/wasmparser/src/validator.rs
+++ b/crates/wasmparser/src/validator.rs
@@ -111,12 +111,12 @@ pub struct Validator {
     /// the last part of the module index space. Consequently we have a second
     /// map `module_code_section_definitions` here which is the same length as
     /// `expected_modules`. This is a map from position in the module code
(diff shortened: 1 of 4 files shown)


alexcrichton:
Note that the first commit comes from https://github.com/bytecodealliance/wasm-tools/pull/150 to avoid dealing with rebase issues later.

PR #161, opened 2020-12-03
Use hex for the Debug output of Float32/Float64

abrown:
When visually comparing the output of failed assertions (e.g. "expected = ..., actual = ..."), it is much easier to compare float values as hex than their unsigned integer representation. Optimally we would use actual floats in the output but this is fraught since sometimes it is necessary to do a bit-by-bit comparison. Instead, a hex representation shows the bits and avoids a manual "decimal-to-hex" conversion.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -935,7 +935,7 @@ dependencies = [
 
 [[package]]
 name = "wasmparser"
-version = "0.69.0"
+version = "0.69.1"
 dependencies = [
  "anyhow",
  "criterion",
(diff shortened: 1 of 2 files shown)


abrown:
@alexcrichton, this is one part of an attempt to fix https://github.com/bytecodealliance/wasmtime/issues/1681 over in Wasmtime. I'm not sure why the Cargo.lock file changed... I didn't really do anything except run `cargo test`.

abrown:
I'll probably need a version bump to import into Wasmtime--should I do that here?

abrown:
I fixed this with https://github.com/bytecodealliance/wasmtime/pull/2475, closing...

PR #163, opened 2020-12-04
Start implementing module-linking in wasm-smith

alexcrichton:
This commit starts generating module-linking-using modules in the
`wasm-smith` module generator. It's going to take some time to implement
everything so I figured it's best to do it piecemeal. First up in this
commit is the type section, which includes generation of module and
instance types. This also applied a few internal refactorings:

* The type/import sections may be generated in an interleaved fashion
  now with module linking.
* Lookups in the `types` array are now indirected through other arrays
  that record type indices of a specific kind.
* Iterators over types/imports are now indirected through side arrays
  which index into `initializers`.

Further refactorings will be necessary to account for adjustments in
the index spaces as aliases are introduced, but this seemed like a good
start!

### DIFF
--- a/crates/wasm-encoder/src/imports.rs
+++ b/crates/wasm-encoder/src/imports.rs
@@ -11,7 +11,7 @@ use std::convert::TryFrom;
 /// let mut imports = ImportSection::new();
 /// imports.import(
 ///     "env",
-///     "memory",
+///     Some("memory"),
 ///     MemoryType {
 ///         limits: Limits {
 ///             min: 1,
@@ -40,27 +40,21 @@ impl ImportSection {
     }
 
     /// Define an import.
-    pub fn import(&mut self, module: &str, name: &str, ty: impl Into<ImportType>) -> &mut Self {
+    pub fn import(
+        &mut self,
+        module: &str,
+        name: Option<&str>,
+        ty: impl Into<EntityType>,
+    ) -> &mut Self {
         self.bytes.extend(encoders::str(module));
-        self.bytes.extend(encoders::str(name));
-        match ty.into() {
-            ImportType::Function(x) => {
+        match name {
+            Some(name) => self.bytes.extend(encoders::str(name)),
+            None => {
                 self.bytes.push(0x00);
-                self.bytes.extend(encoders::u32(x));
-            }
-            ImportType::Table(ty) => {
-                self.bytes.push(0x01);
-                ty.encode(&mut self.bytes);
-            }
-            ImportType::Memory(ty) => {
-                self.bytes.push(0x02);
-                ty.encode(&mut self.bytes);
-            }
-            ImportType::Global(ty) => {
-                self.bytes.push(0x03);
-                ty.encode(&mut self.bytes);
+                self.bytes.push(0xff);
             }
         }
+        ty.into().encode(&mut self.bytes);
         self.num_added += 1;
         self
     }
@@ -85,9 +79,9 @@ impl Section for ImportSection {
     }
 }
 
-/// The type of an import.
-pub enum ImportType {
-    /// The `n`th function type.
+/// The type of an entity.
+pub enum EntityType {
+    /// The `n`th type, which is a function.
     Function(u32),
     /// A table type.
     Table(TableType),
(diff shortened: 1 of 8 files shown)


fitzgen:
(Started reviewing this; will finish reviewing it on monday)

alexcrichton:
No worries! I've got most of the rest of module linking implemented as well, if you'd prefer I can push it here but I can also keep them separate.

fitzgen:
I think it will be easier to review in smaller chunks, thanks.

crates/wasm-smith/src/lib.rs
```diff
@@ -150,12 +159,46 @@ impl Arbitrary for MaybeInvalidModule {
     }
 }
 
+#[derive(Clone, Debug)]
+enum Initializers {
```
fitzgen, line 163: Maybe rename this to `InitialSection`? I think this would make `ConfiguredModule::initializers` (which could be renamed to `initial_sections`) and all the things that index into it easier to understand.

crates/wasm-smith/src/lib.rs
```diff
@@ -94,8 +94,17 @@ where
 {
     config: C,
     valtypes: Vec<ValType>,
-    types: Vec<FuncType>,
-    imports: Vec<(String, String, Import)>,
+    initializers: Vec<Initializers>,
```
fitzgen, line 97: Can we document what this field is and why `Initializers` itself contains nested vecs?

Here is an attempt that you can probably improve upon:

> With the module linking proposal, types, imports, module, instance, and alias sections can come in any order and occur repeatedly at the start of a Wasm module. We want to generate interesting entities -- entities that require multiple, interspersed occurrences of these sections -- and we don't want to always generate the "same shape" of these initial sections. Therefore, we don't represent a Wasm module with a vector for each of types/imports/etc and then run an algorithm to split them up into multiple sections when encoded. Instead, each entry in `initializers` is one of these initial sections, and we will directly encode each entry as a section when we serialize this module to bytes.

crates/wasm-smith/src/lib.rs
```diff
@@ -550,25 +584,137 @@ where
...
+            // If after generating a list of sections we haven't met our
+            // minimum quotas then meet them now.
+            if let Some(min) = self.config.min_types().checked_sub(self.types.len()) {
+                if min > 0 {
+                    self.arbitrary_types(min, u)?;
+                }
+            }
+            if let Some(min) = self.config.min_imports().checked_sub(self.imports.len()) {
+                if min > 0 {
+                    self.arbitrary_imports(min, u)?;
```
fitzgen, line 605: We can simultaneously get rid of a couple lines and make it read clearer (IMO):

```suggestion
            if self.imports.len() < self.config.min_imports() {
                self.arbitrary_imports(self.config.min_imports() - self.imports.len(), u)?;
```

crates/wasm-smith/src/lib.rs
```diff
@@ -550,25 +584,137 @@ where
...
+                    0 => self.arbitrary_types(0, u)?,
+                    1 => self.arbitrary_imports(0, u)?,
+                    _ => break,
+                }
+            }
+            // If after generating a list of sections we haven't met our
+            // minimum quotas then meet them now.
+            if let Some(min) = self.config.min_types().checked_sub(self.types.len()) {
+                if min > 0 {
+                    self.arbitrary_types(min, u)?;
```
fitzgen, line 600: ditto

crates/wasm-encoder/src/imports.rs
```diff
@@ -40,27 +40,21 @@ impl ImportSection {
     }
 
     /// Define an import.
-    pub fn import(&mut self, module: &str, name: &str, ty: impl Into<ImportType>) -> &mut Self {
+    pub fn import(
+        &mut self,
+        module: &str,
+        name: Option<&str>,
```
fitzgen, line 46: The explainer says that it is the module name that is optional now, not the field name. Although with the desugaring of two level imports into a single level instance import, it makes more sense to have the field name optional. I'll file an upstream bug for discussion.

fitzgen (approved):
r=me with suggestions below applied

thanks!

crates/wasm-encoder/src/imports.rs
fitzgen: https://github.com/WebAssembly/module-linking/issues/18

alexcrichton: Ah yeah I've implemented everything so far as optional fields rather than an optional module, but I'll change things depending on the outcome of that issue.

PR #166, opened 2020-12-07
Add initial support for fuzzing the `alias` section

alexcrichton:
This commit adds the bare-bones support o `wasm-smith` to fuzz the
alias section of a module-linking-using module. This enables the fuzzer
to generate `alias` annotations referencing instances and pull more
items into the index space.

The main bulk of this commit is refactoring the internals of `Module` to
keep track in tables the types of all items, regardless of where they
came from. This simplifies some bits and pieces here and there and
obviates a few now no-longer-necessary caches. After the refactoring a
new `AvailableAliases` type was added to keep track of alias candidates
over time which are filtered as soon as a module's maximum requirements
are met.

### DIFF
--- a/crates/wasm-encoder/src/aliases.rs
+++ b/crates/wasm-encoder/src/aliases.rs
@@ -1,0 +1,63 @@
+use super::*;
+
+/// An encoder for the alias section.
+pub struct AliasSection {
+    bytes: Vec<u8>,
+    num_added: u32,
+}
+
+impl AliasSection {
+    /// Construct a new alias section encoder.
+    pub fn new() -> AliasSection {
+        AliasSection {
+            bytes: vec![],
+            num_added: 0,
+        }
+    }
+
+    /// Define an alias that references the export of a defined instance.
+    pub fn instance_export(&mut self, instance: u32, export: crate::Export) -> &mut Self {
+        self.bytes.push(0x00);
+        self.bytes.extend(encoders::u32(instance));
+        export.encode(&mut self.bytes);
+        self.num_added += 1;
+        self
+    }
+
+    /// Define an alias that references a parent's type.
+    pub fn parent_type(&mut self, ty: u32) -> &mut Self {
+        self.bytes.push(0x01);
+        self.bytes.push(0x07);
+        self.bytes.extend(encoders::u32(ty));
+        self.num_added += 1;
+        self
+    }
+
+    /// Define an alias that references a parent's module.
+    pub fn parent_module(&mut self, module: u32) -> &mut Self {
+        self.bytes.push(0x01);
+        self.bytes.push(0x05);
+        self.bytes.extend(encoders::u32(module));
+        self.num_added += 1;
+        self
+    }
+}
+
+impl Section for AliasSection {
+    fn id(&self) -> u8 {
+        SectionId::Alias.into()
+    }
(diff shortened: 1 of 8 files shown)


crates/wasm-smith/src/lib.rs
```diff
@@ -995,25 +1120,17 @@ where
...
-                        EntityType::Global(g) => {
-                            if !g.mutable && g.val_type == ty.val_type {
-                                choices.push(Box::new(move |_, _| {
-                                    Ok(Instruction::GlobalGet(global_idx))
-                                }));
-                            }
-                            global_idx += 1;
-                        }
-                        _ => {}
+                for (i, g) in self.globals.iter().enumerate() {
```
fitzgen, line 1123: Nice to clean this up so that we aren't iterating over all imports anymore :+1:

crates/wasm-smith/src/lib.rs
```diff
@@ -1076,29 +1193,16 @@ where
     }
 
     fn arbitrary_start(&mut self, u: &mut Unstructured) -> Result<()> {
-        let mut choices = Vec::with_capacity(self.total_funcs as usize);
-        let mut func_index = 0;
+        let mut choices = Vec::with_capacity(self.funcs.len() as usize);
 
         if !self.config.allow_start_export() {
             return Ok(());
         }
```
fitzgen, line 1200: Only just noticed this, but can we put `choices` after the check for whether a start is allowed? Especially since we're eagerly reserving capacity.

crates/wasm-smith/src/lib.rs
```diff
@@ -1392,3 +1478,84 @@ fn arbitrary_vec_u8(u: &mut Unstructured) -> Result<Vec<u8>> {
     let size = u.arbitrary_len::<u8>()?;
     Ok(u.get_bytes(size)?.to_vec())
 }
+
+#[derive(Default)]
+struct AvailableAliases {
+    aliases: Vec<Alias>,
+}
```
fitzgen, line 1485: A nice doc comment here would be... nice ;)

> Tracks the entities that are currently available for aliasing.
>
> As we create a Wasm module's initial sections, defining nested modules and instances, we keep track of the nested entities we've created that are available for use in an `alias` production in this structure.

crates/wasm-encoder/src/aliases.rs
```diff
@@ -0,0 +1,63 @@
+use super::*;
+
+/// An encoder for the alias section.
+pub struct AliasSection {
```
fitzgen, line 4: All of the other sections have an example in the doc comment, could you add one here as well? Also please mention that this section is part of the module linking proposal.

crates/wasm-encoder/src/exports.rs
```diff
@@ -80,10 +80,14 @@ pub enum Export {
     Memory(u32),
     /// An export of the `n`th global.
     Global(u32),
+    /// An export of the `n`th instance.
+    Instance(u32),
+    /// An export of the `n`th module.
+    Module(u32),
```
fitzgen, line 86: Can you mention in the doc comments for these that they are part of the module linking proposal?

fitzgen (approved):
Nice! r=me with nitpicks below addressed

Thanks!

PR #167, opened 2020-12-09
Add support for fuzzing the instance section

alexcrichton:
This commit adds support for fuzzing the instance section of a
module-linking-using module. A cache of `AvailableInstantiations` is
kept to keep track of what modules can be instantiated at any point in
time. This cache is then used to, if there's something in it, generate
an `instance` which can then be later used for further instantiations
and/or aliasing.

### DIFF
--- a/crates/wasm-smith/src/lib.rs
+++ b/crates/wasm-smith/src/lib.rs
@@ -60,6 +60,7 @@ mod terminate;
 use crate::code_builder::CodeBuilderAllocations;
 use arbitrary::{Arbitrary, Result, Unstructured};
 use std::collections::HashSet;
+use std::rc::Rc;
 use std::str;
 
 pub use config::{Config, DefaultConfig, SwarmConfig};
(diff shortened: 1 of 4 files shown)


crates/wasm-smith/src/lib.rs
```diff
@@ -1570,3 +1799,56 @@ impl AvailableAliases {
...
+struct Instantiation {
+    /// The module index which can be instantiated.
+    module: u32,
+    /// The argument vector that must be provided to instantiation, where each
+    /// vector element here is a list of module items which match the import's
+    /// type.
+    ///
+    /// Each sub-vector here is guaranteed to have a length of at least 1, and
+    /// if this instantiation candidate is chosen we'll use fuzz input to select
+    /// which item is used for instantiation.
```
fitzgen, line 1828: Why is each sub-vector guaranteed to have a length of at least 1? What about modules with no imports? They shouldn't take any arguments, right?

fitzgen: Ahhhhh because this isn't `[ set_of_complete_arguments, another_set_of_complete_arguments, ... ]` this is `[ options_for_first_arg, options_for_second_arg, ... ]`. Okay, this makes a lot of sense and avoids a combinatorial explosion.

Could we clarify this a little with something like:

> Each sub-vector `args[i]` contains all the valid options for the `i`th instantiation argument of our `module`. The entry is never empty, because that would mean that the module cannot be instantiated, in which case we do not construct this structure. In the following example, both `$g1` and `$g2` are possible arguments for `$nested.imported-global`, so the associated `args[i]` would be `[$g1, $g2]`
>
> ```wat
> (module
>   (module $nested
>     (import $imported-global "g" (mut i32)))
>   (global $g1 (mut i32) (i32.const 0))
>   (global $g2 (mut i32) (i32.const 42)))
> ```

fitzgen (approved):
Nice! r=me with nitpicks below

PR #169, opened 2020-12-10
Add support for fuzzing the module/module code section

alexcrichton:
This commit adds support to wasm-smith and wasm-encoder to generate the
module and module code sections for module-linking modules. The module
section in wasm-smith is generated by using `Arbitrary for
ConfiguredModule<C>` to create a submodule, and then the type of the
submodule is copied out of the child into the parent for insertion into
the type section. Some deduplication is attempted but in general new
module/instance/func types will always be added to type the new module.

### DIFF
--- a/crates/wasm-encoder/src/module_code.rs
+++ b/crates/wasm-encoder/src/module_code.rs
@@ -1,0 +1,78 @@
+use super::*;
+
+/// An encoder for the module code section.
+///
+/// Note that this is part of the [module linking proposal][proposal] and is
+/// not currently part of stable WebAssembly.
+///
+/// [proposal]: https://github.com/webassembly/module-linking
+///
+/// # Example
+///
+/// ```
+/// use wasm_encoder::{
+///     ModuleCodeSection, Function, ModuleSection, Instruction, Module,
+///     TypeSection, ValType
+/// };
+///
+/// let mut types = TypeSection::new();
+/// types.module(vec![], vec![]);
+///
+/// let mut modules = ModuleSection::new();
+/// let type_index = 0;
+/// modules.module(type_index);
+///
+/// let mut module_code = ModuleCodeSection::new();
+/// module_code.module(&Module::new());
+///
+/// let mut module = Module::new();
+/// module
+///     .section(&types)
+///     .section(&modules)
+///     .section(&module_code);
+///
+/// let wasm_bytes = module.finish();
+/// ```
+pub struct ModuleCodeSection {
+    bytes: Vec<u8>,
+    num_added: u32,
+}
+
+impl ModuleCodeSection {
+    /// Create a new code section encoder.
+    pub fn new() -> ModuleCodeSection {
+        ModuleCodeSection {
+            bytes: vec![],
+            num_added: 0,
+        }
+    }
+
+    /// Write a function body into this code section.
+    pub fn module(&mut self, module: &Module) -> &mut Self {
+        self.bytes.extend(
+            encoders::u32(u32::try_from(module.bytes.len()).unwrap())
+                .chain(module.bytes.iter().copied()),
+        );
+        self.num_added += 1;
+        self
+    }
+}
+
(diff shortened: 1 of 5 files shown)


crates/wasm-encoder/src/module_code.rs
```diff
@@ -0,0 +1,78 @@
...
+    /// Create a new code section encoder.
+    pub fn new() -> ModuleCodeSection {
+        ModuleCodeSection {
+            bytes: vec![],
+            num_added: 0,
+        }
+    }
+
+    /// Write a function body into this code section.
+    pub fn module(&mut self, module: &Module) -> &mut Self {
```
fitzgen, line 51: Nice, I like how this nesting worked out at the API level :+1:

crates/wasm-smith/src/lib.rs
```diff
@@ -1042,6 +1057,161 @@ where
...
+            }
+        }
+    }
+
+    fn get_func_type_idx(&mut self, sig: &Rc<FuncType>) -> u32 {
+        // First check to see if our module already defines this function
+        // signature...
+        if let Some((i, _)) = self.func_types().find(|(_, ty)| **ty == **sig) {
+            return i;
+        }
```
fitzgen, line 1135: We may want to make a `HashMap<Rc<FuncType>>, u32>`, although then I suppose we would never generate modules with duplicate function type definitions.

But also, as written, this will always use the first occurrence of a given function type if there are duplicates. We *could* avoid this by starting this search at a random index `i` and checking each type `j` in the range `i..i+self.num_func_types` via `self.func_types()[j % self.num_func_types]`. But this doesn't really seem worth the hassle.

Anyways, I think the hash consing is probably a good thing to do, rather than linear search, since actually doing duplicate function types well is a bit of a pain and also not very useful.

crates/wasm-smith/src/lib.rs
```diff
@@ -1042,6 +1057,161 @@ where
...
+        // ... and otherwises add a new function signature.
+        let type_idx = self.types.len() as u32;
+        self.func_types.push(type_idx);
+        self.push_type(Type::Func(sig.clone()));
+        type_idx
+    }
+
+    fn get_instance_type_idx(&mut self, sig: &Rc<InstanceType>) -> u32 {
+        // For simplicity for now we always inject a new `instance` type in
+        // this situation instead of looking for one that's already defined.
```
fitzgen, line 1145: Hash consing shouldn't make this too hard either.

crates/wasm-smith/src/lib.rs
```diff
@@ -1042,6 +1057,161 @@ where
...
+        });
+        let type_idx = self.types.len() as u32;
+        self.instance_types.push(type_idx);
+        self.push_type(Type::Instance(sig));
+        type_idx
+    }
+
+    fn get_module_type_idx(&mut self, sig: &Rc<ModuleType>) -> u32 {
+        // Similar to instances, we take the simple route of simply always
+        // injecting a new module type instead of trying to deduplicate.
```
fitzgen, line 1161: Ditto

fitzgen (approved):
:taco:

crates/wasm-smith/src/lib.rs
alexcrichton: I'll go ahead and implement `HashMap<Rc<FuncType>, Vec<u32>>` and that should be easy enough to implement, and it makes all possible function signatures a candidate too!

crates/wasm-smith/src/lib.rs
alexcrichton: I can implement this but it's kind of annoying to do so since it's a lot of derives and it doesn't fit super well with how the types are currently represented which also have indices (which aren't hashed). I'll skip this for now since it seems likely that the module linking proposal will soon be modified that the type of a module isn't declared, the module is simply defined, which would make this an unnecessary operation anyway.

PR #174, opened 2020-12-15
Add extending multiply simd instructions

lars-t-hansen:
These instructions were accepted into the SIMD spec at last week's meeting: https://github.com/WebAssembly/simd/pull/376

(A version bump would be lovely, too.)


alexcrichton:
Thanks!

alexcrichton:
Published now as well.

PR #177, opened 2020-12-17
Limit entities created in wasm-smith module/instance types

alexcrichton:
Creation of arbitrary instance and module types did not respect
configuration for limitations on globals/memories/etc, which could lead
to the creation of a module which imports lots and lots of memories.
This commit changes this to instead apply the limits in `Config` on
module/instance types in addition to generated modules themselves.


cfallin (approved):
Looks good!

PR #183, opened 2020-12-30
[wasmparser] Feature guard for Wasm proposal: SIMD

Robbepop:
I just wanted to see what it would take to feature guard an entire Wasm proposal from the `wasmparser` crate set of features.
I am not entirely sure if I eliminated it correctly and fully but all in all it was done pretty quickly and looks promising.

This PR just feature guards the SIMD Wasm proposal from the `wasmparser` crate but more proposals such as `threading` and `modules` could follow piece by piece.
In order to not break the world the new `simd-proposal` crate feature is enabled by default. To disable it users explicitly have to opt-out of the feature via `--default-features=false` and later only enable those features that they actually have a need for.

I am curious about what people behind this project think about this new way of configuring the `wasmparser` crate.
With this feature there are now 2 way of enabling or disabling Wasm proposals:
1. At runtime as before via the `WasmConfiguration` type.
    - This is still useful if you for example build a library or a generic executable.
2. At compiletime via the new crate feature.
    - This is useful if you build a non-generic solution (executable or even library) where you are sure that you do not want to support certain features at all or where it is too costly in terms of processing power, compilation time, binary size etc.


alexcrichton:
Thanks for the PRs! One thing that we'd want to add with this, if we accept it, is CI to verify everything still compiles correctly. I think though that we'd probably want to refactor the code instead of "just add `#[cfg]` to what's already there" because this could very easily just become a bunch off `#[cfg]` soup otherwise.

One example is that this is not how Cargo features are supposed to be used. Cargo features are supposed to be additive but by adding variants to an `enum` that means that it's a breaking change to enable a Cargo feature in this crate, which isn't something that we'd want. One possible solution is for proposals like SIMD to have an `Operator::Simd(SimdOperator)` where `SimdOperator` is either an opaque structure or an `enum`.

Overall though I'm not sure how well something like this would fit into the development of `wasmparser` over time. Do you think this is necessary for understanding the crate? Or is it intended for other compile-time purposes?

Robbepop:
> CI to verify everything still compiles correctly

I also thought about this one and think it is really important. It is definitely not enough to just check whether all crate features enable works but also features need to work in isolation (and maybe combination) as well. But at some number of crate features to configuration space quickly becomes too big.

> easily just become a bunch off #[cfg] soup otherwise.

I agree again. Unfortunately I do not have concrete ideas how to prevent this with the current crate feature proposal of mine.

> adding variants to an enum that means that it's a breaking change to enable a Cargo feature in this crate, which isn't something that we'd want.

That's a good point and I wasn't aware of this problem.

> One possible solution is for proposals like SIMD to have an Operator::Simd(SimdOperator) where SimdOperator is either an opaque structure or an enum.

This is actually a really great idea! Would you be willing to experimenting with that direction? I could try to tackle this next weekend or so with another PoC PR.

> Do you think this is necessary for understanding the crate? Or is it intended for other compile-time purposes?

I don't think this change would help understanding the crate unfortunately. However, it sure would help with binary size of dependencies (and `wasmparser`) is kinda a low-level dependencies where a whole stack of projects are depending on. Also it might help with compile time and runtime, especially in cases where we don't need the majority of its entire feature set.

alexcrichton:
Could you measure this to see the binary impact? I'm curiouos to see how much  it affects the size of the compiled version of this crate. Some proposals like SIMD are pretty easy to exclude as well, but others like module linking are unfortunately much more invasive and would be much harder to exclude.

Robbepop:
I just compiled `wasmparser` from this branch with the `simd-proposal` crate feature disabled in `--release` mode and compared it to current `master/main` branch.
Results:
```

This PR: libwasmparser-109ba5d791b500a8.rlib        : 2.4M
This PR: libwasmparser-109ba5d791b500a8.rmeta       : 1.5M

Current `main`: libwasmparser-21053a8d32fd49a3.rlib : 2.7M
Current `main`: libwasmparser-21053a8d32fd49a3.rmeta: 1.7M
```
I guess the effect is even more severe when disabling much more than just the SIMD feature.

alexcrichton:
Hm ok ~100k code-size shavings doesn't seem too bad, but also not really earth-shattering per-se. It may just be that we could implement other code-size optimization techniques in this crate?

Robbepop:
I will do some more experiments and see how far we can come with this approach but I also see problems with your previously mentioned `cfg`-soup - nobody wants that. So if we actually decide to go down this path we should really only do this if we can make sure not to enter the soup.

Robbepop:
Closing this since I am convinced now that crate features to guard against Wasm proposals won't be a desirable feature to have since it would explode the configuration space of the project significantly and also causes problems with respect to the fact that some Wasm proposals naturally build on others.

PR #184, opened 2020-12-30
[wasmparser] Feature guard for Wasm proposal: Threads

Robbepop:
Read this PR first: https://github.com/bytecodealliance/wasm-tools/pull/183

This follows the same idea for the Wasm threads proposal.
It was equally simple to carve out and feature flag by the `threads-proposal` crate feature.


Robbepop:
Thread for discussion about this kind of `cfg` feature is here: https://github.com/bytecodealliance/wasm-tools/pull/183#issuecomment-754050991

Robbepop:
Closing this since I am convinced now that crate features to guard against Wasm proposals won't be a desirable feature to have since it would explode the configuration space of the project significantly and also causes problems with respect to the fact that some Wasm proposals naturally build on others.

Issue #182, opened 2020-12-28
[wasmparser] Split Operator::BrTable into multiple sub-operators


Robbepop:
Currently the `Operator` enum has a lifetime attached which makes it a bit harder to work with in some contexts.

The only reason for this is its `BrTable { table: BrTable<'a> }` variant with which it is possible as user of this crate to lazily (but necessarily) parse the targets of the branching table.

The solution I wanted to propose here is to split the single `Operator::BrTable` variant into 3 variants:
1. `Operator::BrTableStart { count_targets: u32 }`
2. `Operator::BrTableTarget { depth: u32 }`
3. `Operator::BrTableDefault { depth: u32 }`
These variants will appear in exactly this order for every `BrTable` in the Wasm input if parsed by the `wasmparser` crate.
The `count_targets: u32` field tells the user how many `Operator::BrTableTarget` variants are to be expected directly afterwards, followed by a single `Operator::BrTableDefault` variant.

This proposal would eliminate the lifetime in the very big `Operator` enum. However, it would also be a breaking change and would probably result in potentially significant changes in how users handle `BrTable` inputs.

I am curious if this idea has already been discussed and what others think about it.

PR #185, opened 2020-12-30
Experimental: Split Operator::BrTable into Operator::{BrTableStart, BrTableTarget}

Robbepop:
Implements https://github.com/bytecodealliance/wasm-tools/issues/182.

This PR does not serve as a proposal for the concrete implementation but rather as a place for discussion of #182 .

Tests are failing because I did not yet update the other crates in this repository that are depending on `wasmparser`, e.g.:
```
 error[E0107]: wrong number of lifetime arguments: expected 0, found 1
   --> crates/wasmprinter/src/lib.rs:701:48
    |
701 |     fn print_operator(&mut self, op: &Operator<'_>, nesting_start: u32) -> Result<()> {
    |                                                ^^ unexpected lifetime argument
```

CI dislikes formatting because I wanted to keep the diff of this PR at a bare minimum and formatting would have touched lots of places that are not interesting to this PR apparently.

### DIFF
--- a/crates/wasmparser/src/binary_reader.rs
+++ b/crates/wasmparser/src/binary_reader.rs
@@ -14,13 +14,12 @@
  */
 
 use std::convert::TryInto;
-use std::fmt;
 use std::str;
 
 use crate::limits::*;
 
 use crate::primitives::{
-    BinaryReaderError, BrTable, CustomSectionKind, ExternalKind, FuncType, GlobalType, Ieee32,
+    BinaryReaderError, CustomSectionKind, ExternalKind, FuncType, GlobalType, Ieee32,
     Ieee64, LinkingType, MemoryImmediate, MemoryType, NameType, Operator, RelocType,
     ResizableLimits, ResizableLimits64, Result, SIMDLaneIndex, SectionCode, TableType, Type,
     TypeOrFuncType, V128,
@@ -79,6 +78,7 @@ pub struct BinaryReader<'a> {
     pub(crate) buffer: &'a [u8],
     pub(crate) position: usize,
     pub(crate) original_offset: usize,
+    remaining_br_targets: u32,
 }
 
 impl<'a> BinaryReader<'a> {
@@ -98,6 +98,7 @@ impl<'a> BinaryReader<'a> {
             buffer: data,
             position: 0,
             original_offset: 0,
+            remaining_br_targets: 0,
         }
     }
 
@@ -107,6 +108,7 @@ impl<'a> BinaryReader<'a> {
             buffer: data,
             position: 0,
             original_offset,
+            remaining_br_targets: 0,
         }
     }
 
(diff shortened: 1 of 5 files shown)


alexcrichton:
Thanks for the PR! I've often though that it'd be great to avoiding having to skip ahead in various places inside of `wasmparser`, although this happens in a lot of places currently, not just in `br_table` parsing.

The addition of the state in `BinaryReader` feels a bit odd to me though. One possible fix for this, however, is to avoid updating  `BinaryReader` itself too much and instead document a contract that once `BrTableStart` is seen then `read_var_u32` needs to be called the specified number of times, plus one, in succession. That could be automatically handled by higher-level operator iterators, but it would make `BinaryReader` avoid having any extra state in it.

Robbepop:
> Thanks for the PR! I've often though that it'd be great to avoiding having to skip ahead in various places inside of `wasmparser`, although this happens in a lot of places currently, not just in `br_table` parsing.
> 
> The addition of the state in `BinaryReader` feels a bit odd to me though. One possible fix for this, however, is to avoid updating `BinaryReader` itself too much and instead document a contract that once `BrTableStart` is seen then `read_var_u32` needs to be called the specified number of times, plus one, in succession. That could be automatically handled by higher-level operator iterators, but it would make `BinaryReader` avoid having any extra state in it.

Thanks for the reply and for the new idea about how else we could go forward.
I am mainly interested in starting a discussion about this and whether it is needed at all.
The new state in `BinaryReader` indeed feels wrong but it is fine to get a feeling for the PoC - wouldn't merge it either in the current state tbh.

If you want we could experiment a bit with your proposal and see how it fleshes out.
However, is this generally something you anticipate or do you regard this not very important?

alexcrichton:
I don't feel like I have a great compass for whether this is worthwhile or not. I don't know how slow the skipping of varuint32 actually is. There's a good number of places in wasmparser that "parse" things twice either by literally parsing or skipping bytes, and I'm not sure whether it's actually costly or how it factors "in the large". It might be reasonable to try to see how this affects microbenchmarks for reasonably-sized wasm modules.

Robbepop:
> I don't feel like I have a great compass for whether this is worthwhile or not. I don't know how slow the skipping of varuint32 actually is. There's a good number of places in wasmparser that "parse" things twice either by literally parsing or skipping bytes, and I'm not sure whether it's actually costly or how it factors "in the large". It might be reasonable to try to see how this affects microbenchmarks for reasonably-sized wasm modules.

Fair point and sorry if I falsely expected you to provide me with a compass here! :sweat_smile: 

For me the more important part about this change is what I describe [in this thread](https://github.com/bytecodealliance/wasm-tools/issues/182) under the "What does this solve?" and "Streamlined Design" approach.

Can you point me to some other places where we currently have double parsing?

Robbepop:
That's all very interesting information, thanks!

Before you refactored the entire crate with the genius `Payload` construct I was trying to decouple the validation of the crate using traits that still exist in some way today.
The idea was to have a `Parser<T>` type that is generic over some `T: Validator` where `Validator` is a trait used by the parser to feed and query the validation process. If you only want to parse without validation you'd use something like `Parser<NoopValidator>`. The idea behind this is that oftentimes information required by backends is very similar to what a validator requires and therefore it oftentimes is possible to use the backend as the resource storage for the validator reducing memory footprint and eventually getting rid of double-parsing of the source.

alexcrichton:
Yeah I think that'd work, and another possible alternative I've thought of is to have more specialized methods on `Validator` for "here's the import I just parsed", so the application drives the iteration and then feeds items into the validator, sort of like how the opcode validation currently works.

I haven't worried too too much about all this though yet becaue the code section is almost always the vast majority of the time spent when validating, but it's always useful to plan to fix these sorts of things in the future!

Robbepop:
Closing this since the PR has been stale for quite some time and would require a load of work to get working again. Besides that the most recent changes to the `wasmparser` crate have made the changes introduced in this PR less valuable. If the maintainers wish to proceed with this approach we can re-open or open another PR.

PR #186, opened 2021-01-03
Make wasmparser compile for #![no_std]

tomaka:
This make makes the `wasmparser` crate successfully compile for no-std platforms.
The PR was very mechanical, but feel free to reject if this is out of scope of the repository.

I've added an `std` feature just because of a single `impl std::error::Error for ... {}`.
My personal opinion would be to ditch this `impl` and the `std` feature entirely, but I went for the non-breaking option. Let me know if I should do this.

### DIFF
--- a/crates/wasmparser/Cargo.toml
+++ b/crates/wasmparser/Cargo.toml
@@ -11,6 +11,9 @@ A simple event-driven library for parsing WebAssembly binary files.
 """
 edition = "2018"
 
+[dependencies]
+hashbrown = { version = "0.9.1", features = ["ahash"], default-features = false }
+
 [dev-dependencies]
 anyhow = "1.0"
 criterion = "0.3"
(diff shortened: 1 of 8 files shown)


Robbepop (approved):
LGTM

Having `no_std` support for `wasmparser` would be pretty handy.

alexcrichton:
I [left a comment](https://github.com/bytecodealliance/wasmtime/pull/2537#issuecomment-754057070) on a similar PR as well for Wasmtime, and I'd personally prefer to resolve it there first before figuring out whether or not to merge this.

Robbepop:
@tomaka FYI I published a `no_std` compatible `wasmparser` crate here: https://crates.io/crates/wasmparser-nostd (in case you still need this)

It also switches out `Hash{Map,Set}` with `BTree{Map,Set}` internally so that the implementation is not attackable in `no_std` environment. Enable `no_std` by disabling default features (`--no-default-features`) as usual.

Robbepop:
I think we can close this PR.

- It is outdated.
- We already have an open issue that informs people about `no_std` support in `wasmparser` et. al.
- The issue contains a link to the actual `wasmparser-nostd` fork that I maintain since a while.

alexcrichton:
Sounds good.

PR #191, opened 2021-01-08
Sync syntax/binary of aliases with upstream spec

alexcrichton:
This implements two features from https://github.com/WebAssembly/module-linking/pull/26:

* First shorthand syntax of `(func $instance "export_name")` is implemented everywhere for all items as a syntax sugar for alias references.
* Next a depth argument is added to outer aliases in `wasmparser` and validation to allow referring to any outer module.

### DIFF
--- a/crates/wasmparser/src/readers/alias_section.rs
+++ b/crates/wasmparser/src/readers/alias_section.rs
@@ -11,8 +11,14 @@ pub struct AliasSectionReader<'a> {
 
 #[derive(Debug)]
 pub enum Alias<'a> {
-    ParentType(u32),
-    ParentModule(u32),
+    ParentType {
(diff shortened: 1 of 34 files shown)


crates/wasmparser/src/readers/alias_section.rs
```diff
@@ -11,8 +11,14 @@ pub struct AliasSectionReader<'a> {
 
 #[derive(Debug)]
 pub enum Alias<'a> {
-    ParentType(u32),
-    ParentModule(u32),
+    ParentType {
```
fitzgen, line 14: Maybe name these `OuterType` and `OuterModule` to match the spec more closely?

fitzgen (approved):

PR #195, opened 2021-01-11
Update wasm-smith for recent changes to module-linking

alexcrichton:
* Implement exporting modules/instances
* Implement implicit instances from two-level imports
* Update encoding of instance arguments
* Implement outer aliases to types/modules

### DIFF
--- a/crates/wasm-smith/src/lib.rs
+++ b/crates/wasm-smith/src/lib.rs
@@ -94,8 +94,10 @@ where
     C: Config,
 {
     config: C,
-    depth: usize,
     valtypes: Vec<ValType>,
+
+    /// Parent modules, if any (used for module linking)
+    parents: Vec<Parent>,
 
     /// The initial sections of this wasm module, including types and imports.
     /// This is stored as a list-of-lists where each `InitialSection` represents
@@ -126,12 +128,9 @@ where
     /// Indices into `initializers` which are types. The pair `(i, j)` means
     /// that `initializers[i]` is a type section, and we are referring to the
     /// `j`th type within that type section.
-    types: Vec<(usize, usize)>,
+    types: Vec<LocalType>,
     /// Indices within `types` that are function types.
     func_types: Vec<u32>,
-    /// Map from function signature to indices that have that function
-    /// signature.
-    func_map: HashMap<Rc<FuncType>, Vec<u32>>,
     /// Indices within `types` that are module types.
     module_types: Vec<u32>,
     /// Indices within `types` that are instance types.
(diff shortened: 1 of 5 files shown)


crates/wasm-smith/src/lib.rs
```diff
@@ -94,9 +94,11 @@ where
     C: Config,
 {
     config: C,
-    depth: usize,
     valtypes: Vec<ValType>,
 
+    /// Parent modules, if any (used for module linking)
+    parents: Vec<Parent>,
```
fitzgen, line 100: `outers: Vec<Outer>` etc to match spec?

crates/wasm-smith/src/lib.rs
```diff
@@ -126,12 +128,9 @@ where
     /// Indices into `initializers` which are types. The pair `(i, j)` means
     /// that `initializers[i]` is a type section, and we are referring to the
     /// `j`th type within that type section.
-    types: Vec<(usize, usize)>,
+    types: Vec<LocalType>,
```
fitzgen, line 131: Comment for this member needs to be updated, and I think the existing text can be moved to `LocalType::Defined` (right?)

crates/wasm-smith/src/lib.rs
```diff
@@ -1980,6 +2001,7 @@ fn arbitrary_vec_u8(u: &mut Unstructured) -> Result<Vec<u8>> {
 struct AvailableAliases {
     aliases: Vec<Alias>,
     instances_added: usize,
+    parents_processed: bool,
```
fitzgen, line 2004: `outers_processed`

fitzgen (approved):
LGTM modulo a bunch of little nitpicks -- sorry I am a stickler for capitalizing and punctuating code comments!

crates/wasm-smith/src/lib.rs
```diff
@@ -814,16 +819,33 @@ where
...
+        let mut import_types = indexmap::IndexMap::new();
         let mut names = HashMap::new();
         if !entities.max_reached(&self.config) {
             arbitrary_loop(u, 0, self.config.max_imports(), |u| {
                 let (module, name) = unique_import_strings(1_000, &mut names, true, u)?;
                 let ty = self.arbitrary_entity_type(u, entities)?;
+                if let Some(name) = &name {
+                    let ity = import_types.entry(module.clone()).or_insert_with(|| {
+                        EntityType::Instance(u32::max_value(), Default::default())
+                    });
```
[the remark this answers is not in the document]
alexcrichton: In general there's a fair amount of cloning happening everywhere in wasm-smith wrt the module structure and module linking, so I'm not too worried about this specifically. The default limits in place and `SwarmConfig` I think makes it so we don't have to worry about this too too much. That being said though I'm mostly relying on oss-fuzz's timeouts to tell us where we need to optimize here, because there's definitely more than one place we can optimize in this file!

For now though I think I'll leave this as-is because I think it's the most readable, but if it becomes problematic I'll probably refactor this out so it's not all in one big blob.

alexcrichton:
Heh no worries, it's high time someone calls out my sloppy comments anyway :)

PR #208, opened 2021-01-25
Don't check errors in name section

yurydelendik:
There is an issue with using wasm2wat-rs with new LLVM output. The name section now contains other (non-documented?) name types

https://github.com/llvm/llvm-project/blob/831a143e50cac873ec095fc7139a485173ba8c35/llvm/include/llvm/BinaryFormat/Wasm.h#L321-L322

All invalid custom section shall be ignored -- doing that in this patch. (I think we can keep what's partially read as valid)


yurydelendik:
cc @jlb6740

alexcrichton (approved):

PR #209, opened 2021-01-25
Enable bulk-memory and reference-types by default

alexcrichton:
These proposals moved to stage 4 some time ago, so enable them by
default in the validator.


fitzgen (approved):
Do any docs need updating as well? r=me with those (if needed) too

PR #216, opened 2021-01-28
Optimize `AvailableInstantiations::update`

alexcrichton:
The previous implementation of this function was quite naive and went to
no lengths to be optimized. This is unfortunately causing timeouts on
OSS-Fuzz, however, so this commit applies the requisite optimizations to
get this to run quickly on oss-fuzz.

The purpose of this method is to keep track of all possible
instantiations that a module can make, maintaining a list of modules and
the arguments which can be provided to the module to instantiate it. The
previous implementation simply cleared all state and figured this out
from scratch on every iteration, so if there were lots of items this
would take an exponential amount of time to complete.

This PR changes the `update` method to be incremental. It keeps track of
all processed items and only visits each candidate in a module
once-per-module-to-instantiate, which drives down the cost of this
method significantly. The incremental update should be relatively easily
(hopefully) to reason about as well by simply visiting newer items for
each time `update` is called.

### DIFF
--- a/crates/wasm-smith/src/lib.rs
+++ b/crates/wasm-smith/src/lib.rs
@@ -1151,7 +1151,7 @@ where
             |u| {
                 // We can only instantiate a choice if all of its arguments
                 // have at least one possibility, so filter for those candidates
-                // and break out if there are 0, otherwise selectt one of them.
+                // and break out if there are 0, otherwise select one of them.
                 let choices = available
                     .choices
                     .iter()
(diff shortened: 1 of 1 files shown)


crates/wasm-smith/src/lib.rs
```diff
@@ -1145,15 +1143,24 @@ where
...
         let mut instances = Vec::new();
         arbitrary_loop(
             u,
             0,
             self.config.max_instances() - self.instances.len(),
             |u| {
-                let choice = u.choose(&available.choices)?;
+                // We can only instantiate a choice if all of its arguments
+                // have at least one possibility, so filter for those candidates
+                // and break out if there are 0, otherwise selectt one of them.
```
fitzgen, line 1154:
```suggestion
                // and break out if there are 0, otherwise select one of them.
```

fitzgen (approved):

PR #220, opened 2021-02-05
wasm-encoder: Add support for the linking custom section

fitzgen:
This section is a non-standard convention that is supported by the LLVM
toolchain. It, along with associated "reloc.*" custom sections, allows you to
treat a Wasm module as a low-level object file that can be linked with other
Wasm object files to produce a final, complete Wasm module.

I haven't implemented support for reloc sections yet, nor for encoding always-padded-to-5-bytes-and-therefore-relocatable varuint32 support to the rest of the encoder API yet. Those things will happen sometime in the future.


alexcrichton:
👍

PR #223, opened 2021-02-17
Support `delegate` instruction and update other exception instructions

takikawa:
This PR adds support for the `delegate` instruction that's part of the exception handling proposal. It was previously left out because it didn't have an opcode assignment, but the latest spec explainer (https://github.com/WebAssembly/exception-handling/pull/137#issuecomment-778482300) includes it.

It also updates/fixes support for other instructions to match recent spec updates. For example, `catch_all` is no longer sharing an opcode with `else`.

With these patches, exception support should also match the behavior in wabt's tests (when including these pending PRs: https://github.com/WebAssembly/wabt/pull/1608 and https://github.com/WebAssembly/wabt/pull/1605).  I can submit a PR later to enable exception handling tests that are currently disabled once wabt PRs are merged and the submodule can be updated.

### DIFF
--- a/crates/wasmparser/src/binary_reader.rs
+++ b/crates/wasmparser/src/binary_reader.rs
@@ -1139,6 +1139,7 @@ impl<'a> BinaryReader<'a> {
             0x18 => Operator::Delegate {
                 relative_depth: self.read_var_u32()?,
             },
+            0x19 => Operator::CatchAll,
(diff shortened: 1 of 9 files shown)


alexcrichton:
Thanks!

And yeah a submodule update would be great! That can be tricky sometimes though so let me know if there's any trouble

PR #225, opened 2021-02-17
Implement inline `alias` annotations

alexcrichton:
This implements a text format sugar similar to inline `import`
annotations, but for `alias` directives. This does not implement inline
`alias` annotations for modules just yet because that could currently
require arbitrary lookahead.


peterhuene:
@alexcrichton I'm going to close this PR as it looks like it'll be superseded with the component model sooner than later. Please reopen if you disagree 😄

PR #226, opened 2021-02-18
Update wabt submodule and roundtrip.rs to enable more exception handling tests

takikawa:
This PR enables more exception handling tests by updating the wabt submodule and the roundtrip test runner. These tests should pass now that wabt and wasm-tools are mostly aligned on this proposal.

I left two tests disabled that fail due to minor inconsistencies on the `delegate` instruction (these will be fixed in the near future).

Edit: originally I also had a commit that changed the dependencies of the exception feature flag, but I've removed that as it breaks the tests under the `testsuite/proposals/exception-handling` folder (these tests aren't up-to-date and are all actually reference type tests I think).


alexcrichton (approved):
Nice, thanks so much for this!

PR #227, opened 2021-02-18
Speed up `AvailableInstantiations::update`

alexcrichton:
This fixes a timeout seen on oss-fuzz by applying a few perf
improvements:

* If no entities have been added since `update` was last called we can
  skip all the work entirely.

* When udpating historical candidates we can skip a hash table lookup by
  iterating in lockstep instead since these maps have the same layout.


fitzgen (approved):
Nice!

PR #237, opened 2021-03-01
Account for invalid names in roundtrip fuzzing

alexcrichton:
This fixes a bug in the roundtrip fuzzing target where an input `*.wat`
file can produce a wasm file with an "invalid" name section because the
name might be too large. This means that the roundtrip-ing of that file
would not preserve the name section, causing the previous assertion to
fail.

The fuzzer now validates that the name section, if present, is valid
before proceeding with fuzzing.

### DIFF
--- a/fuzz/fuzz_targets/roundtrip.rs
+++ b/fuzz/fuzz_targets/roundtrip.rs
@@ -24,12 +24,73 @@ fuzz_target!(|data: &[u8]| {
         Ok(bytes) => bytes,
         Err(_) => return,
     };
+
+    // Only roundtrip valid modules for now since invalid modules can often have
+    // bizarre structures which aren't intended to print correctly or roundtrip
+    // well.
     if wasmparser::validate(&wasm).is_err() {
         return;
     }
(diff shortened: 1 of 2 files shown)


fuzz/fuzz_targets/roundtrip.rs
```diff
@@ -24,12 +24,73 @@ fuzz_target!(|data: &[u8]| {
...
+    }
+
+    std::fs::write("wasm1.wasm", &wasm).unwrap();
+    std::fs::write("wasm1.wat", &string).unwrap();
+    std::fs::write("wasm2.wasm", &wasm2).unwrap();
+    std::fs::write("wasm2.wat", &string2).unwrap();
+    panic!("wasm bytes differ on roundtrip");
 });
+
+fn name_section(wasm: &[u8]) -> wasmparser::Result<()> {
```
fitzgen, line 60: I would expect, based on the name, that this return the contents of the parsed name section. Suggest naming this `validate_name_section` or something like that.

fitzgen (approved):

PR #238, opened 2021-03-01
Bump versions

abrown:


crates/wasmprinter/Cargo.toml
```diff
@@ -14,7 +14,7 @@ Rust converter from the WebAssembly binary format to the text format.
 
 [dependencies]
 anyhow = "1.0"
-wasmparser = { path = '../wasmparser', version = '0.75' }
```
abrown, line 17: Should we drop the `version = ...` here? It is not present elsewhere.

crates/wat/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "wat"
-version = "1.0.35"
```
abrown, line 3: I went minor here...

crates/wasmprinter/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "wasmprinter"
-version = "0.2.23"
+version = "0.3.0"
```
alexcrichton, line 3: Since this one didn't have any API-breaking changes we leave wasmprinter on the 0.2 track, so this can just bump it to 0.2.24

crates/wasmprinter/Cargo.toml
alexcrichton: Anything in `[dependencies]` needs a version for publication, but `dev-dependencies` doesn't need it

crates/wat/Cargo.toml
```diff
@@ -1,6 +1,6 @@
...
+version = "1.1.0"
```
alexcrichton, line 3: Similar to `wasmprinter` the API of this crate didn't change, so it's safe to just bump to 1.0.36

PR #239, opened 2021-03-05
Reorder some SIMD opcodes

alexcrichton:
I wanted to do one last final pass to ensure that everything was covered in this repository, and while I was at it I went ahead and reordered the instruction lists of wast/wasmparser to match the spec to easily skim through and see if we're missing opcodes here and there.


alexcrichton:
It's worth noting that it's not my intention to change anything here, just shuffling things around.

PR #242, opened 2021-03-16
Fix emission of exports.

nlewycky:
I think the test for permitting the emission of exports was backwards. Unfortunately I couldn't find a way in wasm-smith/tests to add a test for whether a module has an export or not, but I've tested it myself with a ConfiguredModule with a configuration the requires at least one export (before this patch they have none, afterwards they do) and checking with wasm2wat on the command line.


alexcrichton:
Oops thanks for catching this! I'll merge once CI is green

PR #243, opened 2021-03-18
wasmparser: Add full payload range to custom sections

fitzgen:
Only the range of the raw custom data could be recovered via `data_offset` and
`data.len()` before. This allows easily getting the range of the whole section,
similar to other section-specific readers via the `SectionReader::range` trait
method.

### DIFF
--- a/crates/dump/src/lib.rs
+++ b/crates/dump/src/lib.rs
@@ -273,10 +273,13 @@ impl<'a> Dump<'a> {
 
                 Payload::CustomSection {
                     name,
-                    data,
                     data_offset,
+                    data,
+                    range,
                 } => {
-                    write!(self.state, "custom section: {:?}", name)?;
+                    write!(self.state, "custom section")?;
+                    self.print(range.start)?;
+                    write!(self.state, "name: {:?}", name)?;
                     self.print(data_offset)?;
                     if name == "name" {
                         let mut iter = NameSectionReader::new(data, data_offset)?;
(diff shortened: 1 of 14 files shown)


crates/wasmparser/src/parser.rs
```diff
@@ -151,6 +151,9 @@ pub enum Payload<'a> {
         data_offset: usize,
         /// The actual contents of the custom section.
         data: &'a [u8],
+        /// The range of bytes that specify this whole custom section, specified
+        /// in offsets relative to the start of the byte stream.
```
alexcrichton, line 155: Could this clarify that the intent of this range vs `data_offset` is that this range encompasses the name length and the name in addition to the data?

crates/dump/src/lib.rs
```diff
@@ -271,13 +271,9 @@ impl<'a> Dump<'a> {
...
 
-                Payload::CustomSection {
-                    name,
-                    data,
-                    data_offset,
-                } => {
+                Payload::CustomSection { name, data_offset, data, range } => {
                     write!(self.state, "custom section: {:?}", name)?;
-                    self.print(data_offset)?;
+                    self.print(range.start)?;
```
alexcrichton, line 276: I think `data_offset` here is technically correct given the current code, but that's only because the `write!` above accounts for all of the custom section id, the name, and the length of the whole section.

Could the `write!` above be split  and this calls `self.print` twice? The first `self.print` would simply be "custom section" with `range.start` as the initial start, and then the next `self.print` would be "name: {}" with the offset of `data_offset`

alexcrichton (approved):
Seems reasonable to me!

PR #249, opened 2021-03-29
Continue to tweak type size calculations in `wasm-smith`

alexcrichton:
This commit updates the type size calculation since Wasmtime was getting
requests to instantiate modules with multiple thousands of memories, but
that shouldn't be valid since the maximum type size is 1000.

The fix here is to basically say that when calculating the size of an
instance or module, adding an import/export is where the `+1` happens.
This means that the size of an instance/module is the number of
imports/exports plus the recursive size of each import/export. This fix
is applied to generating instance/module types, but it's also applied
when adding imports/exports (the `+1` happens on the addition of the
import/export.

This also refactors the loop that adds exports to build a list of all
candidate exports and then it's filtered down to only exports whose type
size fits within the remaining budget of type size.

### DIFF
--- a/crates/wasm-smith/src/lib.rs
+++ b/crates/wasm-smith/src/lib.rs
@@ -833,7 +833,7 @@ where
             arbitrary_loop(u, 0, self.config.max_imports(), |u| {
                 let (module, name) = unique_import_strings(1_000, &mut names, true, u)?;
                 let ty = self.arbitrary_entity_type(u, entities)?;
-                match type_size.checked_add(ty.size()) {
+                match type_size.checked_add(ty.size() + 1) {
                     Some(s) if s < self.config.max_type_size() => type_size = s,
                     _ => return Ok(false),
                 }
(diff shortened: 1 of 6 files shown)


crates/wasm-smith/src/lib.rs
```diff
@@ -1476,60 +1476,32 @@ where
...
+            self.funcs.len()
+                + self.tables.len()
+                + self.memories.len()
+                + self.globals.len()
+                + self.instances.len()
+                + self.modules.len(),
+        );
+
+        for i in 0..self.funcs.len() {
+            choices.push(Export::Func(i as u32));
```
fitzgen, line 1489: This is making the size of choices grow from `O(1)` to `O(n)` since we are pushing each function onto `choices` rather than doing a single push for all functions, and similar for other entities.

Can we do the approach we had before, except we only push the export-a-function choice if there exist any functions that won't blow the type size limit, and then do the filtering of which functions are available only if the choice to export a function was made?

alexcrichton: Sure! I've pushed up a version which does that now

fitzgen (approved):

PR #252, opened 2021-03-30
Implement `wat::Error::set_path`.

peterhuene:
This PR exposes a `set_path` method for `Error` that can associate a file
path with the error.

This is similar to `wast::Error::set_path`.  It may be useful for users of the
`wat` crate to be able to associate a file path with the error outside of using
`parse_file`.

When the internal error is an error from `wast`, this will improve the error
message displayed.


alexcrichton (approved):
👍

PR #260, opened 2021-04-16
Add an "exhaustive checking mode" to `wast`

alexcrichton:
This commit adds a conditional compilation mode to `wast` where new
fields are injected into some AST nodes to add assertions later on that
all AST nodes were visited by the resolution passes. There have been
quite a few bugs recently related to the alias pass forgetting to run on
some AST nodes, so this is intended to fix the last of those once and
for all.

When the `--cfg check_exhaustive` flag is used when compiling `wast`
(must be passed via `RUSTFLAGS`) the crate will insert boolean fields
into all item references along with a "visited" flag. This "visited"
flag is then set during alias expansion (and perhaps later we could
check it during resolve expansion too). Resolution and binary encoding
passes which already visit all of these nodes anyway then check this
flag and return a bug or assert if something is encountered.

Another possible solution to this would be some sort of `Visit` trait
but that sort of feels like overkill when just two passes are in
question here (alias expansion and resolution). I'm hoping that we can
leverage this to be a bit lighter weight in the long run, but we'll see!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -61,6 +61,9 @@ jobs:
       shell: bash
 
     - run: cargo test --all
+    - run: cargo test --all
+      env:
+        RUSTFLAGS: --cfg=wast_check_exhaustive
     - run: cargo test --manifest-path crates/wasmparser/Cargo.toml --features deterministic
     - run: cargo build --manifest-path crates/wast/Cargo.toml --no-default-features
     - run: cargo build --manifest-path crates/wast/Cargo.toml --no-default-features --features wasm-module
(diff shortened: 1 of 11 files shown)


.github/workflows/main.yml
```diff
@@ -61,6 +61,9 @@ jobs:
       shell: bash
 
     - run: cargo test --all
+    - run: cargo test --all
+      env:
+        RUSTFLAGS: --cfg=check_exhaustive
```
fitzgen, line 66: Total nitpicking, but: since this isn't "namespaced" as a cargo feature, perhaps it should be `--cfg=wast_check_exhaustive`?

fitzgen (approved):
Nice idea!

PR #264, opened 2021-04-21
Add "unknown" as name subsection type

samsartor:
In principal, the [format of the "name" custom section](https://webassembly.github.io/spec/core/appendix/custom.html#subsections) can support arbitrary custom subsections. I'd be surprised if any tools actually uses this ability for truly custom stuff, but it provides some decent forwards compatibility even once wasmparser gains support for proposals such as "extended name section" (which I am also working on now as a separate PR).

### DIFF
--- a/crates/dump/src/lib.rs
+++ b/crates/dump/src/lib.rs
@@ -359,6 +359,11 @@ impl<'a> Dump<'a> {
                     }
                 }
             }
+            Name::Unknown { ty, range, .. } => {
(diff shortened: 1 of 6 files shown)


samsartor:
I take what I said back about "tools not using this for truly custom stuff". It looks like emscripten takes the extended name section proposal a bit further and uses id 9 to store a data names map.

alexcrichton:
Looks reasonable to me, thanks!

PR #265, opened 2021-04-22
Implement the extended name section proposal

samsartor:


crates/wasmparser/src/validator.rs
```diff
@@ -178,6 +178,8 @@ pub struct WasmFeatures {
     pub exceptions: bool,
     /// The WebAssembly memory64 proposal
     pub memory64: bool,
+    /// The WebAssembly extended name section proposal
+    pub extended_name_section: bool,
```
alexcrichton, line 182: I think it's fine to not list this in `WasmFeatures` since it otherwise only affects custom section parsing which can change at any time (we don't validate custom sections anyway in theory)

samsartor:
> It looks like `wasmprinter` still needs to be updated ... I think ideally this'd also update the `wast` crate

Yep! That is why this PR is a draft. I just wanted to get some visibility on it, since I need to take a break for a few days. I have an actual ticket in JIRA I am neglecting lol.

alexcrichton:
I've updated this and am continuing it in https://github.com/bytecodealliance/wasm-tools/pull/298 to land

alexcrichton:
Merged in https://github.com/bytecodealliance/wasm-tools/pull/298

PR #267, opened 2021-04-27
Bump wasmparser to 0.78.0

fitzgen:


crates/wasmprinter/Cargo.toml
```diff
@@ -14,7 +14,7 @@ Rust converter from the WebAssembly binary format to the text format.
 
 [dependencies]
 anyhow = "1.0"
-wasmparser = { path = '../wasmparser', version = '0.77' }
+wasmparser = { path = '../wasmparser' }
```
alexcrichton, line 17: Since we publish wasmparser we need to keep the dependency version here

alexcrichton (approved):
Mind bumping wasmprinter as well (minor version) so we can publish a version that pulls in the latest wasmparser too?

crates/wasmprinter/Cargo.toml
fitzgen: Ah for some reason I thought that cargo would figure this out manually.

fitzgen:
Tagged and published!

PR #269, opened 2021-05-04
wasmparser: Derive `Hash` for `TypeDef`

fitzgen:
And everything transitively contained inside a `TypeDef`.


fitzgen:
Actually nevermind, this doesn't handle equivalent module types whose entries are provided in different orders.

PR #270, opened 2021-05-10
Add license files in crates

olivierlemasle:
This Pull Request adds the LICENSE files in published crates.

I'm currently packaging crates wat, wast and wasmparser for Fedora, and Fedora packages should have license files.


olivierlemasle:
> A minor logistical question -- I'm not actually sure if the symlinks here will cause unforeseen issues on some platforms. Specifically it looks like git on Windows will just create a text file containing the path; I suppose that's kind-of OK, though it may be better to just copy the file.

I did not test on Windows, and I'm ok with copying the license file if you prefer.

FYI, `wasm-smith` already uses a symlink for license since its re-licensing in commit  ed55b561.

> Separately, as a Fedora user, thank you for packaging these, and let us know if we can help in any way with getting any of the other BA projects into Fedora's repositories!

Thanks! Actually, I plan to package `wasmtime-go` and all its dependencies, which includes `wasmtime`, `wat`, `wast`, `cranelift` and many BA crates...

cfallin:
> FYI, wasm-smith already uses a symlink for license since its re-licensing in commit ed55b56.

Ah, that indeed answers my concern -- "we already do it and it works" is quite a good reason!

alexcrichton:
Ah yeah I haven't had any issues with this myself yet, but I haven't tested it rigorously.

PR #273, opened 2021-05-18
wasmparser: Derive `Hash` for `{Memory,Global,Table}Type` 

fitzgen:
and bump to 0.79.0.

### DIFF
--- a/crates/wasmparser/Cargo.toml
+++ b/crates/wasmparser/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wasmparser"
-version = "0.78.1"
+version = "0.78.2"
 authors = ["Yury Delendik <ydelendik@mozilla.com>"]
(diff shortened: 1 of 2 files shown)


alexcrichton (approved):
I think this can actually be 0.78.2 since it looks API compatible? Otherwise looks good!

fitzgen:
Wait. I might be confusing myself again. I think I need a subset of what's derived here, but let me double check.

fitzgen:
Okay yeah I just needed `Hash` for `{Global, Memory, Table}Type` so I removed the other impls (which are bogus because e.g. `InstanceType` will hash differently depending on order of its exports even though it should be considered equal)

fitzgen:
Also switched to 0.78.2 instead of 0.79.0

alexcrichton (approved):

Issue #207, opened 2021-01-23
How to get the AST information after parse processing


Chehx:
Hello, here is a simple code snippet:

```
use wast::Wat;
use wast::parser::{self, ParseBuffer, Result};
use wast::Module;

fn main() -> Result<(), > {
    let wat = "(module (func $add (param $lhs i32) (param $rhs i32) (result i32)local.get $lhs local.get $rhs i32.add)(export \"add\" (func $add)))";
    let buf = ParseBuffer::new(wat)?;
    let module = parser::parse::<Wat>(&buf)?;
    Ok(())
}
```

After that code, it seems successfully parse the wat text. But, I didn't find any document to how to print or utilize the AST we could get from this parsing.

Hope you can help me figure it out.

Thanks

alexcrichton:
In this example `module` should have type [`Wat`](https://docs.rs/wast/32.0.0/wast/struct.Wat.html) which internally has a `Module` which should have the AST. Is that what you're interested in?

Chehx:
> In this example `module` should have type [`Wat`](https://docs.rs/wast/32.0.0/wast/struct.Wat.html) which internally has a `Module` which should have the AST. Is that what you're interested in?

Many thanks for your reply!
Yeah, but I don't see any public function to print that AST. Is that possible for me to implement by myself?

alexcrichton:
I think you should be able to do `println!("{:?}", module)` perhaps? But yes you can always implement it yourself too.

alexcrichton:
I'm going to close this since I think it's since been addressed.

PR #275, opened 2021-05-29
Derive `Debug` for type `Wat`

9kunal:
This includes a simple change to derive`Debug` for type `Wat`, type `Module` and Type `ModuleKind`

In reference to the issue [207](https://github.com/bytecodealliance/wasm-tools/issues/207) created earlier.
Should be helpful to debug and do a quick print of AST.


alexcrichton:
Thanks!

PR #277, opened 2021-06-04
Enable larger instruction sequences in functions

abrown:
Previously, there was a 50% chance that the instruction sequence would
finish early (i.e., before we run out of random data). This probability
is reduced to 1/256, making it much more likely to create longer
instruction sequences.

Longer instruction sequences in turn seemed to frequently expose an "out
of random data" error. This change makes the more permissive choice to
simply avoid adding new instructions when we run out of data instead of
returning an error.

### DIFF
--- a/crates/wasm-smith/src/code_builder.rs
+++ b/crates/wasm-smith/src/code_builder.rs
@@ -508,18 +508,32 @@ where
         let mut instructions = vec![];
 
         while !self.allocs.controls.is_empty() {
-            let keep_going =
-                instructions.len() < max_instructions && u.arbitrary().unwrap_or(false);
+            let keep_going = instructions.len() < max_instructions
+                && u.arbitrary().map_or(false, |b: u8| b != 0);
             if !keep_going {
                 self.end_active_control_frames(&mut instructions);
                 break;
             }
 
             gather_options(module, &mut self);
+            debug_assert!(self.allocs.options.len() > 0);
 
-            let f = u.choose(&self.allocs.options)?;
-            let inst = f(u, module, &mut self)?;
-            instructions.push(inst);
+            // `u.choose(...)` can fail for two reasons and in both we swallow the error and close
+            // the function:
+            // - it can fail when no options were provided, but this should be impossible since we
+            //   can always emit a `nop`, e.g. (see `assert` above).
+            // - or it can fail because there is not enough underlying data, so we really cannot
+            //   generate any more instructions.
+            match u.choose(&self.allocs.options) {
+                Ok(f) => {
+                    let inst = f(u, module, &mut self)?;
+                    instructions.push(inst);
+                }
+                Err(_) => {
+                    self.end_active_control_frames(&mut instructions);
+                    break;
+                }
+            }
         }
 
         Ok(instructions)


abrown:
cc: @fitzgen, @alexcrichton: it now seems a lot more likely that the functions generated will have a decent number of instructions:

```
$ for i in {0..100}; do head -c 1000 /dev/urandom | cargo run --features structopt -- -o test.wasm 2>/dev/null | grep 'Instructions generated'; done
Instructions generated: 105
Instructions generated: 120
Instructions generated: 109
Instructions generated: 3
Instructions generated: 111
Instructions generated: 38
Instructions generated: 107
Instructions generated: 113
Instructions generated: 114
Instructions generated: 10
Instructions generated: 123
Instructions generated: 35
Instructions generated: 69
Instructions generated: 51
Instructions generated: 113
Instructions generated: 33
Instructions generated: 1
Instructions generated: 1
Instructions generated: 112
...
```

Previously the number of instructions was in the 0-5 range.

crates/wasm-smith/src/code_builder.rs
```diff
@@ -508,18 +508,23 @@ where
...
-
-            let f = u.choose(&self.allocs.options)?;
-            let inst = f(u, module, &mut self)?;
-            instructions.push(inst);
+            if let Ok(f) = u.choose(&self.allocs.options) {
+                // `u.choose` can fail for two reasons and in both we swallow the error:
+                // - it fails when no options were provided, so we continue looping until we run out
+                //   of random data,
+                // - or it fails because there is not enough underlying data, so we really cannot
+                //   generate an instruction.
```
fitzgen, line 524: Note that the first is impossible in this particular case, because we can always add an `unreachable` or a `nop` instruction, for example.

fitzgen: (And I think it is worth clarifying that in this comment)

crates/wasm-smith/src/code_builder.rs
```diff
@@ -508,18 +508,23 @@ where
...
                 self.end_active_control_frames(&mut instructions);
                 break;
             }
 
             gather_options(module, &mut self);
-
-            let f = u.choose(&self.allocs.options)?;
-            let inst = f(u, module, &mut self)?;
-            instructions.push(inst);
+            if let Ok(f) = u.choose(&self.allocs.options) {
```
fitzgen, line 519: Let's change this from an `if let` into a `match` like

```rust
match u.choose(..) {
    Ok(f) => {
        // existing stuff from if let
    }
    Err(_) => {
        self.end_active_control_frames(&mut instructions);
        break;
    }
}
```

This is functionally the same, since the next iteration of the loop will have `keep_going == false` because the `u.arbitrary()` will return an `Err`, but I think better conveys intent and desired control flow.

fitzgen (approved):
Looks good modulo a couple nitpicks. Thanks Andrew!

cfallin:
Oh, wow, good find! In the abstract a `bool::arbitrary()` is just saying "make a decision based on this bit" and there is no particular probability attached, but with a standard fuzzing driver, it's probably going to spend comparable effort on both sides of the branch, so it seems (backed by your data) that it'd be something like p=0.5 in practice. With p=0.5 the mean length will just be 2 instructions.

I wonder if we should audit for other uses of `bool::arbitrary()` as well, and think about better biases to apply?

Another less-biased approach would be to do `u.int_in_range(...)` to get the desired length; this is what I do over in the fuzzing testcase generator in regalloc2.

(I have been meaning to dig into fuzz coverage after the recent CVE in Cranelift, so this is very useful for me to see; thanks again!)

fitzgen:
> (I have been meaning to dig into fuzz coverage after the recent CVE in Cranelift, so this is very useful for me to see; thanks again!)

Yeah, I think this is probably related to why our `wasm-smith`-based fuzzing didn't find that bug.

crates/wasm-smith/src/code_builder.rs
abrown: Ok, perhaps I can add the comment and `assert` that there are always options available?

fitzgen: Perfect :)

fitzgen (approved):
Thanks!

PR #278, opened 2021-06-04
wasm-encoder: encode SIMD instructions

abrown:
This gives the wasm-encoder crate the ability to encode Wasm SIMD instructions. The encodings used are those published in the spec's [BinarySIMD.md](https://github.com/WebAssembly/simd/blob/main/proposals/simd/BinarySIMD.md) file.

I would be much more confident about the correctness of all this (there could be a mistake in here somewhere) if we could compare the output of wasm-encoder against some other toolchain, e.g. WABT, as is done in the top-level roundrip tests. Unfortunately, there is no way currently to do a `.wat/.wast` to `.wasm` conversion with any of the tools present in this repo. If someone could add that I would like to add some test coverage for this library.

### DIFF
--- a/crates/wasm-encoder/src/code.rs
+++ b/crates/wasm-encoder/src/code.rs
@@ -166,6 +166,9 @@ impl MemArg {
     }
 }
 
+/// Describe an unchecked SIMD lane index.
+pub type Lane = u8;
+
 /// The type for a `block`/`if`/`loop`.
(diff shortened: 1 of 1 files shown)


alexcrichton:
Looks good to me! Ideally `wat` would use `wasm-encoder`, but that's a change I haven't gotten around to implementing yet. That would give this a lot more testing.

PR #279, opened 2021-06-08
Fix README

syrusakbary:
README in `wasm-encoder` has an extra "```" which causes to be rendered improperly


alexcrichton:
Thanks!

PR #280, opened 2021-06-08
wasm-smith: generate SIMD instructions

abrown:
This adds the necessary helper functions to generate SIMD instructions
if the Wasm SIMD proposal is enabled in the configuration (it is enabled
by default).


abrown:
Ok, I think this should be ready for review. I have been using the following script to generate a bunch of errors and see where the failures happen--I only now see errors related to "not enough random data" (evidenced by `error: unable to read uint32_t: magic`), alignment issues (#281), and global init issues (#282):

```
#!/bin/bash
# uncomment to stop at first error: set -e
export RUST_BACKTRACE=1
for i in {0..1000}; do 
    head -c 2000 /dev/urandom > test.random
    cat test.random | cargo run --features structopt -- -o test.wasm 2>/dev/null
    wasm2wat --enable-all test.wasm 1>/dev/null
done
```

abrown:
cc: @fitzgen, @alexcrichton

crates/wasm-smith/src/config.rs
```diff
@@ -232,6 +232,12 @@ pub trait Config: for<'a> Arbitrary<'a> + Default + Clone {
         false
     }
 
+    /// Determines whether the SIMD proposal is enabled for
+    /// generating insructions. Defaults to `true`.
+    fn simd_enabled(&self) -> bool {
+        true
```
alexcrichton, line 238: I think this may want to be false for now while it isn't otherwise enabled by default for other tools? One of the main consumers, wasmtime, also hasn't yet finished the implementation so we would need to explicitly turn this off.

alexcrichton (commented):
Looks reasonable to me!

crates/wasm-smith/src/code_builder.rs
```diff
@@ -3171,3 +3409,528 @@ fn table_index<C: Config>(
...
+
+fn lane_index(u: &mut Unstructured, number_of_lanes: u8) -> Result<u8> {
+    u.int_in_range(0..=(number_of_lanes - 1))
+}
+
+fn arbitrary_v128(u: &mut Unstructured) -> Result<[u8; 16]> {
+    let mut buffer = [0; 16];
+    u.fill_buffer(&mut buffer)?;
+    Ok(buffer)
+}
```
fitzgen, line 3421: [`[u8; 16]` implements `Arbitrary`](https://docs.rs/arbitrary/1.0.1/arbitrary/trait.Arbitrary.html#impl-Arbitrary%3C%27a%3E-for-%5BT%3B%20N%5D), so calls to this function can be replaced with `u.arbitrary()?`

fitzgen (approved):

crates/wasm-smith/src/config.rs
abrown: Sure, changed to `false` in 15fd226. I was assuming it was going to be default-enabled soon enough but either way is fine. But I would like to start trying to find SIMD fuzz bugs in Wasmtime sooner rather than later: is there any way to turn it on and then ignore `unimplemented!` errors?

crates/wasm-smith/src/code_builder.rs
abrown: Yeah, that's better... removed.

abrown:
I don't have permissions to merge...

PR #284, opened 2021-06-09
Add configuration options to `wasm-smith` CLI tool

alexcrichton:
This commit moves the `wasm-smith` CLI tool to the top-level with the
other tools in this repository, and then enhances it with CLI options
for all the configuration variables for a wasm module. Additionally a
`-c` json configuration file is now supported as well for commonly
shared options and less giant CLI command lines.

Additionally a flag has been added to generate a possibly-invalid
module, exposing the `MaybeInvalidModule` type from the CLI which can
help fuzz some more as well.


fitzgen:
> This commit moves the `wasm-smith` CLI tool to the top-level with the
> other tools in this repository,

This means that you won't be able to `cargo install wasm-smith` anymore, which seems like a bit of a bummer.

fitzgen (commented):
Actual code changes look great. Still not convinced that this should move to the top level though.

src/bin/wasm-smith.rs
```diff
@@ -0,0 +1,278 @@
...
+    min_uleb_size: Option<u8>,
+    #[structopt(long = "bulk-memory")]
+    #[serde(rename = "bulk-memory")]
+    bulk_memory_enabled: Option<bool>,
+    #[structopt(long = "reference-types")]
+    #[serde(rename = "reference-types")]
+    reference_types_enabled: Option<bool>,
+    #[structopt(long = "module-linking")]
+    #[serde(rename = "module-linking")]
+    module_linking_enabled: Option<bool>,
```
abrown, line 139: Now that #280 is merged, we could use a `simd_enabled` option here.

alexcrichton:
Technically even `cargo install wasm-smith` wasn't supported because you'd have to do `cargo install wasm-smith --features structopt`, but even despite that it's already true that you don't get `wasm2wat` with `cargo install wasmprinter` or similar for the other tools in this repository. There's nothing really stopping us from publishing the `wasm-tools` top-level crate on crates.io, but it just never seemed that pressing.

I figured that most people probably aren't running `wasm-smith` as a binary anyway except for local testing since it's more generally intended to be integrated as a library. Largely I felt that it seemed best to keep all the tools in this repo in one place (and ideally idiomatic with each other, changing the existing ones to use `structopt` instead of `getopts` one day) rather than scatter to the tools to all their separate packages.

fitzgen:
At minimum, `wasm-smith`'s README will need to be updated: https://github.com/bytecodealliance/wasm-tools/tree/main/crates/wasm-smith#as-a-command-line-tool

alexcrichton:
Oops, indeed!

fitzgen (approved):

PR #285, opened 2021-06-09
Generate less `unreachable` in wasm-smith

alexcrichton:
The goal of this commit is to generate fewer explicit `unreachable`
instructions. This is done via two routes:

* Weights are added to instructions to prefer generating some
  instructions over others. The `unreachable` instruction itself, in
  addition to other uninteresting ones like `nop` and `return`, are
  weighted less than all other instructions to avoid them. The
  `unreachable` instruction is always a valid instruction, but it's not
  the most interesting valid instruction, so the goal here is to
  just generate it less.

* When popping control frames from the stack instead of generating
  `unreachable` instructions this forcibly "fixes up" the frame to have
  the right type signature. There's still a small chance that
  `unreachable` is used to finish the frame but otherwise wasm-smith now
  attempts to use what's on the stack, generating arbitrary values for
  mismatched labels.

This should help `unreachable` get emitted less often and ideally make
code a little less likely to trap. The hope is that produces some
slightly more interesting modules from a runtime perspective.

### DIFF
--- a/crates/wasm-smith/src/code_builder.rs
+++ b/crates/wasm-smith/src/code_builder.rs
@@ -7,7 +7,7 @@ use std::collections::{BTreeMap, BTreeSet};
 macro_rules! instructions {
 	(
         $(
-            ($predicate:expr, $generator_fn:ident),
+            ($predicate:expr, $generator_fn:ident $(, $cost:tt)?),
         )*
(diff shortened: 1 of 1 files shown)


fitzgen (approved):
Nice

PR #286, opened 2021-06-11
wip: allow to specify module's interface in wasm-smith

### LINKED ISSUE #276: wasm-smith: generate a Module with specific imports and exports?
I want to generate a wasm module which I can run in a particular environment. That is, I want to specify the set of imports and exports, instead of them being arbitrary.

Context: I want to throw random wasm at the near blockchain, to make sure that our wasm runtime and host functions do not break. I think this should allow me to catch issues a-la https://github.com/wasmerio/wasmer/issues/2329. In that case, an indirect call to an imported function broke things at runtime, and that feels simple enough to be uncovered by a fuzzer. 

Is such functionality reasonable/desirable? If it is, I would appreciate mentoring instructions. I think, API-wise, we want to add something like `fn imports(&self) -> Option<&[Import]>` to the `Config` trait, but I am not familiar with wasm-tools enough to understand what would be the best definition for such an `Import` type.

matklad:
fix #276


matklad:
@fitzgen this turned out to be a bit more compliaceted than I though, for two related reasons.

_First_, when we generate arbitrary imports, we want to pick a type from an already existing set. So, I solved this problem by moving the bulk of the logic to the point where we generate arbitrary types, such that, at the point where we generate the import, we can basically re-use the same logic, and just pick a type from a more restricted set

_Second_, the `EntityType` enum relies on some kind on instances already being in the module. For example, `EntityType::Func` contains an index into a table. The problem here is that in Config we don't have access to those tables, as they don't exist yet. Not sure what's the best path here -- should be just duplicate the `EntityType` hierarchy for config, such that config-flagored version do not rely on indexes? 

The PR in the current state implements the minimal amount of work to make just functions work.

fitzgen:
Regarding the first point: I think we generally want to avoid mixing up the code paths for arbitrary imports vs available imports together too much. I think it will make everything much harder to reason about. To be specific, when we are generating imports from an available set, rather than limiting ourselves to things for which we already have a compatible type definition, we can choose an available import, add its type definition if it is not already defined, and then add the import referencing the maybe-just-defined type. Having all this in a separate code path/method from arbitrary imports should make implementation and maintenance easier for both paths.

Regarding the second point:

> should be just duplicate the `EntityType` hierarchy for config, such that config-flagored version do not rely on indexes?

I think the answer is "yes" here. It is annoying to type out, but that only has to happen once, and we don't expect these definitions to change frequently, so the maintenance/syncing overhead is low. This approach will provide the best user experience for folks writing custom `Config`s with available imports.

matklad:
This all makes sense to me, @fitzgen! Ptal, is this closer to what you have in mind? Want to get :+1: on the design before I get into modules and instances :-) 

One specific question:

* should I remove the import from the set of choices after it is used?
* or should I keep it in (allowing dupes, as per wasm spec), and also keep the cell-based caching of import types (I only create a new type when I use the import for the first type)

crates/wasm-smith/src/lib.rs
```diff
@@ -1373,6 +1390,111 @@ where
...
+                                break;
+                            }
+                        }
+                    }
+                    match idx.get() {
+                        Some(idx) => {
+                            let ty = self.func_type(idx).clone();
+                            EntityType::Func(idx, ty)
+                        }
+                        None => return Ok(false),
```
fitzgen, line 1456: In this case we can create a type section and push it onto `self.initial_sections`.

fitzgen (commented):
This is looking great!

> should I remove the import from the set of choices after it is used?

If module linking is enabled (checkable on the config), then you'll have to, because module linking disallows duplicate imports.

Additionally, with module linking, a series of two-layer imports implicitly create a one-layer import of an instance:

```wat
(import "env" "x" ...)
(import "env" "y" ...)

;; ==>

(import "env" (instance (export "x" ...) (export "y" ...)))
```

But this means that the following is invalid:

```wat
(import "env" "x" ...)
(import "blah" ...)
(import "env" "y" ...)
```

because the `"env"` imports aren't back-to-back and so it desugars into multiple conflicting imports of an `"env"` instance.

So after you add an import with a different module name, you need to remove any imports from the previous module name from the available set.

If module linking is not enabled, then the imports may be reused (although we can leave this as a TODO for now, if you prefer, so that you don't have to implement both code paths right now).

matklad:
Making some progress here! it seems to work without module linking, and module linking currently fails with a bunch of validation errors ("aliased name `g` does not exist in instance" and such), today's quant run out before i figured what's an issue there.

nagisa:
This can be closed now.

PR #287, opened 2021-06-12
Support inline import in event declarations

takikawa:
This adds support for inline `import` for `event` declarations, similar to how it works for `global` and other forms.

(the motivation for this is that some of the proposed spec tests for exception handling uses this syntax, so it's useful for test harnesses in, e.g., SpiderMonkey)


alexcrichton:
Thanks!

PR #288, opened 2021-06-14
Remove `unwind` instruction for exception handling

takikawa:
The `unwind` instruction was recently removed from the exception handling proposal spec (https://github.com/WebAssembly/exception-handling/pull/156), so this PR is for removing it from wasm-tools.

I plan to also make this change in wabt soon, but until then I added all unwind tests to the skipped test list in this PR.


alexcrichton:
Thanks!

PR #289, opened 2021-06-14
Allow bare `try` blocks for exceptions

takikawa:
This PR add support for bare `try` blocks (try blocks that have no `catch`, `delegate`, etc). These are equivalent to a plain `block` and were added to the spec recently (in order to simplify the grammar and avoid special cases): https://github.com/WebAssembly/exception-handling/pull/157

There's a PR pending review for wabt as well: https://github.com/WebAssembly/wabt/pull/1676

I tested the wasm-tools patch against the pending wabt PR and I believe they should agree. It should be ok to merge this before wabt as the only conflicting test is a failure test (`bad-try-no-catch.txt`) that is not run by `roundtrip.rs`.

### DIFF
--- a/crates/wasmparser/src/operators_validator.rs
+++ b/crates/wasmparser/src/operators_validator.rs
@@ -677,9 +677,6 @@ impl OperatorValidator {
                         self.push_ctrl(FrameKind::Else, frame.block_type, resources)?;
                         frame = self.pop_ctrl(resources)?;
                     }
-                    FrameKind::Try => {
-                        bail_op_err!("expected catch block");
-                    }
                     _ => (),
                 }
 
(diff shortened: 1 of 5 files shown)


takikawa:
> I tested the wasm-tools patch against the pending wabt PR and I believe they should agree. It should be ok to merge this before wabt as the only conflicting test is a failure test (bad-try-no-catch.txt) that is not run by roundtrip.rs.

Whoops, this was not quite right as `local/try.wat` with the new case added makes wabt (submodule revision) and wasm-tools disagree. I'll try adjusting the skipped tests.

alexcrichton:
Thanks for this!

If you'd like a new publish with these changes and updates just let me know as well.

takikawa:
@alexcrichton Oh yes, thanks that would be helpful! :) This is the last PR I have for this current batch of changes, so a version bump and publish would be helpful to be able to use this from SpiderMonkey tests.

alexcrichton:
Ok, published now!

PR #291, opened 2021-06-22
Handle the rename of "events" to "tags" in exception handling proposal

takikawa:
The exception handling proposal spec recently renamed "events" to "tags" in order to use a clearer name (https://github.com/WebAssembly/exception-handling/pull/161). This means the event section is now the tag section and so on. This PR handles the renaming for wasm-tools (in the last commit).

This change has already landed in wabt (https://github.com/WebAssembly/wabt/pull/1678) so this PR also updates the wabt test submodule in the first commit.

Updating the submodule also requires some changes to a linking test because of some wabt printing changes (https://github.com/WebAssembly/wabt/pull/1669/files). This is in the second commit.


alexcrichton:
Thanks!

PR #296, opened 2021-06-29
Update wabt submodule and fix try-delegate label resolution to match

takikawa:
This PR updates the wabt submodule to the latest and handles two updates:

  * Fixes `try-delegate` label resolution (the companion to this wabt PR https://github.com/WebAssembly/wabt/pull/1675) so that labels are correctly looked up starting one scope out (i.e., one more than a `br` would target).
  * Removes now outdated test exemptions for exception handling cases where wabt and wasm-tools were out of sync.


alexcrichton:
👍

PR #298, opened 2021-06-30
Finish implementing the extended name section proposal

alexcrichton:
This PR is a continuation of https://github.com/bytecodealliance/wasm-tools/pull/265 and finishes implementing the current state of the [extended name section proposal](https://github.com/WebAssembly/extended-name-section). The changes here were to plumb the custom names throughout `wasmprinter` more thoroughly as well as updating the `wast` crate to encode all of the new names.

In the course of doing this I ended up disabling all tests that compare against the textual output of wabt's `wasm2wat`. Unfortunately it looks like wabt's implementation of the extended name section proposal is somewhat spotty in that the custom names are printed in some places and not others. I don't know why this is and honestly I've spent so much time tracking down issue in wasm2wat that I'm sort of tired of doing this. I think it's best for now to just flat out disable these tests and defer it to later to figure out when to re-enable.


fitzgen (approved):
:+1:

PR #300, opened 2021-07-30
Sync with the upstream memory64 proposal

alexcrichton:
This commit updates the support for the memory64 proposal,
printing/parsing/validation, to match the latest upstream spec. I've
double-checked all the type-checking and such, and the only piece that
this crate was missing was parsing 64-bit offsets in `memarg`
structures. I've threaded through various bits and pieces to parse
32-bit offsets primarily but optionally allow 64-bit offsets when the
memory64 proposal is enabled.

I've also done some updates to the test suite so we can run all the
`*.wast` tests (or at least most) in the upstream wasm64 proposal.

### DIFF
--- a/crates/wasmparser/src/binary_reader.rs
+++ b/crates/wasmparser/src/binary_reader.rs
@@ -21,9 +21,8 @@ use crate::limits::*;
 
 use crate::primitives::{
     BinaryReaderError, BrTable, CustomSectionKind, ExternalKind, FuncType, GlobalType, Ieee32,
-    Ieee64, LinkingType, MemoryImmediate, MemoryType, NameType, Operator, RelocType,
-    ResizableLimits, ResizableLimits64, Result, SIMDLaneIndex, SectionCode, TableType, Type,
-    TypeOrFuncType, V128,
+    Ieee64, LinkingType, MemoryImmediate, MemoryType, NameType, Operator, RelocType, Result,
+    SIMDLaneIndex, SectionCode, TableType, Type, TypeOrFuncType, V128,
 };
 use crate::{ExportType, Import, ImportSectionEntryType, InstanceType, ModuleType, TagType};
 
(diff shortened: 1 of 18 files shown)


crates/wasmparser/src/binary_reader.rs
```diff
@@ -354,20 +361,26 @@ impl<'a> BinaryReader<'a> {
...
+            // here. That seems a bit odd to me at the time of this writing so
+            // I've taken the liberty of reading a 64-bit limits field in those
+            // situations. I suspect that this is a typo in the spec, but if not
+            // we'll need to update this to read a 32-bit limits field when the
+            // shared flag is set.
+            let limits = self.read_resizable_limits64(has_max)?;
             Ok(MemoryType::M64 { limits, shared })
+        } else {
+            let limits = self.read_resizable_limits(has_max)?;
+            Ok(MemoryType::M32 { limits, shared })
```
fitzgen, line 383: It seems strange, API-wise, that we have `enum { M32, M64 }` for `MemoryType` but just have `u64`s inside `MemArg` instead of also having a 32 vs 64 `enum`.

In my experience, the `enum { M32, M64 }` pattern is quite annoying to work with. What do you think of switching `MemoryType` over into a `struct` with a `is_memory64: bool` field and always have 64-bit limits?

Not 100% convinced this is a good idea, but it would be nice to unify `MemoryType` and `MemArg`'s approach to dealing with memory64, and I'm not in love with the idea of making `MemArg` into an `enum` either.

crates/wasmparser/src/primitives.rs
```diff
@@ -257,7 +257,7 @@ pub enum ImportSectionEntryType {
 pub struct MemoryImmediate {
     /// Alignment, stored as `n` where the actual alignment is `2^n`
     pub align: u8,
-    pub offset: u32,
+    pub offset: u64,
```
fitzgen, line 260: At minimum, this field should have a comment documenting memory32 vs memory64 and the guarantee that you'll only get offsets that fit in a `u32` when 64 bit memories aren't enabled.

crates/wasmparser/src/binary_reader.rs
alexcrichton: Nah makes sense and I agree that the `M32`/`M64` business was generally just a pain to deal with.

alexcrichton:
Ok updated! Mind taking another look to confirm this looks as expected @fitzgen?

fitzgen (approved):
Thanks!

PR #301, opened 2021-08-02
Mark `primitives::Operator` as `non_exhaustive`

uint:
This feels right for `primitives::Operator` given how new ones get added over time.

[What `non_exhaustive` does is described here.](https://rust-lang.github.io/rfcs/2008-non-exhaustive.html)

This is a breaking change.


alexcrichton:
Thanks for the PR! I've considered this in the past but I've personally come to the conclusion that this isn't a right fit for this crate. I think that there's value, in consumers of wasm, to know that they've matched everything and don't need a catch-all "unknown instruction". 

It does mean that there's more "breaking changes" from this crate, but since we implement proposals pretty early that's relatively common anyway.

uint:
Ah, alright, fair enough. I probably should have submitted an issue asking about this first. Closing!

PR #306, opened 2021-08-05
Stop being byte-for-byte compatible with wabt

alexcrichton:
This commit removes all reference testing against wabt in the
`roundtrip.rs` suite. Historically this project has attempted to be
byte-for-byte compatible with upstream `wabt`, but over time it's
accrued so many workarounds in both the implementation and while testing
that I don't think that it's worth it any more.

The original motivation for this change is that this module does not
appropriately roundtrip in the Rust tooling:

    (module
      (elem funcref))

The reason for this is that a hack in the implementation was changing
the `funcref` representation to `func`, which means it's round-trip as:

    (module
      (elem func))

and this was found by the fuzzers last night when reference types was
enabled for round-tripping for the first time.

Overall I don't think we have much to gain any more for byte-for-byte
compatibility with wabt. There's a number of long-open issues which
require workarounds in this crate. I think at this point this tooling is
solid enough that comparisons can be made but it isn't necessary to do
so in an automated fashion. This should make it much easier to implement
new features and such because there doesn't need to be so many
workarounds in the test suite for comparing against wabt.

### DIFF
--- a/crates/wast/src/binary.rs
+++ b/crates/wast/src/binary.rs
@@ -644,29 +644,7 @@ impl Encode for ExportKind {
 
 impl Encode for Elem<'_> {
     fn encode(&self, e: &mut Vec<u8>) {
-        // Try to switch element expressions to indices if we can which uses a
-        // more MVP-compatible encoding.
-        //
-        // FIXME(WebAssembly/wabt#1447) ideally we wouldn't do this so we could
-        // be faithful to the original format.
-        let indices;
-        let mut to_encode = &self.payload;
-        if let ElemPayload::Exprs {
-            ty:
-                RefType {
-                    nullable: true,
-                    heap: HeapType::Func,
-                },
-            exprs,
-        } = &to_encode
-        {
-            if let Some(list) = extract_indices(exprs) {
-                indices = ElemPayload::Indices(list);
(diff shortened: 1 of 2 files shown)


fitzgen:
> ```
> (module
>   (elem func))
> ```

Is this even valid? Tables can only hold reference types, and therefore I would assume the same for `elem`s.

Either way, it seems like we should be able to round trip `(elem funcref)`. And if we can do that, does it make sense to merge the rest of this PR? I feel like I'm missing something here...

alexcrichton:
Oh the `elem func` is different `elem funcref` only in syntax, not in semantics where it's storing funcrefs. It's alternative ways to specify either `elem func 0 1 2 3` or `elem funcref (ref.func 0) (ref.func 1) ...` which have different binary encodings.

fitzgen:
Ah okay so we are just doing a little bit of canonicalization that WABT isn't. Gotcha.

fitzgen (approved):

alexcrichton:
Er rather byte-for-byte compatibility with wabt was requiring us to do a form of canonicalization, but this breaks round-tripping and new fuzzing of reference types, so this PR is removing the byte-for-byte compatibility which means we can remove the canonicalization as well

PR #313, opened 2021-08-06
Don't have wasm-smith be entirely generic

alexcrichton:
Working on fuzzers using wasm-smith takes forever because they take so
long to compile. I believe the primary reason for this is that the
entire wasm-smith crate is monomorphized into each individual fuzzer due
to the usage of generics on the `Config` trait. This means each fuzzer
instantiates this entire crate, and the wasm-smith crate is no small
chunk of change.

This commit moves the configuration to being a trait object instead of a
generic, which should not lose that much in runtime and should allow
fuzzers to benefit from separate compilation of wasm-smith itself. If
this becomes a performance bottleneck for whatever reason we can also
remove the trait and simply have struct-based configuration too.

### DIFF
--- a/crates/wasm-smith/src/code_builder.rs
+++ b/crates/wasm-smith/src/code_builder.rs
@@ -1,6 +1,4 @@
-use super::{
-    BlockType, Config, ConfiguredModule, Elements, FuncType, Instruction, MemArg, ValType,
-};
+use super::{BlockType, Elements, FuncType, Instruction, MemArg, Module, ValType};
 use arbitrary::{Result, Unstructured};
 use std::collections::{BTreeMap, BTreeSet};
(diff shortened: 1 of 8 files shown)


fitzgen (approved):
Nice :+1: 

Did you measure compilation time speed ups, by chance?

alexcrichton:
Heh good point I should measure!

I'm mostly concerned about incremental builds here and in release mode (where fuzzers primarily run). 

Before this PR I built the fuzzers, touched `fuzz/src/lib.rs` in this repo, and rebuilding the fuzzers took 42s. After this PR I did the same thing and the build took 6s. A nice improvement!

fitzgen:
Very nice!!!

PR #314, opened 2021-08-09
wasm-smith: Generate fail-to-instantiate modules far less frequently

alexcrichton:
I ran a small program recently where I threw random data at wasm-smith and then tried to instantiate each module. I kept record of statistics about whether or not the module failed to instantiate, and it turns out with a 1GB memory limit and using `wasmtime` we previously failed to instantiate 50%+ modules coming out of wasm-smith. The goal of this commit is to reduce that percentage.

Overall the goal here is to in theory get "more interesting" modules into fuzzers when generated by wasm-smith. The modules that are all failing to instantiate are indeed interesting, but it's mostly more interesting to  get to wasm code rather than constantly checking the bounds-checks of instantiation on memory segments and such.

This commit reimplements limits around memories/tables as well as the offsets of active segments. The goal of this PR is to still leave all the currently-generated modules as *possible* to generate, but just far-less likely. Some compile-time constants have been introduced which can be used to tune the generation of modules and how frequently they should allocate lots of memory and/or trap.

After this commit the paramter values I've chosen means that ~2% of modules will fail instantiation due to exceeding memory limits and/or offsets being out-of-bounds. I figure for now that that's at least a much better starting point for exercising more bits of the wasm runtime than before!

### DIFF
--- a/crates/fuzz-stats/src/bin/failed-instantiations.rs
+++ b/crates/fuzz-stats/src/bin/failed-instantiations.rs
@@ -1,0 +1,207 @@
+use arbitrary::{Arbitrary, Error, Unstructured};
+use rand::RngCore;
+use std::sync::atomic::{AtomicIsize, AtomicUsize, Ordering::SeqCst};
+use std::sync::Arc;
+use wasm_smith::SwarmConfig;
+use wasmtime::*;
+
+struct State {
+    engine: Engine,
+    print: bool,
+    remaining: AtomicIsize,
+    total: AtomicUsize,
+    instantiate_trap: AtomicUsize,
+    instantiate_oom: AtomicUsize,
+}
+
+fn main() {
+    Arc::new(State::new()).run();
+}
+
+// Theoretically this test can fail because it's based on random data. In
+// practice it's expected that the "fails to instantiate" rate is <2%, so if we
+// cross the 10% threshold that's quite bad.
+#[test]
+fn under_10_percent() {
+    let mut state = State::new();
+    state.print = false;
+    state.remaining.store(1000, SeqCst);
+    let state = Arc::new(state);
+    state.run();
+
+    let total = state.total.load(SeqCst);
+    let bad = state.instantiate_trap.load(SeqCst) + state.instantiate_oom.load(SeqCst);
+    assert!(
+        bad < total / 10,
+        "{} modules failed to instantiate out of {}, this failure rate is too high",
+        bad,
+        total
+    );
+}
+
+impl State {
+    fn new() -> State {
+        let mut config = Config::new();
+        config.wasm_multi_memory(true);
+        State {
+            engine: Engine::new(&config).unwrap(),
+            print: true,
+            total: AtomicUsize::new(0),
+            remaining: AtomicIsize::new(isize::max_value()),
+            instantiate_trap: AtomicUsize::new(0),
+            instantiate_oom: AtomicUsize::new(0),
+        }
+    }
+
+    fn run(self: &Arc<Self>) {
+        let threads = (0..num_cpus::get())
+            .map(|_| {
+                let state = self.clone();
+                std::thread::spawn(move || state.run_worker())
+            })
+            .collect::<Vec<_>>();
+        for thread in threads {
+            thread.join().unwrap();
+        }
+    }
+
+    fn run_worker(&self) {
+        let mut rng = rand::thread_rng();
+        let mut data = Vec::new();
+
+        while self.remaining.fetch_sub(1, SeqCst) >= 0 {
+            data.truncate(0);
+            data.resize(1024, 0);
+            rng.fill_bytes(&mut data);
+            loop {
+                match self.run_once(&mut data) {
+                    Ok(()) => break,
+                    Err(Error::NotEnoughData) => {
+                        let cur = data.len();
+                        let extra = 1024;
+                        data.resize(cur + extra, 0);
+                        rng.fill_bytes(&mut data[cur..]);
+                    }
+                    Err(e) => panic!("failed to generated module: {}", e),
+                }
+            }
+        }
+    }
+
+    /// Generates a random modules using `data`, and then attempts to
+    /// instantiate it.
+    ///
+    /// Records when instantiation fails and why it fails.
+    fn run_once(&self, data: &[u8]) -> Result<(), Error> {
+        let mut u = Unstructured::new(data);
+        // Here `SwarmConfig` is used to get hopefully a bit more coverage of
+        // interesting states, and we also forcibly disable all `start`
+        // functions for now. Not much work has gone into minimizing the traps
+        // generated from wasm functions themselves, and this shouldn't be
+        // enabled until that's been worked on.
+        let mut config = SwarmConfig::arbitrary(&mut u)?;
+        config.allow_start_export = false;
+        let mut wasm = wasm_smith::Module::new(config, &mut u)?;
+        wasm.ensure_termination(10_000);
+        let wasm = wasm.to_bytes();
+
+        // We install a resource limiter in the store which limits the store to
+        // 1gb of memory. That's half the default allocation of memory for
+        // libfuzzer-based fuzzers by default, and ideally we're not in a
+        // situation where most of the modules are above this threshold.
+        let module = Module::new(&self.engine, &wasm).expect("failed to compile module");
+        let mut store = Store::new(
+            &self.engine,
+            StoreLimits {
+                remaining_memory: 1 << 30,
+                oom: false,
+            },
+        );
+        store.limiter(|s| s as &mut dyn ResourceLimiter);
+
+        // Synthesize dummy imports based on what the module asked for, and then
+        // instantiate!
+        let instance = fuzz_stats::dummy::dummy_imports(&mut store, &module)
+            .and_then(|imports| Instance::new(&mut store, &module, &imports));
+
+        match instance {
+            // If instantiation succeeded, we're not too interested in anything
+            // else right now. In the future we should probably run exported
+            // functions and record whether a trap happened or not.
+            Ok(_i) => {}
+
+            Err(e) => {
+                // Traps are ok if they happen during instantiation. This is an
+                // expected occurrence we want to account for.
+                if e.downcast_ref::<Trap>().is_some() {
+                    // std::fs::write("trap.wasm", &wasm).unwrap();
+                    self.instantiate_trap.fetch_add(1, SeqCst);
+
+                // Ooms, like traps, are normal during instantiations. This
+                // can happen, for example, if a defined memory is very large.
+                } else if store.data().oom {
+                    self.instantiate_oom.fetch_add(1, SeqCst);
+
+                // In theory nothing else fails to instantiate. If it does, then
+                // panic.
+                } else {
+                    std::fs::write("panic.wasm", &wasm).unwrap();
+                    panic!("unknown: {}", e);
+                }
+            }
+        }
+
+        let prev_total = self.total.fetch_add(1, SeqCst);
+        if prev_total % 10_000 == 0 && self.print {
+            self.print(prev_total + 1);
+        }
+
+        Ok(())
+    }
+
+    /// Prints summary statistics of how many modules have been instantiated so
+    /// far and how many of them have oom'd or trap'd.
+    fn print(&self, total: usize) {
+        print!("total: {:8}", total);
+        let stat = |name: &str, stat: &AtomicUsize| {
+            let stat = stat.load(SeqCst);
+            if stat > 0 {
+                print!(" {} {:5.02}% ", name, (stat as f64) / (total as f64) * 100.);
+            }
+        };
+        stat("i-oom", &self.instantiate_oom);
+        stat("i-trap", &self.instantiate_trap);
+        println!();
+    }
+}
+
+struct StoreLimits {
+    remaining_memory: usize,
+    oom: bool,
+}
+
+impl StoreLimits {
+    fn alloc(&mut self, amt: usize) -> bool {
+        match self.remaining_memory.checked_sub(amt) {
+            Some(mem) => {
+                self.remaining_memory = mem;
+                true
+            }
+            None => {
+                self.oom = true;
+                false
+            }
+        }
+    }
+}
+
+impl ResourceLimiter for StoreLimits {
+    fn memory_growing(&mut self, current: u32, desired: u32, _maximum: Option<u32>) -> bool {
+        self.alloc((desired - current) as usize * 64 * 1024)
+    }
+
+    fn table_growing(&mut self, current: u32, desired: u32, _maximum: Option<u32>) -> bool {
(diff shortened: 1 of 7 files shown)


crates/wasm-smith/src/lib.rs
```diff
@@ -2255,20 +2249,46 @@ impl Module {
...
+    let mut cur_max = min.saturating_add(10);
+    loop {
+        if cur_max >= max {
+            break u.int_in_range(min..=max);
+        }
+        if u.int_in_range(0..=4)? == 0 {
+            break u.int_in_range(min..=cur_max);
+        }
+        cur_max += cur_max / 2;
+    }
```
fitzgen: Instead of this adhoc gradual growth, lets use a cubic easing function:

```rust
/// Map a value from within the input range linearly to the output range.
///
/// For example, mapping `0.5` from the input range `0.0..1.0` to the output
/// range `1.0..3.0` produces `2.0`.
fn map_linear(
    value: f64,
    Range {
        start: in_low,
        end: in_high,
    }: Range<f64>,
    Range {
        start: out_low,
        end: out_high,
    }: Range<f64>,
) -> f64 {
    assert!(!value.is_nan(), "{}", value);
    assert!(value.is_finite(), "{}", value);
    assert!(in_low < in_high, "{} < {}", in_low, in_high);
    assert!(out_low < out_high, "{} < {}", out_low, out_high);
    assert!(value >= in_low, "{} >= {}", value, in_low);
    assert!(value <= in_high, "{} <= {}", value, in_high);

    let dividend = out_high - out_low;
    let divisor = in_high - in_low;
    let slope = dividend / divisor;
    let result = out_low + (slope * (value - in_low));

    assert!(result >= out_low, "{} >= {}", result, out_low);
    assert!(result <= out_high, "{} <= {}", result, out_high);
    result
}

/// Map a value from within the input range cubically to the output range.
///
/// For example, mapping `0.333333` from the input range `0.0..1.0` to the
/// output range `0.0..100.0` will result in approximately `3.7`.
fn map_cubed(
    value: f64,
    Range {
        start: in_low,
        end: in_high,
    }: Range<f64>,
    Range {
        start: out_low,
        end: out_high,
    }: Range<f64>,
) -> f64 {
    assert!(!value.is_nan(), "{}", value);
    assert!(value.is_finite(), "{}", value);
    assert!(in_low < in_high, "{} < {}", in_low, in_high);
    assert!(out_low < out_high, "{} < {}", out_low, out_high);
    assert!(value >= in_low, "{} >= {}", value, in_low);
    assert!(value <= in_high, "{} <= {}", value, in_high);

    let x = map_linear(value, in_low..in_high, 0.0..1.0);
    let unscaled = x * x * x;
    let result = map_linear(unscaled, 0.0..1.0, out_low..out_high);

    assert!(result >= out_low, "{} >= {}", result, out_low);
    assert!(result <= out_high, "{} <= {}", result, out_high);
    result
}

fn gradually_grow(u: &mut Unstructured, min: u64, max: u64) -> Result<u64> {
    let min = f64::try_from(min).unwrap();
    let max = f64::try_from(max).unwrap();
    let x = u.arbitrary::<u32>()?;
    let x = f64::from(x);
    let x = map_cubed(x, f64::from(u32::MIN)..f64::from(u32::MAX), min..max);
    Ok(u64::try_from(x.round()).unwrap())
}
```

Given uniform inputs in the range 0..1000, this results in the following output distribution:

```
=== Inputs ===
# Number of samples = 16384
# Min = 0
# Max = 1000
#
# Mean = 502.8545532226551
# Standard deviation = 290.2145333477139
# Variance = 84224.47536623133
#
# Each ∎ is a count of 17
#
   0 ..   50 [ 834 ]: ∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  50 ..  100 [ 834 ]: ∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
 100 ..  150 [ 772 ]: ∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
 150 ..  200 [ 822 ]: ∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
 200 ..  250 [ 761 ]: ∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
 250 ..  300 [ 819 ]: ∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
 300 ..  350 [ 841 ]: ∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
 350 ..  400 [ 823 ]: ∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
 400 ..  450 [ 821 ]: ∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
 450 ..  500 [ 783 ]: ∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
 500 ..  550 [ 818 ]: ∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
 550 ..  600 [ 812 ]: ∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
 600 ..  650 [ 826 ]: ∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
 650 ..  700 [ 802 ]: ∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
 700 ..  750 [ 832 ]: ∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
 750 ..  800 [ 819 ]: ∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
 800 ..  850 [ 805 ]: ∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
 850 ..  900 [ 842 ]: ∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
 900 ..  950 [ 840 ]: ∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
 950 .. 1000 [ 878 ]: ∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎

=== Outputs ===
# Number of samples = 16384
# Min = 0
# Max = 1000
#
# Mean = 253.90454101562332
# Standard deviation = 286.36370736253974
# Variance = 82004.1728944183
#
# Each ∎ is a count of 119
#
   0 ..   50 [ 5967 ]: ∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  50 ..  100 [ 1552 ]: ∎∎∎∎∎∎∎∎∎∎∎∎∎
 100 ..  150 [ 1092 ]: ∎∎∎∎∎∎∎∎∎
 150 ..  200 [  885 ]: ∎∎∎∎∎∎∎
 200 ..  250 [  736 ]: ∎∎∎∎∎∎
 250 ..  300 [  666 ]: ∎∎∎∎∎
 300 ..  350 [  544 ]: ∎∎∎∎
 350 ..  400 [  548 ]: ∎∎∎∎
 400 ..  450 [  478 ]: ∎∎∎∎
 450 ..  500 [  456 ]: ∎∎∎
 500 ..  550 [  419 ]: ∎∎∎
 550 ..  600 [  388 ]: ∎∎∎
 600 ..  650 [  375 ]: ∎∎∎
 650 ..  700 [  350 ]: ∎∎
 700 ..  750 [  359 ]: ∎∎∎
 750 ..  800 [  346 ]: ∎∎
 800 ..  850 [  306 ]: ∎∎
 850 ..  900 [  294 ]: ∎∎
 900 ..  950 [  341 ]: ∎∎
 950 .. 1000 [  282 ]: ∎∎
```

This should give us roughly our desired distribution shape without consuming any additional bytes from the `Unstructured`.

fitzgen (approved):
Can you add your script that reads random data, tries to instantiate it with wasmtime, and counts successful instantiation rate as an example script in this repo? That will be really useful in the future when tweaking constants and such.

crates/wasm-smith/src/lib.rs
alexcrichton: I'm glad at least one of us is aware of how to actually do this correctly.

alexcrichton: Ok so I implemented this and am getting some interesting results. My goal was to have oom-produced modules happen maybe 1-2% of the time, but doing the cubic thing here cases oom ~30% of the time still. In retrospect this makes sense. I'm considering >1gb to be oom which is 1/4 of the way through the range from the minimum to the maximum of 32-bit memories. Turns out the cube root of 1/4 is ~0.63, which roughly equates to 35% ish of the random choices (in the range of 0-1) being in the oom area.

Thinking more about this though I like the idea of what's going on here, basically generating a random float between 0 and 1 and then graphing some curve from 0-1, using the resulting value to scale up to the range we actually are interested in. I don't think y=x^3 is what we want to do here because at least for memories and oom that still biases 1gb+ memories too much.

According to wolfram alpha if we want 98% of the memories to be in-bounds (e.g. less than 1/4 in the 0-1 range) then the solution to `0.98^n = 1/4` is `n ~= 69`. While possible that seems a little silly to do a 69-th power of a function just to get the desired range.

I wonder if we can still do something similar here? What do you think of doing something like:

* Generate a 32-bit number
* Map it to 0-1
* If the number is less than 0.99, then generate a guaranteed-in-bounds size. This would be new parameters to the `gradually_grow` function where we'd just linearly map 0-0.99 into the output range known to be guaranteed inbounds.
* If the number is 0.99 or above we map the 0.99 to 1 range to the output range which is always out of bounds.

This would largely amount to basically having a function that looks like:

```
f(x) = scale(0..1, x^3, 0..MIN)       if x < 0.99
     = scale(0..1, x, MIN..MAX)     if x >= 0.99
```

where `MIN` is the maximum value that's still in-bounds and `MAX` is the absolute maximum value that we're allowed to generate. This is not a continuous function like x^3 (or x^69) but I think would suffice for our purposes, give us a parameter to tweak (the 0.99 threshold and the MIN), and still consume only the 32-bit integer from the input data?

alexcrichton: Ok I went ahead and pushed up what I was thinking

crates/wasm-smith/src/lib.rs
```diff
@@ -2321,26 +2332,100 @@ impl Module {
...
+        assert!(value.is_finite(), "{}", value);
+        assert!(input.start < input.end, "{} < {}", input.start, input.end);
+        assert!(
+            output.start < output.end,
+            "{} < {}",
+            output.start,
+            output.end
+        );
+        assert!(value >= input.start, "{} >= {}", value, input.start);
+        assert!(value <= input.end, "{} <= {}", value, input.end);
```
fitzgen, line 2380: Maybe also assert that `output.start <= output_inbounds.start` and `output_inbounds.end <= output.end`?

fitzgen (approved):
Sad that we aren't seizing this opportunity for a legendary "nice" comment, but this is the better approach for `wasm-smith` itself. Thanks!

crates/fuzz-stats/src/bin/failed-instantiations.rs
```diff
@@ -134,12 +134,13 @@ impl State {
...
                 // expected occurrence we want to account for.
                 if e.downcast_ref::<Trap>().is_some() {
-                    // std::fs::write("trap.wasm", &wasm).unwrap();
+                    std::fs::write("trap.wasm", &wasm).unwrap();
                     self.instantiate_trap.fetch_add(1, SeqCst);
 
                 // Ooms, like traps, are normal during instantiations. This
                 // can happen, for example, if a defined memory is very large.
                 } else if store.data().oom {
+                    std::fs::write("oom.wasm", &wasm).unwrap();
```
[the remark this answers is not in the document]
alexcrichton: Yeah when I've inspected these they've occasionally been corrupted, but I just do it again and it's likely to not be corrupted. It's also possible in local testing to use only one thread as well to guarantee no races. I didn't want to spray wasm files everywhere though and this worked well enough for me.

PR #319, opened 2021-08-16
Implement a nan canonicalization mode for wasm-smith

alexcrichton:
This commit implements a configuration option for `wasm-smith` where the
generated wasm module will internally do NaN canonicalization to ensure
that the wasm module will produce the exact same result in multiple wasm
runtimes. This should help close one of the gaps of nondeterminism in
wasm, NaN values, to ensure that differential fuzzing between runtimes
works correctly.

This was discovered fuzzing the Wasmtime runtime against the wasm OCaml
spec interpreter where OCaml wasn't canonicalizing NaN but Wasmtime was.
To help assuage any possible difference in NaN representations and
implementation choices in Wasmtime this solution should help ensure
that, when enabled for differential fuzzing, we'll get the same NaN
results in both runtimes.

### DIFF
--- a/crates/wasm-smith/src/code_builder.rs
+++ b/crates/wasm-smith/src/code_builder.rs
@@ -563,8 +563,13 @@ pub(crate) struct CodeBuilderAllocations {
 
 pub(crate) struct CodeBuilder<'a> {
     func_ty: &'a FuncType,
-    locals: &'a Vec<ValType>,
+    locals: &'a mut Vec<ValType>,
     allocs: &'a mut CodeBuilderAllocations,
+
+    // Temporary locals injected and used by nan canonicalization.
+    f32_scratch: Option<usize>,
+    f64_scratch: Option<usize>,
+    v128_scratch: Option<usize>,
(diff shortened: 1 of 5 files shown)


cfallin (approved):
LGTM; thanks for putting this together!

I didn't double-check the opcode list against the latest (e.g. new SIMD opcodes); it's a little suboptimal to duplicate this list here and things would be better IMHO if our opcode-related utilities had a way of providing input/output types, but for now if we missed one (or miss a new one in the future) the worst that happens is that we get fuzzbugs from non-canonicalized NaNs on some opcodes, so I think it's probably OK to have this duplication.

alexcrichton:
Yeah FWIW we have a duplicated list of all wasm opcodes in... 3 places I think now? We've got `wasmparser`, `wast`, and `wasm-encoder`, all of which have different desires for various signatures and such. I'm hoping, though, that you're right in that it'll be not too bad to add things to list list over time as we find them necessary. This is at least the framework to make future additions one-liners.

PR #321, opened 2021-08-18
Optimize name selection in wasmprinter

alexcrichton:
This commit updates how names are chosen for items in wasmprinter.
Printing wasm has the problem where we want to provide human-readable
names from the name section, but the printed wasm text must faithfully
represent the binary itself. This means that the human-readable names
from the name section, which have no restrictions on them beyond utf-8,
cannot be used as-is to represent names of items in the printed text.

Previously the solution that was implemented was to keep a set of names
defined for each particular type and iteratively append integers to the
end of a given name until something wasn't in the set. This technique
was intended to provide human-readable names by basing identifiers off
the names found in the name section, but it also guaranteed uniqueness
by ensuring everything was in the original set. Fuzzing, however, is
showing that this technique can be quite slow because looping can happen
quite a lot and it was additionally implemented pretty inefficiently
with lots of string allocations.

This commit implements a new strategy for naming items based on names
found in the name section. The original name from the name section is
still prioritized if it works (e.g. it's a valid wasm text identifier
and it's the first-of-its-kind), but the fallback is now much more
efficient and involves no kinds of loops. Generated identifiers for an
item are now guaranteed to be unique since they look like:

    #func10<original_name>

would be generated for function index 10 if the original name was
"original name" (because ' ' isn't a valid text identifier character).
This retains the human-readability by having the original name in there,
but uniqueness is guaranteed with the index being factored in.
Additionally `wasmprinter` goes ahead and "reserves" all symbols that
start with a `#` to ensure that wasm-provided names can't collide with
wasmprinter-selected names. This further guarantees uniqueness of
generated names by `wasmprinter`.

This makes the fuzz test case I was looking at much faster, dropping the
printing time from 20 seconds to 10 milliseconds. Additionally the code
is much easier to read as well!

### DIFF
--- a/crates/wasmprinter/src/lib.rs
+++ b/crates/wasmprinter/src/lib.rs
@@ -259,12 +259,15 @@ impl Printer {
     }
 
     fn register_names(&mut self, names: NameSectionReader<'_>) -> Result<()> {
-        fn name_map(into: &mut HashMap<u32, Naming>, names: NameMap<'_>) -> Result<()> {
+        fn name_map(into: &mut HashMap<u32, Naming>, names: NameMap<'_>, name: &str) -> Result<()> {
             let mut used = HashSet::new();
             let mut map = names.get_map()?;
             for _ in 0..map.get_count() {
                 let naming = map.read()?;
-                into.insert(naming.index, Naming::new(naming.name, &mut used));
+                into.insert(
+                    naming.index,
+                    Naming::new(naming.name, naming.index, name, &mut used),
+                );
             }
             Ok(())
         }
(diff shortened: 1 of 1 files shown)


fitzgen:
Just a heads up: I'm a bit focused on some other things at the moment and don't want to context switch to this PR. If you're okay waiting a day or two for review, then no problem, but otherwise maybe ping someone else for review.

alexcrichton:
Nah that's fine there's no rush for this, take your time!

fitzgen (approved):
Sorry this took so long to get to, I assumed it was a larger change, and wanted to have enough time to look at it properly, but it turns out this is relatively small and isolated.

Looks good!

alexcrichton:
No worries, thanks for taking the time regardless :)

PR #322, opened 2021-08-20
Add missing validation for the exceptions proposal

alexcrichton:
Wasmparser's validator would accidentally let a module through with the
tag section or with tag imports, even if the exceptions proposal wasn't
enabled.


fitzgen (approved):
Thanks!

PR #326, opened 2021-08-28
Exclude canonicalize-nan locals from codegen

alexcrichton:
This commit updates wasm-smith to exclude injected locals from the nan
canonicalization insertion from being candidates for generated
`local.{get,set,tee}` instructions. The injected locals contain
non-canonical NaN values which, if used, defeats the purpose of
canonicalizing NaN since it leaks the non-canonical value into the rest
of the program.

The fix implemented here is to accumulate the extra types in a separate
vector that's appened to the normal `locals` vector at the end of code
generation.


cfallin (approved):
Looks good. Might be helpful to include a comment describing why `extra_locals` is separate (i.e. to ensure these locals don't participate in random instruction generation).

PR #329, opened 2021-09-02
Create the initial skeleton for a `wasm-mutate` tool

fitzgen:
`wasm-mutate` will be a new tool for fuzzing Wasm compilers, runtimes,
validators, and other Wasm-consuming programs. `wasm-mutate` will take in an
existing Wasm module and then apply a pseudo-random transformation to it,
producing a new, mutated Wasm module. This new Wasm module will be fed into the
system under test, potentially triggering a bug that users can then diagnose and
fix.

`wasm-mutate` should additionally support the following features:

* `wasm-mutate` should have the ability to restrict mutations to only those that
  shrink the size of the Wasm module, e.g. removing a function from the Wasm
  module. When used in this mode, `wasm-mutate` essentially becomes a Wasm
  test-case reducer.

* `wasm-mutate` should have the ability to only apply semantics-preserving
  changes to the input Wasm module. When used in this mode, the mutated Wasm
  should compute identical results when given the same inputs as the original
  Wasm module.

* `wasm-mutate` should be deterministic. When given the same input Wasm module
  and the same seed, it should always produce the same mutated output Wasm
  module.

* `wasm-mutate` should play nice with mutation-based fuzzers like libFuzzer. It
  should reuse the fuzzer's raw input strings when it makes sense (e.g. when
  adding an export, use the raw fuzzer input when choosing an export name) so
  that fuzzer features like dictionary files can be leveraged. `wasm-mutate`
  should integrate with the `LLVMFuzzerCustomMutator` hook and the
  `libfuzzer_sys::fuzz_mutator!` macro.

--------------------------------------------------------------------------------

cc @jacarte: 

You can run the CLI via `cargo run --bin wasm-mutate` in the root of this repo

You can run its (pretty much non-existant) tests via `cargo test -p wasm-mutate` in the root of the repo.

The library is at `crates/wasm-mutate/src/lib.rs` and the executable CLI that wraps it is at `src/bin/wasm-mutate.rs`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ getopts = "0.2"
 log = "0.4"
 rayon = "1.0"
 structopt = "0.3.16"
+wasm-mutate = { path = "crates/wasm-mutate", features = ["structopt"] }
 wasm-smith = { path = "crates/wasm-smith" }
 wasmparser = { path = "crates/wasmparser" }
 wasmparser-dump = { path = "crates/dump" }
--- a/crates/wasm-mutate/Cargo.toml
+++ b/crates/wasm-mutate/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "wasm-mutate"
+version = "0.1.0"
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+structopt = { optional = true, version = "0.3" }
+thiserror = "1.0.28"
+wasmparser = { version = "0.80", path = "../wasmparser" }
+
+[dev-dependencies]
+anyhow = "1"
+wat = { version = "1", path = "../wat" }
--- a/crates/wasm-mutate/src/error.rs
+++ b/crates/wasm-mutate/src/error.rs
@@ -1,0 +1,10 @@
+/// An error encountered when choosing or applying a Wasm mutation.
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    /// The input Wasm module did not parse or validate okay.
+    #[error("Failed to parse or validate the input Wasm module.")]
+    Parse(#[from] wasmparser::BinaryReaderError),
+}
+
+/// A `Result` type that is either `Ok(T)` or `Err(wasm_mutate::Error)`.
+pub type Result<T> = std::result::Result<T, Error>;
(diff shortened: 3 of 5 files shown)


alexcrichton:
Seems reasonable to me! Should the tool perhaps be developed on a separate branch until it's more fleshed out?

Also in terms of the skeleton, there's a `seed` argument right now but shouldn't that presumably be some sort of list-of-bytes from the fuzz input? It looks like it's the seed to an RNG but I thought that for libfuzzer-based things we tend to prefer "DNAs" to something like that?

fitzgen:
> Should the tool perhaps be developed on a separate branch until it's more fleshed out?

I'd like to avoid waiting on one gigantic PR for the whole tool so that we can review incremental PRs. But I guess we could make a series of incremental PRs to a longer-lived `wasm-mutate` feature branch. Happy to go either way.

> Also in terms of the skeleton, there's a `seed` argument right now but shouldn't that presumably be some sort of list-of-bytes from the fuzz input? It looks like it's the seed to an RNG but I thought that for libfuzzer-based things we tend to prefer "DNAs" to something like that?

That is how we generate a structure from the fuzzer's raw input. This design is based on the custom mutator hooks that `libFuzzer` exposes. The custom fuzz mutator is [given only a `u64` seed](https://docs.rs/libfuzzer-sys/0.4.2/src/libfuzzer_sys/lib.rs.html#363-369) from `libFuzzer`, but then when you are mutating some byte slice, you can [call `LLVMFuzzerMutate`](https://docs.rs/libfuzzer-sys/0.4.2/src/libfuzzer_sys/lib.rs.html#420-426) on a given slice to have `libFuzzer` grow or shrink it (including growing a zero-sized slice, effectively giving you a bunch of new random data). This latter hook is what the `raw_mutate_func` hook in this PR is for integrating with.

alexcrichton:
Nah yeah I also agree it'd be good for incremental reviews, but if you're ok I'd prefer to develop on a separate branch and we can merge to main when it's ready to go.

For `seed`, it appears I need to go off and learn more about how this is integrating into libfuzzer!

alexcrichton (approved):
Oh and this looks good to go to me, so feel free to make a branch and such.

fitzgen:
Okay, I'll push this to the `wasm-mutate` branch. @jacarte: you can send incremental PRs to that branch, rather than `main`.

Jacarte:
> Okay, I'll push this to the `wasm-mutate` branch. @Jacarte: you can send incremental PRs to that branch, rather than `main`.

Gotcha !

fitzgen:
https://github.com/bytecodealliance/wasm-tools/tree/wasm-mutate

PR #331, opened 2021-09-07
Initial data for Wasm module generation

Jacarte:
The main idea is to preset some parts of the structure of the generated Wasm modules. To do so, the `Config` structure was expanded to support initial payloads. For example

```rs
pub struct InitialValuesConfig{
      tpes: Vec<Type>
  }
 impl Config for InitialValuesConfig {
     fn initial_types(&self) -> Option<Vec<Type>> {
         Some(vec![
             wasm_smith::Type::Func(Rc::new(
                FuncType::new(
                   vec![], // params
                   vec![ValType::I32] // result
             )
          )])
     }
 }

```

The previous example, will set the section types in the generated Wasms to have only one type, therefore, the generated functions will have only that signature: non-inputs and an i32 value as result.

The internal structures (Type, FuncType, etc) used in the generation were changed to public.

### DIFF
--- a/crates/wasm-smith/src/config.rs
+++ b/crates/wasm-smith/src/config.rs
@@ -3,7 +3,7 @@
 
 use arbitrary::{Arbitrary, Result, Unstructured};
 
-use crate::{Code, DataSegment, ElementSegment, Export, FuncType, GlobalType, Instruction, MemoryType, TableType, Type};
+use crate::{Code, DataSegment, ElementSegment, EntityType, Export, FuncType, GlobalType, Instruction, MemoryType, TableType, Type};
 
 /// Configuration for a generated module.
 ///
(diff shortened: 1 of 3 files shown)


fitzgen:
I think it might make sense to define the first N mutations we want to support and then let their needs guide us here. I fear that we can really go down a rabbit hole with some of this sharing `wasm-smith` stuff and not make progress on `wasm-mutate` itself.

What about something like:

* replace a function's body with unreachable
* rename an export
* remove an export
* add an export
* swap the left and right operands of a commutative operation
* turn a load offset=n into a an iconst n; iadd; load offset=0
* undo strength reduction by turning x << 1 into x * 2
* take the complement of an if/else's condition and swap the consequent and alternative blocks

This is a mix of module-level mutations and code/instruction-level mutations. Some kind of peephole-y things and some code motion things. I think it is generally representative of the kinds of mutations we want to be able to do eventually.

Does supporting those mutations sound like a good milestone to aim for to you?

I don't think any of these mutations really need any of `wasm-smith`'s guts, so unless I am missing something, we should be able to unask the question of exposing `wasm-smith` stuff more generically and close this PR.

Jacarte:
Gotcha, It sounds perfectly reasonable to me ! Closing PR

PR #333, opened 2021-09-08
Exclude the /benches directory from packaging

choller:
As discussed, it seems reasonable to not package the 1000+ corpus files for easier vendoring of the package.


alexcrichton (approved):
Thanks!

If you want, you could also bump to 0.7.2 here and I'll publish once this merges.

choller:
> If you want, you could also bump to 0.7.2 here and I'll publish once this merges.

Done, thanks! :)

PR #339, opened 2021-09-21
Peephole meta mutator

Jacarte:
General model to support deep code mutations. 

## PeepholeMutator

- The PeepholeMutator iterates over all operators for all defined functions.
- It then generates a "dictionary" of [mutator] -> collection of where it can be applied
- If the "dictionary" is empty, the method returns `NoMatchinPeepholes` error, then a new mutator should be selected
- If the "dictionary" is not empty, it selects a random key and then a random location over all possible places for the deep mutator.
- It calls the deep mutator that returns a mutated function entry.
- Non mutated functions and the newly generated function is encoded into a new code section.
- A mutated module is returned

There is a first implementation for a "deep" mutator that swap two operands for a commutative operator


crates/wasm-mutate/src/mutators/peephole/mod.rs
```diff
@@ -0,0 +1,252 @@
...
+                                // TODO, find better key ?
+                                innerprev
+                                    .entry(peephole.name())
+                                    .or_insert(Vec::new())
+                                    .push((fidx as usize, idx, peephole));
+                            }
+                        }
+                        innerprev
+                    })
+            });
```
fitzgen, line 73: I think as soon as we find an opcode where we have applicable mutators, we want to choose one of them and apply it there.

Because we are choosing a random opcode in the function to start searching for mutations to apply, and we are still checking all applicable mutators against this opcode and choosing a random one of the applicable ones, this should still result in a random mutation, and shouldn't skew our distribution.

But because we aren't checking every single peephole mutator against every single opcode in the function, this should be much faster for large functions.

crates/wasm-mutate/src/mutators/peephole/mod.rs
```diff
@@ -0,0 +1,252 @@
...
+                    .mutate(config, rnd, &mut reader, *operatoridx, &code_section.data)
+                    .unwrap();
+                println!("{:?}", function);
+                codes.function(&function);
+            } else {
+                // Copy exactly the same function to section
+                println!(
+                    "{:?}",
+                    &code_section.data[reader.range().start..reader.range().end]
+                );
```
fitzgen, line 99: Leftover from debugging?

crates/wasm-mutate/src/mutators/peephole/mod.rs
```diff
@@ -0,0 +1,252 @@
...
+    /// Provides the name of the mutator, mostly used for debugging purposes
+    fn name(&self) -> String {
+        return format!("{:?}", std::any::type_name::<Self>());
+    }
+}
+
+// This macro is meant to be used for testing deep mutators
+// It receives the original wat text variable, the expression returning the mutated function and the expected wat
+// For an example, look at SwapCommutativeOperator
+#[macro_export]
```
fitzgen, line 150: If this is only for test code, can we add `#[cfg(test)]` on top of it? That way we can't accidentally use it elsewhere, nor will `rustc` have to do anything other than parse it.

crates/wasm-mutate/src/mutators/peephole/swap_commutative.rs
```diff
@@ -0,0 +1,176 @@
...
+        funcreader: &mut FunctionBody,
+        operator_index: usize,
+        input_wasm: &[u8],
+    ) -> Result<Function> {
+        let opreader = funcreader.get_operators_reader().unwrap();
+        let body_range = opreader.range();
+        let opreaderiterator = opreader.into_iter_with_offsets();
+        let operators = opreaderiterator
+            .collect::<wasmparser::Result<Vec<(Operator, usize)>>>()
+            .unwrap();
```
fitzgen, line 65: Haven't we already parsed the operators by now? E.g. the operators are passed to the `can_mutate` method.

Can we pass them in here as well? And the locals and whatever else is needed. This way we don't need to parse the same thing twice.

crates/wasm-mutate/src/mutators/peephole/swap_commutative.rs
```diff
@@ -0,0 +1,176 @@
...
+        for (_, offset) in operators {
+            newoffset = offset;
+            if idx == operator_index {
+                // Copy previous code to the body
+                let previous = &input_wasm[body_range.start..offset];
+                newf.raw(previous.iter().copied());
+                // Inject new code to swap operands
+                newf.instruction(Instruction::LocalTee(local_count + 1));
+                newf.instruction(Instruction::LocalTee(local_count + 2));
+                break; // this break allows to copy the remaining buffer of the current reader
```
fitzgen, line 95: I think `local.tee`s aren't quite right here, and will result in the following sequence, no?

```wat
...         // Stack: [a b] ; Locals: x = 0, y = 0
local.tee x // Stack: [a b] ; Locals: x = b, y = 0
local.tee y // Stack: [a b] ; Locals: x = b, y = b
```

Instead I think we want to insert this snippet:

```wat
...         // Stack: [a b] ; Locals: x = 0, y = 0
local.set x // Stack: [a]   ; Locals: x = b, y = 0
local.set y // Stack: []    ; Locals: x = b, y = a
local.get x // Stack: [b]   ; Locals: x = b, y = a
local.get y // Stack: [b a] ; Locals: x = b, y = a
```

fitzgen (commented):
This is heading in the right direction, but there are a few things that need to be addressed, see the inline comments below.

Thanks!

crates/wasm-mutate/src/mutators/peephole/mod.rs
```diff
@@ -0,0 +1,252 @@
...
+        rnd: &mut rand::prelude::SmallRng,
+        info: &mut crate::ModuleInfo,
+    ) -> Result<Module> {
+        // Parse the module to get opcodes
+        let mut codes = CodeSection::new();
+        let code_section = info.get_code_section();
+        let mut sectionreader = CodeSectionReader::new(code_section.data, 0)?;
+        let function_count = sectionreader.get_count();
+
+        let peep_optimizers: &Vec<Box<dyn CodeMutator>> = &vec![Box::new(SwapCommutativeOperator)];
```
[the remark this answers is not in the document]
Jacarte: Gotcha

crates/wasm-mutate/src/mutators/peephole/mod.rs
```diff
@@ -0,0 +1,252 @@
...
+
+        // Split where to start looking for mutable function
+        let function_to_mutate = rnd.gen_range(0, function_count);
+        let all_readers = (0..function_count)
+            .map(|fidx| sectionreader.read().unwrap())
```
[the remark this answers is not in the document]
Jacarte: Gotcha

crates/wasm-mutate/src/mutators/peephole/mod.rs
```diff
@@ -0,0 +1,252 @@
...
+
+        // Since we can have several positions for the same mutator it is better to group them by mutator reference
+        let mut applicable: HashMap<String, Vec<(usize, usize, &Box<dyn CodeMutator>)>> =
+            HashMap::new();
+
+        (function_to_mutate..function_count)
+            .chain(0..function_to_mutate)
+            .fold(&mut applicable, |prev, fidx| {
+                let reader = all_readers[fidx as usize];
+                let operatorreader = reader.get_operators_reader().unwrap();
```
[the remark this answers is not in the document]
Jacarte: Gotcha

crates/wasm-mutate/src/mutators/peephole/mod.rs
```diff
@@ -0,0 +1,252 @@
...
+                let operators = &operatorreader
+                    .into_iter()
+                    .collect::<wasmparser::Result<Vec<Operator>>>()
+                    .unwrap();
```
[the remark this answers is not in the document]
Jacarte: Gotcha

crates/wasm-mutate/src/mutators/peephole/swap_commutative.rs
Jacarte: Done

crates/wasm-mutate/src/mutators/peephole/mod.rs
Jacarte: Done

crates/wasm-mutate/src/mutators/peephole/mod.rs
Jacarte: yeah...

crates/wasm-mutate/src/mutators/peephole/swap_commutative.rs
Jacarte: Thats right, but the thing is that I dont want to map each wasmparser operator to wasm-encoder instruction. If we pass the operators vector, we maybe can pass the offset inside the byte stream as well as a vector of tuples (operator, offset). With  the mutation position, the previous byte stream could be copied with this offset, we do the mutation and the we copy the byte stream from the next operator offset. 
I am working in this, I will come with a better solution :)

Jacarte: Ok, now we pass the operators and their respective offset inside the byte stream, besides the function reader, we might need to use locals_reader

crates/wasm-mutate/src/mutators/peephole/mod.rs
Jacarte: Improved, now we mutate as soon as we have at least one mutator for a function at an specific operator

crates/wasm-mutate/src/lib.rs
```diff
@@ -367,15 +372,27 @@ impl WasmMutate {
...
-        Ok(module?.finish())
+        match moduleresult {
+            Ok(module) => Ok(module.finish()),
+            Err(e) => {
+                match e {
+                    Error::NotMatchingPeepholes => {
+                        // TODO, Select another mutator
+                        return Err(e);
+                    }
+                    _ => return Err(e),
```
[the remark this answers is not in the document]
Jacarte: Nice solution :)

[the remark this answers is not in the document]
Jacarte: However, I think that shuffling  the whole collection of mutators cannot be the best solution in terms of performance. What about having a maximum number of random rotations to select the first available mutator?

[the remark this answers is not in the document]
fitzgen: I think we do want to try every mutation once, if every single thing before it is failing, because the alternative is that we return an error, our fuzz mutator integration code does *something*, and then libFuzzer just calls us again with a new random seed. And getting called again with a new random seed means we have to recreate the AST and reparse and all that. So we might as well try and squeeze everything out of the work we've already done before giving up.

But you're right that we don't need to shuffle the whole vec up front, we can do it incrementally and save a handful of calls into the RNG like this:

```rust
while !mutators.is_empty() {
    let i = rng.gen_range(0, mutators.len());
    if let Ok(module) = mutator.mutate(...) {
        return module;
    }
}

return Err(Error::NoApplicableMutations);
```

crates/wasm-mutate/src/mutators/peephole/swap_commutative.rs
```diff
@@ -0,0 +1,184 @@
...
+              i32.const 42
+              i32.const 42
+              i32.add
+              i32.const 56
+              local.set 3
+              local.set 4
+              local.get 3
+              local.get 4
+              i32.add))
+        "#;
```
fitzgen, line 161: Why isn't the function exported anymore after the mutation? Swapping the operands of a commutative operation shouldn't have any effect on whether the operation's function is exported or not.

crates/wasmparser/src/readers/code_section.rs
```diff
@@ -25,6 +25,10 @@ pub struct FunctionBody<'a> {
     allow_memarg64: bool,
 }
 
+impl<'a> Copy for FunctionBody<'a>{
+
+}
+
```
fitzgen, line 31: This should be `#[derive(Copy)]` instead of a handwritten impl.

crates/wasmparser/src/readers/operators.rs
```diff
@@ -76,6 +82,13 @@ impl<'a> OperatorsReader<'a> {
         let pos = self.reader.original_position();
         Ok((self.read()?, pos))
     }
+
+    pub fn range(&self) -> Range {
```
fitzgen, line 86: We shouldn't need to add `offset` and `data` to `OperatorsReader`. The `BinaryReader` has a `range` method that we can delegate too, and it already keeps track of that offset information internally.

crates/wasm-mutate/src/lib.rs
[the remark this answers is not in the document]
Jacarte: Gotcha

crates/wasm-mutate/src/module/mod.rs
```diff
@@ -63,3 +64,13 @@ impl TryFrom<TypeDef<'_>> for TypeInfo {
...
     }
 }
+
+pub fn map_type(tpe: Type) -> super::Result<ValType> {
+    match tpe {
+        Type::I32 => Ok(ValType::I32),
+        Type::I64 => Ok(ValType::I64),
+        Type::F32 => Ok(ValType::F32),
+        Type::F64 => Ok(ValType::F64),
+        _ => Err(super::Error::NoMutationsAplicable),
```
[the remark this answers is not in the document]
Jacarte: Right...

crates/wasm-mutate/src/mutators/peephole/swap_commutative.rs
Jacarte: Right...this is on the test macro..I was not saving all the sections. I will fix this :)

crates/wasmparser/src/readers/code_section.rs
Jacarte: Gotcha

crates/wasm-mutate/src/lib.rs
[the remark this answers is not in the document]
Jacarte: WDYT about having this in the next iteration?

crates/wasm-mutate/src/mutators/peephole/swap_commutative.rs
Jacarte: Done

crates/wasmparser/src/readers/operators.rs
Jacarte: Do you mean making the reader field publicly available?

fitzgen: I just mean returning `self.reader.range()` instead of keeping track of `offset` and `data` and doing the range calculation ourselves here

crates/wasm-mutate/src/mutators/peephole/swap_commutative.rs
```diff
@@ -0,0 +1,184 @@
...
+              (local i32 i32 i32 i32)
+              i32.const 42
+              i32.const 42
+              i32.add
+              i32.const 56
+              local.set 3
+              local.set 4
+              local.get 3
+              local.get 4
+              i32.add))
```
fitzgen, line 160: I just noticed that there are 0 parameters and 4 locals, so the maximum number we should pass to `local.set`/`local.get` is `3`, so I think there is an off-by-one bug in this mutator.

crates/wasm-mutate/src/mutators/peephole/swap_commutative.rs
```diff
@@ -0,0 +1,184 @@
...
+            newoffset = offset;
+            if idx == operator_index {
+                // Copy previous code to the body
+                let previous = &function_stream[body_range.start..offset];
+                newf.raw(previous.iter().copied());
+                // Inject new code to swap operands
+                newf.instruction(Instruction::LocalSet(local_count + 1));
+                newf.instruction(Instruction::LocalSet(local_count + 2));
+                newf.instruction(Instruction::LocalGet(local_count + 1));
+                newf.instruction(Instruction::LocalGet(local_count + 2));
```
fitzgen, line 91: Ah, yeah here it is:

```suggestion
                newf.instruction(Instruction::LocalSet(local_count));
                newf.instruction(Instruction::LocalSet(local_count + 1));
                newf.instruction(Instruction::LocalGet(local_count));
                newf.instruction(Instruction::LocalGet(local_count + 1));
```

fitzgen (approved):
LGTM with the off-by-one fix below!

I think that for mutations that are supposed to leave the Wasm module valid (i.e. all the mutations we have defined thus far) the tests for the mutator should run `wasmparser::validate` on the mutated bytes so that we can prevent bugs like this going forward.

crates/wasm-mutate/src/mutators/peephole/swap_commutative.rs
Jacarte: yep :(

fitzgen (approved):
Looks good! Thanks for your patience with these many back and forths!

PR #343, opened 2021-09-24
First implementation for shift load operation

Jacarte:
Select a random `load` instruction and then place random constant addition before to shift the load offset.


fitzgen:
Is this ready for review or is it obsolete given the upcoming `egg` changes?

Jacarte:
It is obsolete :(, we can close it

PR #344, opened 2021-09-24
Refactoring mutation tests

Jacarte:
Some code deduplication for mutation tests. We have now a macro_rule to asses a concrete mutation.

### DIFF
--- a/crates/wasm-mutate/src/mutators/mod.rs
+++ b/crates/wasm-mutate/src/mutators/mod.rs
@@ -35,28 +35,30 @@ pub mod snip_function;
 // macro for mutation assesment
 #[cfg(test)]
 #[macro_export]
-macro_rules! match_mutation {
-    (
-        $original: tt, $mutator: expr, $expected: tt
-    ) => {{
-        let wasmmutate = WasmMutate::default();
-        let original = &wat::parse_str($original).unwrap();
+pub fn match_mutation(original: &str, mutator: &dyn Mutator, expected: &str) {
+    use rand::SeedableRng;
 
-        let mut info = wasmmutate.get_module_info(original).unwrap();
-        let can_mutate = $mutator.can_mutate(&wasmmutate, &info).unwrap();
+    let wasmmutate = WasmMutate::default();
+    let original = &wat::parse_str(original).unwrap();
 
-        assert_eq!(can_mutate, true);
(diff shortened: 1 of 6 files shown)


crates/wasm-mutate/src/mutators/mod.rs
```diff
@@ -31,3 +31,32 @@ pub mod peephole;
...
+        let mutation = $mutator.mutate(&wasmmutate, &mut rnd, &mut info);
+
+        let mutation_bytes = mutation.unwrap().finish();
+        // If it fails, it is probably an invalid
+        // reformatting expected
+        let text = wasmprinter::print_bytes(mutation_bytes).unwrap();
+        let expected = &wat::parse_str($expected).unwrap();
+        let expected_text = wasmprinter::print_bytes(expected).unwrap();
+
+        assert_eq!(text, expected_text);
```
fitzgen, line 60: Should this also assert that the Wasm is still valid?

crates/wasm-mutate/src/mutators/mod.rs
```diff
@@ -31,3 +31,32 @@ pub mod peephole;
...
 pub mod rename_export;
 pub mod snip_function;
+
+// macro for mutation assesment
+#[cfg(test)]
+#[macro_export]
+macro_rules! match_mutation {
+    (
+        $original: tt, $mutator: expr, $expected: tt
+    ) => {{
```
fitzgen, line 41: I think this can be a function, and doesn't need to be a macro.

```rust
fn assert_mutation(mutator: impl Mutate, original: &str, expected: &str) { ... }

fitzgen (approved):
Looks good with two tiny suggestions below.

crates/wasm-mutate/src/mutators/mod.rs
Jacarte: Yeah but this test code is obsolete as well with the new egg thing

Jacarte: Gotcha

crates/wasm-mutate/src/mutators/mod.rs
Jacarte: Gotcha

Jacarte:
Comments solved 💯

fitzgen (approved):

PR #346, opened 2021-09-29
Add exception-handling proposal to wasm-smith

eqrion:
These commits add support to wasm-smith for the exception-handling proposal. I've manually tested this using `wasm-smith/tests/tests.rs` with the feature enabled (it defaults off) and can confirm it generates exception-handling code and also passes.

### DIFF
--- a/crates/wasm-encoder/src/tags.rs
+++ b/crates/wasm-encoder/src/tags.rs
@@ -2,6 +2,23 @@ use super::*;
 use std::convert::TryFrom;
 
 /// An encoder for the tag section.
+///
+/// # Example
+///
+/// ```
+/// use wasm_encoder::{Module, TagSection, TagType};
+///
+/// let mut tags = TagSection::new();
+/// tags.tag(TagType {
+///     kind: TagKind::Exception,
+///     func_type_idx: 0,
+/// });
+///
+/// let mut module = Module::new();
+/// module.section(&tags);
+///
+/// let wasm_bytes = module.finish();
+/// ```
 #[derive(Clone, Debug)]
 pub struct TagSection {
     bytes: Vec<u8>,
--- a/crates/wasm-encoder/src/tags.rs
+++ b/crates/wasm-encoder/src/tags.rs
@@ -1,0 +1,73 @@
+use super::*;
+use std::convert::TryFrom;
+
+/// An encoder for the tag section.
+#[derive(Clone, Debug)]
+pub struct TagSection {
+    bytes: Vec<u8>,
+    num_added: u32,
+}
+
+impl TagSection {
+    /// Create a new tag section encoder.
+    pub fn new() -> TagSection {
+        TagSection {
+            bytes: vec![],
+            num_added: 0,
+        }
+    }
+
+    /// How many tags have been defined inside this section so far?
+    pub fn len(&self) -> u32 {
+        self.num_added
+    }
+
+    /// Define a tag.
+    pub fn tag(&mut self, tag_type: TagType) -> &mut Self {
+        tag_type.encode(&mut self.bytes);
+        self.num_added += 1;
+        self
+    }
+}
+
+impl Section for TagSection {
+    fn id(&self) -> u8 {
+        SectionId::Tag.into()
+    }
+
+    fn encode<S>(&self, sink: &mut S)
+    where
+        S: Extend<u8>,
+    {
+        let num_added = encoders::u32(self.num_added);
+        let n = num_added.len();
+        sink.extend(
+            encoders::u32(u32::try_from(n + self.bytes.len()).unwrap())
+                .chain(num_added)
+                .chain(self.bytes.iter().copied()),
+        );
+    }
+}
+
+#[allow(missing_docs)]
+#[repr(u8)]
+#[derive(Clone, Copy, Debug)]
+pub enum TagKind {
+    Exception = 0x0,
+}
+
+/// A tag's type.
+#[derive(Clone, Copy, Debug)]
+pub struct TagType {
+    /// The kind of tag
+    pub kind: TagKind,
+    /// The function type this tag uses
+    pub func_type_idx: u32,
+}
+
+impl TagType {
+    pub(crate) fn encode(&self, bytes: &mut Vec<u8>) {
+        bytes.push(self.kind as u8);
+        bytes.extend(encoders::u32(self.func_type_idx));
+    }
+}
--- a/crates/wasm-encoder/src/tags.rs
+++ b/crates/wasm-encoder/src/tags.rs
@@ -6,7 +6,7 @@ use std::convert::TryFrom;
 /// # Example
 ///
 /// ```
-/// use wasm_encoder::{Module, TagSection, TagType};
+/// use wasm_encoder::{Module, TagSection, TagType, TagKind};
 ///
 /// let mut tags = TagSection::new();
 /// tags.tag(TagType {
--- a/crates/wasm-smith/src/code_builder.rs
+++ b/crates/wasm-smith/src/code_builder.rs
@@ -91,7 +91,7 @@ instructions! {
     (Some(return_valid), r#return, 900),
     (Some(call_valid), call),
     (Some(call_indirect_valid), call_indirect),
-    (Some(throw_valid), throw),
+    (Some(throw_valid), throw, 850),
     (Some(rethrow_valid), rethrow),
     // Parametric instructions.
     (Some(drop_valid), drop),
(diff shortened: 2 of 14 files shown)


alexcrichton:
This looks great, thanks!

Could you update a few of the fuzz targets here as well so we can test this while fuzzing? The locations I think are:

* [here](https://github.com/bytecodealliance/wasm-tools/blob/873dd3ef4650fdb7251bd8349190df825ed0265b/fuzz/src/lib.rs#L12-L17)
* [here](https://github.com/bytecodealliance/wasm-tools/blob/873dd3ef4650fdb7251bd8349190df825ed0265b/fuzz/fuzz_targets/validate-valid-module.rs#L16-L25)


and I think that might be it? If you wouldn't mind as well if you could run the `*-valid-module` fuzz targets as well for a minute or so just to make sure nothing is too far off, that'd also help give this some further testing. I think they can all be run with `cargo fuzz run validate-valid-module` (etc)

crates/wasm-smith/src/code_builder.rs
```diff
@@ -87,6 +91,8 @@ instructions! {
     (Some(return_valid), r#return, 900),
     (Some(call_valid), call),
     (Some(call_indirect_valid), call_indirect),
+    (Some(throw_valid), throw),
```
alexcrichton, line 94: I might be worth putting a weight on this one like `nop` and the others since this is probably going to get generated a lot otherwise and it's not too too interesting if modules mostly just immediately `throw`

crates/wasm-encoder/src/tags.rs
```diff
@@ -0,0 +1,73 @@
+use super::*;
+use std::convert::TryFrom;
+
+/// An encoder for the tag section.
+#[derive(Clone, Debug)]
+pub struct TagSection {
```
fitzgen, line 6: Can you add an example here? All of the other sections have examples at the top (e.g. https://github.com/bytecodealliance/wasm-tools/blob/main/crates/wasm-encoder/src/globals.rs#L5-L23) which is nice for people using this crate.

crates/wasm-smith/src/config.rs
```diff
@@ -238,11 +248,17 @@ pub trait Config: 'static + std::fmt::Debug {
     }
 
     /// Determines whether the SIMD proposal is enabled for
-    /// generating insructions. Defaults to `true`.
+    /// generating insructions. Defaults to `false`.
```
fitzgen, line 251: Good catch, thanks!

crates/wasm-smith/src/lib.rs
```diff
@@ -1665,6 +1739,31 @@ impl Module {
...
+        let max = candidate_func_types.len() - 1;
+        let ty = candidate_func_types[u.int_in_range(0..=max)?];
+        Ok(TagType {
+            func_type_idx: ty,
+            func_type: self.func_type(ty).clone(),
+        })
+    }
+
+    fn arbitrary_tags(&mut self, u: &mut Unstructured) -> Result<()> {
+        if !self.config.exceptions_enabled() || self.has_tag_func_types() {
```
fitzgen, line 1753: I'm not familiar with the exceptions proposal yet, but is this supposed to be

```suggestion
        if !self.config.exceptions_enabled() || !self.has_tag_func_types() {
```

?

fitzgen (commented):
Thanks Ryan!

Just a couple small things in addition to what Alex already said.

crates/wasm-smith/src/code_builder.rs
eqrion: Good idea. Thinking on this more, it feels like there are two distinct cases here:
  1. throw with no enclosing catch (acts like an unreachable, basically a panic)
  2. throw to an enclosing catch that can handle it (acts like a br)

It would be nice to generate (2) more frequently than (1) so that we can test exceptional control flow well. Is there a way to make weight conditional on context? Or would that have to be achieved by splitting throw_valid/throw into two different instruction emitters, for the two different cases?

eqrion:
> This looks great, thanks!
> 
> Could you update a few of the fuzz targets here as well so we can test this while fuzzing? The locations I think are:
> 
>     * [here](https://github.com/bytecodealliance/wasm-tools/blob/873dd3ef4650fdb7251bd8349190df825ed0265b/fuzz/src/lib.rs#L12-L17)
> 
>     * [here](https://github.com/bytecodealliance/wasm-tools/blob/873dd3ef4650fdb7251bd8349190df825ed0265b/fuzz/fuzz_targets/validate-valid-module.rs#L16-L25)
> 
> 
> and I think that might be it? If you wouldn't mind as well if you could run the `*-valid-module` fuzz targets as well for a minute or so just to make sure nothing is too far off, that'd also help give this some further testing. I think they can all be run with `cargo fuzz run validate-valid-module` (etc)

I'm probably doing something plainly wrong. Getting the following output.

```
rhunt@rhunt-mac~/s/m/wasm-tools> cargo fuzz run validate-valid-module
warning: unknown pass `sancov-module`, ignoring

warning: 1 warning emitted

warning: unknown pass `sancov-module`, ignoring

warning: 1 warning emitted

warning: unknown pass `sancov-module`, ignoring

warning: 1 warning emitted

warning: unknown pass `sancov-module`, ignoring

warning: 1 warning emitted

warning: unknown pass `sancov-module`, ignoring

warning: 1 warning emitted

warning: unknown pass `sancov-module`, ignoring

warning: 1 warning emitted

warning: unknown pass `sancov-module`, ignoring

warning: 1 warning emitted

warning: unknown pass `sancov-module`, ignoring

warning: 1 warning emitted

    Finished release [optimized] target(s) in 0.08s
warning: unknown pass `sancov-module`, ignoring

warning: 1 warning emitted

warning: unknown pass `sancov-module`, ignoring

warning: 1 warning emitted

warning: unknown pass `sancov-module`, ignoring

warning: 1 warning emitted

warning: unknown pass `sancov-module`, ignoring

warning: 1 warning emitted

warning: unknown pass `sancov-module`, ignoring

warning: 1 warning emitted

warning: unknown pass `sancov-module`, ignoring

warning: 1 warning emitted

warning: unknown pass `sancov-module`, ignoring

warning: 1 warning emitted

warning: unknown pass `sancov-module`, ignoring

warning: 1 warning emitted

    Finished release [optimized] target(s) in 0.08s
     Running `target/x86_64-apple-darwin/release/validate-valid-module -artifact_prefix=/Users/rhunt/src/mozilla/wasm-tools/fuzz/artifacts/validate-valid-module/ /Users/rhunt/src/mozilla/wasm-tools/fuzz/corpus/validate-valid-module`
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 3095649829
INFO:        0 files found in /Users/rhunt/src/mozilla/wasm-tools/fuzz/corpus/validate-valid-module
INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 4096 bytes
INFO: A corpus is not provided, starting from an empty corpus
#2	INITED exec/s: 0 rss: 31Mb
ERROR: no interesting inputs were found. Is the code instrumented for coverage? Exiting.
────────────────────────────────────────────────────────────────────────────────

Error: Fuzz target exited with exit status: 1
```

This is with nightly as the default toolchain on OSX x86_64.

crates/wasm-smith/src/code_builder.rs
alexcrichton: Yeah for that I think it can be done with separate emitters. We haven't had that come up yet so it's breaking new ground!

alexcrichton:
You might need to `cargo install cargo-fuzz` to get the latest version perhaps? I think some changes happened on nightly recently (you'll also probably need the most up-to-date nightly). I can help dig in if those don't work though.

eqrion:
Updating both nightly and cargo-fuzz fixed it. Thanks!

crates/wasm-smith/src/code_builder.rs
eqrion: This is actually a bit challenging, so I'm going to defer it. The problem is that throws in a try block are generated before the catch blocks that describe which tags can be caught in that try. It's solvable, but I think keeping it simple is best to start.

crates/wasm-smith/src/lib.rs
eqrion: Yes, indeed. Good catch.

eqrion:
Okay, I should have addressed all the comments. I also ran fuzzing for the three `-valid-module` fuzz targets for two minutes each and didn't encounter any obvious issue.

alexcrichton:
👍 Thanks!

eqrion:
@alexcrichton Could we get a new version published for this? I'd like to use this in SpiderMonkey CI.

alexcrichton:
Sure yeah, I've posted https://github.com/bytecodealliance/wasm-tools/pull/348 for that.

PR #349, opened 2021-09-30
Define a fuzz target for `wasm-mutate`

fitzgen:
Asserts that if `wasm-mutate` successfully mutates a valid Wasm module, then the
resulting mutated Wasm module should also be valid.

Does not (yet) exercise that the mutated Wasm module has equivalent semantics to
the original module.

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -12,6 +12,7 @@ env_logger = "0.8"
 libfuzzer-sys = "0.4.0"
 log = "0.4"
 tempfile = "3.0"
+wasm-mutate = { path = "../crates/wasm-mutate" }
 wasm-smith = { path = "../crates/wasm-smith" }
 wasmparser = { path = "../crates/wasmparser" }
 wasmprinter = { path = "../crates/wasmprinter" }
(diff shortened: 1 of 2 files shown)


fitzgen:
@jacarte to run this fuzz target, you need `cargo fuzz` and an up-to-date nightly Rust toolchain:

```
$ cargo install cargo-fuzz
$ rustup update nightly
```

Then you can run it like this:

```
$ cargo +nightly fuzz run mutate
```

That should find some bugs relatively quickly. For me it found a bug where we mutate an export and turn it into a duplicate of an existing export, but duplicate exports are not allowed. So our code that chooses export names needs to handle this edge case.

Also, I see now that there are a ton of warnings inside `wasm-mutate` that should be fixed. Also a bunch of public items are missing documentation, which they should also have.

fitzgen:
Going to go ahead and merge this into the `wasm-mutate` branch.

Jacarte:
Tack tack !

fitzgen:
Oh, and once you find a bug, and you want to learn more about what's going on, you can do

```
$ RUST_LOG=debug cargo +nightly fuzz run mutate path/to/failing/input
```

and that will enable some little bits of logging as well as dump `test.wasm` and `test.wat` files based on what `wasm-smith` generated from the fuzzer's random input.

PR #350, opened 2021-09-30
Record  % of modules that we mutate okay in fuzz target

fitzgen:
You can enable this metric by running the fuzz target like this:

```
$ RUST_LOG=info cargo +nightly fuzz run mutate
```

In general, we want to maximize this percentage.

cc @jacarte


fitzgen:
Going to go ahead and merge this.

PR #352, opened 2021-10-03
FIX: rename export mutator fail because of duplicated entry bug

Jacarte:
CC @fitzgen in https://github.com/bytecodealliance/wasm-tools/pull/349

### DIFF
--- a/crates/wasm-mutate/src/mutators/rename_export.rs
+++ b/crates/wasm-mutate/src/mutators/rename_export.rs
@@ -8,19 +8,19 @@ use wasmparser::{CodeSectionReader, ExportSectionReader};
 /// RenameExportMutator generates a random renaming of prexisting exports.
 /// The export entry is selected randmonly and then a new `field` name is generated/
 pub struct RenameExportMutator {
-    /// The maximum length of the generated export entry 
+    /// The maximum length of the generated export entry
     pub max_name_size: u32,
 }
 
 impl RenameExportMutator {
     // Copied and transformed from wasm-smith name generation
-    fn limited_string(&self, rnd: &mut SmallRng, info: &ModuleInfo) -> String {
-        let size = rnd.gen_range(1, self.max_name_size);
-        let size = std::cmp::min(size, self.max_name_size);
+    fn limited_string(&self, rnd: &mut SmallRng, info: &ModuleInfo, max_name_size: u32) -> String {
+        let size = rnd.gen_range(1, max_name_size);
+        let size = std::cmp::min(size, max_name_size);
         let mut str = vec![0u8; size as usize];
         rnd.fill_bytes(&mut str);
 
-        let s = match std::str::from_utf8(&str) {
+        let mut s = match std::str::from_utf8(&str) {
             Ok(s) => String::from(s),
             Err(e) => {
                 let i = e.valid_up_to();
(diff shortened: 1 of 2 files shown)


crates/wasm-mutate/src/mutators/rename_export.rs
```diff
@@ -4,19 +4,23 @@ use rand::prelude::SmallRng;
...
 impl RenameExportMutator {
     // Copied and transformed from wasm-smith name generation
-    fn limited_string(&self, rnd: &mut SmallRng) -> String {
-        let size = rnd.gen_range(1, self.max_name_size);
-        let size = std::cmp::min(size, self.max_name_size);
+    fn limited_string(&self, rnd: &mut SmallRng, info: &ModuleInfo, max_name_size: u32) -> String {
+        let size = rnd.gen_range(1, max_name_size);
+        let size = std::cmp::min(size, max_name_size);
         let mut str = vec![0u8; size as usize];
         rnd.fill_bytes(&mut str);
```
fitzgen, line 21: Unrelated to this PR, but I am just noticing this now: this should call the configured `raw_mutate_func`, if any, and only fallback to using the RNG if there is no `raw_mutate_func`.

crates/wasm-mutate/src/mutators/rename_export.rs
```diff
@@ -27,7 +31,12 @@ impl RenameExportMutator {
                 };
                 String::from(s)
             }
+        };
+        // Add one symbol at a time until it is not contained in the export field names
+        while info.export_names.contains(&s.as_str()) {
+            s.push_str(&self.limited_string(rnd, info, 2));
```
fitzgen, line 37: Two things:

1. We should use a `HashSet<String>` for `export_names` so that we can do `O(1)` lookups instead of `O(n)` lookups.

2. Because `limited_string` returns a `String`, we are doing a heap allocation that we then immediately throw away after pushing onto `s`. And we do that for every iteration of the loop. It would be better if we made a specialized version of `limited_string` that reuses an existing allocation by taking a `s: &mut String` parameter that it mutates directly instead of returning a new `String` allocation.

fitzgen (commented):
Looks good, just a couple suggestions before we merge this.

crates/wasm-mutate/src/mutators/rename_export.rs
Jacarte: Done

crates/wasm-mutate/src/mutators/rename_export.rs
Jacarte: Done

crates/wasm-mutate/src/mutators/rename_export.rs
```diff
@@ -4,19 +4,23 @@ use rand::prelude::SmallRng;
...
 }
 
 impl RenameExportMutator {
     // Copied and transformed from wasm-smith name generation
-    fn limited_string(&self, rnd: &mut SmallRng) -> String {
-        let size = rnd.gen_range(1, self.max_name_size);
-        let size = std::cmp::min(size, self.max_name_size);
+    fn limited_string(&self, rnd: &mut SmallRng, info: &ModuleInfo, max_name_size: u32) -> String {
+        let size = rnd.gen_range(1, max_name_size);
+        let size = std::cmp::min(size, max_name_size);
```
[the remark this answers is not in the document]
Jacarte: Fixed

Jacarte:
Addressed :)

fitzgen (approved):

PR #353, opened 2021-10-04
Peep2egg second iteration

Jacarte:
CC @fitzgen 

Next iteration for the migration of peephole mutators to egg.


crates/wasm-mutate/src/mutators/peephole/eggsy/lang.rs
```diff
@@ -24,6 +25,10 @@ define_language! {
             This operation represent a random number, if its used, every time is should represent the same random number
         */
         "rand" = Rand,
+        /*
+            This instructions is used to define unknown operands, for example when the value can come from the join of several basic blocks in a dfg
+        */
+        "undef" = Undef,
```
fitzgen, line 31: How is this different from symbol, which is also just some opaque value we can manipulate but can't peek at actual value? Also, it seems like there can only be a single `undef` in an egg term as well, since we don't have a way of differentiating them, which is pretty restricting.

crates/wasm-mutate/src/mutators/peephole/eggsy/mod.rs
```diff
@@ -443,79 +463,88 @@ impl Encoder {
...
+                let operatorid = match operator {
+                    Operator::I32Shl => {
+                        // Check this node has only two child
+                        assert_eq!(subexpressions.len(), 2);
+
+                        Ok(expr.add(Lang::I32Shl([subexpressions[0], subexpressions[1]])))
+                    }
+                    Operator::I32Add => {
+                        // Check this node has only two child
+                        assert_eq!(subexpressions.len(), 2);
```
fitzgen: These can probably be `debug_assert_eq!` as they are "internal" implementation assertions and not something we are asserting about user/some-other-crate's input.

fitzgen (approved):
Looks good to me; will hold off on merging until after #352 merges.

fitzgen (approved):
Overall, looking good. I have a few questions/comments, but we can address these in follow up PRs.

crates/wasm-mutate/src/mutators/peephole/eggsy/lang.rs
Jacarte: This symbol is used right now to determine if the DFG is determined, if the generated DFG written as an eterm has the undef node, then the operand for some operator comes from a different basic block.

crates/wasm-mutate/src/mutators/peephole/eggsy/mod.rs
```diff
@@ -110,520 +103,145 @@ where
...
 
         let mut worklist: Vec<_> = rootnode
             .children()
             .iter()
+            .rev()
             .map(|id| (eclass, 0, id, 0)) // (root, operant, depth)
             .collect();
 
-        while let Some((parent, parentidx, &node, depth)) = worklist.pop() {
+        while let Some((_, parentidx, &node, depth)) = worklist.pop() {
```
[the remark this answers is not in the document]
Jacarte: Fixed

Issue #354, opened 2021-10-04
Prototype Relaxed SIMD spec


yurydelendik:
See https://github.com/WebAssembly/relaxed-simd

There are currently few major items to address:
 - [ ] implement `fpenv` section
 - [x] implement operators

There is vague binary format specification at the moment, but looks like "all Relaxed SIMD instructions will take an additional varuint32 immediate, which is an index into the fpenv index space". It requires to specify/read a `varuint32` after SIMD op code. We can assume there will one section defined, and we can default the fpenv index value to 0 for now. There is some churn of symbolic names for WAT, so `relaxed` prefix may go away.

The section code for `fpenv` is hard to define at this phase. We can choose one e.g. `18` or `0`+`'fpenv'` for now. @sunfishcode wdyt? The section body will default to `01` (amount of entries) + `00` (recommended padding). Order somewhere between `Global` and `Export` (?)

/see https://bugzilla.mozilla.org/show_bug.cgi?id=1732996

alexcrichton:
I believe this has since been done, so closing.

PR #359, opened 2021-10-14
Add support for Relaxed SIMD proposal instructions

yurydelendik:
Implements only instructions from https://github.com/WebAssembly/relaxed-simd/blob/main/proposals/relaxed-simd/Overview.md . There is also no support for varuint32 fpenv reference/arg in these instructions.

There is a discussion about exact names at https://github.com/WebAssembly/relaxed-simd/issues/42, I'm going with "_relaxed" suffix for this PR.

(Noticed couple typos in the existing code, I can move it somewhere else from here.)

See #354


crates/wasmparser/src/operators_validator.rs
```diff
@@ -1654,8 +1681,6 @@ impl OperatorValidator {
             | Operator::I32x4Neg
             | Operator::I64x2Abs
             | Operator::I64x2Neg
-            | Operator::I32x4TruncSatF32x4S
```
yurydelendik, line 1657: existing: do there need to be check for non-deterministic?

alexcrichton: AFAIK the deterministic feature isn't heavily used, so yeah I think it's fine to move these to requiring determinism.

PR #366, opened 2021-10-18
Merge the `wasm-mutate` tool into `main`

fitzgen:
So `wasm-mutate` is in a good enough position where I think it makes sense to merge it into `main` and continue development there. Notably its fuzz target is not crashing, meaning that when `wasm-mutate` can find a mutation to apply to the given Wasm module, that mutation always produces a valid Wasm module. Additionally, we have at least one mutation that can be applied to the input Wasm module for somewhere in the high-80s % of random Wasm modules produced by `wasm-smith`.

There are a ton of commits in the `wasm-mutate` branch, they don't all pass tests individually, and it doesn't really make sense to preserve that history. So I'm opening this PR with everything squashed into one commit (this has a few changes/clean ups/renames that aren't present in the `wasm-mutate` branch, as well).

cc @jacarte


fuzz/fuzz_targets/mutate.rs
```diff
@@ -0,0 +1,70 @@
...
+        Err(e) => match e {
+            wasm_mutate::Error::NoMutationsApplicable => return,
+            _ => panic!("Invalid mutation process"),
+        },
+    };
+
+    // Add valiation with all features enabled, like multiple value return
+    let mut features = WasmFeatures::default();
+    // Add other if needed
+    features.multi_value = true;
```
alexcrichton, line 55: I think this is on-by-default so this shouldn't need to get modified?

fuzz/fuzz_targets/mutate.rs
```diff
@@ -0,0 +1,70 @@
+#![no_main]
+
+use libfuzzer_sys::fuzz_target;
+use std::sync::atomic::{AtomicU64, Ordering};
+use wasmparser::WasmFeatures;
+
+static NUM_RUNS: AtomicU64 = AtomicU64::new(0);
+static NUM_SUCCESSFUL_MUTATIONS: AtomicU64 = AtomicU64::new(0);
+
+fuzz_target!(|inputs: (wasm_smith::Module, u64)| {
```
alexcrichton, line 10: Could this use https://github.com/bytecodealliance/wasm-tools/blob/9130f0f0233fe44626368e91579529e646080820/fuzz/src/lib.rs#L5-L8 perhaps using the closure to generate the seed?

fuzz/fuzz_targets/mutate.rs
```diff
@@ -0,0 +1,70 @@
...
+    features.multi_value = true;
+
+    let mut validator = wasmparser::Validator::new();
+    validator.wasm_features(features);
+    let validation_result = validator.validate_all(&mutated_wasm);
+    log::debug!("validation result = {:?}", validation_result);
+    if log::log_enabled!(log::Level::Debug) {
+        let mutated_text = wasmprinter::print_bytes(mutated_wasm).unwrap();
+        std::fs::write("mutated.wat", &mutated_text).expect("should write `mutated.wat` okay");
+    }
```
alexcrichton, line 64: Could this use https://github.com/bytecodealliance/wasm-tools/blob/9130f0f0233fe44626368e91579529e646080820/fuzz/src/lib.rs#L33-L51 or perhaps a variant to write the wasm?

crates/wasmparser/src/readers/operators.rs
```diff
@@ -17,7 +17,7 @@ use super::{BinaryReader, BinaryReaderError, Operator, Result};
 
 #[derive(Clone)]
 pub struct OperatorsReader<'a> {
-    pub(crate) reader: BinaryReader<'a>,
+    pub reader: BinaryReader<'a>,
```
alexcrichton, line 20: Could this be exposed as a method instead of as a public field?

alexcrichton (commented):
Looks reasonable to me! Could this get added to existing fuzz targets for their mutations perhaps? (or a new fuzz target that uses wasm-mutate as a mutator?)

PR #367, opened 2021-10-19
Wasm mutate stats

Jacarte:
Benchmark crate for wasm-mutate, `wasm-mutate-stats`.

It takes the Wasm modules inside the `corpus` folder. Each one is mutated in several iterations where the input module to mutate is the outcome from the previous iteration. The hash of the mutated Wasm module is then saved to check how many different Wasm modules are generated. 

Besides, `wasmtime` is called to compile each generated Wasm module. If the generated Wasm module can be compiled to an object file, the hash of the obj is compared to the object files generated in previous iterations, reporting the preservation of the corresponding Wasm module. This step could help us to identify if the applied mutations are in fact affecting the low-level diversity.


Jacarte:
Done, cc @fitzgen

crates/wasm-mutate-stats/src/bin/wasm-mutate-stats.rs
```diff
@@ -0,0 +1,327 @@
...
+            );
+        }
+    }
+}
+
+const ROTATIONS: usize = 1000000;
+
+fn main() {
+    // Init logs
+    let _ = env_logger::try_init();
```
fitzgen, line 95: This can be `env_logger::init()` instead of the fallible version, because there is no way anyone else has already initialized a logger, as this is the first line of `main`.

crates/wasm-mutate-stats/src/lib.rs
```diff
@@ -0,0 +1,15 @@
...
+    /// The input Wasm cannot be mutated
+    #[error("Failed to mutate input Wasm module.")]
+    WasmMutate(#[from] wasm_mutate::Error),
+    /// IO error
+    #[error("IO operation failure.")]
+    IO(#[from] std::io::Error),
+    /// EnvVar error
+    #[error("Environment variable interaction failure.")]
+    EnvVar(#[from] std::env::VarError),
+}
```
fitzgen, line 15: I think we can just use `anyhow::Error` instead of making custom error variants for this program.

crates/wasm-mutate-stats/src/bin/wasm-mutate-stats.rs
```diff
@@ -0,0 +1,327 @@
...
+                                                    }
+                                                }
+                                            }
+                                            Err(e) => report
+                                                .errors
+                                                .push((report.rotations as u64, format!("{}", e))),
+                                        }
+                                    }
+                                    // All generated Wasm should be valid
+                                    Err(e) => panic!("{}", e),
```
fitzgen, line 254: Can you do `validator.validate_all(&mutated).expect("all generated Wasm should be valid")` instead of `match`ing? This way the code is a little less verbose and we can cut down on the deep indentation here.

crates/wasm-mutate-stats/src/bin/wasm-mutate-stats.rs
```diff
@@ -0,0 +1,327 @@
...
+                None => {
+                    break;
+                }
+            }
+        }
+    }
+
+    fn hash(&self, data: Vec<u8>) -> String {
+        // By default md5 is the hashes here
+        format!("{:x}", Md5::digest(&data))
```
fitzgen, line 272: md5 doesn't seem like the best choice here.

Either

* we don't care a ton about using a good, wide hash to avoid collisions, and can just use the `std` library's default hasher for hash maps like

  ```rust
  use std::collections::hash_map::DefaultHasher;
  use std::hash::{Hash, Hasher};

  let mut h = DefaultHasher::default();
  data.hash(&mut h);
  h.finish()
  ```

* or we do care about using a good, wide hash to avoid collisions, in which case we want something better than md5, like [blake3](https://crates.io/crates/blake3)

crates/wasm-mutate-stats/src/bin/wasm-mutate-stats.rs
```diff
@@ -0,0 +1,327 @@
...
+    ///
+    fn compile(
+        &self,
+        mutated: &Vec<u8>,
+        modulename: &String,
+        rotation: usize,
+    ) -> Result<Vec<String>, wasm_mutate_stats::Error> {
+        // This is suuuper ugly
+        let mut buffer = fs::File::create(format!("{}.mutated", modulename))?;
+        buffer.write(&mutated)?;
```
fitzgen, line 285: You can do

```suggestion
        fs::write(format!("{}.mutated"), mutated)?;
```

crates/wasm-mutate-stats/src/bin/wasm-mutate-stats.rs
```diff
@@ -0,0 +1,327 @@
...
+    fn hash(&self, data: Vec<u8>) -> String {
+        // By default md5 is the hashes here
+        format!("{:x}", Md5::digest(&data))
+    }
+
+    /// Compile a module using `mutated` as the input Wasm
+    ///
+    fn compile(
+        &self,
+        mutated: &Vec<u8>,
```
fitzgen, line 279: Passing `&Vec<u8>` is basically the same as passing `&[u8]` but doesn't allow callers who have a `&[u8]` to call the function, forcing them to allocate a new vec and pass a reference to it unnecessarily.

```suggestion
        mutated: &[u8],
```

crates/wasm-mutate-stats/src/bin/wasm-mutate-stats.rs
```diff
@@ -0,0 +1,327 @@
...
+                let name = format!("{}.{}.{}", modulename, rotation, config.join(""));
+
+                let command = Command::new(&wasmtime_bin)
+                    .arg("compile")
+                    .arg(format!("{}.mutated", modulename))
+                    .args(config)
+                    .arg("-o")
+                    .arg(format!("{}.obj", name))
+                    .spawn()
+                    .expect("Process cannot be launched");
```
fitzgen, line 302: Rather than spawning a `wasmtime` subprocess, which requires configuring which wasmtime CLI to use and env vars and all that, we can use the `wasmtime` API to do it in this process.

```rust
// create the engine outside this function and pass it in as an argument, so that
// it is shared across all modules.
let engine = wasmtime::Engine::default();

// ...

let module = wasmtime::Module::new(&engine, &mutated)
    .expect("the mutated Wasm should be valid");
let obj = module.serialize()?
```

This way, we wouldn't need to write the mutated Wasm out to a file either.

crates/wasm-mutate-stats/src/bin/wasm-mutate-stats.rs
```diff
@@ -0,0 +1,327 @@
...
+                                match validator.validate_all(&mutated) {
+                                    Ok(_) => {
+                                        match self.compile(&mutated, &name, report.rotations) {
+                                            Ok(mhashes) => {
+                                                let hash = self.hash(mutated.clone());
+                                                wasm = mutated;
+                                                report.add_wasm_hash(hash);
+
+                                                for (idx, mhash) in mhashes.iter().enumerate() {
+                                                    report.add_obj_hash(mhash.clone(), idx);
```
fitzgen, line 237: To avoid some of this deep nesting, can you do things like

```rust
let mutated = wasmmutate.run_with_rng(&mut rng, &wasm) {
    Ok(mutated) => mutated,
    Err(e) => ...,
};

// do stuff with `mutated`...
```

instead of putting all of the <code>// do stuff with `mutated`...</code> inside the `match` arm?

crates/wasm-mutate-stats/src/bin/wasm-mutate-stats.rs
```diff
@@ -0,0 +1,327 @@
...
+                                                }
+
+                                                if report.rotations % 100 == 0 {
```
fitzgen, line 240:
```suggestion
                                                if report.rotations % 100 == 99 {
```

So that we don't print a basically-empty report on the first iteration.

But backing up a bit: why do we print out / save to disk multiple reports? Why not just one at the end? I think this would also let us avoid the loading/deserializing of reports and make the code that much simpler.

crates/wasm-mutate-stats/src/bin/wasm-mutate-stats.rs
```diff
@@ -0,0 +1,327 @@
...
+                                                    // Save report
```
fitzgen, line 241: nitpick: it is already pretty self-evident that we are saving a report given that we are calling a method named `save_report` so I don't think this comment is really helping :-p

```suggestion
```

crates/wasm-mutate-stats/src/bin/wasm-mutate-stats.rs
```diff
@@ -0,0 +1,327 @@
...
+            })
+            .expect("Report could not be written");
+    }
+
+    fn run_worker(&self, start: usize, end: usize) {
+        let mut wasmmutate = WasmMutate::default();
+        let mut corpus = self.corpus[start..end].to_vec();
+
+        loop {
+            match corpus.pop() {
```
fitzgen, line 203: Instead you can do

```rust
while let Some((name, data)) = corpus.pop() {
    // ...
}
```

crates/wasm-mutate-stats/src/bin/wasm-mutate-stats.rs
```diff
@@ -0,0 +1,327 @@
...
+                    }
+                    println!("Reached");
+                    while report.rotations <= ROTATIONS {
+                        report.rotations += 1;
+                        match wasmmutate.run_with_rng(&wasm, &mut rng) {
+                            Ok(mutated) => {
+                                let mut validator = wasmparser::Validator::new();
+                                match validator.validate_all(&mutated) {
+                                    Ok(_) => {
+                                        match self.compile(&mutated, &name, report.rotations) {
```
fitzgen, line 230: Do we want to compile in this loop? That means we are measuring how many Wasm modules we can unique mutate *and compile* in a given amount of time, not just how many unique Wasm modules we can mutate in a given amount of time.

I think we should have two separate phases:

1. The first phase is spent solely inside `wasm-mutate`, creating as many mutations as we can within the time/count limit and saving them to disk. *Not* hashing them or anything like that.
2. After the time/count limit is reached, then we do a second post-processing step that 
   * hashes the Wasm files and determines how many are unique
   * compiles the Wasm files to object files with Wasmtime (with and without optimizations)
   * hashes the object files and determines how many of these are unique

crates/wasm-mutate-stats/src/bin/wasm-mutate-stats.rs
```diff
@@ -0,0 +1,327 @@
...
+                    for _ in 0..report.rotations {
+                        match wasmmutate.run_with_rng(&wasm, &mut rng) {
+                            Ok(mutated) => {
+                                wasm = mutated;
+                            }
+                            Err(_) => { /* bypass */ }
+                        }
+                    }
+                    println!("Reached");
+                    while report.rotations <= ROTATIONS {
```
fitzgen, line 223: Supporting time limits, instead of only count limits, would allow us to easily make statements like "`wasm-mutate` generates X unique Wasm modules per hour"

Also, the count should be configurable as well (or missing if we are doing a time limit, and vice versa).

crates/wasm-mutate/src/lib.rs
```diff
@@ -161,18 +160,21 @@ impl WasmMutate {
...
             let mutator = mutators.swap_remove(i);
-            if let Ok(module) = mutator.mutate(&self, &mut rnd, &info) {
+            if let Ok(module) = mutator.mutate(&self, rng, &info) {
                 return Ok(module.finish());
             }
         }
 
         Err(Error::NoMutationsApplicable)
     }
+    /// Run this configured `WasmMutate` on the given input Wasm.
```
fitzgen, line 173: Please keep a newline in between each method.

crates/wasm-mutate/src/lib.rs
```diff
@@ -144,9 +144,8 @@ impl WasmMutate {
         self.raw_mutate_func = raw_mutate_func;
         self
     }
-
-    /// Run this configured `WasmMutate` on the given input Wasm.
-    pub fn run<'a>(&self, input_wasm: &'a [u8]) -> Result<Vec<u8>> {
+    /// Run this configured `WasmMutate` on the given input Wasm by providing a random number generator
+    pub fn run_with_rng<'a>(&self, input_wasm: &'a [u8], rng: &mut SmallRng) -> Result<Vec<u8>> {
```
fitzgen, line 148: Is this method necessary? Can we use the `wasm-mutate-stats`'s RNG to generate a new seed for each of the `WasmMutate` instances it creates instead of making `wasm-mutate`'s API bigger?

fitzgen (commented):
Lots of comments/suggestions, but we are on the right track here. Thanks @Jacarte!

Jacarte:
Thanks @fitzgen  for the patience :) and for the clippy tip, it is quite cool

fitzgen (approved):
Looks good! Thanks for the patience @Jacarte :)

PR #368, opened 2021-10-19
fuzz: Assert that original and mutated Wasm evaluate identically

fitzgen:
cc @Jacarte


fitzgen:
> I'm a little worried about the loss in coverage for other wasm features with the mutate pass

To be clear those config restrictions are there because `wasm-mutate` itself was panicking on those features, and this doesn't affect any other fuzz targets' feature sets or their coverage.

> I'm also worried about the significant increase in compile time this will have for the fuzzer

I can put the new identical-evaluation assertions behind a feature.

PR #378, opened 2021-11-01
Type propagation

Jacarte:
When inferring types for the rewriting egraph tree during the translation from Wasm to the egraph some type information is missing. However, it can be inferred by the type relations between the nodes. For example for the `i32.wrap_i64` operator, we always know that it returns an i32 constant and waits for the operand to be an i64 integer. This information could be propagated to the other nodes in the tree. 

This new implementation does exactly that, it fixes some nodes to expect and receive specific types using fixed operators and function signatures. The type tree is traversed until is converges, if for some reason, a node has not a type (when it should) this type is gathered from the egraph eclass data.

cc @fitzgen

### DIFF
--- a/crates/wasm-mutate/src/module/mod.rs
+++ b/crates/wasm-mutate/src/module/mod.rs
@@ -13,6 +13,13 @@ pub enum PrimitiveTypeInfo {
     F64,
     Empty,
 }
+
+impl PartialEq for PrimitiveTypeInfo {
+    fn eq(&self, other: &Self) -> bool {
+        self == other
+    }
+}
+
 #[derive(Debug, Clone)]
 pub struct FuncInfo {
     pub params: Vec<PrimitiveTypeInfo>,
--- a/crates/wasm-mutate/src/module/mod.rs
+++ b/crates/wasm-mutate/src/module/mod.rs
@@ -5,19 +5,13 @@ use wasmparser::{Type, TypeDef};
 
 use crate::error::EitherType;
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
 pub enum PrimitiveTypeInfo {
     I32,
     I64,
     F32,
     F64,
     Empty,
-}
-
-impl PartialEq for PrimitiveTypeInfo {
-    fn eq(&self, other: &Self) -> bool {
-        self == other
-    }
 }
 
 #[derive(Debug, Clone)]
--- a/crates/wasm-mutate/src/mutators/mod.rs
+++ b/crates/wasm-mutate/src/mutators/mod.rs
@@ -40,7 +40,7 @@ pub(crate) fn match_mutation(original: &str, mutator: &dyn Mutator, expected: &s
     let info = ModuleInfo::new(original).unwrap();
     let can_mutate = mutator.can_mutate(&wasmmutate, &info);
 
-    assert_eq!(can_mutate, true);
+    assert!(can_mutate == true);
 
     let mut rnd = SmallRng::seed_from_u64(0);
     let mutation = mutator.mutate(&wasmmutate, &mut rnd, &info);
--- a/crates/wasm-mutate/src/mutators/mod.rs
+++ b/crates/wasm-mutate/src/mutators/mod.rs
@@ -40,7 +40,7 @@ pub(crate) fn match_mutation(original: &str, mutator: &dyn Mutator, expected: &s
     let info = ModuleInfo::new(original).unwrap();
     let can_mutate = mutator.can_mutate(&wasmmutate, &info);
 
-    assert!(can_mutate == true);
+    assert!(can_mutate);
 
     let mut rnd = SmallRng::seed_from_u64(0);
     let mutation = mutator.mutate(&wasmmutate, &mut rnd, &info);
--- a/crates/wasm-mutate/src/mutators/peephole/eggsy/encoder.rs
+++ b/crates/wasm-mutate/src/mutators/peephole/eggsy/encoder.rs
@@ -1,5 +1,5 @@
 //! Helper methods for encoding eterm expressions to Wasm and back
-use std::cell::RefCell;
+use std::cell::{Cell, RefCell};
 use std::{collections::HashMap, num::Wrapping};
 
 use egg::{Id, RecExpr};
@@ -46,7 +46,6 @@ impl Encoder {
         egraph: &EG,
     ) -> crate::Result<Vec<Option<PrimitiveTypeInfo>>> {
         let nodes = expr.as_ref();
-
         let gettpe = |id: Id| {
             let eclass = node_to_eclass[usize::from(id)];
             let data = &egraph[eclass].data;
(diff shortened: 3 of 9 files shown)


crates/wasm-mutate/src/module/mod.rs
```diff
@@ -13,6 +13,13 @@ pub enum PrimitiveTypeInfo {
     F64,
     Empty,
 }
+
+impl PartialEq for PrimitiveTypeInfo {
+    fn eq(&self, other: &Self) -> bool {
+        self == other
+    }
+}
```
fitzgen, line 21: I think this can be `#[derive(PartialEq)]` on `PrimitiveTypeInfo`'s definition.

crates/wasm-mutate/src/mutators/mod.rs
```diff
@@ -40,7 +40,7 @@ pub(crate) fn match_mutation(original: &str, mutator: &dyn Mutator, expected: &s
     let info = ModuleInfo::new(original).unwrap();
     let can_mutate = mutator.can_mutate(&wasmmutate, &info);
 
-    assert_eq!(can_mutate, true);
+    assert!(can_mutate == true);
```
fitzgen, line 43:
```suggestion
    assert!(can_mutate);
```

crates/wasm-mutate/src/mutators/peephole/eggsy/encoder.rs
```diff
@@ -28,58 +28,23 @@ enum TraversalEvent {
 }
 
 impl Encoder {
-    pub(crate) fn expr2wasm(
+    /// Infers types for the rewriting egraph tree
+    /// During the translation from Wasm to the egraph some type information is missin
```
fitzgen, line 32:
```suggestion
    /// Infers types for the rewriting egraph tree.
    ///
    /// During the translation from Wasm to the egraph some type information is missing.
```

crates/wasm-mutate/src/mutators/peephole/eggsy/encoder.rs
```diff
@@ -177,14 +106,252 @@ impl Encoder {
...
+                    | Lang::Extend8S(_)
+                    | Lang::Extend16S(_)
+                    | Lang::Extend32S(_)
+                    | Lang::ExtendI32S(_)
+                    | Lang::ExtendI32U(_)
+                    | Lang::Call(_) => {}
+                }
+            }
+            // TODO, check a better wway to check if the type inferring converge
+            if format!("{:?}", previous_pass) == format!("{:?}", types) {
```
fitzgen, line 312: Yeah, we shouldn't be (ab)using string formatting for to check if we changed anything.

How about, instead of doing `types[idx] = ...` we define and use this little helper:

```rust
let mut changed = false;

let update = |index, ty| {
    let entry = &mut types[index];
    if entry.is_none() {
        *entry = Some(ty);
         changed = true;
    } else {
        assert_eq!(*entry, ty);
    }
};

// ...

if !changed {
    debug_assert!(types.iter().all(|ty| ty.is_some());
    break;
}
```

crates/wasm-mutate/src/mutators/peephole/eggsy/encoder.rs
```diff
@@ -28,58 +28,23 @@ enum TraversalEvent {
...
-        rnd: &mut rand::prelude::SmallRng,
         expr: &RecExpr<Lang>,
-        node_to_eclass: &Vec<Id>,
-        newfunc: &mut Function,
+        node_to_eclass: &[Id],
         egraph: &EG,
-    ) -> crate::Result<()> {
-        // This is a patch, this logic should be here
-        // If root is Unfold...bypass
+    ) -> crate::Result<Vec<Option<PrimitiveTypeInfo>>> {
```
fitzgen, line 45: We can always find the type of an expression, right? If so, then we should return `Result<Vec<PrimititiveTypeInfo>>` without the `Option` in there.

crates/wasm-mutate/src/mutators/peephole/eggsy/encoder.rs
```diff
@@ -242,23 +392,10 @@ impl Encoder {
                         | Lang::GeS(operands)
                         | Lang::GeU(operands) => {
                             let operands = *operands;
-                            for (idx, operand) in operands.iter().enumerate().rev() {
+                            for (_idx, operand) in operands.iter().enumerate().rev() {
```
fitzgen, line 395: If we aren't using the index anymore, then we shouldn't `.enumerate()`

crates/wasm-mutate/src/mutators/peephole/eggsy/encoder.rs
Jacarte: Done

crates/wasm-mutate/src/mutators/peephole/eggsy/encoder.rs
Jacarte: Not necessarily, some nodes are just helpers, like for example `Arg` nodes, we always should know how to write them down, but the type information is not needed for them when they are in the right operator.

On the other hand, if for some reason they are moved out of the corresponding operator, for example, the `memory-pop-out` mutator (the only case so far), its type is inferred by the fixed point since it corresponds to another operator then.

Jacarte:
ping @fitzgen. Thanks for the feedback :)

crates/wasm-mutate/src/mutators/peephole/eggsy/encoder.rs
```diff
@@ -89,80 +56,61 @@ impl Encoder {
...
+
+        let update = |index, ty| {
+            let mut types = types.borrow_mut();
+            let entry: &Option<PrimitiveTypeInfo> = &types[index];
+            if entry.is_none() {
+                types[index] = ty;
+                *changed.borrow_mut() = true;
+            } else {
+                // reset the fixed point state
+                *changed.borrow_mut() = false;
```
fitzgen, line 81: Within the same iteration, if we call `update` for one node and it sets `changed = true`, but then we call `update` again with another node that is already set, this will reset `changed = false` and that's not what we want here. Instead, we should reset `changed` once per iteration:

```rust
let mut changed = false;

// ... update definition, get definition, etc...

loop {
    changed = false;

    // ...

    if !changed {
        break;
    }
}
```

crates/wasm-mutate/src/mutators/peephole/eggsy/encoder.rs
```diff
@@ -89,80 +56,61 @@ impl Encoder {
...
-                    gettpe(*sibling)
-                }
-                _ => None,
+        let types = nodes
+            .iter()
+            .map(|_| None)
+            .collect::<Vec<Option<PrimitiveTypeInfo>>>();
+        let types = RefCell::new(types);
+        let changed = false;
+        let changed = RefCell::new(changed);
```
fitzgen, line 65: For `Copy` types, we prefer using `std::cell::Cell` over `std::cell::RefCell` (surprised that clippy doesn't warn about this!)

crates/wasm-mutate/src/mutators/peephole/eggsy/encoder.rs
```diff
@@ -89,80 +56,61 @@ impl Encoder {
...
+
+        let get = |index| -> Option<PrimitiveTypeInfo> {
+            let types = types.borrow();
+            let item: &Option<PrimitiveTypeInfo> = &types[index];
+            item.clone()
```
fitzgen, line 70:
```suggestion
            types[index].clone()
```

fitzgen (commented):
Last couple things and this should be ready to merge.

Jacarte:
Done 😄

fitzgen (approved):
Thanks!

PR #381, opened 2021-11-03
Code motion mutator

Jacarte:
Skeleton for code-motion mutators

- First code motion mutation implemented for if-else constructions: 
  - For a random if-else, inverts the expected stack value and  swaps the encoding for `then` and `alternative` branches

cc @fitzgen

### DIFF
--- a/crates/wasm-mutate/src/mutators/codemotion/ir/mod.rs
+++ b/crates/wasm-mutate/src/mutators/codemotion/ir/mod.rs
@@ -1,0 +1,970 @@
+use std::{borrow::Borrow, cell::RefCell, collections::HashMap, io::Write, process::Command, slice::{ChunksExactMut, Iter}, time::{SystemTime, UNIX_EPOCH}};
+
+use wasm_encoder::{BlockType, Function, Instruction};
+use wasmparser::{Operator, Range};
+use crate::{error::EitherType, mutators::OperatorAndByteOffset};
+pub struct ASTBuilder;
+
+#[derive(Debug)]
+struct AST {
+    // Index of the root node
+    root: usize,
+    // Nodes of the tree
+    nodes: Vec<Node>
+}
+
+impl AST {
+
+    /// Build a new empty AST
+    pub fn new(root: usize, nodes: Vec<Node>) -> Self {
+        AST{
+            root,
+            nodes
+        }
+    }
+
+
+    /// Encodes the AST
+    /// It traverses the AST starting from the last root
+    pub fn write_ast<'a>(&self, newfunc: &mut Function, operators: &Vec<OperatorAndByteOffset>, input_wasm: &'a [u8]) -> crate::Result<()> {
+        
+        println!("AST {:?}", self);
+        
+        fn write_ast_rec<'a>(ast: &AST, current_node: &Node, newfunc: &mut Function, operators: &Vec<OperatorAndByteOffset>, input_wasm: &'a [u8]) {
+            match current_node {
+                Node::IfElse(condition, then, alternative) => {
+                    // write condition first
+                    write_ast_rec(ast, &ast.nodes[*condition], newfunc,operators, input_wasm);
+                    newfunc.instruction(&Instruction::If(BlockType::Empty));
+                    for ch in then {
+                        write_ast_rec(ast, &ast.nodes[*ch], newfunc,operators, input_wasm);
+                    }
+
+                    if let Some(alternative) = alternative {
+                        newfunc.instruction(&Instruction::Else);
+
+                        for ch in alternative {
+                            write_ast_rec(ast, &ast.nodes[*ch], newfunc,operators, input_wasm);
+                        }
+                    }
+                    newfunc.instruction(&Instruction::End);
+
+                },
+                Node::Code { range } => {
+                    let operator_range = (range.start, range.end);
+
+
+                    //print operator
+                    for operator in &operators[range.start..range.end]{
+                        println!("{:?}", operator.0)
+                    }
+
+                    let bytes_range = (&operators[operator_range.0].1, &operators[operator_range.1].1);
+                    let piece_of_code = &input_wasm[*bytes_range.0..*bytes_range.1];
+                    
+                    newfunc.raw(piece_of_code.to_vec());
+                },
+                Node::Loop(body) => {
+                    newfunc.instruction(&Instruction::Loop(BlockType::Empty));
+                    for ch in body {
+                        write_ast_rec(ast, &ast.nodes[*ch], newfunc,operators, input_wasm);
+                    }
+                    newfunc.instruction(&Instruction::End);
+
+                },
+                Node::Block(body) => {
+                    newfunc.instruction(&Instruction::Block(BlockType::Empty));
+                    for ch in body {
+                        write_ast_rec(ast, &ast.nodes[*ch], newfunc,operators, input_wasm);
+                    }
+                    newfunc.instruction(&Instruction::End);
+                },
+                Node::Root(body) => {
+
+                    for ch in body {
+                        write_ast_rec(ast, &ast.nodes[*ch], newfunc,operators, input_wasm);
+                    }
+                    // Closing end
+                    newfunc.instruction(&Instruction::End);
+                },
+            }
+        }
+
+        write_ast_rec(self, &self.nodes[self.root], newfunc, operators, input_wasm);
+
+        Ok(())
+    }
+}
+
+#[derive(Debug, Clone)]
+enum Node {
+    IfElse(usize, Vec<usize>, Option<Vec<usize>>),
+    Code{
+        range: Range
+    },
+    Loop(Vec<usize>),
+    Block(Vec<usize>),
+    Root(Vec<usize>)
+}
+
+
+impl ASTBuilder {
+
+    fn dotify_ast<'a>(&self, root: usize, nodes: &[Node], operators: &'a [OperatorAndByteOffset]) -> String {
+        let mut r = String::new();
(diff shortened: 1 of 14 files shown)


crates/wasm-mutate/src/mutators/codemotion/mutators/if_complement.rs
```diff
@@ -0,0 +1,125 @@
...
+        ast: &crate::mutators::codemotion::ir::AST,
+        newfunc: &mut wasm_encoder::Function,
+        operators: &Vec<OperatorAndByteOffset>,
+        input_wasm: &'a [u8],
+    ) -> crate::Result<()> {
+        // Select the if index
+        let if_index = ast
+            .get_ifs()
+            .choose(rnd)
+            .expect("This mutator should check first if the AST contains more than one if");
```
fitzgen, line 118: I think this is supposed to be

```suggestion
            .expect("This mutator should check first if the AST contains at least one if");
```

?

Am I understanding correctly? This doesn't need multiple `if`s, right?

crates/wasm-mutate/src/mutators/mod.rs
```diff
@@ -24,6 +25,11 @@ pub trait Mutator {
     }
 }
 
+// Helper type to return operator and ofsset inside the byte stream
+/// Type helper to wrap operator and the byte offset in the code section of a Wasm module
+pub type OperatorAndByteOffset<'a> = (Operator<'a>, usize);
```
fitzgen, line 30: The `//` comment seems like a bad copy-paste from a function that returns `OperandAndByteOffset`s?

crates/wasm-mutate/src/mutators/codemotion/ir/mod.rs
```diff
@@ -0,0 +1,501 @@
...
+    loops: Vec<usize>,
+    // indexes of block nodes
+    blocks: Vec<usize>,
+}
+
+/// Encodes the Wasm AST
+pub trait ASTWriter {
+    /// Encodes a loop node
+    /// Redefine this in case of a mutator
+    ///
```
fitzgen, line 26: In general, doc comments should have a short overview sentence first (you've done well on this point) and then more details in later paragraphs with empty lines separating paragraphs (this should be fixed in all of these doc comments here).

For example:

```suggestion
    /// Encodes a loop node.
    ///
    /// Redefine this if your implementation mutates loops. For example, if your
    /// implementation unrolls the first iteration of a particular loop, override
    /// this method. For the other loops in the Wasm that your implementation is not
    /// mutating, you can call `write_loop_default`.
```

crates/wasm-mutate/src/mutators/codemotion/ir/mod.rs
```diff
@@ -0,0 +1,501 @@
...
+            for ch in alternative {
+                self.write(ast, *ch, newfunc, operators, input_wasm)?;
+            }
+        }
+        newfunc.instruction(&Instruction::End);
+        Ok(())
+    }
+
+    /// Encodes a code node (pseudo basic block)
+    /// Redefine this in case of a mutator
```
fitzgen, line 159: Examples of when to redefine this would be nice :)

crates/wasm-mutate/src/mutators/codemotion/ir/mod.rs
```diff
@@ -0,0 +1,501 @@
...
+    ///
+    pub fn get_root(&self) -> usize {
+        self.root
+    }
+}
+
+#[derive(Debug, Clone)]
+pub enum Node {
+    /// If-Else node (then branch node indexes, alternative branch node indexes, type of the IfElse node)
+    IfElse(Vec<usize>, Option<Vec<usize>>, TypeOrFuncType),
```
fitzgen, line 299: It is easier to read these variants with named struct fields (a clue that it might be better as struct fields is if you find yourself writing comments like "(description of field 0, descr of field 1, ...)" to disambiguate the fields):

```suggestion
    /// `if`/`else` node.
    IfElse {
        /// Then branch node indices.
        consequent: Vec<usize>,
        /// Else branch node indices.
        alternative: Option<Vec<usize>>,
        /// The block type for the branches.
        ty: TypeOrFuncType
    },
```

crates/wasm-mutate/src/mutators/codemotion/ir/mod.rs
```diff
@@ -0,0 +1,501 @@
...
+            *current_parsing.borrow_mut() = vec![]
+        };
+        let pop_state = || {
+            state_stack
+                .borrow_mut()
+                .pop()
+                .expect("Missing state")
+                .borrow()
+                .clone()
+        };
```
fitzgen, line 354: I think we could remove these `RefCell`s and make this a little more idiomatic by passing in a mutable borrow of the stack state to these closures.

It may also make sense to define something like

```rust
struct ParseContext {
    current: ...,
    stack: Vec<...>,
    frames: Vec<...>,
}
```

and pass mutable references to this to the `push_state`/`pop_state` methods. Even better, `push_state` and `pop_state` could be `&mut self` methods of `ParseContext`.

crates/wasm-mutate/src/mutators/codemotion/ir/mod.rs
```diff
@@ -0,0 +1,501 @@
...
+                        let id = nodes.borrow().len();
+                        nodes.borrow_mut().push(Node::Code {
+                            range: current_code,
+                        });
+                        current_parsing.borrow_mut().push(id);
+                    }
+                    current_code.start = idx + 1;
+                    current_code.end = idx + 1;
+                    push_state();
+                    frame_stack.push((State::Else, None));
```
fitzgen, line 385: It seems like `push_state` should take `State` and `Option<TypeOrFuncType>` arguments, since every `push_state()` call has a corresponding `frame_stack.push(..)` call.

Adding them as arguments to `push_state` helps make sure that the stacks can't get out of sync.

crates/wasm-mutate/src/mutators/codemotion/ir/mod.rs
```diff
@@ -0,0 +1,501 @@
...
+                    frame_stack.push((State::Block, Some(ty)));
+                }
+                Operator::Loop { ty } => {
+                    if current_code.end != current_code.start {
+                        let id = nodes.borrow().len();
+                        nodes.borrow_mut().push(Node::Code {
+                            range: current_code,
+                        });
+                        current_parsing.borrow_mut().push(id);
+                    }
```
fitzgen, line 407: This is repeated a bunch. Perhaps it can also be a shared method on the new `ParseContext` struct? It may make sense for the `ParseContext` to contain the `nodes`, `ifs`, `loops`, etc as well and then have a `finish` method for when the parse is done for creating the root node and the final `Ast` and all that.

fitzgen (commented):
Looks like some good infrastructure for future code motion mutations! A bunch of nitpicks, but they are just that: nitpicks. Fix them all and this should be ready to merge. Thanks!

crates/wasm-mutate/src/mutators/codemotion/mod.rs
```diff
@@ -0,0 +1,335 @@
...
+
+        // Split where to start looking for mutable function
+        // In theory random split will provide a mutable location faster
+        let function_to_mutate = rnd.gen_range(0, function_count);
+        let all_readers = (0..function_count)
+            .map(|_| sectionreader.read().unwrap())
+            .collect::<Vec<FunctionBody>>();
+
+        for fidx in (function_to_mutate..function_count).chain(0..function_to_mutate) {
+            let reader = all_readers[fidx as usize];
```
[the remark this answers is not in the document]
Jacarte: I think this fix will incorrectly assign the reader

```
[reader1, reader2, ..., readern-1, readern]
[fx,         fx-1,         ...,f0,    fn-x]

```

Jacarte:
Thanks @fitzgen !

Comments addressed :)

crates/wasm-mutate/src/mutators/codemotion/ir/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+        newfunc.instruction(&Instruction::Loop(map_block_type(*ty)?));
+        for ch in body {
+            self.write(ast, *ch, newfunc, operators, input_wasm)?;
+        }
+        newfunc.instruction(&Instruction::End);
+        Ok(())
+    }
+
+    /// Default encoding for a loop node
+    /// This function is called by the defaut implementation
```
fitzgen, line 57: It looks like you missed reformatting a couple doc comments on this trait's methods.

crates/wasm-mutate/src/mutators/codemotion/ir/parse_context.rs
```diff
@@ -0,0 +1,249 @@
...
+    pub fn append_instruction_to_current_code(&mut self) {
+        self.current_code_range.end += 1;
+    }
+
+    /// Closes the parsing, creates a `Node::Root` node where the children are the current
+    /// parsed nodes in the current state.
+    ///
+    /// Returns a new Ast instance
+    ///
+    pub fn finish(&mut self) -> Ast {
```
fitzgen, line 238: This can take ownership of `self` and then we can avoid the clones:

```rust
pub fn finish(self) -> Ast {
    // ...
    Ast {
        // ...
        nodes: self.nodes,
        ifs: self.ifs,
        loops: self.loops,
        blocks: self.blocks,
    }
}
```

I think this would also make the `get_nodes` method obsolete, and we could remove that.

fitzgen (commented):
Looking good! Just a couple little things

fitzgen (approved):
Thanks!

PR #382, opened 2021-11-03
Add support for the extended-consts wasm proposal

alexcrichton:
This commit updates the upstream testsuite that we test against and at
the same time implements a new proposal added to the suite, the
extended-const proposal. This proposal adds a few more instructions that
are valid in an init-expr context such as global initialization
expressions.

The internal changes to implement this new proposal are to use an
`OperatorsValidator` in its full glory when validating a global or
element initialization expression. This means that we get all the new
operators supported for free, and an allow-list is checked beforehand
for init-expressions to ensure only valid "const" instructions are
allowed.

This also necessitated some tweaks to parsing and printing of
more-than-one-instruction expressions which are accounted for here as
well.

### DIFF
--- a/crates/wasmparser/src/validator.rs
+++ b/crates/wasmparser/src/validator.rs
@@ -14,6 +14,7 @@
  */
 
 use crate::limits::*;
+use crate::operators_validator::OperatorValidator;
 use crate::WasmModuleResources;
 use crate::{Alias, ExternalKind, Import, ImportSectionEntryType};
 use crate::{BinaryReaderError, GlobalType, MemoryType, Range, Result, TableType, TagType, Type};
@@ -180,6 +181,8 @@ pub struct WasmFeatures {
     pub exceptions: bool,
     /// The WebAssembly memory64 proposal
     pub memory64: bool,
+    /// The WebAssembly extended_const proposal
+    pub extended_const: bool,
 }
 
 impl Default for WasmFeatures {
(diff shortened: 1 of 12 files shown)


crates/wasmparser/src/validator.rs
```diff
@@ -1226,78 +1230,90 @@ impl Validator {
...
+            match &op {
+                Operator::I32Const { .. }
+                | Operator::I64Const { .. }
+                | Operator::F32Const { .. }
+                | Operator::F64Const { .. }
+                | Operator::RefNull { .. }
+                | Operator::V128Const { .. }
+                | Operator::End => {}
+
+                Operator::I32Add
```
fitzgen, line 1250:
```suggestion
                // These are valid const expressions when the extended-const proposal is enabled.
                Operator::I32Add
```

crates/wasmparser/src/validator.rs
```diff
@@ -1226,78 +1230,90 @@ impl Validator {
...
+                | Operator::I32Sub
+                | Operator::I32Mul
+                | Operator::I64Add
+                | Operator::I64Sub
+                | Operator::I64Mul
+                    if self.features.extended_const => {}
+
+                Operator::GlobalGet { global_index } => {
```
fitzgen, line 1258:
```suggestion
                // `global.get` is a valid const expression for imported, immutable globals.
                Operator::GlobalGet { global_index } => {
```

fitzgen (approved):
LGTM, just a couple comment suggestions to help future code readers.

PR #384, opened 2021-11-04
Fuel

### LINKED ISSUE #383: wasm-mutate: Timeout while fuzzing when mutating an medium-sized input
We got a fuzz-bug last night for a timeout on oss-fuzz with wasm-mutate. [This input wasm file](https://github.com/bytecodealliance/wasm-tools/files/7475680/test.wasm.gz) with seed 724246166195997197 takes a fair bit of time to mutate locally:

```
$ time ./target/release/wasm-mutate --seed 724246166195997197 ./test.wasm -o output.wasm --preserve-semantics
./target/release/wasm-mutate --seed 724246166195997197 ./test.wasm -o    8.63s user 3.23s system 99% cpu 11.859 total
```

I noticed that `--preserve-semantics` is required to make this take a fair bit of time.

For reference @Jacarte the timeout on oss-fuzz is 60 seconds for any input, and fuzzing instrumentation typically slows things down by ~15-20x, which means we need to process all inputs in <3s without actual fuzzing instrumentation. Locally with a fuzzing build it looks like it takes ~140s to complete the fuzz run. For reference this is the [fuzz input](https://github.com/bytecodealliance/wasm-tools/files/7475705/input.gz).

Jacarte:
Fixes #383 

The fuel feature will prevent timeout during mutation. Each mutator consumes, depending on the operations they perform, a quantity of "fuel".  For an instance, the `PeepholeMutator` consumes a unit of fuel for each instruction it tries to mutate for each function. When fuel reaches zero, a `NoApplicableMutation` error is raised and is carried on to the root meta mutator.

cc @fitzgen


logs.txt
```diff
@@ -0,0 +1,518 @@
+warning: unused doc comment
```
fitzgen, line 1: I don't think this file was intended to be here

crates/wasm-mutate/src/lib.rs
```diff
@@ -165,7 +215,8 @@ impl WasmMutate {
         while !mutators.is_empty() {
             let i = rng.gen_range(0, mutators.len());
             let mutator = mutators.swap_remove(i);
-            if let Ok(module) = mutator.mutate(self, &mut rng, &info) {
+            let mut resources = Resources::new(self.fuel);
+            if let Ok(module) = mutator.mutate(self, &mut rng, &info, &mut resources) {
```
fitzgen, line 219: Rather than adding another parameter to the `mutate` trait method and defining a new `Resource` type, lets just use the currently-unused `WasmMutate::consume_fuel` method.

crates/wasm-mutate/src/lib.rs
```diff
@@ -145,6 +185,16 @@ impl WasmMutate {
         self
     }
 
+    pub(crate) fn consume_fuel(&mut self, qt: u64) -> bool {
```
fitzgen, line 188: This should return `Result<()>` like `Resources::consume` does.

And it can take `&self` if the `fuel` member becomes a `std::cell::Cell`.

crates/wasm-mutate/src/lib.rs
```diff
@@ -99,10 +103,40 @@ impl Default for WasmMutate {
...
             raw_mutate_func: None,
+            fuel: u64::MAX,
         }
     }
 }
 
+/// Manages resources for the mutation and handles how the resources are consumed
+pub struct Resources {
+    fuel: u64,
+}
```
fitzgen, line 114: And then lets get rid of this separate type.

crates/wasm-mutate/src/mutators/peephole/dfg/mod.rs
```diff
@@ -414,7 +414,7 @@ impl<'a> DFGBuilder {
                     );
                 }
                 Operator::GlobalSet { global_index } => {
-                    let child = self.pop_operand(idx, true);
+                    let child = self.pop_operand(idx, true, 0); // 0 color is undefined
```
fitzgen, line 417: Does the undef color ever change? As far as I can tell it is always passing `0` so this doesn't seem like a desirable change.

Also this seems unrelated to the rest of the PR.

crates/wasm-mutate/src/mutators/peephole/eggsy/encoder.rs
```diff
@@ -1215,8 +1217,17 @@ impl Encoder {
                                 expr,
                             ));
                         }
-                        StackType::GlobalSet(_) => unreachable!(),
-                        StackType::Drop => unreachable!(),
+                        StackType::GlobalSet(_) => return Err(crate::Error::NoMutationsApplicable),
+                        StackType::Drop => {
+                            let id = ids_stack.pop().expect("The dynamic offset operand for the load operation is not in the stack");
```
fitzgen, line 1222:
```suggestion
                            let id = ids_stack.pop().expect("The dynamic offset operand for the drop operation is not in the stack");
```

Jacarte:
Thanks @fitzgen !, comments addressed :)

fitzgen (approved):
Thanks!

PR #389, opened 2021-11-08
Optimize `*.wat` parsing

alexcrichton:
This commit attempts to resolve at least some slowness in text parsing
for wasm. With some profiling the main source of slowdown for parsing
the text format appeared to be the actual utf-8 decoding of the source,
apparent because the source was iterated over `char`-by-`char`. This
decoding led to lots of branch-y code even for simple operations such as
skipping whitespace.

The change in this commit is to change the lexer of the text format to
working with `&[u8]` internally instead of `CharIndices`. This means
that the lexer will utf-8 decode as little as possible from now on and
only reading `char` values as necessary for lexing operations such as
strings. Otherwise the text format is primarily ASCII and doesn't need
to go into the realm of utf-8 decoding. Also note that the lexer still
operates over `&str` at the API-level since wasm source text is required
to be valid utf-8.

For parsing a wasm-smith-generated-module this cause the entire
`wat2wasm` operation to be 40% faster. The vast majority of the
remaining time is still spent in lexing, almost exclusively in skipping
whitespace (since heavily nested blocks lead to lots of whitespace).
Unfortunately I don't really know of an easy way to optimize skipping of
whitespace, but at least the problem is quite tractable where the goal
is to find the first-non-whitespace-byte in a list of bytes, so this
should be ripe for optimization in the future.

### DIFF
--- a/crates/wast/src/lexer.rs
+++ b/crates/wast/src/lexer.rs
@@ -28,7 +28,6 @@ use crate::{Error, Span};
 use std::borrow::Cow;
 use std::char;
 use std::fmt;
-use std::iter;
 use std::str;
 
 /// A structure used to lex the s-expression syntax of WAT files.
@@ -38,7 +37,7 @@ use std::str;
 /// returned for any non-lexable text.
 #[derive(Clone)]
 pub struct Lexer<'a> {
-    it: iter::Peekable<str::CharIndices<'a>>,
+    remaining: &'a str,
     input: &'a str,
     allow_confusing_unicode: bool,
 }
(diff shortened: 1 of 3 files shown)


alexcrichton:
As an interesting tidbit [this godbolt link](https://godbolt.org/z/d8oMzKbrv) is the generated code which is the hottest at this point effectively, what's used to skip whitespace. I've been primarily benchmarking on aarch64 so far so the asm is a bit different, but it turns out that LLVM does a surprisingly good job with optimizing this (even relative to gcc). Making this faster will probably involve simd and unfortunately I don't know of an off-the-shelf crate which would make this faster...

crates/wast/src/lexer.rs
```diff
@@ -506,236 +612,180 @@ impl<'a> Lexer<'a> {
...
-                match it.next() {
-                    None | Some((_, '\n')) => break,
-                    Some((i, ch)) => {
-                        self.check_if_confusing(i, ch)?;
-                        self.it = it.clone();
+        // In an effort to avoid utf-8 decoding the entire `comment` the search
+        // here is a bit more optimized. This checks for the `0xe2` byte because
+        // in the utf-8 encoding that's the leading encoding byte for all
+        // "confusing characters". Each instance of 0xe2 is checked to see if it
+        // starts a confusing character, and if so that's returned.
```
cfallin, line 627: Similar to above, it might be helpful to note here that `0xe2` can only appear as the first byte of a character (all continuing bytes in a utf-8 sequence are `0x80..0xbf`) so there is no need to consider the case that we catch an `0xe2` in the middle of a codepoint.

cfallin (approved):
This looks like a really nice improvement!

All of the new parser code looks reasonable to me, but I also admit I am not intimately familiar with the WAT text-format spec; but this does pass all the tests so that's a reasonable assurance I think.

It's too bad there's not a `mem-not-chr` to look for anything but a space... to whomever tackles the next optimization level you mention, it seems like that would be a generally useful contribution to that crate! (I seem to remember that clang does something with SIMD for this, as you mention; and [this library](https://github.com/lemire/despacer) would probably be a good starting point too...).)

alexcrichton:
> and [this library](https://github.com/lemire/despacer) would probably be a good starting point too

Oh nice! I switched to the lookup table which is indeed a bit faster (albeit code-size-larger) and it looks like that's faster for me locally as well. Perhaps a simd crate for this will exist one day too!

crates/wast/src/lexer.rs
```diff
@@ -266,70 +297,191 @@ impl<'a> Lexer<'a> {
...
 
-    fn token(&mut self) -> Result<Option<Token<'a>>, Error> {
-        // First two are easy, they're just parens
-        if let Some(pos) = self.eat_char('(') {
-            return Ok(Some(Token::LParen(&self.input[pos..pos + 1])));
-        }
-        if let Some(pos) = self.eat_char(')') {
-            return Ok(Some(Token::RParen(&self.input[pos..pos + 1])));
-        }
+        match byte {
```
jcchavezs, line 309: Does this block of change deserve a test? I feel the change is so big that it deserve at least a little happy path test.

alexcrichton: The preexisting test suite of this library was what I intended for that,  but have you found a specific bug with this PR?

PR #394, opened 2021-11-18
Lazy iterator for random expressions in the Egraph

Jacarte:
The extraction of a new equivalent expression is done with a lazy iterator over the nodes of the egraph.

Only by iterating the egraph until a different than the original expression appears (with a maximum depth of 6) seems very promising:

```rust
let expr = lazy_expand(root, egraph.clone(), 6, &rnd_ref);
for expr in expr {
    config.consume_fuel(1)?;
    if expr.eq(&start.to_string()) {
        continue;
    }
    // Let the parser do its job
    let compiled =
        RecExpr::<Lang>::from_str(&expr).expect("Invalid parsing");
...
    return Ok((newfunc, fidx));
}
```

From a previous mutation ratio of `13.50%` it was raised to `21%`

Because we have a few things to do I will open this PR as draft:

- [ ] Better fuel consumption in the peephole iterator
- [ ] Migrate all mutators to be iterators
- [x] The random expression is built by creating the s-expression with Strings. This will help us to use the Language implementation for all kind of generating and compiling tasks. Still it might be not performant and we need to gather some profiling metrics.


Jacarte:
cc @fitzgen

fitzgen:
> * [ ]  Better fuel consumption in the peephole iterator
>
> * [ ] Migrate all mutators to be iterators

These two seem like they could be in follow up PRs.

> * [ ] The random expression is built by creating the s-expression with Strings

This is something we should address before merging this. We should construct the structures directly, instead of creating a string just to then parse the string.

Jacarte:
> > * [ ]   Better fuel consumption in the peephole iterator
> > * [ ]  Migrate all mutators to be iterators
> 
> These two seem like they could be in follow up PRs.
> 
> > * [ ]  The random expression is built by creating the s-expression with Strings
> 
> This is something we should address before merging this. We should construct the structures directly, instead of creating a string just to then parse the string.

Agreed :)

Jacarte:
> > * [ ]   Better fuel consumption in the peephole iterator
> > * [ ]  Migrate all mutators to be iterators
> 
> These two seem like they could be in follow up PRs.
> 
> > * [ ]  The random expression is built by creating the s-expression with Strings
> 
> This is something we should address before merging this. We should construct the structures directly, instead of creating a string just to then parse the string.

Actually, by constructing the structure by ourselves we can missuse the egraph capability of grouping equal nodes. Besides, this will be more likely to have errors

Jacarte:
> > * [ ]   Better fuel consumption in the peephole iterator
> > * [ ]  Migrate all mutators to be iterators
> 
> These two seem like they could be in follow up PRs.
> 
> > * [ ]  The random expression is built by creating the s-expression with Strings
> 
> This is something we should address before merging this. We should construct the structures directly, instead of creating a string just to then parse the string.

Done :)

crates/wasm-mutate/src/mutators/peephole/eggsy/encoder/mod.rs
```diff
@@ -27,7 +29,7 @@ impl Encoder {
     /// Reassembles the mutated function and return a `Function` entry
     pub fn build_function(
         info: &ModuleInfo,
-        rnd: &mut rand::prelude::SmallRng,
+        rnd: &RefCell<&mut rand::prelude::SmallRng>,
```
fitzgen, line 32: Wrapping this in a `RefCell` doesn't seem like it should be necessary, can you explain why we need it here?

crates/wasm-mutate/src/mutators/peephole/eggsy/expr_enumerator.rs
```diff
@@ -0,0 +1,877 @@
...
+    eggsy::{encoder::rebuild::build_expr_inner, RandomExtractor},
+    EG,
+};
+use egg::{AstSize, Id, RecExpr};
+use rand::{prelude::SmallRng, Rng};
+
+use super::lang::Lang;
+
+macro_rules! binop {
+    ($lang:ident, $left:ident, $right: ident, $egraph: ident, $rnd: ident, $depth: ident, $recexpr: ident) => {{
```
fitzgen, line 13: Is there a reason that these need to be macros and can't be functions?

Also, some doc comments would be really appreciated here.

crates/wasm-mutate/src/mutators/peephole/eggsy/expr_enumerator.rs
```diff
@@ -0,0 +1,877 @@
...
+}
+
+/// Returns a lazy iterator over random expressions in the eggraph
+///
+pub fn lazy_expand<'a>(
+    id: Id,
+    egraph: EG,
+    depth: u32,
+    rnd: &'a RefCell<&'a mut SmallRng>,
+    recexpr: &'a RefCell<RecExpr<Lang>>,
```
fitzgen, line 103: It isn't clear to me why this needs to be in a `RefCell` as well.

In general, we should try and avoid `RefCell`s unless there is a really good reason, as it turns compile-time guarantees to run-time checks.

crates/wasm-mutate/src/mutators/peephole/eggsy/expr_enumerator.rs
```diff
@@ -0,0 +1,877 @@
...
+
+        Box::new(t)
+    }};
+}
+
+/// Returns a lazy iterator over random expressions in the eggraph
+///
+pub fn lazy_expand<'a>(
+    id: Id,
+    egraph: EG,
```
fitzgen, line 100: Can we make this parameter `&'a EG` so that we don't have to clone the whole e-graph below?

That does mean that the returned iterator borrows the e-graph, but it is already `+ 'a` so that shouldn't be a problem.

crates/wasm-mutate/src/mutators/peephole/eggsy/mod.rs
```diff
@@ -104,33 +104,24 @@ where
...
-    /// one of its e-nodes at random and then do the same with its children,
-    /// etc. You can imagine this process as a kind of rolling wave function
-    /// collapse, where we choose a concrete expression out of the potentially
-    /// infinite number of equivalent expressions each e-class represents.
+    /// costless with its children.
     ///
-    /// `worklist` constains the operands we still need to process. These are
-    /// currently a pair of the parent node and its operand e-class.
-    pub fn extract_random(
+    pub fn extract_shorter(
```
fitzgen, line 109: Maybe call this `extract_smallest`?

crates/wasm-mutate/src/mutators/peephole/eggsy/mod.rs
```diff
@@ -104,33 +104,24 @@ where
     }
 
     /// Do a pre-order traversal of the e-graph. As we visit each e-class, choose
-    /// one of its e-nodes at random and then do the same with its children,
-    /// etc. You can imagine this process as a kind of rolling wave function
-    /// collapse, where we choose a concrete expression out of the potentially
-    /// infinite number of equivalent expressions each e-class represents.
+    /// costless with its children.
```
fitzgen, line 107:
```suggestion
    /// choose the smallest e-node (according to the cost function).
```

crates/wasm-mutate/src/mutators/peephole/rules.rs
```diff
@@ -25,12 +23,20 @@ impl PeepholeMutator {
...
+            rewrite!("select-reduce";  "(select ?x ?y ?y)" => "?y"),
         ];
         // Use a custom instruction-mutator for this
         // This specific rewriting rule has a condition, it should be appplied if the operand is a constant
         rules.extend(rewrite!("strength-undo";  "(i32.shl ?x 1_i32)" <=> "(i32.mul ?x 2_i32)"));
         rules.extend(rewrite!("strength-undo01";  "(i64.shl ?x 1_i64)" <=> "(i64.mul ?x 2_i64)"));
 
+        // invert select
+        rules.extend(
+            rewrite!("select-invert";  "(select ?x ?y ?z)" <=> "(select (i32.eqz ?x) ?z ?y)"),
```
fitzgen, line 37: Nice

fitzgen (commented):
Looks pretty good! A few comments below that we should resolve before merging.

Also some comments explaining the high-level approach to iteratively extracting expressions from the e-graph would be very appreciated, and help future code readers a ton.

crates/wasm-mutate/src/mutators/peephole/eggsy/encoder/mod.rs
Jacarte: It is passed through several closures, I think this prevents borrowing issues

crates/wasm-mutate/src/mutators/peephole/eggsy/expr_enumerator.rs
Jacarte: Mostly to avoid repetition

crates/wasm-mutate/src/mutators/peephole/eggsy/expr_enumerator.rs
Jacarte: Comments above

crates/wasm-mutate/src/mutators/peephole/eggsy/expr_enumerator.rs
fitzgen: Which repetition can't be avoided with a helper function rather than a helper macro? AFAICT, a function should work here.

crates/wasm-mutate/src/mutators/peephole/eggsy/encoder/mod.rs
fitzgen: Rather than having the closures capture the RNG, it could be a parameter passed to each of them. This should remove the need for the `RefCell`

fitzgen: Or is the issue that the closures that capture the RNG aren't things you can add new parameters to, but instead have to conform to an existing interface, like `Iterator` methods?

Basically this error: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e63e24bfcf10504ff8b3927803479bd9

If so, it should be possible to work around this by refactoring the sequence of `Iterator` methods that are chained into a single `filter_map` or `flat_map` call rather than a series of `filter`s and `map`s.

crates/wasm-mutate/src/mutators/peephole/eggsy/expr_enumerator.rs
Jacarte: Well, not only that, I a using them to generate the right no construction `Lang::$lang`

crates/wasm-mutate/src/mutators/peephole/eggsy/encoder/mod.rs
Jacarte: What if we use `Rc` and `RefCell` to remove the references from the method signature.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=2c6ec744fd1fd8dde78334221aafd369

Issue #391, opened 2021-11-11
wasm-mutate: Mutate use-of-value to instead be use-of-global


alexcrichton:
In Wasmtime one thing we try to optimize is memory operands in the generated code, folding address operations into instructions themselves instead of having a separate load instruction. This is a pretty complex part of Wasmtime and it's a good one to stress while fuzzing, so while talking today with @fitzgen I had the idea where a possible mutator for wasm when preserving semantics could be:

* For any time a value of type `T` is used
* Insert a new `(global $new_global (mut T) (T.const $zero))`
* Just before the usage do `global.set $new_global global.get $new_global`

At least in wasmtime `global.{get,set}` are implemented with raw loads/stores so this could help stress the logic for folding loads/stores into other surrounding instructions.

cc @Jacarte

Jacarte:
Cool ! thanks

fitzgen:
Probably one invocation of `wasm-mutate` would only do this for one type in one function. No need to rewrite the whole binary every time, we can do it incrementally via multiple calls into `wasm-mutate`.

Jacarte:
With the current implementation of the peepholemutator we can create a custom rewriting rule. Something like `?x => use_new_global`. The custom `use_new_global` operator encodes the instructions `global.set $new global.get $new` and also modifies the module structure to add the new global.

fitzgen:
Fixed in #397

PR #397, opened 2021-11-29
"use of global instead of use of value" mutator

Jacarte:
This mutator implements the idea in #391 , cc @alexcrichton.

It is implemented as an special node in the intermediate `eterm` representation, `use_of_global`. This special node allocates for a new global in the module and writes down the corresponding global usage `global.set $allocated global.get $allocated`. The rewriting rule is then `?x => (use_of_global ?x) if returns_something("?x")` . 

I will open this PR as draft for now since it is based on the PR #394 and it is under reviewing now.

cc @fitzgen


alexcrichton:
Ah that's a good point yeah about blowing limits, would it be possible for this to mutator to only be a candidate if there's less than N globals already in the module?

Jacarte:
> Ah that's a good point yeah about blowing limits, would it be possible for this to mutator to only be a candidate if there's less than N globals already in the module?

It is perfectly possible, something like adding another filter to `?x => (use_of_global ?x) if returns_something("?x")`, to get something like `?x => (use_of_global ?x) if returns_something("?x") if no_more_than_n_globals(ASTINFO, MAX_GLOBALS)`

Jacarte:
Closing, rebasing and readdressing in #408

PR #398, opened 2021-11-29
FIX: enode matching

Jacarte:
The matching process during the construction of the enodes was incorrect. It was matching only the `enum` discriminant and this leads to match two any constants as the same enode. The same for immediate having oeprators.

cc @fitzgen

### DIFF
--- a/crates/wasm-mutate/src/mutators/peephole/eggsy/lang.rs
+++ b/crates/wasm-mutate/src/mutators/peephole/eggsy/lang.rs
@@ -310,7 +310,6 @@ pub enum Lang {
         mem: u32,
         value_and_offset: [Id; 2],
     },
-    //Select([Id; 3]),
     // TODO add the others
 
     // Custom mutation operations and instructions
@@ -329,7 +328,6 @@ pub enum Lang {
     */
     UnfoldI32(Id),
     UnfoldI64(Id),
-
     /*
         Just a wrapper of multiple nodes, when encoding to Wasm it is written as nothing
         Its only responsability is to add stack neutral operations (if semantic equivalence is set)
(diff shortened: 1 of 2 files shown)


crates/wasm-mutate/src/mutators/peephole/eggsy/lang.rs
```diff
@@ -878,9 +872,91 @@ impl Lang {
...
+        match_mem!(F64Store, self, other);
+        match_mem!(I32Store8, self, other);
+        match_mem!(I32Store16, self, other);
+        match_mem!(I64Store8, self, other);
+        match_mem!(I64Store16, self, other);
+        match_mem!(I64Store32, self, other);
+
+        match (self, other) {
+            (Lang::I32(v), Lang::I32(v2)) => {
+                ::std::mem::discriminant(self) == ::std::mem::discriminant(other) && v == v2
```
fitzgen, line 929: These discriminant comparisons here and elsewhere will always be true given that we've already entered the match arm where `self` is a `Lang::I32(_)` and `other` is a `Lang::I32(_)` as well.

Same for most of the other cases below.

fitzgen (approved):
LGTM with one nitpick below. Thanks!

crates/wasm-mutate/src/mutators/peephole/eggsy/lang.rs
Jacarte: Done :)

crates/wasm-mutate/src/mutators/peephole/eggsy/lang.rs
```diff
@@ -878,9 +872,80 @@ impl Lang {
...
+            (Lang::I64(v), Lang::I64(v2)) => { v == v2
+            }
+            (Lang::F32(v), Lang::F32(v2)) => { v == v2
+            }
+            (Lang::F64(v), Lang::F64(v2)) => { v == v2
+            }
+            (Lang::GlobalGet(v), Lang::GlobalGet(v2)) => {v == v2
+            }
+            (Lang::LocalGet(v), Lang::LocalGet(v2)) => {v == v2
+            }
```
fitzgen, line 938: This looks like it could use a `rustfmt`

Jacarte: Right...forgot to make a pass 🤦

fitzgen (approved):

PR #400, opened 2021-11-30
wasm-smith: restrict which kinds of instructions can be generated

abrown:
In looking into https://github.com/bytecodealliance/wasmtime/issues/3251, I created a
mechanism for restricting what kinds of instructions wasm-smith can
generate. The [WebAssembly specification](https://webassembly.github.io/spec/core/syntax/instructions.html)
organizes its instructions into several categories (e.g., numeric,
vector, reference, control, etc.) and this change allows the user to
configure the module generation based on these categories:

```
head -c 10000 /dev/urandom | cargo run --bin wasm-smith -- --allowed-instructions memory,parametric -o test.wasm && wasm2wat test.wasm
```

There is some related configuration in wasm-smith to restrict what
instructions are available. Currently, the wasm-smith configuration is
organized around "proposals," which can be enabled or disabled. In
theory, a user could be confused if the proposal was disabled and they
explicitly enabled an instruction kind (e.g. reference)--"why aren't
reference instructions being generated?" But this accident seems
unlikely: `--allowed-instructions` defaults to enabling all kinds, so
the user would have to explicitly filter out some kind, deliberately
shooting themselves in the foot.

Despite some risk of confusion (mitigated by the documentation in this
PR), this filtering of instructions kinds ends up being useful in a
general way: not only is it a start at fixing the issue above, it is
useful for work I am doing to fuzz only parts of the spec.


fitzgen:
What happens if I only enable reference instructions? `table.get` and `table.set` require integers for which table index to operate on. Is it expected that the user needs to also enable numeric instructions then?

Also "numeric" is a pretty big umbrella. I could see value in enabling integers but not floats, for example. Or vice versa.

fitzgen:
Overall though, I do like the kind of thing this is getting at.

I wonder if we should have a method on `Config` to just directly filter whether each instruction is generatable.

fitzgen:
> I wonder if we should have a method on `Config` to just directly filter whether each instruction is generatable.

To clarify: **not** a method per instruction, but something like this:

```rust
pub trait Config {
    // ...

    fn generate_instruction(&self, inst: &Instruction) -> bool { true }
}
```

abrown:
Do you mean that `Instruction` is an enumeration so that one could select the exact set of instructions that are possible to generate? If so, I think that does make sense--I mean, having that level of detailed control would be helpful--but that's probably something that could be in the future, not as a part of this PR.

fitzgen:
`Instruction` would be the actual `crate::Instruction` type, but a default instance that represents all instances of that instruction variant, so the method can say "I am never going to allow `table.set`" without worrying about the particular table immediate.

But this is a bit of an aside and could be useful in addition to / separate from what we have here.

fitzgen (approved):
I think we can land this, and maybe introduce more similar things in follow ups, as discussed in the comments.

alexcrichton:
Is this perhaps something where the swarm testing idea would work well? I could imagine that if a bit of some input indicated whether we could generate a particular instruction we could presumably get a lot more coverage? Not necessarily at the instruction group level but rather individual instruction level.

PR #401, opened 2021-12-01
Introduce the `wasm-shrink` WebAssembly test case reducer

fitzgen:
`wasm-shrink` is a test case shrinker for WebAssembly. It shrinks a Wasm file while preserving an interesting property (such as triggering a bug in your Wasm compiler).

cc @Jacarte

### DIFF
--- a/crates/wasm-shrink/src/lib.rs
+++ b/crates/wasm-shrink/src/lib.rs
@@ -1,0 +1,303 @@
+//! Shrink a Wasm file while maintaining a property of interest (such as
+//! triggering a compiler bug).
+//!
+//! See the [`WasmShrink`][WasmShrink] type for details.
+
+use anyhow::{Context, Result};
+use rand::{rngs::SmallRng, Rng, SeedableRng};
+use wasm_mutate::WasmMutate;
+
+#[rustfmt::skip]
+static EMPTY_WASM: &'static [u8] = &[
+    // Magic.
+    0x00, b'a', b's', b'm',
+    // Version.
+    0x01, 0x00, 0x00, 0x00,
+];
+
+#[cfg_attr(
+    not(feature = "structopt"),
+    doc = r###"
+Shrink a Wasm file while maintaining a property of interest (such as
+triggering a compiler bug).
+
+# Example
+
+```
+use wasm_shrink::WasmShrink;
+
+# fn foo() -> anyhow::Result<()> {
+// Get the Wasm you want to shrink from somewhere.
+let my_input_wasm: Vec<u8> = todo!();
+
+// Configure the shrinking task.
+let shrink = WasmShrink::default()
+    // Give up on shrinking after 999 failed attempts to shrink a given
+    // Wasm test case any further.
+    .attempts(999);
+
+// Run the configured shrinking task.
+let info = shrink.run(
+    my_input_wasm,
+    // Predicate.
+    &mut |wasm| {
+        let is_interesting: bool = todo!(
+            "check for whether the given Wasm is interesting"
+        );
+        Ok(is_interesting)
+    },
+    // Callback called each time we find a new smallest interesting
+    // Wasm.
+    &mut |new_smallest| {
+        // Optionally do something with the new smallest Wasm.
+        Ok(())
+    },
+)?;
+
+// Get the shrunken Wasm and other information about the completed shrink
+// task from the returned `ShrinkInfo`.
+let shrunken_wasm = info.output;
+# Ok(()) }
+```
+"###
+)]
+#[cfg_attr(feature = "structopt", derive(structopt::StructOpt))]
+pub struct WasmShrink {
+    /// The number of shrink attempts to try before considering a Wasm module as
+    /// small as it will ever get.
+    #[cfg_attr(feature = "structopt", structopt(short, long, default_value = "1000"))]
+    attempts: u32,
+
+    /// Allow shrinking the input down to an empty Wasm module.
+    ///
+    /// This is usually not desired and typically reflects a bug in the
+    /// predicate script.
+    #[cfg_attr(feature = "structopt", structopt(long))]
+    allow_empty: bool,
+
+    /// The RNG seed for choosing which size-reducing mutation to attempt next.
+    #[cfg_attr(feature = "structopt", structopt(short, long, default_value = "42"))]
+    seed: u64,
+}
+
+impl Default for WasmShrink {
+    fn default() -> Self {
+        WasmShrink {
+            attempts: 1000,
+            allow_empty: false,
+            seed: 42,
+        }
+    }
+}
+
+impl WasmShrink {
+    /// Set the number of shrink attempts to try before considering a Wasm
+    /// module as small as it will ever get.
+    pub fn attempts(mut self, attempts: u32) -> WasmShrink {
+        self.attempts = attempts;
+        self
+    }
+
+    /// Is it okay to shrink the input down to an empty Wasm module?
+    ///
+    /// This is usually not desired and typically reflects a bug in the
+    /// predicate.
+    pub fn allow_empty(mut self, allow_empty: bool) -> WasmShrink {
+        self.allow_empty = allow_empty;
+        self
+    }
+
+    /// Set the RNG seed for choosing which size-reducing mutation to attempt
+    /// next.
+    pub fn seed(mut self, seed: u64) -> WasmShrink {
+        self.seed = seed;
+        self
+    }
+
+    /// Run this configured Wasm shrinking task.
+    ///
+    /// The `predicate` function is called on each candidate Wasm to determine
+    /// whether the Wasm is interesting or not. Returning `true` means that it
+    /// is interesting, `false` means that it is not.
+    ///
+    /// The `on_new_smallest` function is called whenever we find a new smallest
+    /// interesting Wasm.
+    ///
+    /// Returns information and metrics about the shrink task, such as the
+    /// shrunken Wasm's output file path, the size of the input Wasm, the size
+    /// of the output Wasm, etc.
+    pub fn run<P, I, S>(
+        self,
+        input: Vec<u8>,
+        mut predicate: P,
+        mut on_new_smallest: S,
+    ) -> Result<ShrinkInfo>
+    where
+        P: FnMut(&[u8]) -> Result<I>,
+        I: IsInteresting,
+        S: FnMut(&[u8]) -> Result<()>,
+    {
+        let input_size = input.len() as u64;
+
+        // Prerequisites for the input Wasm.
+        self.validate_wasm(&input)
+            .context("The input is not valid Wasm.")?;
+
+        // First double check that the input Wasm passes the predicate.
+        //
+        // It is fairly common to accidentally write a predicate script that
+        // doesn't consider the input Wasm interesting. Better to surface this
+        // user error as quick as possible than to make them wait until we've
+        // exhausted all the ways we could shrink it further.
+        let result = predicate(&input)?;
+        anyhow::ensure!(
+            result.is_interesting(),
+            "The predicate does not consider the input Wasm interesting: {}",
+            result
+        );
+
+        // The smallest Wasm that passes the predicate.
+        let mut best = input;
+
+        let mut new_best = |old_best: &mut Vec<u8>, new: Vec<u8>| -> Result<()> {
+            debug_assert!(
+                new.len() < old_best.len() || (new == EMPTY_WASM && *old_best == EMPTY_WASM)
+            );
+            on_new_smallest(&new)?;
+            *old_best = new;
+            Ok(())
+        };
+
+        // Next try running the predicate on an empty Wasm module.
+        //
+        // It is fairly common to accidentally write a predicate script that
+        // considers the empty module interesting, and we might as well check
+        // for it eagerly, rather than make the user wait forever until we
+        // finally to reduce the whole Wasm module to nothing.
+        let result = predicate(EMPTY_WASM)?;
+        if result.is_interesting() {
+            if self.allow_empty {
+                new_best(&mut best, EMPTY_WASM.to_vec())?;
+                return Ok(ShrinkInfo {
+                    input_size,
+                    output_size: best.len() as u64,
+                    output: best,
+                });
+            } else {
+                anyhow::bail!(
+                    "The predicate considers the empty Wasm module \
+                     interesting, which is usually not desired and \
+                     is a symptom of a bug in the predicate:\n\
+                     \n\
+                     {}",
+                    result
+                );
+            }
+        }
+
+        // Now we perform the main search. Keep trying to find smaller and
+        // interesting variants of the current smallest interesting Wasm file
+        // until we run out of attempts and get stuck.
+
+        let mut rng = SmallRng::seed_from_u64(self.seed);
+
+        loop {
+            let mut new_smallest = None;
+
+            'attempts: for _ in 0..self.attempts {
+                let mut mutate = WasmMutate::default();
+                mutate.reduce(true).seed(rng.gen());
+
+                let mutations = match mutate.run(&best) {
+                    Ok(m) => m,
+                    Err(_) => {
+                        // This mutation failed, but another randomly chosen
+                        // mutation might succeed, so keep attempting.
+                        continue;
+                    }
+                };
+
+                for mutated_wasm in mutations.filter_map(|w| w.ok()) {
+                    log::debug!("Testing candidate of size {}", mutated_wasm.len());
+
+                    // TODO: This "shouldn't" happen when we set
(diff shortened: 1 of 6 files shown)


crates/wasm-shrink/src/main.rs
```diff
@@ -0,0 +1,27 @@
+use anyhow::Result;
+use structopt::StructOpt;
+use wasm_shrink::WasmShrink;
+
+fn main() -> Result<()> {
```
alexcrichton, line 5: Could this be located at `src/bin/wasm-shrink.rs` with the rest of our binaries?

I was looking at https://github.com/WebAssembly/wasp the other day as well and noticed how the CLI organization there was one tool with subcommands which personally I feel like maps a lot better to what this suite of `wasm-tools` we have in our repository are. One day I'd like to move to that model to reduce duplication between all the tools ideally, but doesn't necessarily have to happen here I think.

crates/wasm-shrink/src/main.rs
```diff
@@ -0,0 +1,27 @@
...
+    let shrink = WasmShrink::from_args();
+
+    let mut config = simplelog::ConfigBuilder::new();
```
alexcrichton, line 8: Ideally I think we'd keep all the various CLI tools consistent, so could the others be switched to `simplelog` or could this use `env_logger`?

crates/wasm-shrink/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+
+            // We'll never enable this here.
+            deterministic_only: false,
+        });
+
+        Ok(validator.validate_all(&wasm).is_ok())
+    }
+
+    fn test_file(&self, file: &Path) -> Result<bool> {
+        let status = std::process::Command::new(&self.predicate)
```
alexcrichton, line 324: Given that this is a library could this interestingness-test be a configuration option that's a closure provided by the caller? I think that the CLI would implement the closure with subprocess spawning but otherwise if used-as-a-crate I could imagine that going through a subprocess would be a bummer.

crates/wasm-shrink/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+            .arg(file)
+            .stdin(std::process::Stdio::null())
+            .stdout(std::process::Stdio::null())
+            .stderr(std::process::Stdio::null())
```
alexcrichton, line 328: FWIW this is one of my major gripes with other shrinking tools, I think that stdin should be closed but stdout/stderr should be captured and presented in error messages. For example if the empty module is "interesting" it's almost always a bug in my script and it'd be nice if `wasm-shrink` went ahead and gave me the output so I didn't have to go figure out how to do it myself to debug why the empty odule is interesting.

crates/wasm-shrink/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+            .with_context(|| {
+                format!(
+                    "Failed to spawn predicate script '{}'",
+                    self.predicate.display()
+                )
+            })?;
+
+        let code = status.code().expect(
+            "`wasm-shrink` doesn't work on platforms where child processes don't have exit codes",
+        );
```
alexcrichton, line 339: I don't think that this should be an `unwrap`, this means that a signal happened on Unix, e.g. a segfault, in which case I believe that should be printed and the stdout/stderr should also be presented as well.

crates/wasm-shrink/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+            log_level: LogLevel::Info,
+            predicate: predicate.into(),
+            input: input_wasm.into(),
+        }
+    }
+
+    /// The output file path to write the shrunken Wasm file to.
+    ///
+    /// By default, a file path based on the input will be generated.
+    pub fn output(mut self, output: impl Into<PathBuf>) -> WasmShrink {
```
alexcrichton, line 152: Obligatory grumble about `&mut -> &mut` being a better builder pattern

crates/wasm-shrink/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+        anyhow::ensure!(
+            self.input.is_file(),
+            "The input Wasm file '{}' does not exist.",
+            self.input.display()
+        );
+        anyhow::ensure!(
+            self.is_valid_wasm(&self.input)?,
+            "The input Wasm file '{}' is not valid Wasm.",
+            self.input.display(),
+        );
```
alexcrichton, line 192: Is this required? I would imagine that reduction could sometimes be useful for even invalid wasm files, e.g. in a wasm file that panics in the validator or something like that.

crates/wasm-shrink/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+        // user error as quick as possible than to make them wait until we've
+        // exhausted all the ways we could shrink it further.
+        anyhow::ensure!(
+            self.test_file(&self.input)?,
+            "The predicate does not consider the input Wasm file interesting."
+        );
+
+        // The smallest Wasm that passes the predicate.
+        let mut best = vec![];
+        read_file(&self.input, &mut best)?;
```
alexcrichton, line 215: This is the only usage of `read_file` so having a special function that takes a destination-as-argument may not be too important? Could this use `std::fs::read`?

crates/wasm-shrink/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+            tail_call: true,
+            multi_memory: true,
+            exceptions: true,
+            memory64: true,
+
+            // We'll never enable this here.
+            deterministic_only: false,
+        });
+
+        Ok(validator.validate_all(&wasm).is_ok())
```
alexcrichton, line 320: FWIW sharing code with `wasm-validate-rs.rs` would enable this to use parallel validation, although it's not the most important thing in the world.

crates/wasm-shrink/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+        // for it eagerly, rather than make the user wait forever until we
+        // finally to reduce the whole Wasm module to nothing.
+        if self.test_bytes(EMPTY_WASM)? {
+            new_best(&mut best, EMPTY_WASM.to_vec())?;
+            log::warn!(
+                "The predicate considers the empty Wasm module \
+                 interesting. This is usually not desired and \
+                 is a symptom of a bug in the predicate script."
+            );
+        } else {
```
alexcrichton, line 244: Could this `else` simply follow the `if` above unconditionally? (avoiding extra indentation)

crates/wasm-shrink/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+    #[cfg_attr(feature = "structopt", structopt(short, long))]
+    output: Option<PathBuf>,
+
+    /// The number of shrink attempts to try before considering a Wasm module as
+    /// small as it will ever get.
+    #[cfg_attr(feature = "structopt", structopt(short, long, default_value = "1000"))]
+    attempts: u32,
+
+    /// The RNG seed for choosing which size-reducing mutation to attempt next.
+    #[cfg_attr(feature = "structopt", structopt(short, long, default_value = "42"))]
```
alexcrichton, line 92: I'm not familiar with general shrinking techniques, but would it make sense for the default value to be a random number instead?

crates/wasm-shrink/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+
+                            // This "shouldn't" happen when we set `.reduce(true)`
+                            // on `WasmMutate`, but we don't want to panic in the
+                            // middle of a long `wasm-shrink` session. Also the
+                            // `wasm-shrink` user isn't in a position to do anything
+                            // useful about this, even if we had a warning message,
+                            // so just ignore it.
+                            if mutated_wasm.len() >= best.len() {
+                                continue;
+                            }
```
alexcrichton, line 270: This seems like something it'd be good to have a `debug_assert!` for which could be tested while fuzzing and on our own CI?

crates/wasm-shrink/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+    let mut file = std::fs::File::open(path)
+        .with_context(|| format!("Failed to open '{}'", path.display()))?;
+    file.read_to_end(into)
+        .with_context(|| format!("Failed to read '{}'", path.display()))?;
+    Ok(())
+}
+
+fn write_file(path: &Path, contents: &[u8]) -> Result<()> {
+    std::fs::write(path, contents)
+        .with_context(|| format!("Failed to write to the file '{}'", path.display()))?;
```
alexcrichton, line 375: Personally I frequently ctrl-c the shrinking program, and given that it might make sense to have a more "atomic" version of this perhaps? Could this be done by writing to a file and then doing a move-and-replace? That way the output file should always be a valid reduction.

crates/wasm-shrink/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+    Ok(())
+}
+
+/// Information about a completed shrinking run.
+pub struct ShrinkInfo {
+    /// The path where the shrunken file was written to.
+    pub output: PathBuf,
```
alexcrichton, line 382: From a library perspective I think this would make more sense as `Vec<u8>`. Ideally the library version of this crate would have zero file operations baked in by default, and only the CLI embedding of the library would bring in the file operations

alexcrichton (commented):
Nice!

Could this add some tests as well to CI? I'm imagining we could have some simple tests like "make sure reduction eventually removes all the functions in this module" or "make sure eventually reduction makes `unreachable` show up somewhere" or something like that.

crates/wasm-shrink/src/main.rs
fitzgen: I wanted to be able to do `cargo install wasm-shrink`, but yeah if we make `wasm-tools {shrink,smith,wasm2wat,etc...}` I would be happy with that.

crates/wasm-shrink/src/main.rs
fitzgen: The thing is that this is using `log::*` macros for *all* output, because I wanted something that would be compatible with either CLI or library usage where you don't necessarily want to write to stdout/stderr, so if the CLI uses `env_logger`, then you get zero output by default, which is pretty confusing for this kind of long-running reduction process.

fitzgen: So, I don't think it makes sense to switch other tools to `simplelog` and I don't think it makes sense to switch this tool to `env_logger`.

crates/wasm-shrink/src/lib.rs
fitzgen: It will be a little funky to get this to work with `structopt` but should be doable.

crates/wasm-shrink/src/lib.rs
fitzgen: Unfortunately, `wasm-mutate` won't really work on invalid Wasm files. We could fix that, but until then I think this should stay.

crates/wasm-shrink/src/lib.rs
fitzgen: I don't think it matters either way, and this way we can get deterministic shrinking by default, which is relatively rare and nice to have.

crates/wasm-shrink/src/lib.rs
fitzgen: I think we would trip the assertion right now, so I'd rather do this in a follow up where we also fix `wasm-mutate`'s `reduce(true)` at the same.

crates/wasm-shrink/src/lib.rs
fitzgen: Good idea!

crates/wasm-shrink/src/lib.rs
fitzgen: Yeah this is just a bit harder to do with the builder-is-the-`structopt` pattern. Doable with a little nesting though.

fitzgen:
> Could this add some tests as well to CI?

We already do `cargo test --all` in CI, fwiw.

> I'm imagining we could have some simple tests like "make sure reduction eventually removes all the functions in this module" or "make sure eventually reduction makes unreachable show up somewhere" or something like that.

I'll see if I can make something that isn't too fragile or takes too long.

alexcrichton:
As a small user-experience report, I wanted to give this a spin on a historical bug from Wasmtime and see what happened. I chose https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=36924 with:

* bytecodealliance/wasmtime@bb85366a3bfab793de960b5be2b069760ab8a6df
* Built with Rust 1.54
* input listed at the top of https://github.com/bytecodealliance/wasmtime/issues/3161

Some things I noticed:

* Could the input be accepted as either `*.wat` or `*.wasm`? (aka could this call `wat::parse_file` or something similar somewhere to get the input?)
* Could the output be automatically disassembled as `*.wat` next to the shrunken `*.wasm`?

Otherwise I didn't get very far:

```
$ cargo run -p wasm-shrink --all-features -- ./test.sh ../wasmtime/fuzz/testcase0.wasm
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s
     Running `target/debug/wasm-shrink ./test.sh ../wasmtime/fuzz/testcase0.wasm`
[INFO] Writing shrunken Wasm file to: ../wasmtime/fuzz/testcase0.shrunken.wasm
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: UnsupportedType(Type(FuncRef))', crates/wasm-mutate/src/module/mod.rs:59:62
```

The wasm-smith generated inputs we use for wasmtime at least are pretty aggressive about enabling lots of wasm proposals, so while the bug isn't with funcref without funcref support this may be a hurdle to reducing wasmtime wasm-smith-generated-inputs.

fitzgen:
Interesting. I had success shrinking sightglass benchmark Wasm files (although `wasm-mutate` is missing lots of things that would help it shrink even further).

`wasm-mutate` needs a bit of polish in things that aren't exercised by the current fuzz target, I think.

crates/wasm-shrink/src/main.rs
fitzgen: Actually once I split the I/O out from the library and only in the CLI this was pretty easy to resolve!

crates/wasm-shrink/src/lib.rs
fitzgen: This is a one-time, up-front check, not something we do repeatedly, so I think we have nothing to worry about here.

fitzgen:
I think I have addressed everything except tests. Will do those tomorrow.

Jacarte:
Cool !

One thing regarding the peephole mutator. Some rewriting rules are applied without taking into account the `reduce` option, for example `?x <=> i32.add ?x 0 if ?x is i32`. One way is perfectly valid to shrink since it remove two nodes and the other is not because it adds two. This can be solved by adding one or the other depending on the tool mode, for example we have some non-semantic preserving rules that are added just in case.


```
 if !config.preserve_semantics {
            // Correctness attraction
            rules.push(rewrite!("correctness-1";  "?x" => "(i32.add ?x 1_i32)" if self.is_const("?x") if self.is_type("?x", PrimitiveTypeInfo::I32)));
            rules.push(rewrite!("correctness-12";  "?x" => "(i64.add ?x 1_i64)" if self.is_const("?x") if self.is_type("?x", PrimitiveTypeInfo::I64)));
        }
```

crates/wasm-shrink/src/lib.rs
alexcrichton: Ah ok, can you write that in the comment as a TODO? (to replace with `debug_assert!` when `wasm-mutate` works better)

src/bin/wasm-shrink.rs
```diff
@@ -0,0 +1,192 @@
...
+    println!(
+        "\n\
+         {} :: {} bytes ({:.02}% smaller)\n\
+         ================================================================================\n\
+         {}\n\
+         ================================================================================",
+        output.display(),
+        shrunken.output.len(),
+        100.0 - (shrunken.output.len() as f64 / initial_size as f64 * 100.0),
+        wat.trim(),
```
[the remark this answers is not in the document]
fitzgen: I actually like it in this final `stdout` output, so I think I'll leave this.

crates/wasm-shrink/src/lib.rs
fitzgen: I think determinism will mostly be nice for testing.

But either way, we have to provide a seed to `wasm-mutate` so we might as well make it customizable. If we want to add parallelism in the future, we can give up the determinism. But also parallelism doesn't help that much for these kinds of brute force, greedy search algorithms in my experience.

fitzgen:
@alexcrichton okay, I think I've addressed everything for real this time, and also have a couple smoke tests.

alexcrichton (approved):
👍

PR #402, opened 2021-12-02
wasm-smith: avoid exporting flagset types in public API

abrown:
In #400, the `FlagSet` type from the `flagset` crate was publicly
exposed by the `Config` trait. This forced users (me) to import
`flagset` anywhere `wasm-smith` was being configured. To avoid this,
`FlagSet` is wrapped in a container structure--`InstructionKinds`.

### DIFF
--- a/crates/wasm-smith/Cargo.toml
+++ b/crates/wasm-smith/Cargo.toml
@@ -21,6 +21,7 @@ flagset = { version = "0.4", features = ["serde"] }
 leb128 = "0.2.4"
 wasm-encoder = { version = "0.8.0", path = "../wasm-encoder" }
 indexmap = "1.6"
+serde = { version = "1", features = ['derive'] }
 
 [dev-dependencies]
 criterion = "0.3.3"
--- a/crates/wasm-smith/Cargo.toml
+++ b/crates/wasm-smith/Cargo.toml
@@ -17,11 +17,11 @@ harness = false
 
 [dependencies]
 arbitrary = { version = "1.0.0", features = ["derive"] }
-flagset = { version = "0.4", features = ["serde"] }
+flagset = "0.4"
 leb128 = "0.2.4"
 wasm-encoder = { version = "0.8.0", path = "../wasm-encoder" }
 indexmap = "1.6"
-serde = { version = "1", features = ['derive'] }
+serde = { version = "1", features = ['derive'], optional = true }
 
 [dev-dependencies]
 criterion = "0.3.3"
@@ -29,3 +29,6 @@ libfuzzer-sys = "0.4.0"
 rand = { version = "0.7.3", features = ["small_rng"] }
 wasmparser = { path = "../wasmparser" }
 wasmprinter = { path = "../wasmprinter" }
+
+[features]
+_internal_cli = ["serde", "flagset/serde"]
--- a/crates/wasm-smith/src/lib.rs
+++ b/crates/wasm-smith/src/lib.rs
@@ -75,6 +75,7 @@ mod terminate;
 use crate::code_builder::CodeBuilderAllocations;
 use arbitrary::{Arbitrary, Result, Unstructured};
 use flagset::{flags, FlagSet};
+use serde::Deserialize;
 use std::collections::{HashMap, HashSet};
 use std::convert::TryFrom;
 use std::marker;
@@ -2458,6 +2459,21 @@ struct Outer {
     modules: Vec<Rc<ModuleType>>,
 }
 
+/// A container for the kinds of instructions that wasm-smith is allowed to
+/// emit. Usage:
+/// ```
+/// # use wasm_smith::{InstructionKinds, InstructionKind};
+/// InstructionKinds::new(InstructionKind::Numeric | InstructionKind::Memory);
+/// ```
+#[derive(Clone, Copy, Debug, Default, Deserialize)]
+pub struct InstructionKinds(pub(crate) FlagSet<InstructionKind>);
+impl InstructionKinds {
+    /// Create a new container.
+    pub fn new(kinds: impl Into<FlagSet<InstructionKind>>) -> Self {
+        Self(kinds.into())
+    }
+}
+
 flags! {
     /// Enumerate the categories of instructions defined in the [WebAssembly
     /// specification](https://webassembly.github.io/spec/core/syntax/instructions.html).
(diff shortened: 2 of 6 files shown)


crates/wasm-smith/src/lib.rs
```diff
@@ -2458,6 +2459,21 @@ struct Outer {
     modules: Vec<Rc<ModuleType>>,
 }
 
+/// A container for the kinds of instructions that wasm-smith is allowed to
+/// emit. Usage:
+/// ```
+/// # use wasm_smith::{InstructionKinds, InstructionKind};
+/// InstructionKinds::new(InstructionKind::Numeric | InstructionKind::Memory);
+/// ```
+#[derive(Clone, Copy, Debug, Default, Deserialize)]
```
alexcrichton, line 2468: Is serde needed for this as a new dependency?

crates/wasm-smith/src/lib.rs
```diff
@@ -2458,6 +2459,21 @@ struct Outer {
...
+pub struct InstructionKinds(pub(crate) FlagSet<InstructionKind>);
+impl InstructionKinds {
+    /// Create a new container.
+    pub fn new(kinds: impl Into<FlagSet<InstructionKind>>) -> Self {
```
alexcrichton, line 2472: Technically `FlagSet` is still in the public API given this bound on the argument

alexcrichton (commented):
Personally I'd prefer to avoid pulling in dependencies for this, it can be a little wordy to support this but this isn't a central API of `wasm-smith` so if using it is slightly wordy I think that's ok. 

I'd personally lean towards an `InstructionKinds` or `InstructionSet` type which has methods to operate on it like `enable_all`, `enable_vector`, `enable_control(false)` or things like that.

crates/wasm-smith/src/lib.rs
abrown: Yes, unfortunately. The CLI and JSON configuration requires serialization and since this type is exposed there it must be `Deserialize`. I think we could avoid this and the `serde` import by doing something like  `pub fn new(kinds: Vec<InstructionKind>) -> Self`--I think serialization would then work and we could still get the benefit of a fast check during instruction generation.

crates/wasm-smith/src/lib.rs
abrown: Yeah, see above.

abrown:
> Personally I'd prefer to avoid pulling in dependencies for this, it can be a little wordy to support this but this isn't a central API of `wasm-smith` so if using it is slightly wordy I think that's ok.
> 
> I'd personally lean towards an `InstructionKinds` or `InstructionSet` type which has methods to operate on it like `enable_all`, `enable_vector`, `enable_control(false)` or things like that.

Well certainly at this point I could have written all of this without `flagset` but let me explain a bit why I used it: I was wary of slowing down instruction generation with this "allowed instruction kinds" check so I wanted some type of quick `&` check. To implement this by hand would involve some methods like you describe plus implementations to serialize/deserialize the structure. Using `flagset` sped all of that up. Now that we have the containing structure, that development benefit is disappearing. How strongly do you want me to re-implement this without `flagset`?

crates/wasm-smith/src/lib.rs
alexcrichton: I think the JSON configuration would make more sense as `Option<Vec<String>>` and the CLI would be a comma-separated string, which would remove the need for serde here.

alexcrichton:
What I'm describing is basically making `FlagSet` an internal implementation detail (100%, not exposing through the constructor). The only public API would be a `InstructionSet` with methods to mutate it. Internally it would still be a bit-vector so testing residence in the set would still be the same.

I suspect that this already has a somewhat-significant regression in generating instructions because getting the instruction set from a `Config` is a virtual-call-per-instruction, where the virtual call should probably be hoisted to the beginning of the "get the set of instructions to generate" function.

abrown:
@alexcrichton, I think the `flagset` parts should be pretty well contained and I think the code should now avoid the call overhead ( (theoretically, using `#[inline]`). I can't get rid of the `serde` dependency though because the JSON parsing wants `InstructionKind` to be `Deserialize`--is there some way around this?

fitzgen:
> I suspect that this already has a somewhat-significant regression in generating instructions because getting the instruction set from a `Config` is a virtual-call-per-instruction, where the virtual call should probably be hoisted to the beginning of the "get the set of instructions to generate" function.

We could reduce this to a single virtual call that returns the set of `InstructionKind`s that are enabled as a `struct`, and then pass that `struct` through the code builder. Probably a good idea to do this.

crates/wasm-smith/src/lib.rs
abrown: `Option<Vec<String>>` makes it hard to fail when an invalid instruction kind is used. I kept it as an `Option<Vec<InstructionKind>>` to fail early on (instead of crashing in `allowed_instructions`) but then we can't get rid of the `serde` dependency. Is there somewhere else I can fail early?

crates/wasm-smith/Cargo.toml
```diff
@@ -21,6 +21,7 @@ flagset = { version = "0.4", features = ["serde"] }
 leb128 = "0.2.4"
 wasm-encoder = { version = "0.8.0", path = "../wasm-encoder" }
 indexmap = "1.6"
+serde = { version = "1", features = ['derive'] }
```
fitzgen, line 24: Can we

* make this an optional dependency,
* make the `flagset` dependency's `"serde"` feature disabled by default,
* add a `"_internal_cli"` feature that enables both previous items,
* and then the CLI tool would enable this feature

This way, regular library usage doesn't need to pull in serde stuff here.

crates/wasm-smith/src/lib.rs
```diff
@@ -2458,10 +2459,43 @@ struct Outer {
     modules: Vec<Rc<ModuleType>>,
 }
 
+/// A container for the kinds of instructions that wasm-smith is allowed to
+/// emit. Usage:
+/// ```
```
fitzgen, line 2464:
```suggestion
/// emit.
///
/// # Example
///
/// ```
```

fitzgen (approved):
LGTM with a minor bit of dependency wrangling -- thanks!

fitzgen (approved):
Thanks!

PR #403, opened 2021-12-02
Lazy iteration over mutations in `wasm-mutate`

Jacarte:
Lazy iteration over mutations. 

The logic of the PR #394 is addressed here, therefore #394 is closed.

This is major change in the code of `wasm-mutate`

cc @fitzgen


fuzz/fuzz_targets/mutate.rs
```diff
@@ -45,45 +45,56 @@ fuzz_target!(|bytes: &[u8]| {
...
+                match mutated_wasm {
+                    Ok(mutated_wasm) => {
+                        // Increase ony once for the same input Wasm
+                        if !found {
+                            // FIXME this metric is somehow undermeasuring the
+                            // effectiveness of the mutator now. Some modules
+                            // can have only one possible mutation meanwhile
+                            // others can have a lot and still each one of them
+                            // is counted as one if they have at least one
+                            // possible mutation
```
fitzgen, line 72: This is working as intended: we are counting the number of modules we successfully mutate, not the number of possible mutations that could have been successful.

fitzgen (approved):
Thanks!

I think this could still be cleaned up a bunch, but at this point it is better to merge it and do clean ups in follow up PRs, so that we can unblock work on `wasm-shrink` and adding more size-reducing mutators, etc.

PR #404, opened 2021-12-02
Add primitive support for v128/funcref/externref in wasm-mutate

alexcrichton:
This only adds a few type definitions and mappings, but it at least gets
the input module ingested. Since none of the new instructions are added
here to wasm-mutate the mutations supported are pretty limited, but
otherwise it helps reducing code from other functions.


fitzgen (approved):
Thanks!

PR #407, opened 2021-12-03
Fix: avoid recycling of validator

Jacarte:
We were using the same validation for all possible mutations in the same module. This leads to an unconsistent  state when we validate one module after the other.

cc @fitzgen


fitzgen (approved):
Thanks!

PR #408, opened 2021-12-03
"use of global instead of use of value" mutator (rebase and squash)

Jacarte:
Readdressing #397


crates/wasm-mutate/src/mutators/peephole/eggsy/encoder/mod.rs
```diff
@@ -53,7 +55,7 @@ impl Encoder {
                 } else {
                     dfg.get_expr(entry.operator_idx)
                 };
-                expr2wasm(config, &to_encode, newfunc, egraph)?;
+                resource_request.push(expr2wasm(config, &to_encode, newfunc, egraph)?);
```
fitzgen, line 58:
```suggestion
                resource_request.extend(expr2wasm(config, &to_encode, newfunc, egraph)?);
```

crates/wasm-mutate/src/mutators/peephole/eggsy/encoder/mod.rs
```diff
@@ -66,6 +68,6 @@ impl Encoder {
         let bytes = &config.info().get_code_section().data[*byterange.0..=*byterange.1];
 
         newfunc.raw(bytes.iter().copied());
-        Ok(())
+        Ok(resource_request.concat())
```
fitzgen, line 71:
```suggestion
        Ok(resource_request)
```

crates/wasm-encoder/src/globals.rs
```diff
@@ -49,6 +49,14 @@ impl GlobalSection {
         self.num_added += 1;
         self
     }
+
+    /// Add a raw byte slice into this code section as a global.
+    ///
```
fitzgen, line 54:
```suggestion
    /// Add a raw byte slice into this code section as a global.
```

fitzgen (approved):
LGTM with the inline suggestions applied -- thanks!

fitzgen:
@Jacarte it looks like this might have introduced an infinite loop when running the tests? `cargo test --all` used to take ~6 minutes in previous CI runs, and we are at ~22 minutes for that task in the current CI and there is nothing being printed to `stdout` or anything like that.

fitzgen:
Actually I think this is `wasm-shrink` getting into an infinite loop when given an infinite iterator from `wasm-mutate`. Fix on the way.

Jacarte:
> Actually I think this is `wasm-shrink` getting into an infinite loop when given an infinite iterator from `wasm-mutate`. Fix on the way.

I could not see this locally

crates/wasm-mutate/src/mutators/peephole/mod.rs
```diff
@@ -212,7 +212,7 @@ impl PeepholeMutator {
                     config.info().function_map.clone(),
                 );
                 let runner = Runner::<Lang, PeepholeMutationAnalysis, ()>::new(analysis)
-                    .with_iter_limit(1) // only one iterations, do not wait for eq saturation, increasing only by one it affects the execution time of the mutator by a lot
+                    .with_iter_limit(200) // FIXME, the iterations should consume fuel from the actual mutator. Be careful with inner set time limits that can lead us to non-deterministic behavior
```
Jacarte, line 215: Here is the source of the long running test. Actually we should link the iteration of the equality saturation to the fuel.

Jacarte: Still, I think this will be solved for `wasm-shrink` by adding only shrinking-based rules if `reduce` is true for the configuration

fitzgen: And also we should switch from using the random extractor to using the `AstSize` extractor when `reduce(true)`

PR #411, opened 2021-12-03
`wasm-shrink`: Fix potential infinite loop and avoid re-testing uninteresting Wasms

fitzgen:
@cfallin mind taking a look at this?


cfallin (approved):
Looks reasonable!

PR #413, opened 2021-12-03
wasm-mutate: Clean up `egg` rule definitions and extract smallest AST size expression when reducing

fitzgen:


fitzgen:
cc @Jacarte

fitzgen:
Oh, the other thing that this does is make the rules unidirectional (most are bidirectional by default) when we are reducing, so that we only make the expression smaller. This could, in theory, miss out on some size-reducing optimizations that take a long, circuitous route to get to their final form, but I think that will be pretty rare in practice and we should get a nice speed up this way since we are dealing with fewer rewrite rules.

crates/wasm-mutate/src/mutators/peephole/eggsy/lang.rs
```diff
@@ -661,11 +661,17 @@ pub enum Lang {
...
 
-    /// Just a wrapper of multiple nodes, when encoding to Wasm it is written as nothing
-    ///    Its only responsability is to add stack neutral operations (if semantic equivalence is set)
-    ///    For example, lets assume we want to insert a nop operation after or before (or both) another node
-    ///    `?x => (container nop ?x nop)`
-    ///    `?x => (container (drop i32.rand) ?x) `
+    /// Just a wrapper of multiple nodes, when encoding to Wasm it is written as
+    /// nothing. Its only responsibility is to express stack-neutral operations.
+    ///
+    /// For example, lets assume we want to insert a nop operation after or
```
cfallin, line 667: pre-existing but s/lets/let's/

cfallin (approved):
LGTM!

PR #414, opened 2021-12-06
`wasm-mutate`: Add a mutator that removes element segments

fitzgen:


cfallin (approved):
Looks reasonable!

alexcrichton:
I think this may not be a valid mutation for the same reason as https://github.com/bytecodealliance/wasm-tools/pull/417#issuecomment-992577451 where modules can `elem.drop` any segment in the module

PR #417, opened 2021-12-08
`wasm-mutate`: Add a mutator to remove data segments

fitzgen:


cfallin (approved):

alexcrichton:
I don't think this is a valid mutation we can do without rewriting the code section unfortunately because you can `data.drop` any data segment (not only passive ones) and otherwise removing data segments means we'll have to update all the indices there

fitzgen:
Ah good catch!

I think we can fix this mutator and element segments removing mutator at the same time when we introduce the code section renumbering helper.

If oss-fuzz finds the bug before then, then we can temporarily disable these mutators.

Issue #415, opened 2021-12-07
`wasm-mutate`: Add/Edit/Remove operations for each Wasm section


fitzgen:
We should have an add, edit, reorder, and remove mutator for each Wasm section (where it makes sense). This is a tracking issue for each of these things.

In general, removing an entity should be something done optimistically, but which can fail if the entity is referenced in other sections (e.g. if we are trying to remove a global, but a function uses it via `global.get` or `global.set`).

### Custom Sections

* [ ] Add new custom section
* [ ] Edit existing custom section
  * [ ] add bytes
  * [ ] remove bytes
  * [ ] edit bytes
* [ ] Remove custom section
* [ ] Move existing custom section

### Type Section

* [ ] Add new type
* [ ] Edit type (if it isn't used)
* [ ] Remove type
* [ ] Reorder types

### Import Section

* [ ] Add new import
* [ ] Edit import name/module
* [ ] Edit import kind (if it isn't used)
* [ ] Remove import
* [ ] Reorder imports

### Function Section

* [ ] Add new function
* [ ] Add a parameter (and update the code section as necessary)
* [ ] Remove a parameter (and update the code section as necessary)
* [ ] Remove function
* [ ] Reorder functions

### Table Section

* [ ] Add a new table (if there aren't any tables or bulk memory is enabled)
* [ ] Edit an existing table
* [ ] Remove a table
* [ ] Reorder tables

### Memory Section

* [ ] Add a new memory (if there aren't any memories or multi-memory is enabled)
* [ ] Edit an existing memory
* [ ] Remove a memory
* [ ] Reorder memories

### Global Section

* [ ] Add a new global
* [ ] Edit an existing global
* [ ] Remove a global
* [ ] Reorder globals

### Export Section

* [ ] Add a new export
* [X] Rename an existing export
* [ ] Edit an existing export to export a different entity
* [X] Remove an export

### Start Section

* [ ] Add a start section (if there isn't one, and there is a function of the right type)
* [ ] Change which function is the start function in an existing start section (if there is another of the right type)
* [ ] Remove the start section

### Element Section

* [ ] Add a new element segment
* [ ] Edit an existing element segment
  * [ ] Reorder functions in the segment
  * [ ] Add a new function to the segment
  * [ ] Remove a function from the segment
* [ ] Remove an element segment
* [ ] Reorder element segments

### Code Section

[...]

PR #420, opened 2021-12-13
wasm-mutate: Add a mutator to remove items in a module

alexcrichton:
This commit adds a new mutator to `wasm-mutate` which can remove any of
these items in a module:

* Functions
* Tables
* Memories
* Tags
* Globals
* Types
* Data
* Elements

This pass is not currently a "gc" pass where a set of used references is
generated, but rather the mutator considers itself applicable if a wasm
module contains any of the above items. If, during removal of an item,
the item is actually used then the mutation aborts with "not
applicable". It's expected that drivers like `wasm-shrink` will try
again with different seeds which will try a different mutator or will
otherwise try to remove a different item.

The main complexity in this mutator is that almost the entire module
(including the code section) is entirely rewritten. This is because item
references via index are everywhere in a wasm module and all of them are
candidate for being updated.

Overall my original test case for when `wasm-shrink` was first added
does indeed get successfully reduced down further from before, removing
extraneous functions and globals. The function-in-question still isn't
shrunk much but I believe that's because the peephole and code motion
mutator don't understand simd yet so those mutations are not applicable.

This commit also removes the recently added data segment and element
segment mutators. Their functionality is folded into this new
`RemoveItemMutator` mutator and the code section is now also updated to
account for updating `{elem,data}.drop` references.

cc #415


crates/wasm-mutate/src/mutators/remove_item.rs
```diff
@@ -0,0 +1,1358 @@
...
+        }
+        s.function(&func);
+        Ok(())
+    }
+
+    /// This is a pretty gnarly function that translates from `wasmparser`
+    /// operators to `wasm_encoder` operators. It's quite large because there's
+    /// quite a few wasm instructions. The theory though is that at least each
+    /// individual case is pretty self-contained.
+    fn translate_op(&mut self, op: &Operator<'_>) -> Result<Instruction<'static>> {
```
fitzgen, line 647: We've implemented this kind of thing a bunch of times now. Do you want to make a `wasmparser` cargo feature for `wasm-encoder` and move all these `translate_foo` methods to be a bunch of `[Try]From<wasmparser::Foo> for Foo` implementations inside `wasm-encoder` that are gated with `cfg(feature = "wasmparser")`?

fitzgen (commented):
Is there a way we could turn `RemoveItemMutator` into `RemoveItemMutator<Item>` that only removes `Item`s, and then instantiate it once for `Global`s, once for `Memory`s, one for `Table`s, etc.? Having a removal mutator per kind of item has a certain aesthetic appeal to me, and once we have weights for mutations, which I think we will inevitably want the same way we weigh instructions in `wasm-smith`, then we can have different weights for different items.

Also, I think it is worth making `filter_out` a little more generic so that we can use it for adding new items and reordering items. I think if the renumbering was a function argument that takes the old index and returns the new index, then it would be generic enough to work for these use cases as well. Without this change, then implementing adding new items and reordering items would require reimplementing all of this `wasmparser` to `wasm_encoder` translating and renumbering all over again.

alexcrichton:
> Is there a way we could turn RemoveItemMutator into RemoveItemMutator<Item> that only removes Items

Sounds reasonable to me! I opted not for generics but rather a field of a struct since `wasm-mutate` already takes forever to compile in fuzzing mode.

> Also, I think it is worth making filter_out a little more generic

I'm not sure I understand how you're thinking it should be more generic? Happy to do so, but I would be hesitant to make too many things too generic just yet. My hope is that `wasmparser => wasm_encoder` is only written once so whatever an "add an item" pass does would presumably reuse and refactor most of this code instead of duplicating it. Without actually having another pass to design with though I'd fear that I wouldn't actually know the constraints.

crates/wasm-mutate/src/mutators/remove_item.rs
alexcrichton: I think this might make sense but I double-checked to ensure that at least for now this is the only translation point for `Operator`. Even if a shared translation were added this module would still have to "walk" everything to remap indices for items and such, so it's not like it'll make this module trivial. I think, though, this module could roughly look like "update a `wasmparser` thing in place" and then use shared code to translate that version to a `wasm_encoder` version. I'd prefer to make `wasmparser => wasm_encoder` infallible though so I'd want to refactor other things in `wasmparser` first like `Type` which has weird variants that are rarely used.

fitzgen:
> Without actually having another pass to design with though I'd fear that I wouldn't actually know the constraints.

Happy to wait until we start writing add-new-entity mutators.

crates/wasm-mutate/src/mutators/remove_item.rs
fitzgen: Yeah, `wasmparser::Type` could definitely be cleaned up.

fitzgen (approved):
Thanks!

PR #421, opened 2021-12-14
Enable the SIMD proposal by default

alexcrichton:
This commit updates wasmparser and related tooling to enable the SIMD
proposal by default. The SIMD proposal has been merged into the formal
specification at this point as well.


fitzgen (commented):
:tada:

PR #422, opened 2021-12-14
wasm-mutate: Refactor and simplify dfg module

alexcrichton:
I was curious to add support for some simd instructions and noticed that
the dfg module could do from some simplifications. This should make it a
bit easier to add new instructions in the future hopefully. This commit
is intended to be a pure refactoring and not actually change anything
(e.g. no simd support added here).

### DIFF
--- a/crates/wasm-mutate/src/mutators/peephole/dfg.rs
+++ b/crates/wasm-mutate/src/mutators/peephole/dfg.rs
@@ -7,6 +7,7 @@ use wasmparser::{Operator, Range};
 
 use crate::mutators::peephole::Lang;
 use crate::{module::PrimitiveTypeInfo, ModuleInfo};
+use PrimitiveTypeInfo::*;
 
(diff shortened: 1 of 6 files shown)


fitzgen (approved):
Very nice! Thank you!

Jacarte:
I agree that the peephole mutator could be refactored even more. Right now to add new `operators` should follow the next mechanical steps:

1 - Register/Implement the new operator and its "symbolic" execution in the `dfg` module
2 -  Register/Implement the new operator as a `Lang` enode with its parsing, formatting and children interaction in the `lang` module (wasm 2 enode).
3 - Add the returning type info in the `analysis` module
4 - Add the `enode` to wasm translation in the `expr2wasm` module
5 - Add the `enode` to the `rebuild` module
6 - Add the lazy iteration implementation in the `expr_enumerator` module.

It seems overcomplicated, but this steps are purely mechanical. I am pretty sure that the full process can be refactored probably with some macro_rules.

PR #425, opened 2021-12-15
wasm-mutate: Simplify the lazy_expand function 

alexcrichton:
I originally wanted to clean up this function to make it a bit easier to
add new cases too but when studying it I realized that I don't think
that the function actually need to match on `Lang` at all but rather
instead can work entirely on the `children_mut` accessor to update
in-place the nodes of `Lang`. This update basically inlines the 0/1/2/3+
operand cases depending on how many children nodes a `Lang` item has,
and in doing so avoids the need to match on `Lang` and duplicate the
calls to `lazy_expand` in a number of places.

This change inherits the limitation where 3-or-more children means that
a cartesian product is not actually formed but instead the first
expansion of each child is used once. This should be a bit more
straightforward to fix in the future, however, if so desired.


fitzgen (approved):
Thanks!

PR #430, opened 2021-12-16
Consolidate all binaries into one tool

alexcrichton:
This commit moves all of our various `src/bin/*.rs` utilities into one
tool as `wasm-tools`. Each previous binary is present now as a
subcommand from the main binary.

The goal of this commit is to provide a few benefits:

* It's a bit easier to think about this project if it has one "swiss
  army knife" instead of lots of little knives all over the place. It's
  also a little less cognitive overhead adding a new subcommand to an
  existing tool than it is to add a new binary.

* All binaries now follow the same style, notably using `structopt`
  instead of a mixture of `getopts` and `structopt`.

* In the future it should be easier to continue to ensure that all
  binaries have similar configuration, for example the same logger,
  since there's a shared entry point for everything now at the top-level
  `main` function.

Overall this isn't intended to really change much functionality, just
move some things around.


src/bin/wasm-tools/validate.rs
```diff
@@ -0,0 +1,130 @@
...
+/// valid, or nonzero and an error message on stderr if the binary is not valid.
+///
+/// Examples:
+///
+/// ```sh
+/// $ wasm-tools validate foo.wasm
+///
+/// $ wasm-tools validate --features all fancy.wasm
+///
+/// $ wasm-tools validate --features=-all mvp.wasm
```
fitzgen, line 22:
```suggestion
/// # Validate `foo.wasm` with the default Wasm feature proposals.
/// $ wasm-tools validate foo.wasm
///
/// # Validate `foo.wasm` with all Wasm feature proposals enabled.
/// $ wasm-tools validate --features all fancy.wasm
///
/// # Validate `foo.wasm` without any Wasm feature proposals enabled.
/// $ wasm-tools validate --features=-all mvp.wasm
```

fitzgen (approved):
Nice! We should probably start publishing this root crate now, so it is easy to `cargo install`.

PR #431, opened 2021-12-17
Implement encoding of wasm components. 

peterhuene:
This PR implements encoders for adapter modules and components.

The encoders added here are in accordance with the (very much)
in-progress [wasm component model proposal](https://github.com/WebAssembly/interface-types/blob/component-rebase/design/proposals/interface-types/Binary.md).

It is expected that this will be updated when the component model proposal
changes.

### DIFF
--- a/crates/wasm-encoder/Cargo.toml
+++ b/crates/wasm-encoder/Cargo.toml
@@ -22,4 +22,7 @@ wasmparser = { path = "../wasmparser" }
 
 [features]
 default = []
+# The feature to support encoding of WebAssembly components.
+# Currently the component model spec is being worked on, so support is gated
+# behind this feature because the API may have substantial breaking changes.
 component-model = []
--- a/crates/wasm-encoder/Cargo.toml
+++ b/crates/wasm-encoder/Cargo.toml
@@ -19,3 +19,7 @@ leb128 = "0.2.4"
 anyhow = "1.0.38"
 tempfile = "3.2.0"
 wasmparser = { path = "../wasmparser" }
+
+[features]
+default = []
+component-model = []
--- a/crates/wasm-encoder/Cargo.toml
+++ b/crates/wasm-encoder/Cargo.toml
@@ -19,10 +19,3 @@ leb128 = "0.2.4"
 anyhow = "1.0.38"
 tempfile = "3.2.0"
 wasmparser = { path = "../wasmparser" }
-
-[features]
-default = []
-# The feature to support encoding of WebAssembly components.
-# Currently the component model spec is being worked on, so support is gated
-# behind this feature because the API may have substantial breaking changes.
-component-model = []
--- a/crates/wasm-encoder/Cargo.toml
+++ b/crates/wasm-encoder/Cargo.toml
@@ -2,7 +2,7 @@
 name = "wasm-encoder"
 version = "0.8.0"
 authors = ["Nick Fitzgerald <fitzgen@gmail.com>"]
-edition = "2018"
+edition = "2021"
 license = "Apache-2.0 WITH LLVM-exception"
 readme = "README.md"
 repository = "https://github.com/bytecodealliance/wasm-tools/tree/main/crates/wasm-encoder"
@@ -11,8 +11,6 @@ documentation = "https://docs.rs/wasm-encoder"
 description = """
 A low-level WebAssembly encoder.
 """
-
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
 leb128 = "0.2.4"
--- a/crates/wasm-encoder/src/component.rs
+++ b/crates/wasm-encoder/src/component.rs
@@ -1,0 +1,371 @@
+//! The module for encoding a WebAssembly component.
+//!
+//! This is an implementation of the in-progress [component
+//! model proposal](https://github.com/WebAssembly/component-model/).
+
+use crate::{encoders, GlobalType, MemoryType, RawSection, Section, TableType};
+
+mod adapters;
+mod aliases;
+mod exports;
+mod functions;
+mod imports;
+mod instances;
+mod modules;
+mod types;
+
+pub use adapters::*;
+pub use aliases::*;
+pub use exports::*;
+pub use functions::*;
+pub use imports::*;
+pub use instances::*;
+pub use modules::*;
+pub use types::*;
+
+const INDEX_REF_INSTANCE: u8 = 0x00;
+const INDEX_REF_MODULE: u8 = 0x01;
+const INDEX_REF_FUNCTION: u8 = 0x02;
+const INDEX_REF_TABLE: u8 = 0x03;
+const INDEX_REF_MEMORY: u8 = 0x04;
+const INDEX_REF_GLOBAL: u8 = 0x05;
+
+const TYPE_REF_INSTANCE: u8 = 0x00;
+const TYPE_REF_MODULE: u8 = 0x01;
+const TYPE_REF_FUNCTION: u8 = 0x02;
+const TYPE_REF_TABLE: u8 = 0x03;
+const TYPE_REF_MEMORY: u8 = 0x04;
+const TYPE_REF_GLOBAL: u8 = 0x05;
+const TYPE_REF_ADAPTER_FUNCTION: u8 = 0x06;
+
+const CANONICAL_OPTION_UTF8: u8 = 0x01;
+const CANONICAL_OPTION_UTF16: u8 = 0x02;
+const CANONICAL_OPTION_COMPACT_UTF16: u8 = 0x03;
+const CANONICAL_OPTION_WITH_REALLOC: u8 = 0x04;
+const CANONICAL_OPTION_WITH_FREE: u8 = 0x05;
+
+/// A WebAssembly component section.
+///
+/// This trait marks sections that can be written to a `Component`.
+///
+/// Various builders defined in this crate already implement this trait, but you
+/// can also implement it yourself for your own custom section builders, or use
+/// `RawSection` to use a bunch of raw bytes as a section.
+pub trait ComponentSection {
+    /// This section's id.
+    ///
+    /// See `SectionId` for known section ids.
+    fn id(&self) -> u8;
+
+    /// Write this section's data and data length prefix into the given sink.
+    fn encode<S>(&self, sink: &mut S)
+    where
+        S: Extend<u8>;
+}
+
+impl ComponentSection for RawSection<'_> {
+    fn id(&self) -> u8 {
+        self.id
+    }
+
+    fn encode<S>(&self, sink: &mut S)
+    where
+        S: Extend<u8>,
+    {
+        <Self as Section>::encode(self, sink);
+    }
+}
+
+/// A Wasm adapter module that is being encoded.
+#[derive(Clone, Debug)]
+pub struct AdapterModule {
+    bytes: Vec<u8>,
+}
+
+impl AdapterModule {
+    /// Begin writing a new `AdapterModule`.
+    pub fn new() -> Self {
+        Self {
+            bytes: vec![
+                0x00, 0x61, 0x73, 0x6D, // magic (`\0asm`)
+                0x0a, 0x00, 0x01, 0x00, // version
+            ],
+        }
+    }
+
+    /// Finish writing this adapter module and extract ownership of the encoded bytes.
+    pub fn finish(self) -> Vec<u8> {
+        self.bytes
+    }
+
+    /// Write a section into this adapter module.
+    pub fn section(&mut self, section: &impl ComponentSection) -> &mut Self {
+        self.bytes.push(section.id());
+        section.encode(&mut self.bytes);
+        self
+    }
+}
+
+impl Default for AdapterModule {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+/// A Wasm component that is being encoded.
+#[derive(Clone, Debug)]
+pub struct Component {
+    bytes: Vec<u8>,
+}
+
+impl Component {
+    /// Begin writing a new `Component`.
+    pub fn new() -> Self {
+        Self {
+            bytes: vec![
+                0x00, 0x61, 0x73, 0x6D, // magic (`\0asm`)
+                0x0a, 0x00, 0x02, 0x00, // version
+            ],
+        }
+    }
+
+    /// Finish writing this component and extract ownership of the encoded bytes.
+    pub fn finish(self) -> Vec<u8> {
+        self.bytes
+    }
+
+    /// Write a section into this component.
+    pub fn section(&mut self, section: &impl ComponentSection) -> &mut Self {
+        self.bytes.push(section.id());
+        section.encode(&mut self.bytes);
+        self
+    }
+}
+
+impl Default for Component {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+/// Known component section IDs.
+///
+/// Useful for implementing the `ComponentSection` trait, or for setting
+/// `RawSection::id`.
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Ord, PartialOrd)]
+#[repr(u8)]
+pub enum SectionId {
+    /// The section is a custom section.
+    Custom = 0,
+    /// The section is a type section.
+    Type = 1,
+    /// The section is an import section.
+    Import = 2,
+    /// The section is a module section.
+    Module = 3,
+    /// The section is an instance section.
+    Instance = 4,
+    /// The section is an alias section.
+    Alias = 5,
+    /// The section is an export section.
+    Export = 6,
+    /// THe section is a function section.
+    Function = 7,
+    /// The section is an adapter function section.
+    AdapterFunction = 8,
+}
+
+impl From<SectionId> for u8 {
+    #[inline]
+    fn from(id: SectionId) -> u8 {
+        id as u8
+    }
+}
+
+/// Represents a reference to an index in a WebAssembly section.
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Ord, PartialOrd)]
+pub enum IndexRef {
+    /// The reference is to an instance in the instance section.
+    Instance(u32),
+    /// The reference is to a module in the module section.
+    Module(u32),
+    /// The reference is to a function in the function section.
+    Function(u32),
+    /// The reference is to a table in the table section.
+    Table(u32),
+    /// The reference is to a memory in the memory section.
+    Memory(u32),
+    /// The reference is to a global in the global section.
+    Global(u32),
+}
+
+impl IndexRef {
+    pub(crate) fn encode(&self, bytes: &mut Vec<u8>) {
+        match self {
+            IndexRef::Instance(index) => {
+                bytes.push(INDEX_REF_INSTANCE);
+                bytes.extend(encoders::u32(*index));
+            }
+            IndexRef::Module(index) => {
+                bytes.push(INDEX_REF_MODULE);
+                bytes.extend(encoders::u32(*index));
+            }
+            IndexRef::Function(index) => {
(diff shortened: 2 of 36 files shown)


peterhuene:
Marking this as a draft PR so that I can include some more unit tests and to get a few remaining questions clarified about the binary format that is currently proposed.

crates/wasm-encoder/src/linking.rs
```diff
@@ -97,7 +97,7 @@ const WASM_SYMBOL_TABLE: u8 = 8;
 /// A subsection of the [linking custom section][crate::LinkingSection] that
 /// provides extra information about the symbols present in this Wasm object
 /// file.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Default)]
```
alexcrichton, line 100: With `Default` being added could the existing `new` functions defer to `Default`?

alexcrichton: (ideally done for all the other `Default` additions in this PR as well)

crates/wasm-encoder/src/elements.rs
```diff
@@ -113,7 +118,7 @@ impl ElementSection {
                 table: None,
                 offset,
             } => {
-                self.bytes.extend(encoders::u32(0x00 | expr_bit));
+                self.bytes.extend(encoders::u32(/* 0x00 | */ expr_bit));
```
alexcrichton, line 121: I realize that clippy is the one telling you to do something about this, but personally I do sort of like the symmetry of the cases here with the leading byte. I also don't want `#[allow(some_clippy_thing)]` though, so I suppose this is the next best thing.

crates/wasm-encoder/Cargo.toml
```diff
@@ -12,12 +12,14 @@ description = """
...
 leb128 = "0.2.4"
 
 [dev-dependencies]
 anyhow = "1.0.38"
 tempfile = "3.2.0"
 wasmparser = { path = "../wasmparser" }
+
+[features]
+default = []
+component-model = []
```
alexcrichton, line 25: I'd naively expect that this would be compiled in by default, but is there a reason you wanted this gated behind a feature in particular?

crates/wasm-encoder/src/component.rs
```diff
@@ -0,0 +1,371 @@
...
+
+impl Default for AdapterModule {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+/// A Wasm component that is being encoded.
+#[derive(Clone, Debug)]
+pub struct Component {
```
alexcrichton, line 117: Oh I thought that adapter modules and components were one-and-the-same with components possibly being a refinement where no core items were exported, only interface-types-looking-items?

crates/wasm-encoder/src/component.rs
```diff
@@ -0,0 +1,371 @@
...
+    /// The section is an instance section.
+    Instance = 4,
+    /// The section is an alias section.
+    Alias = 5,
+    /// The section is an export section.
+    Export = 6,
+    /// THe section is a function section.
+    Function = 7,
+    /// The section is an adapter function section.
+    AdapterFunction = 8,
```
alexcrichton, line 175: I would sort of naively expect that these would be part of the same section, but I also haven't caught up on the encoding in awhile.

crates/wasm-encoder/src/component/types.rs
```diff
@@ -0,0 +1,574 @@
...
+    /// The definition is a memory.
+    Memory,
+    /// The definition is a global.
+    Global,
+}
+
+impl DefinitionKind {
+    pub(crate) fn encode(&self, bytes: &mut Vec<u8>) {
+        match self {
+            Self::Instance => bytes.push(INSTANCE_DEFINITION_KIND),
```
alexcrichton, line 82: Instead of a separate constant for these could the byte value be the discriminant for the `enum`?

crates/wasm-encoder/src/component/types.rs
```diff
@@ -0,0 +1,574 @@
...
+                    ty.encode(bytes);
+                }
+            }
+        }
+    }
+}
+
+/// Represents a type.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Type<'a> {
```
alexcrichton, line 209: Given the recursive and index-bound nature of a `Type` as well as the number of lifetimes involved here, this might be more deserving of an encode-style interface where each variant here, if necessary, has a builder which internally builds up the encoding but can be incrementally added to. That would mean that you don't have to have all the `InstanceTypeDef` items resident at the same time, nor do there need to be vectors-in-vectors and such in case the most end up working around the lifetimes.

crates/wasm-encoder/src/component/types.rs
```diff
@@ -0,0 +1,574 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum AdapterFunctionResults<'a> {
+    /// The adapter function returns zero results.
+    None,
+    /// The adapter function returns one unnamed result.
+    Single(InterfaceType),
+    /// The function returns multiple named results.
+    ///
+    /// The length of the slice must be greater than one.
+    Multiple(&'a [NamedInterfaceType<'a>]),
```
alexcrichton, line 185: I would naively expect that the names here would be part of a name subsection elsewhere (like the wasm `name` section) rather than directly in the type section itself.

Otherwise though encoding-wise it seems a bit odd here that >1 return value require names and are encoded differently (with names) as opposed to 0 or 1 return values.

crates/wasm-encoder/src/component/types.rs
```diff
@@ -0,0 +1,574 @@
...
+    },
+    /// The alias is to an index in an outer module.
+    Outer {
+        /// The count of the outer module.
+        ///
+        /// The count starts at 0 which denotes the current module.
+        count: u32,
+        /// The index from the outer module's index space being aliased.
+        index: OuterAliasIndex,
+    },
```
alexcrichton, line 136: Since these aliases are only in the type section I think only the "outer type" alias is needed here, right? Everything else would be invalid anyway?

crates/wasm-encoder/src/component/types.rs
```diff
@@ -0,0 +1,574 @@
...
+    /// A variant interface type.
+    Variant(&'a [(&'a str, InterfaceType)]),
+    /// A tuple interface type.
+    Tuple(&'a [InterfaceType]),
+    /// A flags interface type.
+    Flags(&'a [&'a str]),
+    /// An enumeration interface type.
+    Enum(&'a [&'a str]),
+    /// A union interface type.
+    Union(&'a [InterfaceType]),
```
alexcrichton, line 419: I sort of forget where we landed on all these but the original intention was that most of these are not present, for example enum/union/optional/expected would all be folded into `Variant` where tuple/flags would be folded into `Record`.

crates/wasm-encoder/src/component/types.rs
```diff
@@ -0,0 +1,574 @@
...
+
+/// Represents a compound interface type.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum CompoundInterfaceType<'a> {
+    /// A list interface type.
+    List(InterfaceType),
+    /// A record interface type.
+    Record(&'a [(&'a str, InterfaceType)]),
+    /// A variant interface type.
+    Variant(&'a [(&'a str, InterfaceType)]),
```
alexcrichton, line 411: I think that the interface type here is optional?

crates/wasm-encoder/src/component/types.rs
```diff
@@ -0,0 +1,574 @@
...
+        /// The type representing success.
+        ok: InterfaceType,
+        /// The type representing failure.
+        error: Option<InterfaceType>,
+    },
+    /// A named interface type.
+    ///
+    /// Named types are either part of a type's declaration (i.e. a record)
+    /// or may be a type alias.
+    Named(NamedInterfaceType<'a>),
```
alexcrichton, line 435: I'm a little confused what a named variant is here in this case? Is this something like the `name` section but baked directly into the type section?

alexcrichton (commented):
Seems reasonable enough to me! 

One of my main questions about this though is what we're going to do with the existing implementation of module linking in this repository. It feels somewhat odd to have two implementations, one for the component model and one for the historical "part of core wasm model". I would personally prefer to only have one at a time, but naturally we have a number of things built up on the module linking implementation in this crate and others.

Is there an expected story for how we'll phase from one to the other?

crates/wasm-encoder/Cargo.toml
peterhuene: No reason in particular other than I like to have a non-default feature as a warning that this is in flux and basing anything on top of this is likely to break in the future.

crates/wasm-encoder/src/component.rs
peterhuene: I suspect you are right and `AdapterModule` will be deleted and one will be able to encode `Component`s in the module section along with core modules.

As it stands right now, the IT binary format proposal extends the updated module linking proposal's format, resulting in two different file formats (one for adapter modules from the linking spec and one for components from the IT spec).

This was the clarification I needed to get resolved before putting this up for review.

crates/wasm-encoder/src/component.rs
peterhuene: I think it makes sense to encode as two different sections as their expected types (IT-based vs. core based) as well as the "bodies" differ (currently in components, the bodies are simply pointing at what function is the target of the canonical lifting/lowering).

crates/wasm-encoder/src/component/types.rs
peterhuene: I'll make that change.

crates/wasm-encoder/src/component/types.rs
peterhuene: I agree, this API is a bit hard to use.

I'll replace with a builder pattern.

crates/wasm-encoder/src/component/types.rs
peterhuene: The names are encoded along with the types so that we can effectively encode everything from a `wit` definition, thus we can generate bindings from the types section when needed.

The reason for not putting it in a custom section is that custom sections are generally seen as optional (and potentially strip-able) for correct behavior; the names obviously don't impact runtime behavior, but they would the DX for users consuming the component.

Regarding the encoding of results from adapter functions this way, that's a very recent change (pushed yesterday) and the intent behind it is to potentially support named return values, but I think there might be more to work out here (for example, we can't encode a single return value with a name). I _believe_ the intention for multiple unnamed results would be to return a single tuple? I'll get more clarification on this too.

crates/wasm-encoder/src/component.rs
peterhuene: Fixed the typo in the doc string on `Function` too.

crates/wasm-encoder/src/component/types.rs
peterhuene: I'm not entirely sure I follow your question, but my understanding of the new format, and that could explain why there's no attempt to bolt the component's format on top of the core module format, is that sections may be defined in any order. The only validation of indexes is that, at the point they are read, the index exists (i.e. a prior section or the current section has defined it). Note that the outer alias count can also refer to the _current_ module at index 0, not just starting from the parent.  Thus types that refer to those indexes must be defined _after_ those referred-to indexes exist.

crates/wasm-encoder/src/component/types.rs
peterhuene: I believe the reason for these specializations not being folded in like that is to have the components be self-describing to the degree that bindings could be generated from binary format.

crates/wasm-encoder/src/component/types.rs
peterhuene: Good point. I think that needs a change to the binary format to support, as we should be able to support both case `Foo` and case `Foo(())` right?

crates/wasm-encoder/src/component/types.rs
peterhuene: That's correct. This can be used to both define a type alias and also give a "name" to a type (e.g. naming a record type).

It isn't strictly *required* to give types a name as it won't effect the runtime behavior of the component, but the intention behind including it here in the types section rather than some other section is to really promote it as something that's not really optional for a good developer experience when consuming the component (bindings generators can always derive a name of some sort, but _using_ the binding won't be pleasant).

crates/wasm-encoder/src/elements.rs
peterhuene: I'm happy to put it back and live with the warning too.

crates/wasm-encoder/src/linking.rs
peterhuene: I'll update all places that now derive `Default` and change their new impls.

crates/wasm-encoder/src/component/types.rs
peterhuene: Actually, I don't think we can actually represent `Foo(())` in `wit` currently, right? Still, can't hurt to change the format to make the interface type optional.

crates/wasm-encoder/src/component.rs
peterhuene: So I talked more to Luke about this.

The updated spec for module linking will define its own module format to define "adapter modules" that may be used to adapt. This is what `AdapterModule` will be. It will support a subset of what components support in their sections.

Thus components may have three different types of modules encoded in its modules section: a core wasm module, a module adapter, or another component.

I'll refactor this a little to align with the proposal better.

crates/wasm-encoder/src/component/types.rs
peterhuene: We've updated the spec to make the types in variant optional, plus unified the expected representation so that both `ok` and `error` cases have optional types. The unit type was removed from the spec for now.

crates/wasm-encoder/Cargo.toml
alexcrichton: Ok in that case I think it's fine to avoid features for now. We already expose work-in-progress things (such as module linking!) without features and we revision breaking changes like that through major versions.

crates/wasm-encoder/src/component.rs
alexcrichton: Hm ok I'm confused now. I was under the impression at the binary layer we'd only have adapter modules and core wasm modules, but from a viewpoint-perspective we'd also have components which just-so-happened to only import/export interface-type looking things. 

I can open a separate issue for this, though. I didn't know though that there was a proposal yet for the binary encoding of interface types?

crates/wasm-encoder/src/component.rs
alexcrichton: IIRC historically I concluded that having one notion of "function" was the way to go for interface types, as opposed to having two categories of functions as you're alluding to I think. I only vaguely remember now but I think it had to do with how to all embedders these things are just functions with fancier types, and it gets sort of redundant and repetitive to deal with two kinds of functions otherwise.

Also I was under the impression that components would want to manufacture core-wasm functions as adaptations of imported interface-types-using-functions to serve as imports for core-wasm imports?

crates/wasm-encoder/src/component/types.rs
alexcrichton: I was personally assuming that we would go the route of "an auxiliary name-like section must be present for good DX" with the assumption that it still technically works otherwise but it won't be nearly as nice. There's other things from the input `*.wit` file like comments I think we may also want to include (which don't easily have places here). Additionally I think that having a separate section for names would improve symmetry with the core wasm format which personally I thought would be a nice goal to have for interface types and components.

crates/wasm-encoder/src/component/types.rs
alexcrichton: Oh I mean that this `Alias`, which is distinct from aliases in the alias section, is only intended for pulling in outer types into an instance or module type. In that sense this for this specific `Alias` variant, which is only inteded for types, is always invalid for anything other than the "outer type" variant, right?

crates/wasm-encoder/src/component/types.rs
alexcrichton: Hm ok, this is something then I think that we never really nailed down in my mind. I'm under the impression that at some "type checking layer" all these things are purely records and variants. At some "ABI layer" they are distinct, however, such as flags getting represented as bitflags. That being said those that have the same type checking and ABI checking I think can go ahead and get shoved into the same variant, such as unions/enums becoming variants.

crates/wasm-encoder/src/component.rs
peterhuene: It's a work in progress in this branch: https://github.com/WebAssembly/interface-types/blob/component-rebase/design/proposals/interface-types/Binary.md

crates/wasm-encoder/src/component/types.rs
peterhuene: Ah, I understand now. Yes, you are correct, the only expected case of an alias here is the outer form.

I've removed the `Alias` enum here with better builder patterns anyway, but I'll ensure that only outer aliases can be included in instance/module types.

crates/wasm-encoder/Cargo.toml
peterhuene: I've removed the feature.

crates/wasm-encoder/src/component.rs
peterhuene: I've pushed a new `adapter` module that defines `AdapterModule` and its relevant sections based on the binary format outlined in the proposal link above.

It doesn't currently share much code with the `component` module as the differences between the two formats can be subtle so sharing sections could be problematic.

crates/wasm-encoder/src/component.rs
peterhuene: Perhaps we should discuss this more with Luke re: the format outlined [here](https://github.com/WebAssembly/interface-types/blob/component-rebase/design/proposals/interface-types/Binary.md)?

crates/wasm-encoder/src/component/types.rs
peterhuene: I've replaced this enum (and all the lifetime-involved types) with encoders for the various types; there are nested vectors when defining instance and module types because they need to know how many definitions are being defined _before_ their encoding, but the singular builders borrow their parent vector to encode the types.

I'm open to a different approach to this too.

crates/wasm-encoder/src/component/types.rs
peterhuene: I'm open to having a name section too.

Luke and I discussed having a custom section for doc strings since they only improve the DX, whereas the names for types and parameters are pretty essential for even a remotely decent DX.

Let's all hop on a call in the new year to talk this design over.

crates/wasm-encoder/src/component/types.rs
peterhuene: This should now be fixed (and the confusing `Alias` enum here removed entirely).

crates/wasm-encoder/src/component/types.rs
peterhuene: Resolving this one so we can continue the "names encoded in the types section" discussion in the other comment.

peterhuene:
Regarding the phasing of the implementation of the "old" module linking format, I think once we're far enough along in supporting the component model proposal in wasm-tools, the proposal has reached a certain level of maturity, and any runtimes that implement the old module linking design have implemented support for the component model (or at least adapter modules from the updated module linking proposal), then we can rip out anything that deals with core wasm modules in encoder/parser/wat that is not in the core wasm spec.

I agree that it's not ideal to have multiple (not-completely-baked) module linking implementations in the various wasm-tools crates, but I'd hate to break downstream consumers (especially Wasmtime) until the component model hits a usable milestone and we provide enough notice to any downstream users that the support for the "old" module linking sections in core wasm modules will be removed.

alexcrichton:
Ok let's go ahead and merge this and we can update it later as necessary.

PR #439, opened 2022-01-07
wasm-mutate: Remove a hot vector creation found during fuzzing

alexcrichton:
This commit fixes a hot loop found during fuzzing which cause fuzzing to
time out. Previously allocated vectors are now replaced with closures to
avoid all of the allocations entirely. This is intended to purely
refactor how this functionality is expressed, no changes in behavior is
intended here.

Previously a fuzzer running in ~60s now runs in 3s.


fitzgen (approved):
Nice!

PR #441, opened 2022-01-10
wasmparser: Improve FuncType

Robbepop:
This reduces the number of separate allocations per FuncType instance by 1.
Also this adds some public API and hides the implementation details.
Other crates of the workspace have been adjusted.
The `Debug` implementation has been adjusted to output the same output as before.


Robbepop:
Sorry, somehow I got confused at my own local workspace and accidentally merged some separate works ... :facepalm:

PR #442, opened 2022-01-10
wasmparser: Improve FuncType

Robbepop:
This reduces the number of separate allocations per FuncType instance by 1.
Also this adds some public API and hides the implementation details.
Other crates of the workspace have been adjusted.
The `Debug` implementation has been adjusted to output the same output as before.

The same optimization could be applied to Wasmtime's `FuncType`, too.

### DIFF
--- a/crates/wasmparser/src/binary_reader.rs
+++ b/crates/wasmparser/src/binary_reader.rs
@@ -236,10 +236,7 @@ impl<'a> BinaryReader<'a> {
         for _ in 0..returns_len {
             returns.push(self.read_type()?);
         }
-        Ok(FuncType {
-            params: params.into_boxed_slice(),
-            returns: returns.into_boxed_slice(),
-        })
+        Ok(FuncType::new(params, returns))
     }
 
     pub(crate) fn read_module_type(&mut self) -> Result<ModuleType<'a>> {
--- a/crates/wasmparser/src/primitives.rs
+++ b/crates/wasmparser/src/primitives.rs
@@ -200,17 +200,16 @@ impl Debug for FuncType {
 
 impl FuncType {
     /// Creates a new function type.
-    pub fn new<I, O>(inputs: I, outputs: O) -> Self
+    pub fn new<I, O>(params: I, results: O) -> Self
     where
         I: IntoIterator<Item = Type>,
         O: IntoIterator<Item = Type>,
-        I::IntoIter: ExactSizeIterator,
     {
-        let inputs = inputs.into_iter();
-        let len_params = inputs.len();
-        let inputs_outputs = inputs.chain(outputs).collect::<Vec<_>>().into_boxed_slice();
+        let mut params_results = params.into_iter().collect::<Vec<_>>();
+        let len_params = params_results.len();
+        params_results.extend(results);
         Self {
-            params_results: inputs_outputs,
+            params_results: params_results.into_boxed_slice(),
             len_params,
         }
     }
--- a/crates/wasmparser/src/primitives.rs
+++ b/crates/wasmparser/src/primitives.rs
@@ -223,11 +223,6 @@ impl FuncType {
     pub fn returns(&self) -> &[Type] {
         &self.params_results[self.len_params..]
     }
-
-    /// Returns the pair of parameter and result types of the function type.
-    pub fn params_returns(&self) -> (&[Type], &[Type]) {
-        self.params_results.split_at(self.len_params)
-    }
 }
 
 #[derive(Debug, Clone)]
--- a/crates/wasmparser/src/primitives.rs
+++ b/crates/wasmparser/src/primitives.rs
@@ -15,6 +15,7 @@
 
 use std::error::Error;
 use std::fmt;
+use std::fmt::Debug;
 use std::result;
 
 #[derive(Debug, Clone)]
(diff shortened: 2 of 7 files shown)


crates/wasmparser/src/binary_reader.rs
```diff
@@ -236,10 +236,7 @@ impl<'a> BinaryReader<'a> {
         for _ in 0..returns_len {
             returns.push(self.read_type()?);
         }
-        Ok(FuncType {
-            params: params.into_boxed_slice(),
-            returns: returns.into_boxed_slice(),
-        })
+        Ok(FuncType::new(params, returns))
```
Robbepop, line 239: Technically we could improve this implementation and reuse allocations but that would have required more code changes and I wanted to keep them minimal.

crates/wasmparser/src/primitives.rs
```diff
@@ -169,10 +170,65 @@ pub enum TypeDef<'a> {
...
+
+    /// Returns the result types of the function type.
+    pub fn returns(&self) -> &[Type] {
+        &self.params_results[self.len_params..]
+    }
+
+    /// Returns the pair of parameter and result types of the function type.
+    pub fn params_returns(&self) -> (&[Type], &[Type]) {
+        self.params_results.split_at(self.len_params)
+    }
```
alexcrichton, line 231: I think this method can be omitted since the params/returns methods can be called together?

crates/wasmparser/src/primitives.rs
```diff
@@ -169,10 +170,65 @@ pub enum TypeDef<'a> {
...
+}
+
+impl FuncType {
+    /// Creates a new function type.
+    pub fn new<I, O>(inputs: I, outputs: O) -> Self
+    where
+        I: IntoIterator<Item = Type>,
+        O: IntoIterator<Item = Type>,
+        I::IntoIter: ExactSizeIterator,
```
alexcrichton, line 207: I think it'd be best to remove this and instead use `.push` or similar to look at the number of parameters.

alexcrichton (commented):
Thanks for this! Do you have some profiling information which has been guiding this or some performance numbers for before/after? This is one of the less-efficient parts of wasmparser right now and is indeed ripe for improvement.

crates/wasmparser/src/primitives.rs
Robbepop: We can remove this API indeed. We have this in `wasmi` because of some lifetime issues but there we additionally operate via a `Store` as done in Wasmtime. So maybe only the Wasmtime definition would probably benefit from this.

Robbepop:
> Thanks for this! Do you have some profiling information which has been guiding this or some performance numbers for before/after? This is one of the less-efficient parts of wasmparser right now and is indeed ripe for improvement.

I executed the `compare-with-main.sh` script but it found no performance differences.
However, in theory the new implementation has less pressure on the heap allocator (if we actually make use of that in the parser) and parameter and result types are more data local and therefore should technically be more cache efficient, i.e. less pointer chasing. 

```
~/p/w/c/wasmparser> ./compare-with-main.sh
Already on 'main'
Your branch is up to date with 'origin/main'.
   Compiling wasmparser v0.82.0 (/home/robrob/projects/wasm-tools/crates/wasmparser)
   Compiling wasmparser-dump v0.1.0 (/home/robrob/projects/wasm-tools/crates/dump)
    Finished bench [optimized] target(s) in 7.67s
     Running unittests (/home/robrob/projects/wasm-tools/target/release/deps/benchmark-5530eeb73a768754)
WARNING: HTML report generation will become a non-default optional feature in Criterion.rs 0.4.0.
This feature is being moved to cargo-criterion (https://github.com/bheisler/cargo-criterion) and will be optional in a future version of Criterion.rs. To silence this warning, either switch to cargo-criterion or enable the 'html_reports' feature in your Cargo.toml.

it works benchmark      time:   [7.5139 ms 7.5460 ms 7.5824 ms]                               
                        change: [-0.4231% +0.1704% +0.7660%] (p = 0.59 > 0.05)
                        No change in performance detected.
Found 10 outliers among 100 measurements (10.00%)
  4 (4.00%) high mild
  6 (6.00%) high severe

validate benchmark      time:   [21.502 ms 21.636 ms 21.798 ms]                               
                        change: [-1.0421% -0.1693% +0.6935%] (p = 0.71 > 0.05)
                        No change in performance detected.
Found 10 outliers among 100 measurements (10.00%)
  4 (4.00%) high mild
  6 (6.00%) high severe

Already on 'main'
Your branch is up to date with 'origin/main'.
    Finished bench [optimized] target(s) in 0.05s
     Running unittests (/home/robrob/projects/wasm-tools/target/release/deps/benchmark-5530eeb73a768754)
WARNING: HTML report generation will become a non-default optional feature in Criterion.rs 0.4.0.
This feature is being moved to cargo-criterion (https://github.com/bheisler/cargo-criterion) and will be optional in a future version of Criterion.rs. To silence this warning, either switch to cargo-criterion or enable the 'html_reports' feature in your Cargo.toml.

it works benchmark      time:   [7.5560 ms 7.5860 ms 7.6192 ms]                               
                        change: [-0.0888% +0.5304% +1.1422%] (p = 0.10 > 0.05)
                        No change in performance detected.
  8 (8.00%) high mild
  1 (1.00%) high severe

validate benchmark      time:   [21.822 ms 22.008 ms 22.280 ms]                               
                        change: [+0.5872% +1.7173% +3.1834%] (p = 0.00 < 0.05)
  7 (7.00%) high mild
  2 (2.00%) high severe
```

crates/wasmparser/src/primitives.rs
Robbepop: done

crates/wasmparser/src/binary_reader.rs
Robbepop: @alexcrichton If you are okay with introducing a `pub(crate) fn FuncType::from_raw_parts(..)` constructor we could implement this micro optimization in this PR. It should reduce the amount of heap allocations while parsing `FuncType`. In practice I was not able to measure a difference. Probably the benchmarks do not have too many big `FuncType` instances.

alexcrichton:
Oh so is this not motivated by a real-world module? I understand how this is theoretically better but I would prefer that there would be something concrete showing it before merging.

Robbepop:
> Oh so is this not motivated by a real-world module? I understand how this is theoretically better but I would prefer that there would be something concrete showing it before merging.

I guess I cannot deliver on such a module then. Feel free to either merge or close this PR then.

PR #443, opened 2022-01-10
wasmparser: Apply clippy suggestions + add clippy CI job

Robbepop:
This PR adds a `clippy` CI job for `wasmparser` that can be extended to other crates of this workspace in later PRs.
Also the PR applies all `clippy` suggestions to the `wasmparser` crate.

### DIFF
--- a/crates/wasmparser/src/validator.rs
+++ b/crates/wasmparser/src/validator.rs
@@ -374,7 +374,7 @@ impl Validator {
             } => self.code_section_start(*count, range)?,
             CodeSectionEntry(body) => {
                 let func_validator = self.code_section_entry()?;
-                return Ok(ValidPayload::Func(func_validator, body.clone()));
+                return Ok(ValidPayload::Func(func_validator, *body));
             }
             ModuleSectionStart {
                 count,
@@ -649,11 +649,11 @@ impl Validator {
             }
             ImportSectionEntryType::Table(t) => {
                 self.table_type(t)?;
-                Ok(EntityType::Table(t.clone()))
+                Ok(EntityType::Table(*t))
             }
             ImportSectionEntryType::Memory(t) => {
                 self.memory_type(t)?;
-                Ok(EntityType::Memory(t.clone()))
+                Ok(EntityType::Memory(*t))
             }
             ImportSectionEntryType::Tag(t) => {
                 self.tag_type(t)?;
@@ -661,7 +661,7 @@ impl Validator {
             }
             ImportSectionEntryType::Global(t) => {
                 self.global_type(t)?;
-                Ok(EntityType::Global(t.clone()))
+                Ok(EntityType::Global(*t))
             }
             ImportSectionEntryType::Module(type_index) => {
                 self.module_type_at(*type_index)?;
@@ -840,7 +840,6 @@ impl Validator {
 
     /// Validates [`Payload::ModuleSectionStart`](crate::Payload)
     pub fn module_section_start(&mut self, count: u32, range: &Range) -> Result<()> {
-        drop(count);
         if !self.features.module_linking {
             return self.create_error("module linking proposal not enabled");
         }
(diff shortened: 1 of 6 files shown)


crates/wasmparser/src/validator.rs
```diff
@@ -1792,7 +1791,7 @@ mod arc {
                 return None;
             }
             debug_assert!(Arc::get_mut(&mut self.arc).is_some());
-            Some(unsafe { &mut *(&*self.arc as *const T as *mut T) })
+            Some(self.assert_mut())
```
Robbepop, line 1794: The downside of this change is that it introduces a runtime check.
The ideal replacement was [`Arc::get_mut_unchecked`](https://doc.rust-lang.org/alloc/sync/struct.Arc.html#method.get_mut_unchecked) but that's still unstable as of now.

Robbepop:
I don't understand why the tests are failing. The `wasmparser` tests work fine locally.

crates/wasmparser/src/validator.rs
alexcrichton: This needs to stay as it was, otherwise this is a mutally recursive function that never returns.

Robbepop:
@alexcrichton Thanks for the review! Applied your suggestions.

Robbepop:
@alexcrichton Would you be interested for a similar PR for the `wast` crate? I'd be down to fix `clippy` lints there as well as adding it to the CI when everything is cleaned up.

crates/wasmparser/src/validator.rs
Robbepop: Applied your suggestion and tests are running now again, too. :tada:

crates/wasmparser/src/validator.rs
```diff
@@ -1792,7 +1791,19 @@ mod arc {
...
-            Some(unsafe { &mut *(&*self.arc as *const T as *mut T) })
+            Some(
+                // # Safety
+                //
+                // This operation is safe since the API asserts that
+                // `self.owned` is only ever `true` if there is just
+                // one owner of the `Arc`.
+                // As soon as `Arc::get_mut_unchecked` we should
+                // make use of it instead.
+                #[allow(clippy::cast_ref_to_mut)]
```
alexcrichton, line 1802: I would prefer to have no `#[clippy]` attributes in this crate, can this be removed?

crates/wasmparser/src/validator.rs
```diff
@@ -1792,7 +1791,19 @@ mod arc {
...
             debug_assert!(Arc::get_mut(&mut self.arc).is_some());
-            Some(unsafe { &mut *(&*self.arc as *const T as *mut T) })
+            Some(
+                // # Safety
+                //
+                // This operation is safe since the API asserts that
+                // `self.owned` is only ever `true` if there is just
+                // one owner of the `Arc`.
+                // As soon as `Arc::get_mut_unchecked` we should
+                // make use of it instead.
```
alexcrichton, line 1801: This `# Safety` comment doesn't really add anything over the structure of the code that's already here, so I don't think it should be added.

crates/wasmparser/src/validator.rs
Robbepop: How are we going to satisfy the `clippy` CI then?
Are those `#[clippy]` attributes okay in the crate root? Less ideal though ...

alexcrichton: We don't run clippy in CI in this repository, so there's nothing to satisfy.

Robbepop: Is there a reason for that? With this PR the `wasmparser` crate could profit from having a `clippy` CI job at hand. In fact this PR adds a `clippy` CI job for `wasmparser`.

alexcrichton: Personally I do not believe that clippy adds value to a project. That's my personal opinion, others may feel differently. I don't really want to debate the merits of clippy here at this time, my point is that we don't run clippy and that is not something I would like to change off-hand in a PR.

Robbepop:
I applied your review comments:

- Removed `clippy` CI job for `wasmparser`.
- Removed `Safety` comment.
- Remove attribute to silence a `clippy` warning for unsafe code.

PR #445, opened 2022-01-13
wasm-mutate: Add bulk memory support to DFG

alexcrichton:
This commit adds support to the wasm-mutate DFG for the bulk-memory
proposal, notable table/memory copying instructions as well as table
operation instructions. No rules are added yet for rewriting, for
example simply removing `elem.drop` or `data.drop`, but I'm hoping that
will be easier to add later with the DFG nodes already added.


src/bin/wasm-tools/mutate.rs
```diff
@@ -75,17 +75,17 @@ impl Opts {
...
                 );
-                (output, f.display().to_string())
+                (output, f.display().to_string(), false)
             }
             None => {
                 let output = Box::new(stdout.lock());
-                (output, "<stdout>".to_string())
+                (output, "<stdout>".to_string(), true)
             }
         };
```
fitzgen, line 90: Maybe it's just early in the morning, but it took me a second to understand what was happening here and why we were sometimes outputting WAT instead of Wasm. I think a little comment in the `stdout` match arm would be really helpful, something like:

> When we are writing to `stdout` and presumably a human is reading the output, don't emit Wasm, as that is not human readable. Instead, emit the WAT disassembly.

Although this does make me wonder whether this should be a dedicated `--wat` flag instead, since it means that piping `stdout` somewhere else might not work as expected. Thoughts?

fitzgen (approved):

src/bin/wasm-tools/mutate.rs
alexcrichton: `--wat` seems reasonable to me!

PR #446, opened 2022-01-13
Implement component parsing in `wasmparser`. 

peterhuene:
This PR implements WebAssembly component parsing in `wasmparser` per the in-progress component model proposal.

Similar to the wasm module parser, `component::Parser` is implemented to
incrementally parse the sections of a WebAssembly component.

This PR also fixes `wasm-encoder` parsing of canonical options to fit the
current component model spec and adds encoding of custom sections for WebAssembly components.

The changes to `wasmparser` do not yet include a validator for WebAssembly components; that will be coming soon.

### DIFF
--- a/crates/wasmparser/src/parser.rs
+++ b/crates/wasmparser/src/parser.rs
@@ -20,14 +20,14 @@ pub(crate) const WASM_MODULE_VERSION: u32 = 0x1;
 /// This primary function for a parser is the [`Parser::parse`] function which
 /// will incrementally consume input. You can also use the [`Parser::parse_all`]
 /// function to parse a module that is entirely resident in memory.
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone)]
 pub struct Parser {
     state: State,
     offset: u64,
     max_size: u64,
 }
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone)]
 enum State {
     ModuleHeader,
     SectionStart,
@@ -741,8 +741,8 @@ impl Parser {
                 // Afterwards we turn the loop again to recurse in parsing the
                 // nested module.
                 Payload::ModuleSectionEntry { parser, range: _ } => {
-                    stack.push(cur);
-                    cur = *parser;
+                    stack.push(cur.clone());
+                    cur = parser.clone();
                 }
 
                 _ => {}
--- a/crates/wasmparser/src/parser.rs
+++ b/crates/wasmparser/src/parser.rs
@@ -8,6 +8,9 @@ use std::convert::TryInto;
 use std::fmt;
 use std::iter;
 
+const WASM_EXPERIMENTAL_VERSION: u32 = 0xd;
+pub(crate) const WASM_MODULE_VERSION: u32 = 0x1;
+
 /// An incremental parser of a binary WebAssembly module.
 ///
 /// This type is intended to be used to incrementally parse a WebAssembly module
@@ -17,14 +20,14 @@ use std::iter;
 /// This primary function for a parser is the [`Parser::parse`] function which
 /// will incrementally consume input. You can also use the [`Parser::parse_all`]
 /// function to parse a module that is entirely resident in memory.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Copy)]
 pub struct Parser {
     state: State,
     offset: u64,
     max_size: u64,
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Copy)]
 enum State {
     ModuleHeader,
     SectionStart,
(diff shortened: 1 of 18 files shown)


crates/wasmparser/src/parser.rs
```diff
@@ -17,14 +20,14 @@ use std::iter;
 /// This primary function for a parser is the [`Parser::parse`] function which
 /// will incrementally consume input. You can also use the [`Parser::parse_all`]
 /// function to parse a module that is entirely resident in memory.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Copy)]
```
alexcrichton, line 23: Although this is indeed `Copy` under the hood it's a stateful type where subtle bugs can happen due to where rustc may emit copies or not. In the same sense that iterators are generally not supposed to be `Copy` I think it might be best for this to remain not-`Copy`, although I'm not 100% sure why this was added so I could be convinced otherwise!

alexcrichton (commented):
One of my original hopes for the component model was to avoid having a parallel API surface area for comopnents as well as for core wasm modules. Where possible I think it would be best to use the same interfaces and avoid having to switch types based on the use case. In that sense would it be possible to move this all into the existing `Parser` type?

One thing that jumps to mind is that all the section ids collide, but this might be a good time to perhaps choose entirely disjoint section ids from core wasm to clearly indicate that they're component sections instead of core wasm sections.

Additionally as well I'd envision that `Validator` would be updated for components and it would still be the case that outputs of `Parser` would basically be fed directly into `Validator`.

peterhuene:
I think one of the primary motivators for not extending the core module format is that wasm component sections are not ordered and will allow duplicate sections of the same kind (consequently, their index spaces are incrementally built up). The move away from the core module spec is very much intentional and Luke probably has more reasons as to its motivations.

Overlapping section identifiers notwithstanding, implementing a parser/validator that can parse/validate both core modules and components seems like, to me, that it would result in pretty complex code that needs to check at every place what is actually being parsed/validated. Plus, a unified `Payload` would not be pleasant to deal with as a user as non-trivial subset of the variants simply aren't ever returned depending on what's being parsed.

I'd prefer to keep the parser and validator implementations separate for now, but I'm entirely open to unifying them should the component model change to be defined in terms of the core module spec.

crates/wasmparser/src/parser.rs
peterhuene: I'll remove Copy from this and the component parser.

peterhuene:
I'm willing to try to merge the parser/validator implementations, although I still believe it's going to be quite messy.

Are we willing to take a breaking change to at least rename the existing `Payload` variants and `Validator` methods so that they are more clearly intended for core wasm modules?

peterhuene:
If we unify `Payload`, `Parser`, `Validator` (etc.), I'd also like to shuffle existing readers/primitives into a `core` module to keep at least some separation between core and component types. Thoughts on that level of breaking change?

alexcrichton:
Oh sure yeah breaking changes are always fine in `wasmparser`, we're on like 0.80 now at this point and are making good use of the infinite version space.

To clarify though I don't want to force things to be "unified in one" if it make the codebase truly awful. If that's the case then I think that's feedback to take to the interface types proposal and work on it there. Basically if we find ourselves implementing a completely parallel surface area for everything then I would fear that module linking and interface types would face a very steep uphill battle in any other engine anywhere. Although we might be willing to invest in doubling the surface area other engines may not be so amenable if it's not seen as an additional feature like simd which fits within the existing framework.


Also yeah for reorganizing the crate personally I feel that we haven't ever really scrutinized `wasmparser`'s API so movement around is fine. That being said it's a fairly heavily used crate (especially by Wasmtime and various other things we have) so movement and shuffling would ideally happen once.

peterhuene:
I think one of my biggest problems is that the component model is based on an updated module linking proposal and it isn't compatible with the older module linking implementation already present in `wasmparser`, which makes unifying the underlying types returned by the various readers that much more difficult.

I've already implemented the updated module-linking proposal to `wasm-encoder` in the form of `AdapterModule` (although that also needs unification a bit in that crate as well).

I wonder if it would make sense to remove the former module-linking proposal implementation in `wasm-encoder`/`wasmparser` now (replaced with support for the current module linking and component model proposals) and switch `wast` over to using the `wasm-encoder` impl for encoding (however, that also needs wast AST updates per the updated module linking proposal). It's a bunch of work, but I think that would result in cleaner interfaces across the board, but the peril being a much larger breaking change for downstream crates.

alexcrichton:
Personally I'm fine ripping out all the existing module linking support, it's not really going anywhere and AFAIK the preview isn't all that useful at this point. At minimum we have historical versions of Wasmtime with module linking enabled and otherwise we've kinda learned what we can from it.

Also FWIW I don't really review `wasm-encoder` too too closely since it's not as critical of a dep as `wasmparser`, but I do think that ideally top-level encoding would be the same across core wasm and component modules.

It's a long-term intent for `wast` to use `wasm-encoder` but they weren't developed in that order so it's a fair bit of work to get that done which hasn't been prioritized yet. I agree it'd be good to get there eventually but I think it's ok to duplicate the encoding bits for now. If you'd like to move `wast` to `wasm-encoder` though I think that's fine but would be best done as an isolated PR (and `wasm-encoder` probably needs updates for a proposal here-or-there, `wast` typically "implements" the most proposals).

peterhuene:
I put up #448 to unify the various section types in `wasm-encoder`. Those changes impacted `wasm-smith` the most as it's now in an in-between state where we can write the new formats but not parse them. I will now start work on updating this PR to unify the parser implementation and updating the module linking support.

peterhuene:
Closing this PR for now so we don't accidentally merge it prior to it being updated with a unified parsing API.

PR #448, opened 2022-01-20
Reimplement component encoding in `wasm-encoder`.

peterhuene:
This PR reimplements encoding of components in `wasm-encoder`.

The goal of this PR is to share the various section encoders between the two supported file formats: modules and components. This greatly reduces the duplicated code from the previous implementation.

It also removes the support for previous adapter modules based on changes to the component model proposal.

This removes all the duplicate types currently present in `wasm-encoder` from the (now-removed) `adapter` and `component` submodules. In doing so, support for encoding the previous module linking proposal has been removed entirely. This is, understandably, a large breaking change.

`wasm-mutate` was easily updated to the `wasm-encoder` changes because it does not implement support for the previous iteration of the module linking proposal.

However, `wasm-smith` implemented support for the earlier module linking proposal and thus has been largely impacted by these changes. I decided to remove support for the module linking proposal from `wasm-smith` while I continue work on updating `wasmparser`. The hope is to update `wasm-smith` eventually to support the newer component model proposal by generating components instead of just core modules.

I'm not convinced this is the right approach; that said, I also anticipate breaking the existing module linking support in `wasmparser` soon in favor of a unified way to parse modules and components.


crates/wasm-encoder/src/custom.rs
```diff
@@ -30,6 +23,45 @@ impl Section for CustomSection<'_> {
...
+        ComponentSectionId::Custom
+    }
+
+    fn encode<S>(&self, sink: &mut S)
+    where
+        S: Extend<u8>,
+    {
+        self.encode(sink);
+    }
+}
```
alexcrichton, line 63: I've got two questions from this:

* Why is there a distinction between component sections and adapter module sections? I thought we only had adapter modules? I may be tripping myself up again in perceptions around what I thought this was supposed to be...
* The intention, I thought, was that the component model would share the binary encoding with core wasm wherever possible, so I would hope that we don't need to parameterize the `Section` trait and have three implementations for `CustomSection`, instead just one?

crates/wasm-encoder/src/functions.rs
```diff
@@ -42,23 +49,43 @@ impl FunctionSection {
...
     }
 
-    /// Define a function that uses the given type.
+    /// Define a function in a module's function section.
+    ///
+    /// This method is only supported for encoding modules.
     pub fn function(&mut self, type_index: u32) -> &mut Self {
+        if self.format != SectionEncodingFormat::Module {
+            panic!("functions can only be encoded for modules");
+        }
```
alexcrichton, line 58: I briefly mentioned this above, but I don't think that these panics are necessary and would hope it's ok to remove them. Trying to assert that only a valid wasm module is emitted is I think out of the scope of the `wasm-encoder` crate.

crates/wasm-encoder/src/imports.rs
```diff
@@ -46,36 +53,38 @@ impl ImportSection {
...
+                    "module name cannot be provided for adapter module or component encoding"
+                );
+                self.bytes.extend(encoders::str(name));
             }
         }
-        ty.into().encode(&mut self.bytes);
+        ty.into().encode(self.format, &mut self.bytes);
         self.num_added += 1;
         self
     }
```
alexcrichton, line 84: For the import section I would imagine that because the encoding is different between core wasm and the component model that this would be two methods, one for each encoding choice, instead of one method with various optional parameters.

crates/wasm-encoder/src/custom.rs
peterhuene: Right now the module linking proposal defines _adapter modules_ that are limited to describing how core/adapter modules link together; they only have a small subset of the sections from a core module (type, import, and export) and thus can't really define their own items other than modules and instances (or alias core items from instances).

The component model proposal extends the module linking proposal by overlaying support for interface types; it is a true extension, but with a different version in the header, so they could technically be combined and indeed this is something that Luke was contemplating doing as the utility of having the less capable _adapter module_ format is questionable (perhaps useful for engines that don't want to support the component model at all but want a mechanism to link core modules together?).

The current intention of the binary encoding of both proposals is to make a break from the core binary encoding. Perhaps [this explainer](https://github.com/WebAssembly/interface-types/blob/component-rebase/design/proposals/module-linking/Explainer.md#binary-format-considerations) needs more explanation as to why module linking proposal (and by extension the component model proposal) is departing from the core module spec entirely?

crates/wasm-encoder/src/exports.rs
```diff
@@ -1,44 +1,108 @@
...
+            },
+            SectionEncodingFormat::Component => match self {
+                Self::Instance(i) => (0x00, *i),
+                Self::Module(i) => (0x01, *i),
+                Self::Function(i) => (0x02, *i),
+                Self::Table(i) => (0x03, *i),
+                Self::Memory(i) => (0x04, *i),
+                Self::Global(i) => (0x05, *i),
+                Self::AdapterFunction(i) => (0x06, *i),
+                _ => panic!("cannot encode {:?} for a WebAssembly component", self),
```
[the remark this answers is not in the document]
peterhuene: Let me review the places where these things don't line up with Luke and see if we can at least reduce the overhead here of having an unnecessary differences from the core module format (like this example here).

If we can get these things to better align, perhaps some of these sections can encode entirely the same.

crates/wasm-encoder/src/functions.rs
peterhuene: I would normally agree, but the intention here is to put the error as close as possible to the problem (e.g. you're using a method on a generic writer type meant for a different format that you specified); relying on `wasmparser` to validate the output means you could get an entirely unhelpful error (e.g. something was incorrectly decoded to a length and now the validator gives an extremely unhelpful "unexpected EOF at byte offset X" error).

Perhaps along with reducing the unnecessary differences in the formats (where possible) the section writers can automatically determine what format they're using based on how they are used and perhaps limit the check to when the section itself gets written to the module/adapter module/component?

crates/wasm-encoder/src/imports.rs
peterhuene: Agreed, this can be two different methods. I'll fix.

peterhuene:
Let me get back to you after talking a little with Luke to reduce the overhead in these formats; with that perhaps in place I can at least make the section encoding format _internal_ to the crate and automatically determined based on API usage. I think that would reduce the cognitive overhead of using this API.

peterhuene:
Talked with Luke, who spoke with the other proposal authors regarding merging adapter modules and components. It sounds like we're going to do away with adapter modules entirely, so there's only components and core modules, but limit module types to describe core modules only with _component types_ being added to describe components (i.e. effectively a module type but with interface types).

We also spoke regarding reusing core enumerations for common sections like imports and exports, so hopefully this will make this much cleaner.

An updated binary grammar is forthcoming. I'm going to review it and then ingest it into this PR with the hope of eliminating the complexity of specifying the intended section formats up front.

crates/wasm-encoder/src/custom.rs
peterhuene: We now have just `Section` (for modules) and `ComponentSection` (for components).

I've removed having to explicitly specify the format for each of the section encoders; the format is now determined by how the API is used.

Some more work is needed once the binary grammar of the spec has been updated, but the new commit at least makes the API more reasonable.

crates/wasm-encoder/src/exports.rs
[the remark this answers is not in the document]
peterhuene: I've merged the encoding of `Export` and `TypeRef` in the hopes they will be better aligned once the binary grammar has been updated in the component model spec, but once the spec is updated I'll revisit these encodings to make sure they are correct.

crates/wasm-encoder/src/functions.rs
peterhuene: Some of these checks were removed where possible.

crates/wasm-encoder/src/imports.rs
peterhuene: I've added an `import_by_name` function specifically for encoding functions (totally open to a better name than that one).

crates/wasm-encoder/src/exports.rs
```diff
@@ -48,24 +86,32 @@ impl ExportSection {
...
+    /// Define an export in the export section.
     pub fn export(&mut self, name: &str, export: Export) -> &mut Self {
+        self.uses_component_exports |= match export {
+            Export::Function(_)
+            | Export::Table(_)
+            | Export::Memory(_)
+            | Export::Global(_)
+            | Export::Tag(_) => false,
+            Export::Instance(_) | Export::Module(_) | Export::AdapterFunction(_) => true,
+        };
```
alexcrichton, line 98: Personally I still feel that this sort of guard isn't necessary. It looks like `uses_component_exports` is purely used for an assert, and otherwise there wouldn't need to be any handling of components-vs-not.

I mentioned this before, but I still feel that `wasm-encoder` shouldn't be in the business of going the extra mile to ensure that only a valid module is emitted. There's naturally a large number of properties that aren't checked during encoding so there's a line to draw at what's checked and what isn't. I think it's reasonable to have some guarantees about the static structure via the API design and properties we can get trivially from that, but otherwise going above-and-beyond what the APIs are providing feels out of scope of `wasm-encoder` to me at least.

crates/wasm-encoder/src/instances.rs
```diff
@@ -1,28 +1,30 @@
-use super::*;
+use crate::{encoders, ComponentSection, ComponentSectionId, Export};
```
alexcrichton, line 1: FWIW this file is sort of what I would naively expect to be the shape of most of the module linking transition to the new component model format. Most of the previous module linking implementation is still applicable, everything just has a slightly different shape with a few updates here and there.

crates/wasm-encoder/src/lib.rs
```diff
@@ -209,31 +237,79 @@ impl Default for Module {
...
 
-/// Known section IDs.
+/// Represents a WebAssembly component that is being encoded.
 ///
-/// Useful for implementing the `Section` trait, or for setting
-/// `RawSection::id`.
+/// Unlike core WebAssembly modules, the sections of a component
+/// may appear in any order and may be repeated.
+#[derive(Clone, Debug)]
+pub struct Component {
```
alexcrichton, line 245: My personal preference would be to get to the point where we can do away with this type. It looks like the component/module distinction is purely used for assertions right now and doesn't actually otherwise affect encoding (beyond the version byte). In that sense I commented above about how I personally think that the assertions aren't necessary, and if we go that route then I think there's room for:

```rust
impl Module {
    fn new() -> Module; // creates a core wasm module
    fn component() -> Module; // creates a component module
    // ...
}
```

I think that would also allow the removal of `ComponentSection`? There might need to be a new layer of abstraction for encoding the section id where some section ids all encode as the same value if the value clashes between core wasm and not, but I don't think that poses an API issue for this crate.

Basically I feel that this is a good opportunity to have core wasm and the component model be closely aligned in terms of API hierarchy and encourage maximal reuse between the two.

crates/wasm-smith/src/lib.rs
```diff
@@ -104,9 +104,6 @@ pub struct Module {
     config: Rc<dyn Config>,
     valtypes: Vec<ValType>,
 
-    /// Outer modules, if any (used for module linking)
-    outers: Vec<Outer>,
```
alexcrichton, line 108: We talked a bit at the wasmtime meeting yesterday, but to copy over some of what we concluded I believe the current thinking is that we may want to hold off on the deletion of the current module linking support until there's a better sense of the timeline of what it takes to replace it with the current component model writeup.

I don't think that means that you should implement module-linking for the new component model here-and-now but I do think that it may delay landing this PR. That being said I also think that all this support in `wasm-smith` would be somewhat useful to not throw away since I'm at least naively under the impression that it should be largely reusable for the next iteration of wasm-smith-module-linking-using-modules. The main difference between the historical module linking implementation and the current one is what types of items can be in a module and the representation of types, which feels like it should be a smaller delta from the current `wasm-smith` implementation than deleting everything and bringing it all back.

In any case I think the basic tl;dr; is that we may want to figure out a timeline for implementing all the various components since there was some pushback to having module linking support removed for now.

crates/wasm-encoder/src/lib.rs
peterhuene: If we can align the section ids, I will remove `ComponentSection` and `Component` entirely.

crates/wasm-smith/src/lib.rs
peterhuene: That's fine, there's no rush to get this merged in, the component tooling being built on top of these changes can work off my branch for now.

peterhuene:
This is blocked on spec changes needed to simplify the API.

peterhuene:
@alexcrichton so this has now been updated with the latest proposal changes.

Unfortunately, the updates ended up actually making the component and module formats _more_ disjoint, which actually makes a lot of sense to be honest (for example, the addition of a _component type_ made it such that components can now only import or export non-core items). As a result, I left `ComponentSection` and `Component` in place and broke apart a few "commonly-named" encoders (e.g. `TypeSection` as not only are the encodings entirely different, but there's now zero overlap between the two encoders in terms of _what types_ can be encoded).

As we're now targeting the `component-model` branch with this PR, which will serve as a staging ground for all the changes, what are your thoughts on reviewing this and merging it even with perhaps not the most ideal API? I'm still planning on having a single `Parser` and `Validator` for `wasmparser`, if that's any consolation.

I'm starting to think that component and module encoding _should_ be different (where appropriate) as at least the encoders now have a little more correctness in terms of what they permit; obviously it's still possible to emit an invalid encoding based on validation rules, but now users won't easily and accidentally encode a core function into a component function section, or vice versa, for example.

alexcrichton:
Ok nice! This looks good to me (only skimming, I think "real review" for this will largely come out from integration elsewhere and/or I can look more closely at merge-to-`main` time). 

Given that this is on the `component-model` branch instead of `main` I think it's good to get the ball rolling there, so I'm going to merge!

PR #452, opened 2022-01-21
wasm-shrink: Don't rename files across devices

### LINKED ISSUE #451: Failed to rename the result file
When running `wasm-tools shrink` I receive the following error:

```
Error: Failed to rename /run/user/1000/.tmpciSPbc to ./e.shrunken.wasm

Caused by:
    Invalid cross-device link (os error 18)
```

Neither of these directories are on a remote share or anything like that. `/run/user/1000` is a tmpfs and `./e.shrunken.wasm` is on a zfs pool.

The error message would suggest that rather than rename, a hard link is being attempted.

Calling `wasm-tools shrink` as such:

```
wasm-tools shrink ./wasm_shrink.sh e.wasm -o /run/user/1000/shrunken.wasm
```

makes the error go away.

alexcrichton:
Instead make sure that the temporary file used to create the output is
in the same directory as the final destination file so the `rename`
should always succeed.

Closes #451

### DIFF
--- a/src/bin/wasm-tools/shrink.rs
+++ b/src/bin/wasm-tools/shrink.rs
@@ -2,6 +2,7 @@ use anyhow::{Context, Result};
 use clap::Parser;
 use is_executable::IsExecutable;
 use std::path::{Path, PathBuf};
+use tempfile::NamedTempFile;
 use wasm_shrink::{IsInteresting, WasmShrink};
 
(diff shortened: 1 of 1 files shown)


fitzgen (approved):
LGTM, not really something that is easy/worth testing, but maybe add a comment here documenting why we aren't just always using `NamedTempFile::new()`?

PR #453, opened 2022-01-21
Synchronize ops syntax for relaxed SIMD proposal

yurydelendik:
Changes:

* matches syntax with LLVM repo, see e.g. https://github.com/llvm/llvm-project/blob/main/llvm/test/MC/WebAssembly/simd-encodings.s
* fix wrong amount of argument for fma/fms instructions


alexcrichton:
Thanks! Looks good to me, but would you also be up for renaming the enum variants in Rust to match the new names in the text format?

yurydelendik:
> renaming the enum variants in Rust to match the new names

yep, will do that today

alexcrichton (approved):

PR #454, opened 2022-01-22
wasmparser: minimize diff for no_std support mirrors

Robbepop:
This PR does **_not_** introduce `no_std` support to the `wasmparser` crate.
Instead it just simplifies the process of supporting a `no_std` mirror of the `wasmparser` crate by minimizing the required diff.

This also replaces `Hash{Map,Set}` usage in the `wasmparser` validator with `BTree{Map,Set}` since those are not attackable in `no_std` environments. I saw absolutely no difference in the benchmarks. Using `BTree{Map,Set}` also leads to more deterministic runtime behavior in general.

@alexcrichton Many projects would love to use this high quality and battle tested Wasm parsing and validation crate in a `no_std` environment so I thought about a solution that fits those projects while not putting a burden on the maintainers (you). This PR just minimizes the diff needed to add trivial `no_std` support to `wasmparser` but does not actually introduce `no_std` support. With this PR it will become very easy to maintain a `no_std` fork of `wasmparser ` and keep it in sync. I'd be happy to know how you feel about this idea.

Link to `no_std` `wasmparser` fork: https://crates.io/crates/wasmparser-nostd

### DIFF
--- a/crates/wasmparser/src/binary_reader.rs
+++ b/crates/wasmparser/src/binary_reader.rs
@@ -13,10 +13,10 @@
  * limitations under the License.
  */
 
+use alloc::{boxed::Box, format, vec::Vec};
 use core::convert::TryInto;
 use core::fmt;
 use core::str;
-use alloc::{vec::Vec, format, boxed::Box};
 
 use crate::limits::*;
 
--- a/crates/wasmparser/src/binary_reader.rs
+++ b/crates/wasmparser/src/binary_reader.rs
@@ -13,9 +13,10 @@
  * limitations under the License.
  */
 
-use std::convert::TryInto;
-use std::fmt;
-use std::str;
+use core::convert::TryInto;
+use core::fmt;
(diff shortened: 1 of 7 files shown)


alexcrichton:
Personally I do not think a PR like this is right to merge into `wasmparser`, it's basically getting all the downsides of "no-std support" without actually supporting it in this crate, which seems like an odd place to be for me. The downsides I personally see are:

* Non-standard import paths (e.g. take-your-pick of `alloc` or `core`)
* Non-standard choices of collections (e.g. switching `Hash*` to `BTree*`)
* No testing on CI that this state is maintained
* Non-standard imports of standard types like `String` and `Vec` which are otherwise expected to be in the prelude

I understand how "supporting no-std" would basically be putting `#![no_std]` at the top of the crate with this patch, but that's a project level decision which currently AFAIK has said won't be supported in `wasmparser` at this time.

Issue #459, opened 2022-01-29
wasmparser: Return concrete type for BrTable::targets


Robbepop:
Currently `BrTable::targets` returns an [`impl Trait` type: `impl Iterator<Item = Result<u32>`](https://github.com/bytecodealliance/wasm-tools/blob/ac9028a4c52dbf1b0e78b24be33b7e5ce9618ca6/crates/wasmparser/src/binary_reader.rs#L1985).
This can be unflexible for user code, e.g. when user code wants to efficiently wrap the `BrTable` for a cleaner API in the client code.
However, with an `impl Trait` this is not really possible to my knowledge.

# Example

```rust
/// A thin-wrapper around a Wasm `br_table` construct.
///
/// # Note
///
/// This wrapper assumes that Wasm validation of the `br_table` has already
/// happened and therefore panics upon encountering errors instead of forwarding
/// those.
#[derive(Debug)]
pub struct WasmBrTable<'a> {
    /// The underlying inner `wasmparser` branching table.
    br_table: wasmparser::BrTable<'a>,
}

impl<'a> BrTable for WasmBrTable<'a> {
    type Targets = WasmBrTableTargets; // ERROR: missing generic argument

    fn len(&self) -> u32 {
        self.br_table.len()
    }

    fn is_empty(&self) -> bool {
        self.br_table.is_empty()
    }

    fn default(&self) -> RelativeDepth {
        RelativeDepth::from_u32(self.br_table.default())
    }

    fn targets(&self) -> Self::Targets {
        WasmBrTableTargets {
            iter: self.br_table.targets(),
        }
    }
}

/// Thin wrapper around the targets of a `br_table` coming from `wasmparser`.
///
/// # Note
///
/// This wrapper assumes that Wasm validation of the `br_table` has already
/// happened and therefore panics upon encountering errors instead of forwarding
/// those.
pub struct WasmBrTableTargets<T> {
    /// The underlying iterator of the target relative depths.
    iter: T,
}

impl<T> Iterator for WasmBrTableTargets<T>
where
    T: Iterator<Item = Result<u32, wasmparser::BinaryReaderError>>,
{
    type Item = RelativeDepth;

    fn next(&mut self) -> Option<Self::Item> {
        self.iter
            .next()
            .map(|item| {
                item.unwrap_or_else(|error| {
                    panic!(
                        "encountered unexpected invalid relative depth of `br_table`: {}",
                        error
                    )
                })
            })
[...]

PR #460, opened 2022-01-31
Return a concrete iterator type for BrTable::targets

Robbepop:
Closes https://github.com/bytecodealliance/wasm-tools/issues/459

This PR does not implement `ExactSizeIterator` since I was not sure if it actually matches the semantics of an `ExactSizeIterator` due to the fact that the iterator might not yield as many items as expected since the `reader` is not checked prior.


alexcrichton:
Thanks!

PR #462, opened 2022-02-01
Begin adding support for components to `wast`

sunfishcode:
This begins adding support for `*.cat` files, which are like `*.wat` files, but hold `(component ...)` s-expressions representing wasm components, following the AST design [here]. This has several `todo` markers left to be done, but has enough code to illustrate the high-level shape of this support: `Module` has an `is_component` field to indicate whether it was parsed from `(component ...)` versus `(module ...)` indicating a core-wasm component.

[here]: https://github.com/WebAssembly/component-model/blob/ast-and-binary/design/MVP/Explainer.md

This makes `ValType` non-`Copy`, as it can now contain interface types which can contain `Vec`s.

This doesn't yet contain any tests; I'll add tests as I implement the `todo`s.

### DIFF
--- a/crates/wast/README.md
+++ b/crates/wast/README.md
@@ -29,6 +29,7 @@ types to parse anything that looks like a WebAssembly s-expression.
 
 * Need to parse a `*.wat` file?
 * Need to parse a `*.wast` file?
+* Need to parse a `*.cat` file?
 * Need to run test suite assertions from the official wasm test suite?
 * Want to write an extension do the WebAssembly text format?
 
@@ -38,12 +39,12 @@ interface if all you'd like to do is convert `*.wat` to `*.wasm`.
 
 ## Cargo features
 
-By default this crate enables and exports support necessary to parse `*.wat` and
-`*.wast` files, or in other words entire wasm modules. If you're using this
-crate, however, to parse simply an s-expression wasm-related format (like
-`*.witx` or `*.wit` perhaps) then you can disable the default set of features to
-only include a lexer, the parsing framework, and a few basic token-related
-parsers.
+By default this crate enables and exports support necessary to parse `*.wat`,
+`*.wast`, and `*.cat` files, or in other words entire wasm modules. If you're
+using this crate, however, to parse simply an s-expression wasm-related format
+(like `*.witx` or `*.wit` perhaps) then you can disable the default set of
+features to only include a lexer, the parsing framework, and a few basic
+token-related parsers.
 
 ```toml
 [dependencies]
--- a/crates/wast/src/ast/mod.rs
+++ b/crates/wast/src/ast/mod.rs
@@ -350,6 +350,7 @@ pub mod kw {
     custom_keyword!(catch);
     custom_keyword!(catch_all);
     custom_keyword!(code);
+    custom_keyword!(component);
     custom_keyword!(data);
     custom_keyword!(dataref);
     custom_keyword!(declare);
--- a/crates/wast/src/ast/mod.rs
+++ b/crates/wast/src/ast/mod.rs
@@ -62,7 +62,7 @@ macro_rules! custom_keyword {
         }
 
         impl $crate::parser::Peek for $name {
-            fn peek(cursor: $crate::parser::Cursor<'_>) -> bool {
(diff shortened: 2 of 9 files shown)


crates/wast/README.md
```diff
@@ -29,6 +29,7 @@ types to parse anything that looks like a WebAssembly s-expression.
 
 * Need to parse a `*.wat` file?
 * Need to parse a `*.wast` file?
+* Need to parse a `*.cat` file?
```
alexcrichton, line 32: Could we possibly use the `*.wat` extension instead of creating a new one? I was originally under the impression that components were "just another kind of wasm" so if they still have `*.wasm` as the binary format extension then it seems like we'd still want to use `*.wat` as the text extension. This doesn't really affect this PR too much, mainly just the documentation, but I'd ideally like to avoid components feeling like a separate world if we can.

crates/wast/src/ast/mod.rs
```diff
@@ -62,7 +62,7 @@ macro_rules! custom_keyword {
         }
 
         impl $crate::parser::Peek for $name {
-            fn peek(cursor: $crate::parser::Cursor<'_>) -> bool {
+            fn peek(cursor: $crate::parser::Cursor<'_>) -> std::primitive::bool {
```
alexcrichton, line 65: FWIW I think it's fine below to do something like `bool_ = "bool"` to avoid the need for this. I don't think the usage of raw identifiers to avoid clashes with rust identifiers has really worked out all that well and at some point I should probably move everything that's a rust identifier to an underscore after it as well. Anyway minor point, could go either way here.

crates/wast/src/ast/module.rs
```diff
@@ -36,9 +37,45 @@ impl<'a> Parse<'a> for Wat<'a> {
...
+        module.validate(parser)?;
+        Ok(Cat { module })
+    }
+}
+
 /// A parsed WebAssembly module.
 #[derive(Debug)]
 pub struct Module<'a> {
+    /// Is this a component module, rather than a core module?
+    pub is_component: bool,
```
alexcrichton, line 78: One possible downside to this implementation is that this may run the risk of making the internals more complicated than they otherwise need to be or make encoding more difficult. For example a components and modules have a sort of a venn-diagram of supported fields. For example exports/imports are pretty similar (but imports are notably different with one-or-two levels) but otherwise modules allow for table definitions while components only allow for aliases. Similarly data segments and element segments are allowed in modules but not components.

This probably mostly adds complexity to the resolver in the `wast` crate. Today it handles nested modules as if they could contain everything because that's what module linking originally was, but we actually in theory should be able to remove a ton of that complexity since we modules are always leaves and components can only refer, in a limited fashion, to other items.

Overall I'm not sure if this is the best representation or whether it would be better to have this split (a `Module` type and a `Component` type). You mentioned that having `Component` might introduce a good deal of duplication but I feel like a lot of complexity, especially around the resolver, may end up getting shifted largely to components in such a design. 

I wanted to write this down, but again I'm not sure what the right answer is. It's probably best to just pick one you find reasonable and go with that but I think we should keep our eyes peeled for places where it may be better to go with an alternate design.

crates/wast/src/ast/types.rs
```diff
@@ -66,6 +69,244 @@ impl<'a> Peek for ValType<'a> {
...
+        } else if l.peek::<kw::string>() {
+            parser.parse::<kw::string>()?;
+            Ok(InterType::String)
+        } else if l.peek::<kw::record>() {
+            let record = parser.parse::<Record>()?;
+            Ok(InterType::Record(record))
+        } else if l.peek::<kw::variant>() {
+            let variant = parser.parse::<Variant>()?;
+            Ok(InterType::Variant(variant))
+        } else if l.peek::<kw::tuple>() {
```
alexcrichton, line 152: The parsing here may end up being a little different (for everything else here that isn't just a plain keyword as well) because I suspect we'd want this represented as `(variant ...)` or `(tuple ...)` surrounded in parens.

alexcrichton (commented):
Seems like a reasonable start to me!

peterhuene:
Can we retarget this PR at `component-model`?

sunfishcode:
This is superseded by #475.

PR #465, opened 2022-02-03
wasmparser: no functype blocks without multi-value

nagisa:
Block types have been generalzed from `resulttype := valtype?` to
`functype` with the integration of the multi-value proposal. Before then
producing a wasm with functype-typed blocks would constitute a malformed
WASM.

This adjusts wasmparser to reject blocks with functype results unless
multi-value is enabled. This also results in somewhat worse error
message, however.

### DIFF
--- a/crates/wasmparser/src/operators_validator.rs
+++ b/crates/wasmparser/src/operators_validator.rs
@@ -484,24 +484,14 @@ impl OperatorValidator {
                 self.check_reference_types_enabled()
             }
             TypeOrFuncType::Type(Type::V128) => self.check_simd_enabled(),
-            TypeOrFuncType::FuncType(idx) => {
-                let ty = func_type_at(&resources, idx)?;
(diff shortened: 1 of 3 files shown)


nagisa:
(TODO: needs a test)

I have discovered this during development of #464. wasm2wat from wabt would reject some of the wasm modules that were autogenerated that wasm-parser would continue accepting. I believe this specific issue to have been the cause.

nagisa:
I would like some advise on what the best way would be to construct a test file that uses functype block(s) and yet are minimal. I'm not really finding any tooling to build wasm structure from scratch.

nagisa:
Added the test. I'm not sure if we had any tests for blocks with parameters or multiple returns in the past, but structurally wasm requires to use a functype in those situations anyway and so the newly added test should cover those situations as well.

alexcrichton:
Thanks!

PR #466, opened 2022-02-04
wasm-mutate: avoid mutating empty code sections

nagisa:
The `Rng::gen_range`, which both `SnipMutator` and
`FunctionBodyUnreachable` mutators utilize to pick a function to mutate,
has a panicking failure mode when the range provided to it is empty.

This failure mode may be triggered if the mutation ends up removing all
of the functions, thus leaving an empty code section.

One could argue that removing the section as it becomes empty would be a
cleaner fix for this particular problem. Being able to verify operation
of the predicate when fed modules with empty sections is an important
feature that should be preserved.

### DIFF
--- a/crates/wasm-mutate/src/info.rs
+++ b/crates/wasm-mutate/src/info.rs
@@ -230,6 +230,17 @@ impl<'a> ModuleInfo<'a> {
         Ok(info)
     }
 
+    pub fn has_nonempty_code(&self) -> bool {
+        if let Some(section) = self.code {
+            let section_data = self.raw_sections[section].data;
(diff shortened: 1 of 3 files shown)


nagisa:
(TODO: this wants a regression test… its gonna be another tricky one to construct)

alexcrichton:
Thanks!

fitzgen:
> One could argue that removing the section as it becomes empty would be a
> cleaner fix for this particular problem.

I think this should be a separate mutation, fwiw, since there could be bugs relating to present-but-empty sections.

nagisa:
That's exactly what I was trying to say, maybe poorly ^^

PR #469, opened 2022-02-05
wasm-mutate: simplify initializer const expressions

nagisa:
This can replace WASM such as:

     (elem (global.get 0) func)
     => (elem (i32.const 0) func)

or

    (global funcref (func $f))
    => (global funcref (ref.null))

which then can lead to further reduction opportunities such as removing
entities no longer referenced. That said, today this doesn't always help
wasm-shrink, since some of these reductions have the same encoded
byte-length, and therefore `wasm-shrink` decides the arguably simpler
result is not actually interesting for further consideration…

For that reason I'm marking this as a draft and I guess the intent I've here
is to kick off a discussion on whether there is a better metric for wasm-shrink
to use during reduction.

In particular a question I have is why `wasm-shrink` doesn't blindly trust that
`wasm-mutate` will produce a “simpler” wasm after every step. Sure, I imagine a 
`Mutator` could forget to check for `config.reduce` and actually fail to produce
an objectively simpler WASM at times, but ultimately even in presence of such a 
bug the mistake would most likely be undone by some other invocation of some 
`Mutator`. Not to mention wasm-smith itself could remember the smallest wasm 
file it has seen and present that to the user, even if it wasn't the exact same
as the working state at the end of the reduction process.

### DIFF
--- a/crates/wasm-mutate/src/mutators/modify_init_exprs.rs
+++ b/crates/wasm-mutate/src/mutators/modify_init_exprs.rs
@@ -1,0 +1,190 @@
+//! This mutator modifies the constant initializer expressions between various valid forms in
+//! entities which require constant initializers.
+
+use crate::mutators::translate::{self, InitExprContext, Item, Translator};
+use crate::{Error, Mutator, Result};
+
+use rand::Rng;
+use wasm_encoder::{ElementSection, GlobalSection, Instruction};
+use wasmparser::{ElementSectionReader, GlobalSectionReader, InitExpr, Operator, Type};
+
+#[derive(Copy, Clone)]
+pub enum InitExpressionMutator {
+    Global,
+    ElementOffset,
+    ElementFunc,
+}
+
+struct InitTranslator {
+    new_value: u128,
+    skip_inits: u32,
+    context: InitExprContext,
+}
+
+impl Translator for InitTranslator {
+    fn as_obj(&mut self) -> &mut dyn Translator {
+        self
+    }
+
+    fn remap(&mut self, item: Item, idx: u32) -> Result<u32> {
+        match (self.context, item) {
+            // Element's function items can either be presented to `Translator` via `remap` here or
+            // via expressions (the function below). The API here is insufficiently flexible to
+            // produce null funcrefs so at least lets try pointing at the lowest numbered function
+            // instead.
+            //
+            // TODO: may make sense to actually generate the functions pointed at more randomly.
+            (InitExprContext::ElementFunction, Item::Function) => {
+                if self.skip_inits != 0 {
+                    self.skip_inits -= 1;
+                    Ok(idx)
+                } else {
+                    log::trace!("... replacing referenced function index with 0");
+                    self.skip_inits = u32::MAX;
+                    Ok(0)
+                }
+            }
+            _ => Ok(idx),
+        }
+    }
+
+    fn translate_init_expr(
+        &mut self,
+        e: &InitExpr<'_>,
+        ty: &Type,
+        ctx: InitExprContext,
+    ) -> Result<Instruction<'static>> {
+        use {Instruction as I, Type as T};
+        if ctx != self.context {
+            return translate::init_expr(self.as_obj(), e);
+        } else if self.skip_inits != 0 {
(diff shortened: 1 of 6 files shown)


alexcrichton:
Seems reasonable to me! I think we could update wasm-shrink to allow for a little slop of generating slightly larger modules or something like it gets up to 10 steps of mutations to generate a smaller module to solve the issue you're running into.

crates/wasm-mutate/src/mutators/modify_init_exprs.rs
```diff
@@ -0,0 +1,190 @@
...
+                    Ok(idx)
+                } else {
+                    log::trace!("... replacing referenced function index with 0");
+                    self.skip_inits = u32::MAX;
+                    Ok(0)
+                }
+            }
+            _ => Ok(idx),
+        }
+    }
```
fitzgen, line 49: Shouldn't `remap` never get called for the to-be-mutated entity inside an init expression since you override `tanslate_init_expr` (and never call `remap` inside it for the target entity) below?

fitzgen (commented):
Thanks! This will be really nice to have, just a few things that need to get resolved first.

crates/wasm-mutate/src/mutators/modify_init_exprs.rs
nagisa: The `Mutator::mutate` calls `Translator::translate_element` which calls `translate::element` which in turn can call either `Translator::remap` or `Translator::translate_init_expr` depending on what shape the `elem`'s initializers have.

fitzgen: Ah okay -- do you want to override `translate::element` then so that you can produce null funcrefs there? Also fine to just stick with this as is for now.

nagisa: I'll investigate this, yeah. I wasn't sure initially if `ElementItem::Func` style elements can actually encode nulls and I'm also not sure if encoding between `ElementItem::Func` and `ElementItem::Expr` is meaningfully different – if it is, I think there's likely a fair amount of value in not retaining `ElementItem::Func` encoding over producing `ref.null`s.

fitzgen: Ah that's a good point!

I think actually what you have now is good then, because the default table elements are null, so we can simplify `ElementItem::Func`s to null via removing elements from the segment.

But we want to generate/support/mutate both kinds of segment encodings just to exercise those code paths in parsers/runtimes/etc.

crates/wasm-mutate/src/mutators/modify_init_exprs.rs
```diff
@@ -0,0 +1,190 @@
...
+            },
+        };
+        match self {
+            Self::Global => {
+                let num_total = config.info().num_local_globals();
+                translator.new_value = match config.rng().gen::<u8>() {
+                    0..=99 => 0,
+                    100..=199 => 1,
+                    200.. => config.rng().gen::<u128>(),
+                };
```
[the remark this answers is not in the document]
nagisa: I went ahead and made this mutator only work when in reduction mode. I don't really feel like I'll be able to dedicate the attention necessary to come up with a generic `Operator -> Instruction` mapping that handles all these cases well. If there's already one, though, I'm happy to change the code to use it.

crates/wasm-mutate/src/mutators/modify_init_exprs.rs
```diff
@@ -0,0 +1,334 @@
...
+            }
+        }
+    }
+
+    fn can_mutate(&self, config: &crate::WasmMutate) -> bool {
+        // the implementation here can only reduce for now,
+        // but could be extended to mutate arbitrarily.
+        if !config.reduce {
+            return false;
+        }
```
fitzgen, line 242: Yeah this is fine for now.

fitzgen (approved):
Looks good, thanks! Just needs a tiny rebase and then we can merge this.

nagisa:
Rebased

fitzgen:
Thanks!

PR #470, opened 2022-02-05
wasm-mutate: tell peephole about float and ref const replacements

nagisa:
Some time ago I was reducing a test case that utilized floating point numbers and eventually managed to reduce the test case further by hand by replacing all sorts of floating point operations, primarily those dealing with function locals, with their `*.const` versions.

Much like the #469, in many cases these rules are prone to producing WASM that's equivalent or larger byte-wise even if it sets up for more opportunities for further reduction.

### DIFF
--- a/crates/wasm-mutate/src/mutators/peephole/eggsy/lang.rs
+++ b/crates/wasm-mutate/src/mutators/peephole/eggsy/lang.rs
@@ -4,6 +4,7 @@ use egg::Id;
 use std::convert::TryInto;
 use std::fmt::{self, Display};
 use std::str::FromStr;
+use wasm_encoder::ValType;
 
 /// This is a macro used to define the `Lang` enum.
 ///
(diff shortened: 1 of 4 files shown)


crates/wasm-mutate/src/mutators/peephole/rules.rs
```diff
@@ -415,6 +415,16 @@ impl PeepholeMutator {
...
                 ),
+                rewrite!(
+                    "replace-with-f32-1";
+                    "?x" => "f32.const.1"
+                        if self.is_type("?x", PrimitiveTypeInfo::F32)
+                ),
+                rewrite!(
+                    "replace-with-f64-1";
+                    "?x" => "f64.const.1"
+                        if self.is_type("?x", PrimitiveTypeInfo::F64)
```
alexcrichton, line 426: I think this may not work as intended since this produces a float equivalent to `f64::from_bits(1)`, so for this to actually use `1.0` I think it'd need to modify the AST to store a `f64` or a newtype wrapper or something like that.

crates/wasm-mutate/src/mutators/peephole/eggsy/lang.rs
```diff
@@ -700,6 +704,10 @@ lang! {
         F32(u32) = "f32.const",
         /// F64 constant node
         F64(u64) = "f64.const",
+        /// constant ref.null funcref node
+        FuncRefNull = "funcref.null",
+        /// constant ref.null externref node
+        ExternRefNull = "externref.null",
```
alexcrichton, line 710: Oh I appear to have forgotten `ref.null` in the language definition here when I added bulk-memory support. Would you be up for switching this to `RefNull(ValType)` or similar so this can be `ref.null func` and `ref.null extern` (or similar)

alexcrichton (commented):
Thanks!

crates/wasm-mutate/src/mutators/peephole/eggsy/lang.rs
nagisa: `ValType` doesn't really fit here all that well – `ValType` would need a `Display` and a `FromStr` implementation, and also it covers more distinct types than `ref.null` can handle either way.

I went ahead and added a `RefType` enum instead with a conversion to `ValType`.

crates/wasm-mutate/src/mutators/peephole/rules.rs
nagisa: I… went with hardcoding the bit pattern for 1.0 instead. Not the prettiest solution, but it isn't clear to me if we'll be interested in any other constants and the rewrite rule name acts as a pretty good documentation as to what is happening here.

nagisa:
I was hoping to add some tests as well, but peephole test suite seems to be focused around testing the operation of separate rules (it even had a test with a `=> f32.const 0` replacement rule!) rather than the operation of the mutator with its default rules as a whole.

alexcrichton:
Yeah I think this is fine, a lot of testing will happen on oss-fuzz with fuzzing as well and if necesary we can add full end-to-end tests eventually too.

PR #474, opened 2022-02-08
`wasm-smith`: Start generating arbitrary Wasm components

fitzgen:
This adds initial support for Wasm components in `wasm-smith`. This support for
components is incomplete:

* Only custom and type sections are currently generated.

* This commit doesn't add support for encoding components, ensuring termination,
  etc... yet.

Regardless, this commit provides a kernel of functionality and establishes a
skeleton for adding the rest of the missing functionality.


crates/wasm-smith/src/component.rs
```diff
@@ -0,0 +1,802 @@
...
+            7 => Ok(ValueType::U32),
+            8 => Ok(ValueType::S64),
+            9 => Ok(ValueType::U64),
+            10 => Ok(ValueType::F32),
+            11 => Ok(ValueType::F64),
+            12 => Ok(ValueType::Char),
+            13 => Ok(ValueType::String),
+            14 => {
+                let max_type = u32::try_from(self.current_type_scope().len() - 1).unwrap();
+                Ok(ValueType::Indexed(u.int_in_range(0..=max_type)?))
```
alexcrichton, line 416: Naively this feels like it might generate an invalid module, because I think types can point to things like functions and modules, right? (and we can't have value types for those)

crates/wasm-smith/src/component.rs
```diff
@@ -0,0 +1,802 @@
...
+                    // Imports.
+                    let name = crate::unique_string(100, &mut imports, u)?;
+                    let ty = u.int_in_range(0..=me.current_type_scope().len() - 1)?;
+                    let ty = u32::try_from(ty).unwrap();
+                    defs.push(ComponentTypeDef::Import(Import { name, ty }));
+                } else {
+                    // Instance type definitions.
+                    let def = me.arbitrary_instance_type_def(u, &mut exports)?;
+                    defs.push(ComponentTypeDef::Instance(def));
+                }
```
alexcrichton, line 293: My ignorance of the current spec is probably showing but I thought components were similar to modules where they have imports, exports, aliases, and possible type definitions themselves. That makes me naively think like things are missing here but wanted to confirm.

crates/wasm-smith/src/component.rs
```diff
@@ -0,0 +1,802 @@
...
+            Ok(true)
+        })?;
+        Ok(RecordType { fields })
+    }
+
+    fn arbitrary_variant_type(&mut self, u: &mut Unstructured<'a>) -> Result<VariantType> {
+        let mut cases = vec![];
+        let mut case_names = HashSet::new();
+        arbitrary_loop(u, 0, 100, |u| {
+            cases.push(self.arbitrary_named_type(u, &mut case_names)?);
```
alexcrichton, line 446: This might be more of a question for the spec, but I thought that the type payload for variants was optional? (only the name would be required)

crates/wasm-smith/src/component.rs
```diff
@@ -0,0 +1,802 @@
...
+            Ok(true)
+        })?;
+        Ok(EnumType { variants })
+    }
+
+    fn arbitrary_union_type(&mut self, u: &mut Unstructured<'a>) -> Result<UnionType> {
+        let mut variants = vec![];
+        let mut variant_names = HashSet::new();
+        arbitrary_loop(u, 0, 100, |u| {
+            variants.push(crate::unique_string(100, &mut variant_names, u)?);
```
alexcrichton, line 499: I may again be missing things from the spec, but I thought unions were unions of types rather than unions of strings? (basically the union/enum cases may be duplicated here by accident)

crates/wasm-smith/src/component.rs
```diff
@@ -0,0 +1,802 @@
...
+#[derive(Clone, Debug)]
+struct ComponentType {
+    defs: Vec<ComponentTypeDef>,
+}
+
+#[derive(Clone, Debug)]
+enum ComponentTypeDef {
+    Instance(InstanceTypeDef),
+    Import(Import),
+}
```
alexcrichton, line 666: Expanding on my comment from above, I thought that this would have an import, export, type, and alias case. I didn't think that instances would be singled-out here

crates/wasm-smith/src/component.rs
```diff
@@ -0,0 +1,802 @@
...
+    Record(RecordType),
+    Variant(VariantType),
+    List(ListType),
+    Tuple(TupleType),
+    Flags(FlagsType),
+    Enum(EnumType),
+    Union(UnionType),
+    Optional(OptionalType),
+    Expected(ExpectedType),
+    Named(NamedType),
```
alexcrichton, line 722: This is probably more of a spec question, but what is a named type at the spec level? (or why have named types?)

Also, this is notably missing resources which is a major complexity point in my mind, so I was curious if that was missing from the spec or intentionally missing from here

crates/wasm-smith/src/component.rs
```diff
@@ -0,0 +1,802 @@
...
+struct InstanceSection {}
+
+#[derive(Debug)]
+struct ExportSection {}
+
+#[derive(Debug)]
+struct StartSection {}
+
+#[derive(Debug)]
+struct AliasSection {}
```
alexcrichton, line 802: FWIW I think it's fine to remove these empty structs and the empty methods with `todo!()` for now and we can add them back in as they get implemented

alexcrichton (commented):
Seems reasonable to me!

One thing that I think may be worth considering though is that in the original implementation of module linking I had to have special affordance for creating small-ish types instead of possibly-hugely-recursive types. I'm sort of fuzzy on as to why but wasm-smith would often time out on fuzzing and/or validation and/or type-checking and/or anything else along the line. I ended up fixing this by having validation assert the "effective size" of any item's type was reasonably sized, something like a few thousand.

As written though each layer of type nesting can be up to 100 layers deep (as mostly loops are arbitrary from 0 to 100). This means you could have an instance with 100 instance types, each with 100 instance types each ... and so on up to 100 layers deep. I think that's up to 100^100 types or something like that, but we definitely want to protect against that.

I forget precisely how this was handled, but I think a fuel-like approach will be necessary and/or a maximum depth of type nesting. Validation will end up wanting to double-check all this to prevent something like Wasmtime from having to process too-big types, but even with that in place we don't want wasm-smith to time out trying to generate these massive modules.

fitzgen:
Regarding type sizes: yes we will have to do something like that before this merges into `main`. I was hoping that we could land this skeleton in the `component-model` branch and then do follow up PRs also targeting the `component-model` branch, rather than keeping this PR open for many iterations of review.

Does that sound alright?

crates/wasm-smith/src/component.rs
fitzgen: This would just be churn on my end. I'll have to add them back as soon as I merge this skeleton PR, and this PR isn't going into `main`, so I figured I could leave them as placeholders for now.

crates/wasm-smith/src/component.rs
```diff
@@ -0,0 +1,802 @@
...
+    Core(CoreSection),
+    Component(ComponentSection),
+    Instance(InstanceSection),
+    Export(ExportSection),
+    Start(StartSection),
+    Alias(AliasSection),
+}
+
+#[derive(Debug)]
+struct CustomSection<'a> {
```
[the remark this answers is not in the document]
fitzgen: No strong feelings either way. I haven't found the `'a` to be annoying yet, and figured we might as well avoid a copy if it is easy. But I also don't feel strongly enough to push back on copying.

crates/wasm-smith/src/component.rs
```diff
@@ -0,0 +1,802 @@
...
+        self.core_valtypes.push(ValType::I64);
+        self.core_valtypes.push(ValType::F32);
+        self.core_valtypes.push(ValType::F64);
+        if self.config.simd_enabled() {
+            self.core_valtypes.push(ValType::V128);
+        }
+        if self.config.reference_types_enabled() {
+            self.core_valtypes.push(ValType::ExternRef);
+            self.core_valtypes.push(ValType::FuncRef);
+        }
```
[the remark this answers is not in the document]
fitzgen: Yeah, I can factor it out into a helper.

crates/wasm-smith/src/component.rs
```diff
@@ -0,0 +1,802 @@
...
+        loop {
+            // Keep going while the fuzzer tells us to / has more data for us.
+            if !u.arbitrary()? {
+                break;
+            }
+
+            choices.clear();
+            choices.push(Self::arbitrary_custom_section);
+            if self.types.len() < self.config.max_types() {
+                choices.push(Self::arbitrary_type_section);
```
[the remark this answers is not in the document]
fitzgen: That's a good idea.

crates/wasm-smith/src/component.rs
```diff
@@ -0,0 +1,802 @@
...
+    fn arbitrary_module_type(&mut self, u: &mut Unstructured<'a>) -> Result<ModuleType> {
+        // TODO: this currently only supports function type definitions,
+        // function imports, and function exports.
+
+        let mut defs = vec![];
+        let mut types: Vec<Rc<crate::core::FuncType>> = vec![];
+        let mut imports = HashMap::new();
+        let mut exports = HashSet::new();
+
+        arbitrary_loop(u, 0, 100, |u| {
```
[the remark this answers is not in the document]
fitzgen: In the core module generation code we have a bunch of hard coded bounds for various string sizes and loop iterations already. I think we can do a follow up PR to clean up all of them at once.

crates/wasm-smith/src/component.rs
```diff
@@ -0,0 +1,802 @@
...
+            3 => Ok(Type::Func(self.arbitrary_func_type(u)?)),
+            4 => Ok(Type::Value(self.arbitrary_value_type(u)?)),
+            5 => Ok(Type::Compound(self.arbitrary_compound_type(u)?)),
+            _ => unreachable!(),
+        }
+    }
+
+    fn arbitrary_module_type(&mut self, u: &mut Unstructured<'a>) -> Result<ModuleType> {
+        // TODO: this currently only supports function type definitions,
+        // function imports, and function exports.
```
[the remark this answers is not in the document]
fitzgen: It's a bit funky since module types have their own scope in the component model spec and this is different from the scopes in core Wasm (and the old core wasm module linking). And since the representations of these scopes are different, I think we would need some kind of trait to abstract the differences away while still dealing with things like type sizes, etc. It kinda seems like all that machinery would be more of a headache than some sorta-but-not-quite duplicated code here. Thoughts?

crates/wasm-smith/src/component.rs
```diff
@@ -0,0 +1,802 @@
...
+            // Export.
+            0 if !self.current_type_scope().is_empty() => Ok(InstanceTypeDef::Export {
+                name: crate::unique_string(100, exports, u)?,
+                ty: u.int_in_range(
+                    0..=u32::try_from(self.current_type_scope().len()).unwrap() - 1,
+                )?,
+            }),
+
+            // Outer type alias.
+            1 if self.types_scopes.iter().any(|scope| !scope.is_empty()) => {
```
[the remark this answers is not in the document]
fitzgen: I think I was thinking that this was equivalent to the `choices`-vec pattern, and so this could avoid the heap allocations, but I think you're right that because the conditions are different it will skew the distributions. I'll switch to using a `choices` vec.

crates/wasm-smith/src/component.rs
fitzgen: I don't think that types can point to functions and modules? These types, here and elsewhere, are transcribed from the grammar in the explainer and binary documents.

crates/wasm-smith/src/component.rs
fitzgen: The components have these things, but the component *types* don't, IIUC.

crates/wasm-smith/src/component.rs
fitzgen: According to the grammars in the binary and explainer docs, the payload is not optional.

crates/wasm-smith/src/component.rs
fitzgen: They are, this is a bad copy paste. Good eye!

crates/wasm-smith/src/component.rs
fitzgen: `InstanceTypeDef` has exports, types, and aliases. I could probably flatten it into this to make the types a bit more clear, but that would also mean duplicating the code to generate this stuff.

crates/wasm-smith/src/component.rs
fitzgen: I'm not totally sure. Semantically they're identical to the underlying type.

crates/wasm-smith/src/component.rs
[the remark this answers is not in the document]
alexcrichton: I'm mostly worried about consumers like wasmtime where threading `'a` around when it's not already present can be pretty onerous and otherwise it doesn't seem all that important for perf here.

crates/wasm-smith/src/component.rs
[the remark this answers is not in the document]
alexcrichton: seems reasonable to me!

crates/wasm-smith/src/component.rs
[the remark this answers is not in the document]
alexcrichton: Ah yeah I don't know for sure or anything like that that this will be nicely deduplicated with some core helper. I'm mostly wondering if there's a commonality between the imports of a core wasm module and the imports here which could get all these other kinds of imports "for free". Seems fine to just implement it and see what happens though.

crates/wasm-smith/src/component.rs
alexcrichton: Hm ok I think that's something to bring up with the spec, I don't think that matches how we've been thinking about variants historically.

crates/wasm-smith/src/component.rs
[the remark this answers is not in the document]
fitzgen: Makes sense, I'll remove the `'a`.

crates/wasm-smith/src/component.rs
fitzgen: FWIW, `enum`s are variant types that don't have payloads, and we can always have the unit type payload as well. But, by all means, bring it up with the spec.

crates/wasm-smith/src/component.rs
fitzgen: I've flattened this now.

fitzgen:
Okay I think I've addressed the most immediate concerns in your review @alexcrichton. I think I will merge this into the `component-model` branch assuming CI comes back green and no one has any lingering concerns (speak up now!)

crates/wasm-smith/src/component.rs
peterhuene: The spec was updated earlier this week to remove named compound types in favor of being able to export types as a way to "name" them.

I haven't yet updated `wasm-encoder`, but we can probably remove this variant for now.

peterhuene: Oh and Luke is working on getting resources and handles into the spec.

peterhuene:
👍  I'm good with landing this WIP with or without the `Name` compound type variant (I need to update `wasm-encoder` anyway).

PR #475, opened 2022-02-08
Add minimal recognition for the `component` keyword.

sunfishcode:
This is an updated version of #462, based on the component-model branch, using `.wat` instead of `.cat`, with a separate `Component` type rather than an `is_component` field in `Module`, with fewer `r#` identifiers, and with more infrastructure in place, though several of the major pieces are still stubbed out.


crates/wast/src/ast/expr.rs
```diff
@@ -1144,6 +1144,9 @@ instructions! {
         F32x4RelaxedMax : [0xfd, 0xe2]: "f32x4.relaxed_max",
         F64x2RelaxedMin : [0xfd, 0xd4]: "f64x2.relaxed_min",
         F64x2RelaxedMax : [0xfd, 0xee]: "f64x2.relaxed_max",
+
+        CanonLift(CanonLift<'a>) : [/*TODO: encoding*/]: "canon.lift",
+        CanonLower(CanonLower<'a>) : [/*TODO: encoding*/]: "canon.lower",
```
alexcrichton, line 1149: Given how separate these are from all the other instructions I think it'd be better to not put this in the instruction enum and have a dedicated enum for component-defined functions (e.g. `ComponentFunc`)

crates/wast/src/ast/component.rs
```diff
@@ -0,0 +1,188 @@
...
+
+/// A listing of all possible fields that can make up a WebAssembly component.
+#[allow(missing_docs)]
+#[derive(Debug)]
+pub enum ComponentField<'a> {
+    Type(ast::Type<'a>),
+    Import(ast::Import<'a>),
+    Func(ast::Func<'a>),
+    Export(ast::Export<'a>),
+    Start(ast::ItemRef<'a, kw::func>),
```
alexcrichton, line 135: TIL components have a `start` function...

crates/wast/src/ast/component.rs
```diff
@@ -0,0 +1,188 @@
...
+            kind,
+        })
+    }
+}
+
+/// A listing of all possible fields that can make up a WebAssembly component.
+#[allow(missing_docs)]
+#[derive(Debug)]
+pub enum ComponentField<'a> {
+    Type(ast::Type<'a>),
```
alexcrichton, line 131: I think that this'll want to be `ComponentType<'a>` or something like that in the near future, and then the module/instance types would come out of `ast::Type`

I think this actually applies to most of the fields below, given the disjoint-ness of components and modules right now. There probably won't be all that much that's shared between the two parsing-wise.

tests/roundtrip.rs
```diff
@@ -431,7 +431,10 @@ impl TestState {
         if ret.contains("offset=4294967296") && !test.iter().any(|t| t == "memory64") {
             bail!("i32 constant out of bounds");
         }
-        Ok(wat.module.encode()?)
+        match wat {
+            Wat::Module(mut module) => Ok(module.encode()?),
+            Wat::Component(mut component) => Ok(component.encode()?),
+        }
```
alexcrichton, line 437: Could the `encode` method be a method on `Wat` as well?

crates/wast/src/binary.rs
```diff
@@ -1178,6 +1450,51 @@ impl Encode for Names<'_> {
...
+            && self.funcs.is_empty()
+            && self.locals.is_empty()
+            && self.labels.is_empty()
+            && self.types.is_empty()
+        // NB: specifically don't check modules/components/instances since they're
+        // not encoded for now.
+    }
+}
+
+impl Encode for ComponentNames<'_> {
```
alexcrichton, line 1465: AFAIK there's no spec or thinking yet for how the name section will work out in the binary encoding, so it might be good to ensure that something is written up before implementing something here

crates/wast/src/binary.rs
```diff
@@ -1108,7 +1228,159 @@ fn find_names<'a>(
...
+            Name::Component => (&mut ret.components, &mut ret.component_idx),
+            Name::Instance => (&mut ret.instances, &mut ret.instance_idx),
+            Name::Type => (&mut ret.types, &mut ret.type_idx),
+        };
+        if let Some(name) = get_name(id, name) {
+            list.push((*idx, name));
+        }
+
+        // Handle module locals separately from above
+        if let ComponentField::Func(f) = field {
```
alexcrichton, line 1324: I don't think components can have functions with locals and such, so this may not be necessary?

crates/wast/src/binary.rs
```diff
@@ -1108,7 +1228,159 @@ fn find_names<'a>(
...
+            ComponentField::NestedComponent(c) => (Name::Component, &c.id, &c.name),
+            ComponentField::Instance(i) => (Name::Instance, &i.id, &i.name),
+            ComponentField::Type(t) => (Name::Type, &t.id, &t.name),
+            ComponentField::Func(f) => (Name::Func, &f.id, &f.name),
+            ComponentField::Alias(a) => (
+                match a.kind {
+                    ExportKind::Func => Name::Func,
+                    ExportKind::Table => panic!("table inside component"),
+                    ExportKind::Memory => panic!("memory inside component"),
+                    ExportKind::Global => panic!("global inside component"),
```
alexcrichton, line 1297: I think these are valid cases for components, right? In that I think this should be handled rather than panicking?

crates/wast/src/binary.rs
```diff
@@ -1108,7 +1228,159 @@ fn find_names<'a>(
...
+) -> ComponentNames<'a> {
+    fn get_name<'a>(id: &Option<Id<'a>>, name: &Option<NameAnnotation<'a>>) -> Option<&'a str> {
+        name.as_ref().map(|n| n.name).or(id.and_then(|id| {
+            if id.is_gensym() {
+                None
+            } else {
+                Some(id.name())
+            }
+        }))
+    }
```
alexcrichton, line 1259: I suspect that this is the same implementation as something for modules, so could that be deduplicated?

crates/wast/src/binary.rs
```diff
@@ -144,6 +153,110 @@ fn encode_fields(
...
+        })
+        .peekable();
+
+    // A special path is used for now to handle non-module-linking modules to
+    // work around WebAssembly/annotations#11
+    if aliases.len() == 0 && components.len() == 0 && instances.len() == 0 {
+        e.section_list(1, Type, &types);
+        e.section_list(2, Import, &imports);
+    } else {
+        while let Some(field) = items.next() {
```
alexcrichton, line 217: This is basically a weird affordance for module linking today, but I suspect that this will all go away for components. IIRC components have no internal orderings so I think that we'd create a component by simply iterating over the fields and emitting the items as we see them.

crates/wast/src/binary.rs
```diff
@@ -144,6 +153,110 @@ fn encode_fields(
...
+            ComponentField::Func(i) => funcs.push(i),
+            ComponentField::Export(i) => exports.push(i),
+            ComponentField::Start(i) => start.push(i),
+            ComponentField::Custom(i) => customs.push(i),
+            ComponentField::Instance(i) => instances.push(i),
+            ComponentField::NestedModule(i) => modules.push(i),
+            ComponentField::NestedComponent(i) => components.push(i),
+            ComponentField::Alias(a) => aliases.push(a),
+        }
+    }
```
alexcrichton, line 187: Similar to below, while this is done for core wasm modules I don't think this will be applicable for components.

alexcrichton:
In https://github.com/bytecodealliance/wasm-tools/pull/519 the `component-model` branch was merge to main and I just deleted the `component-model` branch which auto-closed this PR (as a heads up)

PR #485, opened 2022-02-17
Bump `wasm-mutate` to 0.2.0

fitzgen:


crates/wasm-mutate/src/lib.rs
```diff
@@ -316,8 +321,14 @@ impl<'wasm> WasmMutate<'wasm> {
         // subslice of data with a random slice of other data.
         //
         // First up start/end indices are picked.
-        let a = self.rng().gen_range(0..=data.len());
-        let b = self.rng().gen_range(0..=data.len());
+        let (a, b) = if data.is_empty() {
```
alexcrichton, line 324: For data here specifically it seems surprising that this would panic because `0..=0` is a range with one element to choose from so it seems that `gen_range` shouldn't panic.

crates/wasm-mutate/src/mutators/remove_export.rs
```diff
@@ -18,7 +18,12 @@ impl Mutator for RemoveExportMutator {
     ) -> Result<Box<dyn Iterator<Item = Result<Module>> + 'a>> {
         let mut exports = ExportSection::new();
         let mut reader = ExportSectionReader::new(config.info().get_exports_section().data, 0)?;
+
         let max_exports = reader.get_count() as u64;
+        if max_exports == 0 {
+            return Err(Error::no_mutations_applicable());
+        }
```
alexcrichton, line 25: I thought that this was guarded by `can_mutate` below? Otherwise how could `mutate` get called?

crates/wasm-mutate/src/mutators/function_body_unreachable.rs
```diff
@@ -23,6 +23,10 @@ impl Mutator for FunctionBodyUnreachable {
         let mut reader = CodeSectionReader::new(code_section.data, 0)?;
 
         let count = reader.get_count();
+        if count == 0 {
+            return Err(Error::no_mutations_applicable());
+        }
```
alexcrichton, line 28: Instead of adding a condition here could the `can_mutate` function be updated below to test that there are actual functions in the module? I think right now `has_code` only tests for the presence of the code section, and the code section could be empty?

alexcrichton (commented):
I'm a little confused as to how fuzzing hasn't caught any of these on oss-fuzz, but I was also wondering if you're a bit overeager in updating `gen_range`? I commented on some places below but doesn't `can_mutate` guard most of these `gen_range` instances to ensure non-empty ranges are randomly chosen from?

Otherwise for the version bump managing all the crates in this repo is pretty tedious so I'm hopeful that we can do something consistent. Unfortunately though the current consistency is to run `./publish` which requires publishing everything in the repository that's had changes, which I suspect you don't want to do here. The thing I'm worried about is that this should also force a minor bump of `wasm-shrink` and `wasm-tools`, but it currently doesn't do that. While you could achieve that by running the `./publish` script it would bump up to `wasm-tools` but to actually publish that you'd also need to bump unrelated crates like `wasmparser`. Overall I don't know the best answer to this, but it's been awhile since we published things so would you be up for running the publish script to bump everything? (it should ideally also be sufficient for actually publishing all the crates to crates.io)

fitzgen:
> I'm a little confused as to how fuzzing hasn't caught any of these on oss-fuzz, but I was also wondering if you're a bit overeager in updating gen_range? I commented on some places below but doesn't can_mutate guard most of these gen_range instances to ensure non-empty ranges are randomly chosen from?

I think `wasm-smith` will never generate an empty function section, which is how I found the panic inside the function-snip mutator.

I figured I would do a survey of all of our `gen_range` calls and add checks to all the ones that weren't immediately obviously never going to panic. It is certainly an over-estimate.

I can either go back and be more precise / double check the call sites, or we can just accept that there are benign guards for cases that can't happen. Thoughts?

crates/wasm-mutate/src/lib.rs
fitzgen: Yeah this is my b, I didn't think about how this is an inclusive range.

crates/wasm-mutate/src/mutators/remove_export.rs
fitzgen: I didn't check super carefully, mostly just trying to cover all `gen_range` calls.

crates/wasm-mutate/src/mutators/function_body_unreachable.rs
fitzgen: Yeah I can do that instead.

fitzgen:
Okay I moved all these checks into `can_mutate` where they didn't already exist.

fitzgen:
Hmm seems like something similar landed on `main` recently and this last commit is basically a no-op.

fitzgen:
Ah 5380b16c523ac3b56235d3254522504c2efedd3b also fixed that panic.

fitzgen:
Dropping that last commit as it's no longer necessary.

alexcrichton (approved):
Ah ok this makes sense, code changes look good. 

That being said I think you'll actually need to release the other crates as well because `wasm-mutate` no longer builds with `wasmparser` from crates.io. Would you be up for running `./publish bump` with the appropriate arguments to bump all the crates?

fitzgen:
Ah I didn't realize we added a publish script to this repo. Will do!

PR #490, opened 2022-02-26
Allow exporting of component function types.

peterhuene:
This PR allows the exporting of component function types in the validator.

See [this discussion](https://github.com/WebAssembly/component-model/pull/6#discussion_r815185164) as to why it's needed.

Marking this as a draft until the request spec change is incorporated.


alexcrichton (approved):

peterhuene:
Per the spec update, the restriction on exporting only interface types was lifted to allow any type.

PR #491, opened 2022-02-28
`wasm-smith`: Generate arbitrary import sections for components

fitzgen:


fitzgen:
@peterhuene a second pair of eyes on the `<deftype>` stuff would be very appreciated, want to make sure that I got that right.

crates/wasm-smith/src/component.rs
```diff
@@ -54,13 +54,23 @@ pub struct Component {
...
 }
 
 #[derive(Debug, Default)]
 struct TypesScope {
     // All types in this index space, regardless of kind.
     types: Vec<Rc<Type>>,
 
+    // The indices of all the entries in `types` that describe things that can
+    // be imported or exported at instantiation time.
+    def_types: Vec<u32>,
```
peterhuene, line 72: One thing I do want to call out as we eventually implement the export and instance sections, any type itself (including `intertype`) can be exported or passed as an argument to component instantiation, but types can't be imported by design currently (that will probably change once resources are a thing in the spec). With respect to component instantiation, type arguments are simply ignored right now as instantiation arguments are allowed to be a superset of the component's imports.

So `def_types` is really tracking the item types that are allowable for import. I don't know if another name would better reflect that or not, but I'm also good with `def_types` as it matches the spec.

peterhuene (approved):
Looks good, just the one comment that doesn't have to be addressed 👍

crates/wasm-smith/src/component.rs
fitzgen: Gotcha, thanks for the clarification.

I think sticking with names that match the spec is the way to go, makes it easier for folks to dive in and see the correspondence.

PR #496, opened 2022-03-04
Implement printing of components in `wasmprinter`.

peterhuene:
This PR implements printing of components in `wasmprinter`.

The output text format is based on the current spec's textual representation.

This does not alter the name section parsing implementation to update it for
the component model; it is assumed that will be done when the `wat`
implementation is completed.


alexcrichton (approved):
Seems reasonable to me! Most of `wasmprinter`, as I'm sure you've seen, is exactly highly tuned or anything like that, it's mostly whatever works for coming out of wasmparser.

PR #498, opened 2022-03-08
Allow specifying fuel for `wasm-tools mutate`

alexcrichton:
Adds a CLI option to the `wasm-tools` command to manually specify fuel
in case that's interesting for a particular reduction or something like
that.

### DIFF
--- a/crates/wasm-mutate/src/lib.rs
+++ b/crates/wasm-mutate/src/lib.rs
@@ -160,8 +160,9 @@ pub struct WasmMutate<'wasm> {
     preserve_semantics: bool,
 
     /// Fuel to control the time of the mutation.
-    #[cfg_attr(feature = "clap", clap(skip = Cell::new(u64::MAX)))]
-    fuel: Cell<u64>,
+    #[cfg_attr(feature = "clap", clap(long, parse(try_from_str = parse_fuel)))]
+    fuel: Option<Cell<u64>>,
+
     /// Only perform size-reducing transformations on the Wasm module. This
     /// allows `wasm-mutate` to be used as a test case reducer.
     #[cfg_attr(feature = "clap", clap(long))]
@@ -179,6 +180,11 @@ pub struct WasmMutate<'wasm> {
     info: Option<ModuleInfo<'wasm>>,
 }
 
+#[cfg(feature = "clap")]
+fn parse_fuel(s: &str) -> Result<Cell<u64>, String> {
+    s.parse::<u64>().map(Cell::new).map_err(|s| s.to_string())
+}
+
 impl Default for WasmMutate<'_> {
     fn default() -> Self {
         let seed = 3;
@@ -187,7 +193,7 @@ impl Default for WasmMutate<'_> {
             preserve_semantics: false,
             reduce: false,
             raw_mutate_func: None,
-            fuel: Cell::new(u64::MAX),
+            fuel: None,
             rng: None,
             info: None,
         }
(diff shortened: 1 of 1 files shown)


crates/wasm-mutate/src/lib.rs
```diff
@@ -160,8 +160,9 @@ pub struct WasmMutate<'wasm> {
     preserve_semantics: bool,
 
     /// Fuel to control the time of the mutation.
-    #[cfg_attr(feature = "clap", clap(skip = Cell::new(u64::MAX)))]
-    fuel: Cell<u64>,
+    #[cfg_attr(feature = "clap", clap(long, parse(try_from_str = parse_fuel)))]
+    fuel: Option<Cell<u64>>,
+
```
fitzgen, line 165: Seems like keeping `fuel: Cell<u64>` but doing `clap(default_value = u64::MAX)` instead of skipping would have the same effect but decrease the diff here and probably also be more performant inside `consume_fuel`.

alexcrichton: Ah yeah I originally tried that but it failed because `Display for Cell` is missing and it tries to print out the default value of `Cell::new(u64::MAX)` otherwise, but I figured out I could relax that requirement by using `None` here instead for the no-fuel case

fitzgen: This works for me locally:

```diff
diff --git a/crates/wasm-mutate/src/lib.rs b/crates/wasm-mutate/src/lib.rs
index e7e44af2..344b7b06 100644
--- a/crates/wasm-mutate/src/lib.rs
+++ b/crates/wasm-mutate/src/lib.rs
@@ -155,18 +155,27 @@ pub struct WasmMutate<'wasm> {
     #[cfg_attr(feature = "clap", clap(short, long, default_value = "42"))]
     seed: u64,
 
     /// Only perform semantics-preserving transformations on the Wasm module.
     #[cfg_attr(feature = "clap", clap(long))]
     preserve_semantics: bool,
 
     /// Fuel to control the time of the mutation.
-    #[cfg_attr(feature = "clap", clap(skip = Cell::new(u64::MAX)))]
+    #[cfg_attr(
+        feature = "clap",
+        clap(
+            short,
+            long,
+            default_value = "18446744073709551615", // u64::MAX
+            parse(try_from_str = parse_fuel),
+        )
+    )]
     fuel: Cell<u64>,
+
     /// Only perform size-reducing transformations on the Wasm module. This
     /// allows `wasm-mutate` to be used as a test case reducer.
     #[cfg_attr(feature = "clap", clap(long))]
     reduce: bool,
 
     // Note: this is only exposed via the programmatic interface, not via the
     // CLI.
     #[cfg_attr(feature = "clap", clap(skip = None))]
@@ -174,16 +183,21 @@ pub struct WasmMutate<'wasm> {
 
     #[cfg_attr(feature = "clap", clap(skip = None))]
     rng: Option<SmallRng>,
 
     #[cfg_attr(feature = "clap", clap(skip = None))]
     info: Option<ModuleInfo<'wasm>>,
 }
 
+#[cfg(feature = "clap")]
+fn parse_fuel(s: &str) -> Result<Cell<u64>, String> {
+    s.parse::<u64>().map(Cell::new).map_err(|s| s.to_string())
+}
+
 impl Default for WasmMutate<'_> {
     fn default() -> Self {
         let seed = 3;
         WasmMutate {
             seed,
             preserve_semantics: false,
             reduce: false,
             raw_mutate_func: None,
```

Issue #499, opened 2022-03-08
mutate: Invalid wasm module produced from wasm-mutate


alexcrichton:
With this input file:

<details>

```
(module
  (type (;0;) (func (param f64) (result i64)))
  (type (;1;) (func (result f32 f32 f32 f32 f32 f32 f32 funcref)))
  (type (;2;) (func (param f32 f32 f32) (result f32)))
  (func (;0;) (type 0) (param f64) (result i64)
    (local f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 funcref f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f
32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32 f32)
    ref.null func
    call 2
    call 2
    i64.const 2314885530818453536
    table.size 2
    i64.const 2314885530818453536
    drop
    loop (result f32)  ;; label = @1
      i64.const 2314885530818453536
      f64.const 0x1.020202020202p-509 (;=0.0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000006013470016999068;)
      br 0 (;@1;)
      br 0 (;@1;)
      i32.trunc_sat_f64_u
      memory.size
      i32.add
      i32.load offset=35659796 align=1
      i64.extend_i32_u
      i32.const 538976288
      i64.extend_i32_s
      i32.const 538976288
      i64.extend_i32_s
      i32.const 538976288
      i64.extend_i32_s
      i32.const 538976288
      i64.extend_i32_s
      i32.const 538976288
      i64.extend_i32_s
      i32.const 538976288
      i64.extend_i32_s
      i32.const 538976288
      i64.extend_i32_s
      i64.shl
      i64.shl
      i64.shl
      i64.shl
      i64.shl
      drop
      i64.shl
      i64.shl
      i32.const 538976288
      local.get 42
      table.set 2
      i32.const 538976288
[...]

PR #500, opened 2022-03-08
wasm-mutate: `table.set` doesn't produce a value

fitzgen:
So use `empty_node` rather than `push_node`.

Fixes https://github.com/bytecodealliance/wasm-tools/issues/499

### DIFF
--- a/crates/wasm-mutate/src/lib.rs
+++ b/crates/wasm-mutate/src/lib.rs
@@ -160,8 +160,17 @@ pub struct WasmMutate<'wasm> {
     preserve_semantics: bool,
 
     /// Fuel to control the time of the mutation.
-    #[cfg_attr(feature = "clap", clap(skip = Cell::new(u64::MAX)))]
+    #[cfg_attr(
+        feature = "clap",
+        clap(
+            short,
+            long,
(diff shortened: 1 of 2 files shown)


alexcrichton (approved):
For future me, I realize now I forgot to `=trace` I was only looking at debug logs, but other than that is there anything else you did to reduce this? (basically how did you get to the point that you could spot the `table.set`?)

fitzgen:
This is the command I was repeatedly running while debugging:

```
RUST_LOG=trace,egg=debug cargo run mutate --seed 8319607238179319672 ~/scratch/test.wat -o ~/scratch/test.mutated.wasm && cargo run -- validate ~/scratch/test.mutated.wasm
```

In particular, I was looking at the input and output expressions in the `wasm-mutate` logs, as well as the set of rules that `egg` was applying to the e-graph.

The breakthrough was when I looked at the input expression a little harder and realized that it wasn't valid because `table.set` doesn't produce a value. Then I went hunting through all references to our `TableSet` instruction in the code base and found that bug (once I pulled the AST parsing stuff back into my working memory).

fitzgen:
I also quickly found that `--fuel 2` was enough to repro the bug

PR #501, opened 2022-03-10
wasmprinter: fix nested module/component printing.

peterhuene:
This commit fixes `wasmprinter` to properly skip over nested modules and
components when locating name sections in the current component.

Additionally, closing parens for non-empty functions and modules/components now appear on
a new line which makes it easier to spot where they close.

### DIFF
--- a/crates/wasmparser/src/validator/component.rs
+++ b/crates/wasmparser/src/validator/component.rs
@@ -732,7 +732,7 @@ impl ComponentState {
     }
 
     pub(super) fn add_alias(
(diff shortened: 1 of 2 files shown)


peterhuene:
@alexcrichton sorry for the re-review. The setting of the "should we end on a newline flag" in `print_code` bugged me so I pushed up a new commit that simply tracks what lines groups start on and ends the group on a new line if the group itself spanned multiple lines.

alexcrichton (approved):
Seems fine by me!

PR #503, opened 2022-03-16
wasmparser: return type information built up from validation. 

peterhuene:
This PR changes `Validator::end` to return type information for the module
or component that was just validated.

The type information will help other tools so that they don't have to duplicate
what the validator is doing to determine the types of things in a module or
component's index spaces.


crates/wasmparser/src/validator/types.rs
```diff
@@ -0,0 +1,1157 @@
...
+    pub default_to: Option<String>,
+}
+
+/// Represents an interface type.
+#[derive(Debug, Clone)]
+pub enum InterfaceType {
+    /// The type is a primitive interface type.
+    Primitive(PrimitiveInterfaceType),
+    /// The type is a record.
+    Record(HashMap<String, InterfaceTypeRef>),
```
peterhuene, line 483: These types need to be changed as they can no longer use unstable associative collections so that tools that generate interface definitions from the validator's type information can use the correct ordering. We still need fast lookup for subtype checking in the validator, though.

I'm working on that now.

peterhuene: Not entirely happy with the new (and only) dependency on `indexmap` from `wasmparser` to solve this, but it is a very minimal dependency at least. Resolving this for now and we can revisit whether or not it's a good thing in the PR review.

crates/wasmparser/src/validator/types.rs
```diff
@@ -480,17 +482,17 @@ pub enum InterfaceType {
     /// The type is a primitive interface type.
     Primitive(PrimitiveInterfaceType),
     /// The type is a record.
-    Record(HashMap<String, InterfaceTypeRef>),
+    Record(IndexMap<String, InterfaceTypeRef>),
     /// The type is a variant.
-    Variant(HashMap<String, VariantCase>),
+    Variant(IndexMap<String, VariantCase>),
```
fitzgen, line 487: Is there a reason not to use `BTreeMap` here? Do we actually need to sort by insertion order here, or does ordering by key work?

fitzgen (approved):
Overall it looks good to me, but maybe @alexcrichton wants to take a look too, since he is a bit more familiar with the inner workings of `wasmparser`'s validation impl.

crates/wasmparser/src/validator/types.rs
peterhuene: I'm updating `wit-component` to directly use the type information from validation (this PR) and it's creating `wit_parser::Interface`s directly from that type information; so it needs a way to ensure that the declaration order in the original encoding is also respected in any wit variants it generates.

alexcrichton (approved):
This all seems fine to me. We can always tweak as necessary if issues arise.

PR #505, opened 2022-03-21
wasm-mutate: Add support for simd

alexcrichton:
Reducing a fuzz bug for Wasmtime recently I found that all the test
cases were using simd so I figured I'd go ahead and add simd support to
wasm-mutate to try to help reduce test cases. While this all appears to
work it unfortunately didn't help really reduce many test cases further,
and I'm not entirely sure why but figured this is still useful in its
own right.


fitzgen (approved):
Nice!

PR #506, opened 2022-03-21
wasmparser: remove enforcement that `into` instances must export a memory.

peterhuene:
Currently, the component model validator requires that an instance pointed at
by the `into` canonical option export a memory named `memory`.

However, we have a chicken-and-the-egg problem when lowering imported
functions: the functions' `into` options must point at an instance, but we
don't have the instance yet because we're lowering the functions that the inner
core module will be importing.

To work around this problem, a small "shim" instance is used that has a
`funcref` table pointing at the canonical ABI functions expected by the `into`
option. The table gets populated after the inner core module gets instantiated.

Thus the `into` options for lowering imported functions point at this shim
instance rather than the actual core module's instance for the component.

As the shim instance can't export the same memory as the actual core module,
this validation enforcement prevents this from working.

### DIFF
--- a/crates/wasmparser/src/validator/component.rs
+++ b/crates/wasmparser/src/validator/component.rs
@@ -485,16 +485,6 @@ impl ComponentState {
             Some(idx) => {
                 let into_ty = self.module_instance_at(idx, types, offset)?;
 
-                match into_ty.exports.get("memory") {
-                    Some(EntityType::Memory(_)) => {}
-                    _ => {
-                        return Err(BinaryReaderError::new(
-                            "instance specified by `into` option does not export a memory named `memory`",
-                            offset,
-                        ));
-                    }
-                }
-
                 check_into_func(
                     into_ty,
                     "canonical_abi_realloc",


peterhuene:
To better illustrate the issue, here's what the wat looks like for a component that imports an interface named `markdown` with a single function named `render` of signature `[string] -> [string]`:

```wat
(component
  ...
  (module (;0;) ... ) ;; actual core module!
  (type (;0;) (func (param "markdown" string) (result string)))
  (type (;1;) 
    (instance
      (alias outer 1 (type (;0;) 0))
      (export "render" (type 0))
    )
  )
  (import "markdown" (instance (;0;) (type 1)))
  (module (;1;)
    (type (;0;) (func (param i32 i32 i32 i32) (result i32)))
    (type (;1;) (func (param i32 i32 i32)))
    (func (;0;) (type 0) (param i32 i32 i32 i32) (result i32)
      local.get 0
      local.get 1
      local.get 2
      local.get 3
      i32.const 0
      call_indirect (type 0)
    )
    (func (;1;) (type 1) (param i32 i32 i32)
      local.get 0
      local.get 1
      local.get 2
      i32.const 1
      call_indirect (type 1)
    )
    (table (;0;) 2 2 funcref)
    (export "table" (table 0))
    (export "canonical_abi_realloc" (func 0))
    (export "canonical_abi_free" (func 1))
  )
  (instance (;1;) (instantiate (module 1)))
  (alias export (instance 0) "render" (func (;0;)))
  (func (;1;) (canon.lower utf8 (into (instance 1)) (func 0)))
  ...
)
```

Here `(into (instance 1))` is to the shim instance that will indirectly call the canonical ABI functions from the actual core module (module 0), which later gets instantiated and the canonical ABI functions put into the shim's table.

It's not possible for this module to export the same memory as the real core module, so we shouldn't be checking for a `memory` export on the `into` canonical option.

peterhuene:
Once I have this all working in the component tool, I think I'm going to file an issue against the component model proposal to make lowering _instances_ a thing as there really is too much boiler plate to lower instance imports to something usable in a module instantiation. It should be easier to describe a lowering of a component instance (that _only_ exports functions) into a module instance that exports the lowered function types (which can then be directly passed as an argument to module instantiation).

alexcrichton (approved):
FWIW @lukewagner and @fitzgen and I have been discussing this a bit as well, personally I think that we need some deeper changes at the spec layer to avoid the need for shim instances because right now basically everyone needs it and it doesn't seem like that's going to change any time soon. That's a spec thing though and I should probably file an issue or similar upstream at some point.

alexcrichton:
Coming back to this, this doesn't actually feel quite right. Without a `memory` how does the canonical abi know what to lower into? I didn't really review this too closely at the time but thinking about this in terms of a Wasmtime implementation my first goal is to implement exported functions but if the underlying instance doesn't have a `memory` then... something needs to happen but it's not clear to me what.

peterhuene:
I personally think this needs to be redesigned in the component model, which I've talked to Luke about a little bit.

I think we should definitely add this back in once the underlying spec issue is solved; for now, this validation check prevents me from validating `wit-component` output that uses imported interfaces.

Issue #511, opened 2022-03-23
wasm-shrink not effectively shrinking straight-line code


alexcrichton:
With the simd support in https://github.com/bytecodealliance/wasm-tools/pull/505 I was trying to shrink the reproduction test cases from https://github.com/bytecodealliance/wasmtime/issues/3951 as sourced from oss-fuzz to get something along the lines of the reproduction that I ended up getting to in the issue itself. One example original file was [e.wasm.gz](https://github.com/bytecodealliance/wasm-tools/files/8336023/e.wasm.gz) which when shrunk with:

```
$ cat test.sh
#!/bin/sh
$HOME/code/wasmtime/target/debug/wasmtime \
  compile --wasm-features memory64,multi-memory \
  --target x86_64 $1 -o /dev/null 2>&1 | \
  grep "register allocation: "
$ cargo run --release shrink ./test.sh e.wasm
    Finished release [optimized] target(s) in 0.05s
     Running `/home/acrichto/code/wasm-tools/target/release/wasm-tools shrink ./test.sh e.wasm`
4886 bytes (0.33% smaller)
4875 bytes (0.55% smaller)
4871 bytes (0.63% smaller)
4868 bytes (0.69% smaller)
4830 bytes (1.47% smaller)
4829 bytes (1.49% smaller)
4828 bytes (1.51% smaller)
4796 bytes (2.16% smaller)
4793 bytes (2.22% smaller)

e.shrunken.wasm :: 4793 bytes (2.22% smaller)
================================================================================
(module 
...
```

Here there reduction really didn't do anything at all unfortunately. The function itself is one giant basic block with a ton of simd-looking things in the middle. There aren't a lot of simd reduction rules but even the one we have that replaces v128 expressions with a const-0 should do something here. 

The reduction has a ton of `f64x2.nearest` instructions which is a unop that can simply be removed at any time in wasm, so I added a reduction rule:

```diff
diff --git a/crates/wasm-mutate/src/mutators/peephole/rules.rs b/crates/wasm-mutate/src/mutators/peephole/rules.rs
index b8f240a5..125fdb5f 100644
--- a/crates/wasm-mutate/src/mutators/peephole/rules.rs
+++ b/crates/wasm-mutate/src/mutators/peephole/rules.rs
@@ -579,6 +579,11 @@ impl PeepholeMutator {
             }
         }

+        rules.push(rewrite!(
+            "f64x2.nearest-remove";
[...]

PR #513, opened 2022-03-23
Add initial memory64 and multi-memory support to wasm-mutate

alexcrichton:
This fixes at least one issue I was encountering on #511 where I forgot
that the modules were using memory64 which meant that the
`allow_memarg64` method needed to be enabled to enable parsing 64-bit
offsets in memargs. This also touches up a few other locations to
support memory64 and multi-memory, enabling fuzzing as well. This gets
the exact test case on #511 to reduce much further in the shape of the
outer module (e.g. elements, imports, etc all removed), although the
function itself causing the issue is still not being reduced.


fitzgen (approved):
Nice!

PR #514, opened 2022-03-23
wasm-mutate: construct DFG for `v128.*` arithmetic operations

fitzgen:
Part of #511

### DIFF
--- a/crates/wasm-mutate/src/mutators/peephole/dfg.rs
+++ b/crates/wasm-mutate/src/mutators/peephole/dfg.rs
@@ -598,6 +598,14 @@ impl<'a> DFGBuilder {
                 Operator::I64Rotl => self.binop(idx, Lang::I64RotL),
                 Operator::I64Rotr => self.binop(idx, Lang::I64RotR),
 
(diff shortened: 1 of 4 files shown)


crates/wasm-mutate/src/mutators/peephole/dfg.rs
```diff
@@ -1126,6 +1134,19 @@ impl<'a> DFGBuilder {
...
+        let a = self.pop_operand(idx, false);
+        let b = self.pop_operand(idx, false);
+        let c = self.pop_operand(idx, false);
+
+        // The operands should not be the same.
+        assert_ne!(a, b);
+        assert_ne!(a, c);
+        assert_ne!(b, c);
+
+        self.push_node(op([Id::from(a), Id::from(b), Id::from(c)]), idx);
```
alexcrichton, line 1147: I think this order may need to be reversed (judging by the other methods that do this)

alexcrichton (approved):
Gah of course! I think then it was all related to decoding issues probably, I'll try a re-reduction on all the test cases and see what happens with this merged in as well as the memory64 bits.

PR #518, opened 2022-03-24
mutate: Fix logic removing element segments

alexcrichton:
If there were `ref.func` instructions present in a module this would
accidentally prevent all element segments from removal typically because
the functions referenced from element segments were accidentally not
included in the `referenced_functions` set. This updates the insertion
into that set to handle all function references, not just those with
`ref.func`, which enables more removal of element segments.


fitzgen (approved):
Good catch

PR #521, opened 2022-03-24
Another round of small wasm-mutate changes

alexcrichton:
I was debugging #511 a bit more with @fitzgen and these are some tweaks which helped reduce the files I was looking at even further. These changes are all oriented at opening up more opportunities for wasm-shrink to continue to mutate and delete code.


fitzgen (approved):
:+1:

Issue #517, opened 2022-03-24
wasm-encoder 0.10.0: elements section generated not as per the spec?


Ekleog:
Hello,

I'm coming after noticing that wasm-encoder generates an elements section that seems to me like it diverges from the spec. `wabt`'s `wasm2wat` does manage to parse it, but making it back with `wat2wasm` gives another result, and `pwasm-utils` fails to parse the resulting module.

The diff of the elements section is the following:
```
(0909) 01 02 00 41000b 00 0100 # result of wasm-encoder
(0907) 01    00 41000b    0100 # after a wasm2wat2wasm round-trip
# to the best of my understanding of https://www.w3.org/TR/wasm-core-1/#binary-elemsec
^^^^^^-||-||-||-||||||-||-||||- element section, length
       ^^-||-||-||||||-||-||||- number of elem items
          ^^-||-||||||-||-||||- ??? something added by wasm-encoder ???
             ^^-||||||-||-||||- table index
                ^^^^^^-||-||||- offset (i32.const 0; end)
                       ^^-||||- ??? something added by wasm-encoder ???
                          ^^^^- vec of actual items, here one funcref index 0
```

The first 02 appears to come from https://github.com/bytecodealliance/wasm-tools/blob/main/crates/wasm-encoder/src/elements.rs#L137, and the second 00 from https://github.com/bytecodealliance/wasm-tools/blob/main/crates/wasm-encoder/src/elements.rs#L142 ; but I cannot find any reference to these magic numbers in https://www.w3.org/TR/wasm-core-1/#binary-elemsec ; and the git blame is not very helpful as it appears to come from the day wasm-encoder was added to this repository.

Is this a bug in wasm-encoder, or is it on purpose and eg. a new version of the spec that hasn't been published yet? And in the second case how could I get wasm-encoder to output code that conforms to the currently published version of the wasm spec?

alexcrichton:
I believe these are different ways to encode the same thing. It's a known issue with wabt where it canonicalizes the first encoding here to the second, whereas the tooling in this repository tries to avoid changing the encoding when it can.

If external crates are failing to parse the output then they may not have full support for the reference types proposal which added more encodings of element segments. It looks like you're reading a published version of the specification but that doesn't track the latest version, which is available at https://webassembly.github.io/spec/. I believe that the specs there will show that the extra unaccounted-for-bytes here are a way of encoding an "active" element segment for a table that has a nonzero index (although in this case it does indeed have a zero index).

[...]

PR #524, opened 2022-03-25
wasm-encoder: make documentation more explicit about about elements' table encoding

Ekleog:
Fixes https://github.com/bytecodealliance/wasm-tools/issues/517


alexcrichton:
Thanks!

PR #525, opened 2022-03-28
Support smithing 1.0 WebAssembly modules

nagisa:
Add an ability to disable saturating_float_to_int and sign_extension_ops
proposals.


alexcrichton:
Looks good! I think rustfmt may be needed but otherwise I'm happy to merge.

nagisa:
The test suite has passed.

PR #529, opened 2022-03-28
Initial wast support for components

sunfishcode:
This implements parsing for the [component text format](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md), and partial support for name resolution and encoding. There are several TODOs that I'll submit follow-up incremental PRs for once this base PR has landed.


crates/wast/tests/parse-component.rs
```diff
@@ -0,0 +1,67 @@
...
+            find_tests(&f.path(), tests);
+            continue;
+        }
+        match f.path().extension().and_then(|s| s.to_str()) {
+            Some("wat") => {}
+            _ => continue,
+        }
+        tests.push(f.path());
+    }
+}
```
alexcrichton, line 67: Could all of these tests be added to `tests/roundtrip.rs` and/or something like `tests/local/component-model`? They ideally would all exercise wasmparser's validation and wasmprinter's printing for round-trip-ness as well.

crates/wast/src/ast/alias.rs
```diff
@@ -11,50 +11,98 @@ pub struct Alias<'a> {
...
-                index: parser.parse()?,
-            }
-        } else {
-            AliasSource::InstanceExport {
-                instance: parser.parse::<ast::IndexOrRef<_>>()?.0,
-                export: parser.parse()?,
-            }
-        };
-        let (kind, id, name) = parser.parens(|p| Ok((p.parse()?, p.parse()?, p.parse()?)))?;
+        parser.parens(|parser| {
```
alexcrichton, line 83: Conventionally most of the parsers in `wast` delegate surrounding parenthesis parsing to the outer parsing context rather than the inner parsing context, so whomever parses aliases would parse these parens in theory.

alexcrichton: (this is true for a number of the parsers I see below but I'll just comment here)

I should also point out that this isn't 100% true throughout `wast` but IIRC it's the style I tried to adhere to where possible

crates/wast/src/ast/token.rs
```diff
@@ -726,6 +696,22 @@ impl Peek for LParen {
     }
 }
 
+/// A convenience type to use with [`Parser::peek`](crate::parser::Parser::peek)
+/// to see if the next token ends an s-expression.
+pub struct RParen {
```
alexcrichton, line 701: I think in theory this shouldn't be necessary since the parser should otherwise report eof (or so I thought), but did the need for this arise?

alexcrichton, line 701: I think in theory this shouldn't be necessary since the parser should otherwise report eof (or so I thought), but did the need for this arise?

crates/wast/src/parser.rs
```diff
@@ -568,6 +578,38 @@ impl<'a> Parser<'a> {
...
+        if cursor.advance_token().is_some() && cursor.advance_token().is_some() {
+            T::peek(cursor)
+        } else {
+            false
+        }
+    }
+
+    /// Same as the [`Parser::peek3`] method, except checks the next next next
+    /// token, not the next next token.
+    pub fn peek4<T: Peek>(self) -> bool {
```
alexcrichton, line 594: Were these added opportunistically or was this needed for parsing somewhere internally? It wasn't my intention to really go infinitely long here with the peeking, and ideally we wouldn't add the desire to go too much further than the current spec needs.

crates/wast/src/resolve/component_names.rs
```diff
@@ -0,0 +1,379 @@
...
+    value => Value
+    table => Table
+    memory => Memory
+    global => Global
+    tag => Tag
+    r#type => Type
+}
+
+#[derive(Default)]
+struct Namespace<'a> {
```
alexcrichton, line 331: This looks pretty similar to the `Namespace` helper for modules, so could that be reused?

crates/wast/src/resolve/deinline_import_export.rs
```diff
@@ -1,16 +1,16 @@
 use crate::ast::*;
-use crate::resolve::gensym;
+use crate::resolve::gensym::Gensym;
 use std::mem;
 
-pub fn run(fields: &mut Vec<ModuleField>) {
+pub(crate) fn run(fields: &mut Vec<ModuleField>, gensym: &mut Gensym) {
```
alexcrichton, line 5: How come the gensym bits went from something global to a struct-to-pass-around?

(IIRC I originally passed it around but it was such a pain to do so I stopped doing that)

alexcrichton (commented):
Seems all pretty reasonable to me, but I think that we're going to want to disable something at compile-time or otherwise the fuzzers are likely going to get real angry at this. I think we may want a Cargo feature to control the internals of `Wat` parsing so the fuzzers don't try to fuzz all this (also so it's not exposed by default if we publish since it's in an incomplete state)

Otherwise the main stylistic comment is that most parsers are starting off by parsing their parens when I typically try where possible to have it the other way around and parsers don't parse their own parens and parents parse the outer parens of the item they're sub-parsing.

crates/wast/src/parser.rs
sunfishcode: It was used in earlier version of this PR, as the text format has some tricky parsing ambiguities, but it's not needed anymore. I've now removed it.

crates/wast/src/resolve/deinline_import_export.rs
sunfishcode: That was just me trying to understand name resolution for nested modules and wanting to understand what state was kept where. I can reinstate the thread-local if it's clearer that way, but the current code doesn't need to pass it around that many places.

crates/wast/src/ast/alias.rs
sunfishcode: I personally would like to replace manual paren-consuming and manual token-lookahead with an s-expression parser that produces a tree with strings at the leaves, and then parse that :-).

crates/wast/src/ast/token.rs
sunfishcode: It comes in handy when resolving some of the parsing ambiguities in the component-model text format.

crates/wast/src/ast/alias.rs
alexcrichton: In the meantime would you be ok sticking to to general idioms of the rest of the library as-is?

crates/wast/src/ast/token.rs
alexcrichton: Are those locations where sub-parsers or eof don't work? (or places where the text format could be changed to avoid the need for such lookahead?)

crates/wast/src/ast/alias.rs
sunfishcode: I've now moved most of the parse implementations to this style. The type rules still consume the paren though, because intertype doesn't always start with a paren, and it seems simpler to have it optionally parse a paren than to split the parsing into two places.

crates/wast/src/ast/token.rs
sunfishcode: I've now reorganized the parsing to avoid the need for this, so `RParen` is now removed.

crates/wast/tests/parse-component.rs
sunfishcode: Yes, that will be good to do once everything roundtrips, but wasmparser doesn't appear to support everything that this new wast encoder is encoding yet. I'll move things to roundtrip tests once we work these out.

crates/wast/src/resolve/mod.rs
```diff
@@ -73,14 +97,26 @@ pub fn resolve<'a>(module: &mut Module<'a>) -> Result<Names<'a>, Error> {
...
+    let resolver = names::resolve(fields)?;
     Ok(Names { resolver })
 }
 
+pub fn resolve_component<'a>(component: &mut Component<'a>) -> Result<(), Error> {
+    let fields = match &mut component.kind {
+        ComponentKind::Text(fields) => fields,
+        _ => return Ok(Default::default()),
+    };
+
```
alexcrichton, line 113: I think one thing this is missing is the de-inlining/elaboration, right? 

Or, more generally, would you be up for making a checklist of remaining TODO items after this?

crates/wast/src/resolve/component_names.rs
```diff
@@ -0,0 +1,817 @@
...
+/// `Index::Id` to `Index::Num` and expanding alias syntax sugar.
+pub fn resolve<'a>(id: Option<Id<'a>>, fields: &mut Vec<ComponentField<'a>>) -> Result<(), Error> {
+    let mut resolve_stack = Vec::new();
+    resolve_fields(id, fields, &mut resolve_stack)
+}
+
+fn resolve_fields<'a, 'b>(
+    id: Option<Id<'a>>,
+    fields: &mut Vec<ComponentField<'a>>,
+    resolve_stack: &'b mut Vec<ComponentResolver<'a>>,
```
alexcrichton, line 17: This option seems to be passed around to most functions here, along with the redefinitions of `'a` and `'b`. Would you be up for using a style similar to the core module resolver where a `struct` is used to contain this `Vec`?

Also reaidng over this I'm not sure if the `'b` lifetime parameter throughout this module is still necessary, although I may be missing where it's needed. At least for this function though it's not required.

crates/wast/src/resolve/component_names.rs
```diff
@@ -0,0 +1,817 @@
...
+    let mut found = 0;
+    for resolver in resolve_stack.iter_mut().rev() {
+        if let Ok(resolved) = resolver.resolve(ns, idx) {
+            found = resolved;
+            break;
+        }
+        depth += 1;
+    }
+    if depth as usize == resolve_stack.len() {
+        todo!("name resolution failed")
```
alexcrichton, line 661: I think this can return a normal resolve error here perhaps?

crates/wast/src/binary.rs
```diff
@@ -620,10 +852,36 @@ impl Encode for Global<'_> {
...
-        if let ItemRef::Item { kind, .. } = &self.index {
-            kind.encode(e);
-        }
+        let ItemRef {
+            kind, extra_names, ..
+        } = &self.index;
+        kind.encode(e);
         self.index.encode(e);
+        if !extra_names.is_empty() {
+            eprintln!("TODO: implement alias sugar for encoding `Export`");
```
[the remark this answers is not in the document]
alexcrichton: Or also in general could the `eprintln` be removed from this module?

crates/wast/src/resolve/component_names.rs
```diff
@@ -0,0 +1,817 @@
...
+            }
+            InstanceTypeField::Type(ty) => {
+                resolver.types.register(ty.id, "type")?;
+            }
+            InstanceTypeField::Export(_export) => {}
+        }
+        i += 1;
+    }
+
+    resolve_stack.pop();
```
alexcrichton, line 539: This pattern of push, loop, add aliases, etc shows up in a few places, could this be wrapped up into a helper method to avoid duplicating the logic to insert aliases based on the results of resolution?

crates/wast/src/resolve/component_names.rs
```diff
@@ -0,0 +1,817 @@
...
+                    }
+                    DefType::Component(c) => {
+                        resolver.components.register(c.id, "component")?;
+                    }
+                    DefType::Instance(i) => {
+                        resolver.instances.register(i.id, "instance")?;
+                    }
+                    DefType::Value(v) => {
+                        resolver.values.register(v.id, "value")?;
+                    }
```
alexcrichton, line 478: Is it necessary to maintain these index spaces for component types? I thought that the only thing that mattered at this layer was names-of-types?

alexcrichton (approved):
Looking reasonable to me! One thing I'm not certain about is what's remaining after this, do you feel like you've got a handle on that to write up an issue with remaining work-items?

alexcrichton:
I'm going to be helping to push this over the finish line now and for now I plan on continuing to wean down the size of this by splitting out refactorings such as https://github.com/bytecodealliance/wasm-tools/pull/577. I've done an initial merge of this and that branch to the point where this builds, but there's still quite a lot of `todo!`, `TODO`, `eprintln`, `panic!`, and features that need implementing here before I finish landing this. My hope is to get this to a point which is land-able in the near future which is well-tested too (e.g. roundtrip testing) and then work on subsequent features and sugar required by the text format.

alexcrichton:
Ok I have made a bunch more progress on this but there is yet more still to be made. The branch is in a pretty good place passing a good chunk of tests and most of the remaining failing tests are failing because they need updates from module linking. My current plan with this PR is:

* Probably don't land this behind a feature gate, just land it now.
* Split out more minor refactorings and bugfixes I've been accmulating here and land separately
* Postpone refactoring the `resolve.rs` module as per [this comment](https://github.com/bytecodealliance/wasm-tools/pull/529#discussion_r858787833) and [this comment](https://github.com/bytecodealliance/wasm-tools/pull/529#discussion_r858805377)
* Postpone fleshing out tests beyond what's already included


Beyond all that this PR is, uh, still quite chonky at 6k+ lines. I feel like a big chunk is just the verbosity of the `wast` crate itself, but the component model is also quite large and encompasses a lot (and everything here is brand new, not much ended up being able to be shared).

I will re-review the size of this PR once I get all the refactorings out and see if we can't slim it down. But that's where I'm at! (won't be back at this again until Monday)

alexcrichton:
Oh also the history via git here is an absolute mess now thanks to my endless merging of `main`, my plan is to just squash this all down when merging.

alexcrichton:
Ok this is the final state and I think this is ready for merging. @peterhuene or @fitzgen did one of y'all want to take a look at this? I reviewed Dan's original work here but I've made a number of nontrivial changes here myself. Unfortunately I can't really easily generate a diff between what I last reviewed personally and what's here now due to all the internal refactorings.

Overall I don't think that we need to hold `wast` to too-high of a standard of correctness (as opposed to `wasmparser`) since it's primarily intended for developers. In that sense if y'all don't have time to look at this I think it's ok to land and fixup later with more tests and such.

My next follow-up to this will be to use coverage information to find holes in what we're not parsing and add test for that, followed by holes in what we're not validating and adding tests for that as well.

crates/wast/src/resolve/component_names.rs
alexcrichton: I'm splitting this out to https://github.com/bytecodealliance/wasm-tools/issues/599

crates/wast/src/resolve/component_names.rs
alexcrichton: This is split out to https://github.com/bytecodealliance/wasm-tools/issues/599

crates/wast/src/component/module.rs
```diff
@@ -0,0 +1,118 @@
+use crate::component::*;
+use crate::core;
+use crate::kw;
+use crate::parser::{Cursor, Parse, Parser, Peek, Result};
+use crate::token::{Id, NameAnnotation, Span};
+
+/// A nested WebAssembly nested module to be created as part of a module.
```
peterhuene, line 7:
```suggestion
/// A nested WebAssembly module to be created as part of a component.
```

crates/wast/src/parser.rs
```diff
@@ -264,6 +264,16 @@ pub trait Peek {
...
     }
 
+    /// The same as `peek2`, except it checks the token immediately following
+    /// the token immediately following the current token.
+    fn peek3(mut cursor: Cursor<'_>) -> bool {
+        if cursor.advance_token().is_some() {
+            Self::peek(cursor)
+        } else {
+            false
+        }
```
peterhuene, line 274: Am I misreading this or is this function identical to `peek2`?

peterhuene (commented):
Looks good, just a few comments :+1:. I skipped over most of the test case changes.

Definitely don't need to block on my feedback since we're about to change things here quite a bit with the upcoming text/binary format changes.

crates/wast/src/component/binary.rs
```diff
@@ -0,0 +1,612 @@
...
+                    }
+                    AliasKind::Instance => {
+                        e.push(0x00);
+                        e.push(0x02);
+                    }
+                    AliasKind::Value => {
+                        e.push(0x00);
+                        e.push(0x04);
+                    }
+                    AliasKind::ExportKind(export_kind) => {
```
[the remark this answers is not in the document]
alexcrichton: Yeah I think this will all need refactoring with the new index space splits, and I'm hoping that the index space split will lend a natural structure to the AST here once we implement it. There's a whole mess of test cases that should be passing but are failing tagged with #588 which I hope to get back to once we redo the binary format.

crates/wast/src/component/deftype.rs
```diff
@@ -0,0 +1,424 @@
...
+}
+
+/// A parameter of a [`ComponentFunctionType`].
+#[derive(Clone, Debug)]
+pub struct ComponentFunctionParam<'a> {
+    /// An optional identifer to refer to this `param` by as part of
+    /// name resolution.
+    pub id: Option<Id<'a>>,
+    /// An optional `@name` annotation for this parameter
+    pub name: Option<NameAnnotation<'a>>,
```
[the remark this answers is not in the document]
alexcrichton: Aha yeah this is actually something that turned up really quickly in my "let's use coverage to figure out what's not being tested" and it turns out this wasn't being tested in `wast`. I'll leave this as-is for now but will follow-up shortly with a refactoring to bring this in line with the model's expectation of encodings.

crates/wast/src/parser.rs
alexcrichton: Ah turns out this was dead code!

alexcrichton:
Ok I'm going to go ahead and merge this. Thanks again @sunfishcode for all the initial work here and thanks @peterhuene for the review and putting up with my issues that are sometimes overeager!

crates/wast/src/component/deftype.rs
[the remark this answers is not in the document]
alexcrichton: (specifically I'm changing this in https://github.com/bytecodealliance/wasm-tools/pull/602)

PR #532, opened 2022-03-29
wasmparser: Tweak how features work in the validator

alexcrichton:
This commit removes the `Validator::wasm_features` setter method and
instead replaces it with a getter, allowing access to the activated
features of a `Validator`. At the same time a new constructor,
`Validator::new_with_features` is added. Upon reflection being able to
configure features after construction seems like it can run the risk of
being confusing where one section is validated with one set of features
and another section could be validated with a different set of features.
This API, however, forces the contract that a module is always validated
with the same set of features.


fitzgen (approved):
Nice little API clean up

PR #539, opened 2022-03-30
wasm-smith: Allow specifying the set of valid imports

nagisa:
This is a revival of https://github.com/bytecodealliance/wasm-tools/pull/286. The original PR seems to have gotten blocked on the module linking proposal support, but that has been since removed from wasm-smith and will eventually be replaced by components. My understanding is that the limitations imposed on imports in the original module linking proposal have been removed in the component model, making it a non-issue for this implementation either way.

These are the changes I made over the original PR:

* The available imports are specified by the way of example WebAssembly binary module. This module is parsed and imports are selected from it.
* I changed the logic around how we decide whether to introduce an import or not (basically a boolean check on every iteration of the loop.)
    * I wonder if it would be better to weigh this check towards including more imports. Or perhaps some other approach?
    * In particular, I had a hard time reconciling `min_imports` and `available_imports`; e.g. what happens if `min_imports > available_imports.len()`? The implementation right now ignores `min_imports` altogether.

Fixes https://github.com/bytecodealliance/wasm-tools/issues/276


fitzgen:
> * In particular, I had a hard time reconciling `min_imports` and `available_imports`; e.g. what happens if `min_imports > available_imports.len()`? The implementation right now ignores `min_imports` altogether.

I think this is fine. In general, we consider max limits as hard limits and min limits as best effort limits.

fitzgen (commented):
Thanks for reviving this!

alexcrichton:
One thing I'm wary of in support like this is that we're duplicating the type system of wasm modules in yet-another-location. This is somewhat ok today when the type system is simple but I wouldn't want to implement this for the component model for example. 

I don't really have any silver bullets about how to solve that though, and in the meantime it's not so big as to rebally be that big of an issue. Otherwise though one possible idea is that the set of available imports is specified as a wasm module which allows us to reuse `wasmparser`'s type system. Other than that though I don't know how we'd deduplicate, but I definitely don't think that this approach is scalable as the type system gets more complex.

crates/wasm-smith/src/config.rs
```diff
@@ -2,6 +2,80 @@
...
+    /// Construct a new static `ImportType::Function`.
+    pub const fn function(params: &'static [ValType], results: &'static [ValType]) -> Self {
+        Self::Function {
+            params: Cow::Borrowed(params),
+            results: Cow::Borrowed(results),
+        }
+    }
+
+    /// Construct a new static `ImportType::Tag`.
+    pub const fn tag(params: &'static [ValType]) -> Self {
```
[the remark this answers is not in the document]
nagisa: Alas `Into` is not `const`. I was trying to make it easier to construct `'static` arrays of `AvailableImports` which requires that the array's constituents are all const-evaluatable.

nagisa:
> One thing I'm wary of in support like this is that we're duplicating the type system of wasm modules in yet-another-location.

Do I understand correctly that by “type system” you mean largely the `AvailableImport` and `ImportType` ADTs? In that case, yeah, it is definitely sub-par.

> Otherwise though one possible idea is that the set of available imports is specified as a wasm module which allows us to reuse wasmparser's type system.

Do I understand correctly that, were this idea implemented, the imports would be provided as something like a `Vec<u8>` of binary encoding of a WebAssembly module, possibly corresponding to something along the lines of

```wat
(module
    (import ...)
    (import ...)
)
```

which `wasm-smith` would then parse and go from there? That definitely seems like a reasonable approach to try implementing. Definitely seems like having an ability to write down valid imports “by-example” would be more convenient than constructing ad-hoc arrays of some random ADTs.

alexcrichton:
Yeah `ImportType` and its structure is what I'm worried about as the type system gets more complex (e.g. more pointers-to-things).

My rough idea, which to be clear I suspect is still nontrivial work and not guaranteed to be a win, is what you mentioned probably with the `wat` crate thrown in the mix to support the text format as well. Roughly `wasmparser` would be used to parse the type section and that would be used as the source of the set of available imports. That would still involve writing a translation from `wasmparser`'s type hierarchy to `wasm-smith`, but it would at least avoid writing another type hierarchy. I dunno the best way to finagle this though.

crates/wasm-smith/tests/core.rs
```diff
@@ -139,66 +139,25 @@ impl Config for ImportConfig {
...
-                    },
-                },
-                wasm_smith::AvailableImport {
-                    module: "tags".into(),
-                    field: "tag1".into(),
-                    import_type: ImportType::tag(&[ValType::I32]),
-                },
-            ]
+            wat::parse_str(
+                r#"
```
nagisa, line 145: Nit to self: ew, rustfmt.

crates/wasm-smith/src/core.rs
```diff
@@ -509,6 +562,194 @@ impl Module {
...
+        let mut make_func_type = |parsed_sig_idx: u32| {
+            let serialized_sig_idx = match available_types.get_mut(parsed_sig_idx as usize) {
+                None => panic!("signature index refers to a type out of bounds"),
+                Some((_, Some(idx))) => *idx as usize,
+                Some((wasmparser::TypeDef::Func(func_type), index_store)) => {
+                    let new_index = first_type_index + new_types.len();
+                    if new_index < max_types {
+                        let func_type = Rc::new(FuncType {
+                            params: func_type.params.iter().map(|t| convert_type(*t)).collect(),
+                            results: func_type.returns.iter().map(|t| convert_type(*t)).collect(),
```
nagisa, line 622: Nit to self: we should disallow function types with multiple results without the multi_value feature.

crates/wasm-smith/src/config.rs
```diff
@@ -59,6 +60,41 @@ pub trait Config: 'static + std::fmt::Debug {
...
+    ///         (import "env" "ping" (func (param i32)))
+    ///         (import "env" "pong" (func (result i32)))
+    ///         (import "env" "memory" (memory 1))
+    ///         (import "env" "table" (table 1))
+    ///         (import "env" "tag" (tag (param i32)))
+    ///     )
+    /// "#))
+    /// # ;
+    /// ```
+    fn available_imports(&self) -> Option<Cow<'static, [u8]>> {
```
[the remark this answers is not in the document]
nagisa: Not super straightforward to do so.

The reason for `'static` is that `self` needs to be borrowed mutably throughout the construction of a module, and allowing `available_imports` to return a reference to an implementer of `Config` would effectively cause a hold of another reference reference to `self`'s contents. There is probably a way to write down the `arbitrary_initial_sections` that avoids the issue, but it'll probably have significant hit on architecture of the code...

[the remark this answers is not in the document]
nagisa: Actually never mind, I just figured out a way to move the check for `available_imports` into the function that uses said available imports. rustc is smart enough to do the rest.

crates/wasm-smith/src/core.rs
```diff
@@ -509,6 +561,201 @@ impl Module {
...
+            new_imports.push(Import {
+                module: import.module.to_string(),
+                field: import.name.to_string(),
+                entity_type,
+            });
+            self.num_imports += 1;
+        }
+
+        // Finally, add the sections we just generated.
+        if !new_imports.is_empty() || u.arbitrary().unwrap_or(false) {
```
fitzgen, line 737: Is `u.arbitrary()` call effectively only making it so we add all available imports or none of them? It seems to me like we should be calling `u.arbitrary()` (or `u.ratio(...)`) for each import instead.

fitzgen (commented):
Thanks! I like the direction of using a wasm module as a dsescription of available imports, that's a neat trick.

crates/wasm-smith/src/core.rs
nagisa: The `arbitrary` here makes a decision as to whether an empty section should be added or not. This is a pattern that is replicated throughout the this module, so perhaps it calls for some sort of abstraction.

nagisa:
Addressed the comments.

crates/wasm-smith/src/core.rs
```diff
@@ -509,6 +561,201 @@ impl Module {
...
+                wasmparser::Payload::ImportSection(mut import_reader) => {
+                    for _ in 0..import_reader.get_count() {
+                        let im = import_reader.read().expect("could not read import");
+                        // We can immediately filter whether this is an import we want to
+                        // use.
+                        let use_import = u.arbitrary().unwrap_or(false);
+                        if !use_import {
+                            continue;
+                        }
+                        available_imports.push(im);
```
fitzgen, line 601: Ah okay and here is where the filtering happens, not below.

fitzgen (approved):
Thanks!

PR #542, opened 2022-04-01
wasm-mutate: remove empty sections

nagisa:
I have found that `wasm-shrink` would reduce the reproducer case to 1 or
two relevant sections and leave a bunch of other empty sections laying
around. This change introduces a mutation to `wasm-mutate` which is able
to clean-up these sections. Since the removed sections are empty,
removing them should have no impact on the validity of the module.


crates/wasm-mutate/src/mutators/remove_section.rs
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    fn mutate<'a>(
+        self,
+        config: &'a mut WasmMutate,
+    ) -> Result<Box<dyn Iterator<Item = Result<Module>> + 'a>>
+    where
+        Self: Copy,
+    {
+        let custom_sections = config
```
alexcrichton, line 82: I think with this change this may be better named something like `removal_candidates` perhaps?

alexcrichton (commented):
Looks good to me, thanks!

PR #543, opened 2022-04-01
Swap order of alias index printing

alexcrichton:
From bytecodealliance/wit-bindgen#183 I realized that the current
explainer has the printing of indices here slightly different where the
outer index/name is outside in the `alias` annotation where the local
name/index is printed inside of the `type` or item container.

This also updates the printing of the local index to attempt to use the
name section, if one is present.


peterhuene (approved):

peterhuene:
Thanks, beat me to it :)

PR #548, opened 2022-04-05
wasm-smith: no flag encoding in elems w/o feats

nagisa:
The encoding used by the `Some` table indices requires the bulk_memory
proposal. This proposal introduced the flag-based encoding, including
the `02` flag.

Sadly I don't think I can write an actual test here, because wasmparser's
binary reader will accept either encoding regardless of the features AFAIK.

### DIFF
--- a/crates/wasm-encoder/src/core/elements.rs
+++ b/crates/wasm-encoder/src/core/elements.rs
@@ -74,8 +74,9 @@ pub enum ElementMode<'a> {
     Active {
         /// The table index.
         ///
-        /// `None` is implicitly table `0`. Non-`None` tables are part of the
-        /// reference types proposal, including `Some(0)`.
+        /// `Active` element specifying a `None` table forces the MVP encoding and refers to the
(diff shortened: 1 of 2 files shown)


crates/wasm-encoder/src/core/elements.rs
```diff
@@ -74,8 +74,9 @@ pub enum ElementMode<'a> {
     Active {
         /// The table index.
         ///
-        /// `None` is implicitly table `0`. Non-`None` tables are part of the
-        /// reference types proposal, including `Some(0)`.
+        /// `Active` element specifying a `None` table forces the MVP encoding and refers to the
+        /// 0th table holding `funcref`s. Non-`None` tables use the encoding introduced with the
+        /// bulk memory proposal and can refer to tables with any valid reference type.
```
nagisa, line 79: Ref: https://github.com/WebAssembly/spec/blob/main/proposals/bulk-memory-operations/Overview.md#element-segments

PR #549, opened 2022-04-05
Start support for component model in `wasm-tools dump`

alexcrichton:
I found a need to do this when debugging a module for Wasmtime, so I
went ahead and filled out the basics of `wasm-tools dump` to help
debugging in the future as well.

### DIFF
--- a/crates/dump/src/lib.rs
+++ b/crates/dump/src/lib.rs
@@ -14,6 +14,7 @@ struct Dump<'a> {
     state: String,
     dst: String,
     nesting: u32,
+    offset_width: usize,
 }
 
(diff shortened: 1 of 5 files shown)


fitzgen (approved):
Add a component test dump?

alexcrichton:
That's an excellent point, I'll wait to merge this until https://github.com/bytecodealliance/wasm-tools/pull/529 is ready

fitzgen:
I'm going to go ahead and merge this now, because debugging invalid components generated by `wasm-smith` is a huge pain without this. We can always land tests in a follow up PR after #529 merges.

PR #553, opened 2022-04-07
`wasm-mutate`: Fix the `AddTypeMutator` when the first section is a custom section

fitzgen:
And the type section follows that. Before we would accidentally replace the custom section with the new type section and then incorrectly have two type sections. Now we leave the custom section and correctly replace the old type section.


alexcrichton (approved):
Could a test be added for this as well? I think the `(@custom ..)` syntax can be used to insert a custom section before the type section in the text format.

PR #554, opened 2022-04-13
`wasm-mutate`: Add a mutator for custom sections

fitzgen:


crates/wasmparser/src/parser.rs
```diff
@@ -538,27 +527,13 @@ impl Parser {
...
-                        data_offset: content.original_position(),
-                        data: content.remaining_buffer(),
-                        range,
-                    });
-                }
+                if id == 0 {}
 
                 match (self.encoding, id) {
                     // Module sections
+                    (Encoding::Module, 0) => {
```
alexcrichton, line 534: I think this needs to match either encoding, right? (both components and core wasm modules)

alexcrichton (approved):
Looks good to me! Can `CustomSectionReader` have private fields though field with accessors?

crates/wasmparser/src/parser.rs
fitzgen: Done.

fitzgen:
> Can CustomSectionReader have private fields though field with accessors?

Done. Made them `pub(crate)` tho so that I didn't have to update all the `assert_matches!` parse tests.

Issue #450, opened 2022-01-20
[meta] Support for the component model


fitzgen:
Here is a list of what is required to migrate Bytecode Alliance projects to the new module linking spec that is based on the component model, rather than the old module linking spec that was based on core Wasm. Putting this list in the `wasm-tools` repo because most of the crates are here, and the work needs to start here. Let me know if I'm missing something and/or just edit this issue to add a checkbox for the missing item!

* [x] `wasmparser` (@peterhuene) PR: #484

* [x] `wat` and `wast` (@sunfishcode) PR: #621

* [x] `wasm-encoder` (@peterhuene) PR: #448 

* [x] `wasmprinter` (@peterhuene) PR: #496 
  * (depends on `wasmparser`) 

* [ ] `wasm-smith` (@fitzgen)
  * (depends on `wasm-encoder`)

* [ ] `wasm-mutate`
  * (depends on `wasm-encoder`)

* [x] `wasm-tools` local tests
  * (depends on `wat`, `wasmparser`)

* [ ] `wasm-tools` fuzz targets
  * (depends on `wat`, `wasmparser`, `wasmprinter`, and `wasm-smith`)

* [x] `wasm-tools dump` PR: https://github.com/bytecodealliance/wasm-tools/pull/549
  * (depends on `wat`, `wasmparser`)

* [x] `wasm-tools objdump` PR: https://github.com/bytecodealliance/wasm-tools/pull/555
  * (depends on `wat`, `wasmparser`)

* [ ] Wasmtime: https://github.com/bytecodealliance/wasmtime/issues/4185
  * (depends on `wat`, `wasm-encoder`, and `wasmparser`)

* [ ] `wit-bindgen`
  * (depends on `wat`, `wasm-encoder`, `wasmparser`, and Wasmtime)

* [ ] Wizer (https://github.com/bytecodealliance/wizer/issues/48)
  * (depends on `wat`, `wasm-encoder`, `wasmparser`, and Wasmtime)

* [ ] Interface Types + Canonical ABI test case generator

cc @peterhuene @alexcrichton @tschneidereit @lukewagner

peterhuene:
Thanks for putting this together, Nick!

I have a PR up for both `wasm-encoder` and `wasmparser` currently. An updated component model binary grammar is forthcoming soon, so I will revise both of those PRs with the new grammar as it should help reduce the complexity of those implementations greatly.

peterhuene:
[...]

PR #555, opened 2022-04-14
feat: Initial Component Model support in objdump

saulecabrera:
Part of https://github.com/bytecodealliance/wasm-tools/issues/450

But as in https://github.com/bytecodealliance/wasm-tools/pull/549 I think that it might be best to wait until https://github.com/bytecodealliance/wasm-tools/pull/529 lands to confidently test and merge. 

---

This commit modifies the original implementation of `objdump` to support
the Component Model.

The main difference with the previous implementation is that it augments
the properties of the `Printer` implementation to track the component
and module indices as well as the encoding that is currently being
processed `(Module | Component)`

--- 
Results:

Encoding [this component](https://github.com/bytecodealliance/wit-bindgen/blob/main/crates/wit-component/tests/components/import-conflict/component.wat) with `wit-component` and then applying `wasm-tools objdump`, this is the result:

<details>

```
  types                                  |        0xa -       0x45 |        59 bytes | 7 count
  module                                 |       0x48 -       0xd5 |       141 bytes | 1 count
    ------ start module 0 -------------
    types                                |       0x52 -       0x6a |        24 bytes | 4 count
    imports                              |       0x6c -       0x87 |        27 bytes | 3 count
    functions                            |       0x89 -       0x8c |         3 bytes | 2 count
    memories                             |       0x8e -       0x91 |         3 bytes | 1 count
    exports                              |       0x93 -       0xca |        55 bytes | 3 count
    code                                 |       0xcc -       0xd5 |         9 bytes | 2 count
    ------ end module 0 -------------
  imports                                |       0xd7 -       0xe7 |        16 bytes | 3 count
  module                                 |       0xe9 -      0x141 |        88 bytes | 1 count
    ------ start module 1 -------------
    types                                |       0xf3 -      0x100 |        13 bytes | 2 count
    functions                            |      0x102 -      0x105 |         3 bytes | 2 count
    tables                               |      0x107 -      0x10c |         5 bytes | 1 count
    exports                              |      0x10e -      0x122 |        20 bytes | 3 count
    code                                 |      0x124 -      0x141 |        29 bytes | 2 count
    ------ end module 1 -------------
  module                                 |      0x143 -      0x180 |        61 bytes | 1 count
    ------ start module 2 -------------
    types                                |      0x14d -      0x15a |        13 bytes | 2 count
    imports                              |      0x15c -      0x176 |        26 bytes | 3 count
    elements                             |      0x178 -      0x180 |         8 bytes | 1 count
    ------ end module 2 -------------
  instances                              |      0x182 -      0x187 |         5 bytes | 1 count
  alias                                  |      0x189 -      0x199 |        16 bytes | 3 count
  functions                              |      0x19b -      0x19f |         4 bytes | 1 count
  instances                              |      0x1a1 -      0x1cc |        43 bytes | 4 count
  alias                                  |      0x1ce -      0x1e7 |        25 bytes | 3 count
  functions                              |      0x1e9 -      0x1f5 |        12 bytes | 2 count
  instances                              |      0x1f7 -      0x214 |        29 bytes | 2 count
  exports                                |      0x216 -      0x217 |         1 bytes | 0 count

```

</details>

Similarly running `objdump` on a core wasm module, for example [this one](https://github.com/bytecodealliance/wasm-tools/blob/main/tests/local/br_if-loop-stack.wasm), gives:

<details>

```
  types                                  |        0xa -        0xf |         5 bytes | 1 count
  functions                              |       0x11 -       0x13 |         2 bytes | 1 count
  exports                                |       0x15 -       0x1d |         8 bytes | 1 count
  code                                   |       0x1f -       0x32 |        19 bytes | 1 count
  custom "name"                          |       0x39 -       0x47 |        14 bytes | 1 count
```
</details>

---

Some open questions:

1. wasmparser's new `Payload`, doesn't return a `count` for `Payload::ComponentSection` and `Payload::ModuleSection` so I've defaulted to adding a count of 1 when processing either or. Not sure if we'd like to change that approach, since it might be a bit confusing for someone reading and expecting the total count as it was before. Although that information can be easily derived with the counts and information about each module/component.
2. I'm not able to find any roundtrip tests for `objdump`; is this right or am I missing something? If this is right, would there be any value on adding some tests for this, either on this PR or a subsequent one? (Ideally once https://github.com/bytecodealliance/wasm-tools/pull/529 lands I guess)
3. As of this change I'm keeping the component section names without an explicit differentiator in the name (e.g. "types" instead of "component types"). I'm marking the start/end of each component/module so IMO that should be difference enough to avoid having different names for the component sections. Another reason why I went this way is because the spec doesn't seem to give these sections different names. Let me know if I'm missing something here or if this should be changed.  

Thoughts? @peterhuene @alexcrichton

### DIFF
--- a/src/bin/wasm-tools/objdump.rs
+++ b/src/bin/wasm-tools/objdump.rs
@@ -18,9 +18,11 @@ impl Opts {
         let input = wat::parse_file(&self.input)?;
 
         let mut printer = Printer::default();
+        printer.indices.push(IndexSpace::default());
+
         for payload in Parser::new(0).parse_all(&input) {
             match payload? {
-                Version { .. } => printer.start(),
+                Version { encoding, .. } => printer.start(encoding),
 
                 TypeSection(s) => printer.section(s, "types"),
                 ImportSection(s) => printer.section(s, "imports"),
@@ -37,15 +39,27 @@ impl Opts {
                 CodeSectionStart { range, count, .. } => printer.section_raw(range, count, "code"),
                 CodeSectionEntry(_) => {}
 
-                ComponentTypeSection(_) => todo!("component-model"),
-                ComponentImportSection(_) => todo!("component-model"),
-                ComponentFunctionSection(_) => todo!("component-model"),
-                ModuleSection { .. } => todo!("component-model"),
-                ComponentSection { .. } => todo!("component-model"),
-                InstanceSection(_) => todo!("component-model"),
-                ComponentExportSection(_) => todo!("component-model"),
-                ComponentStartSection { .. } => todo!("component-model"),
-                AliasSection(_) => todo!("component-model"),
+                ComponentTypeSection(s) => printer.section(s, "types"),
+                ComponentImportSection(s) => printer.section(s, "imports"),
+                ComponentFunctionSection(s) => printer.section(s, "functions"),
+                ModuleSection { range, .. } => {
+                    printer.section_raw(range, 1, "module");
+                    if let Some(space) = printer.indices.last_mut() {
+                        space.inc_module();
+                    }
+                }
+
+                ComponentSection { range, .. } => {
+                    let mut index_space = IndexSpace::default();
+                    index_space.inc_component();
+                    printer.indices.push(index_space);
+                    printer.section_raw(range, 1, "component")
+                }
+
+                InstanceSection(s) => printer.section(s, "instances"),
+                ComponentExportSection(s) => printer.section(s, "exports"),
+                ComponentStartSection(_) => {}
+                AliasSection(s) => printer.section(s, "alias"),
 
                 CustomSection(c) => printer.section_raw(
                     c.data_offset()..c.data_offset() + c.data().len(),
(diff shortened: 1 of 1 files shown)


saulecabrera:
I'm not sure about the nightly failure, it seems that wasm-mutate is failing to compile.

alexcrichton:
Seems reasonable to me! The failure on nightly was a SIGKILL which typically means something was interrupted or OOM'd, so I restarted to see what happens.

> wasmparser's new Payload, doesn't return a count for Payload::ComponentSection and Payload::ModuleSection so I've defaulted to adding a count of 1 when processing either

Oh that's because the old module linking binary encoding proposal is different from the current component model where in the old module linking the module section was a list of modules but in the component module each section contains exactly one module, so there's not actually any section counts for component or module sections.

> I'm not able to find any roundtrip tests for objdump; is this right or am I missing something?

Nah you're right that there's no tests. This was mostly just a utility crate I found helpful in debugging at some point and isn't really super polished or anything like that. If you'd be up for adding tests that would be awesome, otherwise I think it's ok to just manually verify things look ok.

> As of this change I'm keeping the component section names without an explicit differentiator in the name (e.g. "types" instead of "component types").

Seems reasonable to me!

saulecabrera:
I just tried locally on nightly and it works, the difference between machine specs could explain why it works locally though, for reference:

```
MacBook Pro (14-inch, 2021)
Chip Apple M1 Pro
Memory 16 GB
```

peterhuene:
I'm also unable to reproduce with nightly on either macos or linux, so likely just the compiler eating up too much memory for the constrained CI system.

If this doesn't resolve soon, I propose we disable nightly CI for the near term.

src/bin/wasm-tools/objdump.rs
```diff
@@ -36,15 +38,27 @@ impl Opts {
...
+                ModuleSection { range, .. } => { 
+                    printer.section_raw(range, 1, "module");
+                    if let Some(space) = printer.indices.last_mut() {
+                        space.inc_module();
+                    }
+                },
+
+                ComponentSection { range, .. } => {
+                    let mut index_space = IndexSpace::default();
+                    index_space.inc_component();
```
peterhuene, line 53: Should this be incremented on the parent index space similar to how it's performed for `ModuleSection`?

src/bin/wasm-tools/objdump.rs
```diff
@@ -36,15 +38,27 @@ impl Opts {
...
+                    printer.indices.push(index_space);
+                    printer.section_raw(range, 1, "component")
+                },
+
+                InstanceSection(s) => printer.section(s, "instances"),
+                ComponentExportSection(s) => printer.section(s, "exports"),
+                ComponentStartSection(_) => {},
```
peterhuene, line 60: Should we treat this the same as a module start section?

src/bin/wasm-tools/objdump.rs
```diff
@@ -70,40 +84,91 @@ impl Opts {
...
-            *a += 1;
-            *a == *b
-        } else {
-            false
-        };
-        if pop {
-            self.module_code_counts.pop();
+
+        if let Some(space) = self.indices.last_mut() {
+            match space.processing.last() {
```
peterhuene, line 141: Can this just pop here to condense this to something like:

```rust
match space.processing.pop() {
  Some(wasmparser::Encoding::Module) => {
    println!("{}------ end module {} -------------", header, space.modules.unwrap());
  }
  Some(wasmparser::Encoding::Component) => {
    println!("{}------ end component {} ----------", header, space.components.unwrap());
  }
  None => self.indices.pop(),
}
```

src/bin/wasm-tools/objdump.rs
```diff
@@ -70,40 +84,91 @@ impl Opts {
     }
 }
 
+#[derive(Default, Debug)]
+struct IndexSpace {
+    modules: Option<u32>,
```
peterhuene, line 89: Do these need to be `Option`? What if we didn't increment these values upon encountering a module or component section, but instead in `end` (i.e. we've just finished a nested module/component, increment the parent index space's counters)? That would also simplify the increment logic.

peterhuene (commented):
This looks great, thanks for taking this on @saulecabrera!

Just a few comments I have to discuss.

src/bin/wasm-tools/objdump.rs
saulecabrera: Good catch, yeah! I applied your suggestion to simplify the increment logic and this is now done when ending a component.

src/bin/wasm-tools/objdump.rs
saulecabrera: Agreed, added it in https://github.com/bytecodealliance/wasm-tools/pull/555/commits/44d1d0e04cd7c6ff6fd6779c05b164beeba0a8e6

src/bin/wasm-tools/objdump.rs
saulecabrera: Yeah, agreed that this is better, I also defer calling `printer.start` when encountering a `Component` or `Module` rather than when encountering a `Version`, that makes it easier to handle the "top-level" case.

src/bin/wasm-tools/objdump.rs
```diff
@@ -65,40 +76,75 @@ impl Opts {
...
+                        space.inc_component();
+                    }
+                },
+                None => { self.indices.pop(); },
+            }
         }
     }
 
-    fn section<T>(&self, section: T, name: &str)
+    fn section<T>(&mut self, section: T, name: &str)
```
peterhuene, line 140: Does this change (and the change below to `section_raw`) still need to be made?

src/bin/wasm-tools/objdump.rs
```diff
@@ -65,40 +76,75 @@ impl Opts {
...
 
+#[derive(Debug, Default)]
+struct IndexSpace {
+    modules: u32,
+    components: u32,
+    processing: Vec<Encoding>,
+}
+
+impl IndexSpace {
+    pub fn inc_module(&mut self) {
```
peterhuene, line 87: Nit: could probably just delete these methods since they're no longer complex enough to justify them?

peterhuene (approved):
Looks great, just two minor nits.

Thanks for taking this change on!

peterhuene:
@alexcrichton I think we can land this without wat tests for now. What's your take?

alexcrichton:
Yeah that sounds reasonable to me, especially because there's no tests already!

PR #556, opened 2022-04-15
Use std::ops::Range instead of a custom type

alexcrichton:
This commit removes `wasmparser::Range` and instead uses
`std::ops::Range<usize>` instead. This is inspired by review on
bytecodealliance/wasmtime#4005 where it's more convenience to have all
the libstd methods available.

### DIFF
--- a/crates/wasm-mutate/src/info.rs
+++ b/crates/wasm-mutate/src/info.rs
@@ -4,6 +4,7 @@ use crate::{
 };
 use std::collections::HashSet;
 use std::convert::TryFrom;
+use std::ops::Range;
 use wasm_encoder::{RawSection, SectionId};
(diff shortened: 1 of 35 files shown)


fitzgen:
What if we just had

```rust
pub type Range = std::ops::Range<usize>;
```

in `src/lib.rs`?

alexcrichton:
I personally find typedefs like that confusing unless they're well-known which I figure this one probably won't be. I would hope that there wouldn't be too much churn for downstream consumers since this type isn't so common to work with, but are you thinking it'll have a bigger impact?

fitzgen:
Nah it was just an idea. We can go ahead with this.

fitzgen (approved):

PR #560, opened 2022-04-19
`wasm-smith`: Generate module types with canonical ABI functions

fitzgen:
It is highly unlikely that we would ever generate module types that export the
canonical ABI's expected `realloc` and `free` functions by chance, and passing
certain types across an interface types boundary requires these functions, so
special case them and generate them most of the time for most modules. Do the
same for the `memory` export.

### DIFF
--- a/crates/wasm-smith/Cargo.toml
+++ b/crates/wasm-smith/Cargo.toml
@@ -16,7 +16,7 @@ name = "corpus"
 harness = false
 
 [dependencies]
-arbitrary = { version = "1.0.0", features = ["derive"] }
+arbitrary = { version = "1.1.0", features = ["derive"] }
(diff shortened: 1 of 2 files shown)


alexcrichton (approved):

alexcrichton:
FWIW the business around `canonical_abi_free` is a bit of a moving target where with the next draft of the canonical ABI it goes away and is instead replaced with post-return functions. This'll eventually need an update but matching what wasmparser validates for now seems reasonable too.

fitzgen:
Yeah, I actually had this written weeks ago and forgot about it.

Agreed that it makes sense to match `wasmparser` for now, even tho they will both need updates Soon(tm)

PR #562, opened 2022-04-21
wasm-smith: Generate arbitrary component function sections

fitzgen:


alexcrichton (approved):
Do we have a test that enables the component model support at this time? It may be a bit early to enable in the fuzzers but having it on in one of the wasm-smith tests might be good (we may have this already too)

fitzgen:
> Do we have a test that enables the component model support at this time? It may be a bit early to enable in the fuzzers but having it on in one of the wasm-smith tests might be good (we may have this already too)

We have a smoke test that uses `rand` to generate a bunch of random byte sequences, feeds them into `wasm-smith` to generate a component, and then checks that we can successfully generate and encode a component.

My next PR is going to be enabling `wasmparser`'s validation on the output, which didn't exist yet when that test was written.

PR #565, opened 2022-04-26
Add a fuzz target for validating components generated by wasm-smith

itsrainy:
This was created based on the equivalent fuzz target for modules

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -80,3 +80,9 @@ name = "mutate"
 path = "fuzz_targets/mutate.rs"
 test = false
 doc = false
+
+[[bin]]
+name = "validate-component"
+path = "fuzz_targets/validate-component.rs"
+test = false
+doc = false
--- a/fuzz/fuzz_targets/validate-component.rs
+++ b/fuzz/fuzz_targets/validate-component.rs
@@ -1,0 +1,26 @@
+#![no_main]
+use libfuzzer_sys::fuzz_target;
+
+fuzz_target!(|data: &[u8]| {
+    let (bytes, config) = match wasm_tools_fuzz::generate_valid_component(data, |_, _| Ok(())) {
+        Ok(c) => c,
+        Err(_) => return,
+    };
+
+    // Validate the component and assert that it passes validation.
+    let mut validator = wasmparser::Validator::new_with_features(wasmparser::WasmFeatures {
+        component_model: true,
+        multi_value: config.multi_value_enabled,
+        multi_memory: config.max_memories > 1,
+        bulk_memory: true,
+        reference_types: true,
+        simd: config.simd_enabled,
+        relaxed_simd: config.relaxed_simd_enabled,
+        memory64: config.memory64_enabled,
+        exceptions: config.exceptions_enabled,
+        ..wasmparser::WasmFeatures::default()
+    });
+    if let Err(e) = validator.validate_all(&bytes) {
+        panic!("Invalid component: {}", e);
+    }
+});
--- a/fuzz/src/lib.rs
+++ b/fuzz/src/lib.rs
@@ -1,6 +1,6 @@
 use libfuzzer_sys::arbitrary::{Result, Unstructured};
 use std::fmt::Debug;
-use wasm_smith::{Module, SwarmConfig};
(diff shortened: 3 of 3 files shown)


alexcrichton:
Seems reasonable to me! Given how resources are allocated though on oss-fuzz for our fuzzing I think it might be better to instead fold this into the `validate-valid-module` fuzzer (despite the name at this point). When doing that we could use something like [`Unstructured::ratio`](https://docs.rs/arbitrary/latest/arbitrary/struct.Unstructured.html#method.ratio) to bias generating more modules than components (since that's the highest-value for now).

Also, @fitzgen and @peterhuene, do y'all feel that wasm-smith/wasmparser are up for fuzzing at this point?

fitzgen:
> Also, @fitzgen and @peterhuene, do y'all feel that wasm-smith/wasmparser are up for fuzzing at this point?

The parts of components in `wasm-smith` that are implemented are worth fuzzing, for sure. At this point the only way to get it better is to actually start throwing it at the fuzzers.

fitzgen:
> Seems reasonable to me! Given how resources are allocated though on oss-fuzz for our fuzzing I think it might be better to instead fold this into the `validate-valid-module` fuzzer (despite the name at this point). When doing that we could use something like [`Unstructured::ratio`](https://docs.rs/arbitrary/latest/arbitrary/struct.Unstructured.html#method.ratio) to bias generating more modules than components (since that's the highest-value for now).

This seems reasonable.

@itsrainy, this would involve making `generate_valid_{module,component}` take a `u: &mut Unstructured` rather than `&[u8]` and then basically doing something like this

```rust
use arbitrary::Unstructured;

fuzz_target!(|data: &[u8]| {
    let mut u = Unstructured::new(data);
    let (wasm_bytes, config) = if u.ratio::<u8>(1, 10)? {
        // generate an arbitrary component 1/10
        // times via `generate_valid_component`
    } else {
        // 9/10 times generate an arbitrary core module here
        // via `generate_valid_module.
    };

    let mut validator = wasmparser::Validator::new_with_features(wasmparser::WasmFeatures {
        component_model: config.component_model,
        // etc...
    });
    if let Err(e) = validator.validate_all(&bytes) {
        panic!("Invalid module: {}", e);
    }
});
```

Does that make sense?

itsrainy:
@fitzgen yep that makes sense!

> Given how resources are allocated though on oss-fuzz for our fuzzing I think it might be better to instead fold this into the validate-valid-module fuzzer (despite the name at this point).

@alexcrichton just for my understanding, can you elaborate on how oss-fuzz allocates resources for us? Is it splitting some set amount of compute time equally between each of our fuzz targets or something?

alexcrichton:
Oh oops right yes sorry, what you're thinking is how it works, we get a fixed amount of time and it's split between all of our fuzzers.

crates/wasm-smith/src/config.rs
```diff
@@ -429,6 +429,7 @@ pub struct SwarmConfig {
     pub available_imports: Option<Vec<u8>>,
     pub bulk_memory_enabled: bool,
     pub canonicalize_nans: bool,
+    pub component_model: bool,
```
fitzgen, line 432: Sorry I forgot that we don't actually want to create this `bool` flag because nothing inside `wasm-smith` will ever use it in practice since we statically know whether we are doing the component model or not based on whether we call into `wasm-smith` via `Module::new` or via `Component::new`.

So let's remove this flag, and instead...

fuzz/fuzz_targets/validate-valid-module.rs
```diff
@@ -1,30 +1,47 @@
...
-fuzz_target!(|m: &[u8]| {
-    let (bytes, config) = match wasm_tools_fuzz::generate_valid_module(m, |_, _| Ok(())) {
-        Ok(m) => m,
-        Err(_) => return,
+fuzz_target!(|data: &[u8]| {
+    let mut u = Unstructured::new(data);
+    let generate_component = match u.ratio::<u8>(1, 10) {
+        Ok(b) => b,
+        Err(_) => false,
+    };
```
fitzgen, line 13: ...and instead use the existing `generate_component` variable to turn `WasmFeatures::component_model` on or off in the validator down below.

PR #566, opened 2022-04-26
wasm-smith: fix bug in tag type generation for components

itsrainy:
This fixes a bug found by the fuzz target introduced in https://github.com/bytecodealliance/wasm-tools/pull/565. Filtering for empty tag results *before* enumerating was causing the following exception:

```thread '<unnamed>' panicked at 'Invalid module: invalid exception type: non-empty tag result type (at offset 0xb)', fuzz/fuzz_targets/validate-component.rs:25:9```

The call to enumerate should come first and then we can successfully filter for empty tag result types


fitzgen (approved):
Thanks!

PR #568, opened 2022-04-27
`wasm-smith`: Add support for generating nested components

fitzgen:


peterhuene (approved):
👍 Thanks for fixing that validation bug too.

PR #574, opened 2022-05-06
wasmparser: delay lowered function type calculation in component validation

peterhuene:
This PR delays the calculation of the lowered function type of a component function type until referenced from a component function definition.

This will remove unnecessary lowered function type calculations for things like component type definitions that may contain function types never referenced from a function definition.

This PR also updates `wasmparser` to validate with the latest expected canonical ABI signatures.

### DIFF
--- a/crates/wasmparser/Cargo.toml
+++ b/crates/wasmparser/Cargo.toml
@@ -13,6 +13,7 @@ edition = "2021"
 
 [dependencies]
 indexmap = "1.8.0"
+once_cell = "1.10.0"
 
 [dev-dependencies]
 anyhow = "1.0"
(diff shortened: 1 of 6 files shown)


alexcrichton:
Would it be possible to instead structure this where instead of calculating the function type and using `PartialEq` instead validation is done over a particular function type as input? That would avoid the next to lazily calculate something here and would additionally enforce that we only ever do the work when it's actually necessary.

I don't think there's anything wrong with this approach per-se but I think it would be best if we could keep the dependencies here slim and avoid allocations altogether if we can.

peterhuene:
> validation is done over a particular function type as input

I think so. I'll implement it that way instead.

peterhuene:
@alexcrichton Okay, I pushed up a commit that removes allocations when determining the lowered function signatures, also bailing out early when the maximum flat parameters/results are reached; removed the need for the memoization of the lowered signature.

alexcrichton (approved):

PR #575, opened 2022-05-09
wasmparser: fix variant lowering calculation.

peterhuene:
This PR fixes variant lowering when calculating lowered function
signatures.

A `get_mut` call was being made on underlying storage space for the
`LoweredTypes` helper type, which always returned success, causing the case
types to never be appended on to the original list of lowered types.

This commit fixes this by implementing `get_mut` on `LoweredTypes` and also
improves the error message for when there is a lowered signature mismatch when
validating lifting core functions.


peterhuene:
Follow-up to the recent merge, caught the problem via unit tests in wit-component in the wit-bindgen repo, that I should have run before merging the wasm-tools PR (sorry).

alexcrichton (approved):

PR #582, opened 2022-05-11
components: Print out the item-type of export types

alexcrichton:
This commit updates `wasmprinter` to print the exports of
instance/module/etc types as `(export (component (type 0)))` instead of
the previous `(export (type 0))`. Although the binary encoding only
specifies the type the text format is much more useful with the kind of
the type in there as well and this mirrors what's done for imports.

Additionally while I'm here I turned off index printing for
imports-in-types since those indices don't matter like normal core wasm
indices do.

### DIFF
--- a/crates/wasmprinter/src/lib.rs
+++ b/crates/wasmprinter/src/lib.rs
@@ -2399,11 +2399,11 @@ impl Printer {
                     self.start_group("export ");
                     self.print_str(name)?;
                     self.result.push(' ');
-                    self.print_import_ty(&state, types, &ty, true)?;
+                    self.print_import_ty(&state, types, &ty, false)?;
                     self.end_group();
                 }
                 ModuleType::Import(import) => {
-                    self.print_import(&state, types, &import, true)?;
+                    self.print_import(&state, types, &import, false)?;
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


alexcrichton:
Actually I'm not sure that this is the change to make. @lukewagner could you perhaps weigh in on the intention here? Given a module like:

```wasm
(component 
  (type (component
    (import "" (func))
    (export "" (func))
  ))
)
```

How is this expected to be printed without the sugar? Should it be this:

```wasm
;; option A
(component 
  (type (component
    (type (func))
    (import "" (func (type 0)))
    (export "" (func (type 0)))
  ))
)
```

or this:

```wasm
;; option B
(component 
  (type (component
    (type (func))
    (import "" (type 0))
    (export "" (type 0))
  ))
)
```

Prior to this PR `wasmprinter` implements (A) for imports but (B) for exports. Additionally (B) matches the binary encoding where there is no information about what "kind" an import/export type has, only that it points to an index. I would say, however, that option (A) is more readable to a human. All that being said though I'm personally a fan of the text format following the binary format as close as possible which would imply that (B) is the desired state here. I wanted to confirm with you Luke, though, that this is intended.

lukewagner:
Sorry, this has changed in the component-model explainer a few times due to misunderstandings on my part.  As currently captured by the grammar in the core-split branch and shown in the examples at the end of the [type definitions section](https://github.com/WebAssembly/component-model/blob/core-split/design/mvp/Explainer.md#type-definitions), it's option A (which also matches core wasm).

alexcrichton:
Ok thanks for the info, @peterhuene in that case I think we should probably land something like this and when your updates to the binary format are finished the output here won't change but the implementation will be simpler because the binary format will have the "kind" as opposed to right now where side tables are maintained for that.

peterhuene (approved):

PR #583, opened 2022-05-12
wasmparser: enforce effective type size limits for validation.

### LINKED ISSUE #580: Excessive time spent validating deeply recursive component
This is a bit of a placeholder issue as it's difficult to run this for now, but I wanted to keep this on file that right now the implementation of validation in `wasmparser` for the component model can spend excessive time doing subtyping checks for deeply recursive instances such as [this test case](https://github.com/bytecodealliance/wasm-tools/blob/d659932d35c4dd79e26298ec510fd9c4944a9eaa/tests/local/component-model/very-nested.wast#L1568-L1685). I found this through my work in #529 where I was enabling that test and it ended up taking a very long time to complete in the validator.

cc @peterhuene

peterhuene:
This PR reintroduces the effective type size limit in the validator that
existed with the prior module linking proposal.

With the limit enforced, an upper bound is placed on how much work the
validator will do when performing subtype checks.

This also fixes type, import, and export limit enforcement in module,
component, and instance types.

Fixes #580.

### DIFF
--- a/crates/wasmparser/src/limits.rs
+++ b/crates/wasmparser/src/limits.rs
@@ -54,3 +54,4 @@ pub const MAX_WASM_INSTANTIATION_EXPORTS: usize = 1000;
 pub const MAX_WASM_FUNCTION_OPTIONS: usize = 10;
 pub const MAX_WASM_INSTANTIATION_ARGS: usize = 1000;
 pub const MAX_WASM_START_ARGS: usize = 1000;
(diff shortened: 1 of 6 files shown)


peterhuene:
@alexcrichton As this is difficult to test without `wast` updated, do you mind validating the failing test case with this change?

alexcrichton (approved):
Can confirm, this fixes the issue locally for me 👍

PR #591, opened 2022-05-12
wasmparser: fix instance validation.

### LINKED ISSUE #586: components: Component instances don't think they are component instances
Given this input:

```wasm
(component
  (component $c
    (component (export ""))
  )
  (instance $x (instantiate (component $c)))
  (alias export $x "" (component $f))
)
```

which is:

```
  0x0 | 00 61 73 6d | version 65546 (Component)
      | 0a 00 01 00
  0x8 | 05 18       | [component 0] inline size
    0xa | 00 61 73 6d | version 65546 (Component)
        | 0a 00 01 00
   0x12 | 05 08       | [component 0] inline size
     0x14 | 00 61 73 6d | version 65546 (Component)
          | 0a 00 01 00
   0x1c | 07 04       | component-export section
   0x1e | 01          | 1 count
   0x1f | 00 01 00    | export ComponentExport { name: "", kind: Component(0) }
 0x22 | 06 05       | instance section
 0x24 | 01          | 1 count
 0x25 | 00 01 00 00 | [instance 0] Component { index: 0, args: [] }
 0x29 | 09 05       | alias section
 0x2b | 01          | 1 count
 0x2c | 00 01 00 00 | alias [component 1] InstanceExport { kind: Component, instance: 0, name: "" }
```

this yields from validation:

```
Error: instance 0 is not a component instance (at offset 0x2c)
```

I believe this is because the type registered for the component is `TypeDef::Component` rather than `TypeDef::Instance` since the type originated from a component type.

cc @peterhuene

peterhuene:
This PR fixes instance validation so that only module and instance types
exist within the instance index space.

Previously, the instance index space in the validator might have had module,
component, or instance types. This lead to confusing validation.

To fix this, a new `ModuleInstanceType` is introduced to represent an instance
of a module. Instantiation of a module now only inserts a `ModuleInstanceType`
into the instance type space. Additionally, instantiation of a component only
inserts an `InstanceType` in the instance type space.

This makes it much easier to reason with what types might be present in the
instance index space, at the cost of having to clone the exports from a
module/component type during instantiation.

Fixes #586.


crates/wasmparser/src/validator/types.rs
```diff
@@ -392,6 +404,16 @@ impl ModuleType {
...
     }
 
+    pub(crate) fn to_instance_type(&self) -> ModuleInstanceType {
+        ModuleInstanceType {
+            type_size: self
+                .exports
+                .iter()
+                .fold(1, |acc, (_, ty)| acc + ty.type_size()),
+            exports: self.exports.clone(),
+        }
```
alexcrichton, line 414: Instead of cloning things around here, I think something that the old module linking implementation did was to basically say "this instance is an instantaition of this module" which means we'd chase a pointer in a sense but I think would avoid having to clone anything here related to the type information.

alexcrichton (commented):
Looks good to me, but I think it'd be better to have variants of the instance types which are pointers to the module/component types instead of cloning the information. That would keep the number of clones and such down and I think helps better model how it's derived from another previously found entry perhaps?

peterhuene:
That approach seems reasonable to me, I'll update this PR.

crates/wasmparser/src/validator/types.rs
peterhuene: I've removed the clone calls in favor of storing a reference (via a `TypeId`) to the underlying module or component type for the instance.

crates/wasmparser/src/validator/component.rs
```diff
@@ -670,9 +677,23 @@ impl ComponentState {
 
         let state = components.pop().unwrap();
 
-        Ok(InstanceType {
+        // Push a component type for the instance
+        let component_ty = TypeDef::Component(ComponentType {
```
alexcrichton, line 681: I think it would be better to have a new enum of `TypeDef` which is something like `TypeDef::Instance` or something like that. Otherwise I'm afraid in the future that a function will be implemented that iterates over `types` or somehow relies on the information in the `types` array and this model could mistakenly classify an instance as a component. Overall I'd prefer to future-proof this and add a separate variant for module instance and component instance types which have a pointer (as you've implemented right now) back to the component it was instantiated from.

peterhuene: Rather than expanding `TypeDef`, I added a "kind" enum to `ModuleInstanceType` and `InstanceType` that better represents how the instance type ended up in the type list.

Removed having to insert a component / module type to reference.

alexcrichton (approved):

PR #609, opened 2022-05-17
wasm-encoder: introduce an `Encode` trait implemented on all encodable types

peterhuene:
This PR refactors `wasm-encoder` such that all "encodable" types implement a simple `Encode` trait, effectively making their encoding implementations public.

Encoding of most module and component sections has been delegated to an `encode_section` function to remove the duplicated implementations throughout the crate.

Moved `ComponentImportSection` out of the `core` module where it didn't belong.

Implemented a few custom section encodings in terms of `CustomSection`.


peterhuene:
@alexcrichton @fitzgen this is just a clean-up PR for `wasm-encoder`; I'm not wed to any of these changes, so if there's anything you all disagree with, I'll happily back out.

crates/wasm-encoder/src/component.rs
```diff
@@ -104,3 +97,19 @@ impl Default for Component {
         Self::new()
     }
 }
+
+impl Encode for Component {
+    fn encode<S>(&self, sink: &mut S)
```
alexcrichton, line 102: This seems somewhat surprising to me in that I would naively expect `Encode for Component` to look the same as `Component::finish`. Or otherwise `Component` here is "top-level" so I'd expect it to produce the top-level encoding as opposed to the nested encoding. Not exactly a huge issue but wanted to point this out.

crates/wasm-encoder/src/core/code.rs
```diff
@@ -772,1547 +775,1550 @@ pub enum Instruction<'a> {
     F64x2RelaxedMax,
 }
 
-impl Instruction<'_> {
-    pub(crate) fn encode(&self, bytes: &mut Vec<u8>) {
+impl Encode for Instruction<'_> {
+    fn encode<S>(&self, sink: &mut S)
```
alexcrichton, line 779: This is somewhat preexisting but I think this refactor makes things a bit worse. I'm pretty wary of the `S` type parameter on `Encode` because it means this entire crate is monomorphized on use, meaning that this ends up increasing compile time for consumers. Additionally I feel like the usage of iterators and adapters has increased with this change which is quite a lot for LLVM to chew through. 

Overall I would personally prefer if the trait simply took `&mut Vec<u8>` (as many encoders did internally as as did this `Instruction::encode` method. While it's not super-optimal for everyone I think it strikes a better balance in terms of compile time and development overhead.

alexcrichton (commented):
Personally I think a design like this works well because you can sort of pick and choose the layer of encoding as necessary if you need to. Othrewise though I left some comments below, the main one being how I think this pushes me over the edge to stop trying to use iterators everywhere and use something concrete instead to avoid all the compile-time overhead of iterators everywhere.

crates/wasm-encoder/src/core/code.rs
peterhuene: Happy to make it `&mut Vec<u8>`.

peterhuene: I've pushed up that change.

crates/wasm-encoder/src/component.rs
peterhuene: Indeed, this might be a point of confusion. Open to suggestions as how best to make it less so.

alexcrichton: I don't particularly feel strong enough that this is the "one blocker" or anything like that. Personally I thought the previous solution of something like `struct NestedComponentSection<'a>(pub &'a Component)` would work well. It's an extra type or two for modules and components but it felt like the best way to communicate intent.

alexcrichton (approved):

crates/wasm-encoder/src/component.rs
peterhuene: As I also didn't feel too strongly about it, I've restored `ModuleSection` and `NestedComponentSection`.

PR #624, opened 2022-06-07
wasm-smith: generate shared memory when `--threads` is present

abrown:
One of the parts to the threads proposal is the addition of shared
memory. This change allows `wasm-smith` to generate Wasm code using
memory that is shared; it does not include the other parts of the
threads proposal (atomics, `wait/notify`).

### DIFF
--- a/crates/wasm-smith/src/config.rs
+++ b/crates/wasm-smith/src/config.rs
@@ -418,6 +418,18 @@ pub trait Config: 'static + std::fmt::Debug {
     fn generate_custom_sections(&self) -> bool {
         false
     }
+
+    /// Determines whether the threads proposal is enabled.
+    ///
+    /// The [threads proposal] involves shared linear memory, new atomic
+    /// instructions, and new `wait` and `notify` instructions.
+    ///
+    /// [threads proposal]: https://github.com/WebAssembly/threads/blob/master/proposals/threads/Overview.md
+    ///
+    /// Defaults to `false`.
+    fn threads_enabled(&self) -> bool {
+        false
+    }
 }
 
 /// The default configuration.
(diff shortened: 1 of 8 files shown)


crates/wasm-smith/src/core.rs
```diff
@@ -1341,6 +1342,7 @@ pub(crate) fn arbitrary_table_type(u: &mut Unstructured, config: &dyn Config) ->
 }
 
 pub(crate) fn arbitrary_memtype(u: &mut Unstructured, config: &dyn Config) -> Result<MemoryType> {
+    let threads = config.threads_enabled() && u.arbitrary()?;
```
alexcrichton, line 1345: I think the `u.arbitrary()` here can be removed since the arbitrary branch for whether or not this memory is shared is generated below in `let shared = ...`

alexcrichton (approved):
Seems reasonable to me!

crates/wasm-smith/src/core.rs
alexcrichton: For whether or not a feature is enabled we typically leave that exclusively to the `config` rather than factoring in further probability to consider it disabled. To weight the decision of making a shared memory less often I think it might be better to use the `Unstructured::ratio` method perhaps to explicitly indicate that the purpose is to weight the odds not 50/50 but something like 25/75 shared/not-shared

fitzgen (commented):
Thanks! Looks good, modulo what Alex already pointed out, and another little thing below.

crates/wasm-smith/src/config.rs
```diff
@@ -519,6 +532,7 @@ impl<'a> Arbitrary<'a> for SwarmConfig {
             max_nesting_depth: u.int_in_range(0..=10)?,
             saturating_float_to_int_enabled: u.arbitrary()?,
             sign_extension_enabled: u.arbitrary()?,
+            threads_enabled: u.arbitrary()?,
```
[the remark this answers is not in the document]
abrown: Guess I should have read the doc comments two lines down...

PR #626, opened 2022-06-08
Add PartialEq to `ExternalKind`

alexcrichton:
Used in Wasmtime so far and figured it'd be easiest to add here.


peterhuene (approved):
Should we also slap on `Eq` too?

PR #627, opened 2022-06-08
Implement module/component alias sugar when instantiating

alexcrichton:
The syntax for instantiation changed slightly in #621 where a bare index
is now required, but this syntax prevents the inline injection and usage
of an alias of an export of another component/instance. This commit
soups up the support here by bringing back the `IndexOrRef` structure
which can be used to parse one of the two, allowing both a bare
reference plus a parenthesized reference that enables alias injection.


alexcrichton:
I grepped around and we're not using the `Index<'a>` field in too many places so I think we should be good, but I think it's fine to fix these on a case-by-case basis as well.

PR #628, opened 2022-06-08
Reduce repetition in the text format of canonical options

alexcrichton:
Currently canonical options have a textual syntax that looks like:

    (memory (core memory $name))

but personally this seems a bit wordy and repetitive to me. The "core"
part is already implied by the canonical lifting and lowering, and
otherwise the "memory" part is repeated. Alternatively I think it might
be a bit easier to read/write syntax of the form:

    (memory $name)
    (realloc $realloc)

without any loss in clarity. Additionally though this would still keep
the ability to reference and instance along the lines of:

    (memory $libc "memory")


peterhuene:
I'm all for it. I agree it is needlessly verbose and there's precedent in other parts of the grammar for this syntax, e.g. instantiation of modules and components are done by index too.

peterhuene (approved):
Love it; makes the canonical options easier to use.

Good to merge once Luke approves the spec change.

alexcrichton:
Adjusted for the syntax settled on https://github.com/WebAssembly/component-model/pull/42

PR #633, opened 2022-06-09
wasmparser: improve component validation.

### LINKED ISSUE #622: components: additional validation needed in `wasmparser`
In addition to #619, the following needs to be validated for components in `wasmparser`:

* record field names should be unique.
* variant case names should be unique.
* enum names should be unique.
* flag names should be unique.
* variants must have at least one case.
* variant case refinements should validate that the index is less than the current case index.
* values must be "used" exactly once (right now we check for more than once, but don't require every value to be used).

peterhuene:
This PR improves component validation by checking:

* Record field names, variant case names, enum tag names, and flag names should
  be unique per type definition.
* Variants must have at least one case.
* Variant case refinement index must be less than the current case index.
* Every value in the component must be used at exactly once.

Due to the last bullet point above, this commit also temporarily halts
generating value imports in `wasm-smith` because there currently is no way to
use them via instantiation, start function, or export.

Fixes #622.

### DIFF
--- a/crates/wasm-smith/src/component.rs
+++ b/crates/wasm-smith/src/component.rs
@@ -628,9 +628,11 @@ impl ComponentBuilder {
             });
         }
 
-        if for_type_def || !for_import || self.total_values < self.config.max_values() {
-            choices.push(|me, u| Ok(ComponentTypeRef::Value(me.arbitrary_component_val_type(u)?)));
-        }
+        // TODO: wasm-smith needs to ensure that every arbitrary value gets used exactly once.
+        //       until that time, don't import values
+        // if for_type_def || !for_import || self.total_values < self.config.max_values() {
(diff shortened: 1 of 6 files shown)


crates/wasm-smith/src/component.rs
```diff
@@ -628,9 +628,11 @@ impl ComponentBuilder {
             });
         }
 
-        if for_type_def || !for_import || self.total_values < self.config.max_values() {
-            choices.push(|me, u| Ok(ComponentTypeRef::Value(me.arbitrary_component_val_type(u)?)));
-        }
+        // TODO: wasm-smith needs to ensure that every arbitrary value gets used exactly once.
```
peterhuene, line 631: @fitzgen here's the disabling of arbitrary value imports I pinged you about for your review.

alexcrichton (approved):

PR #635, opened 2022-06-09
Add a missing case to the incremental parse fuzzer

alexcrichton:


peterhuene:
Thanks for fixing this.

Missed it with the last few changes to that PR to split out the core type section.

PR #638, opened 2022-06-13
wasm-smith: generate all shared memories with a maximum

abrown:
According to the threads proposal specification, all shared memories must have a declared maximum. This change adds a maximum if shared memories are enabled. Also, this enables the threads proposal in the wasm-smith fuzzing to catch this kind of issue in the future.


abrown:
cc: @alexcrichton

alexcrichton (approved):

alexcrichton:
For a version bump here for wasm smith, could you run `rustc publish.rs` then `./publish bump wasm-smith:minor`? After that PR merges I'll do the `./publish publish` and get this on crates.io

PR #640, opened 2022-06-14
Implemement outer core aliases in wasm-tools.

peterhuene:
This PR updates the various tools in `wasm-tools` to support outer core aliases, based on [changes to the component model spec](https://github.com/WebAssembly/component-model/pull/44).

To support fuzzing these changes, `wasm-smith` was also updated to encode arbitrary core function types in the core type section of components.

### DIFF
--- a/crates/wast/src/component/alias.rs
+++ b/crates/wast/src/component/alias.rs
@@ -36,25 +36,64 @@ pub struct CoreAlias<'a> {
     pub target: CoreAliasTarget<'a>,
 }
 
-impl<'a> Parse<'a> for CoreAlias<'a> {
-    fn parse(parser: Parser<'a>) -> Result<Self> {
-        let span = parser.parse::<kw::core>()?.0;
-        parser.parse::<kw::alias>()?;
+impl<'a> CoreAlias<'a> {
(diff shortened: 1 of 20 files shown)


crates/wast/src/component/alias.rs
```diff
@@ -36,25 +36,64 @@ pub struct CoreAlias<'a> {
     pub target: CoreAliasTarget<'a>,
 }
 
+impl<'a> CoreAlias<'a> {
+    /// Parses only an outer type alias.
+    pub fn parse_outer_type_alias(parser: Parser<'a>) -> Result<Self> {
```
alexcrichton, line 41: I think this may be dead code at this point? If not though could this be deduplicated with the implementation of `Parse` below?

alexcrichton (approved):

peterhuene:
@fitzgen when you get a moment, do you mind looking over the [wasm-smith changes](https://github.com/bytecodealliance/wasm-tools/pull/640/commits/a3c955f937d7bb19a8e17cb754ff7699d456307c). I believe it to be pretty straight forward, but I'm loathe to step on your toes as you continue your work to rebase your changes on all this churn.

PR #643, opened 2022-06-17
wasmprinter: remove variant case identifiers.

peterhuene:
This PR removes the identifiers from variant cases; instead, a `refines`
clause references the case by index.


alexcrichton (approved):
Oops, thanks for catching this!

Issue #630, opened 2022-06-08
components: Expanded aliases should be deduplicated


alexcrichton:
Currently this component:

```wasm
(component
  (core module $m
    (global (export "f") i32 i32.const 0)
  )
  (core instance $m (instantiate $m))
  (core instance
    (export "a" (global $m "f"))
    (export "b" (global $m "f"))
  )
)
```

dumps as:

```wasm
$ cargo run -q dump foo.wat
...
 0x32 | 03 0b       | core alias section
 0x34 | 02          | 2 count
 0x35 | 03 00 00 01 | core alias [global 0] InstanceExport { kind: Global, instance_index: 0, name: "f" }
      | 66
 0x3a | 03 00 00 01 | core alias [global 1] InstanceExport { kind: Global, instance_index: 0, name: "f" }
      | 66
...
```

This should ideally only create one core alias annotation instead of two, recognizing that the alias was already added.

alexcrichton:
This is basically the same as https://github.com/bytecodealliance/wasm-tools/issues/598 except for aliases instead of types

macovedj:
Was poking around and was wondering if it would make sense to create a list of aliases that have already been encountered when parsing that could be compared against when new ones are encountered, so as to prevent them from making it into the ast.  Thought [here](https://github.com/bytecodealliance/wasm-tools/pull/637/files#diff-154ee93e092342f2f38473cbc114e5d864ebf78c4a302e5988805f4f659e45f3R666) could make sense in the binary_reader?  Probably could do it with a smaller piece of data though.

macovedj:
Cool, I reverted my commit so my draft PR closed, but I'll try and check out trying one of those options when I get a chance.

macovedj:
New draft PR exists.  I went with the first approach that you referenced above @alexcrichton.  I also don't get to work with rust all that often... so if there are obvious ways for me to be more idiomatic, would love to hear about it.

PR #644, opened 2022-06-18
check for duplicate alias before adding

macovedj:
Potential solve for [this](https://github.com/bytecodealliance/wasm-tools/issues/630)

### DIFF
--- a/crates/wasm-compose/src/encoding.rs
+++ b/crates/wasm-compose/src/encoding.rs
@@ -120,8 +120,8 @@ impl<'a> TypeEncoder<'a> {
                         let ty = self.0.type_from_id(id).unwrap().as_func_type().unwrap();
                         let index = encodable.type_count();
                         encodable.ty().function(
-                            ty.params.iter().copied().map(Self::val_type),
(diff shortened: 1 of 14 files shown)


alexcrichton:
Thanks for the PR! Checking `aliases_to_insert` isn't quite right though since that's just pending aliases to insert during the processing of the current item. Instead I think a new map needs to be introduced which is populated as aliases are visited during expansion/resolution. 

Additionally it would be good to add a test to `tests/dump/*` which ensures that given to aliases it only generates one actual alias.

macovedj:
Didn't see a clear way to move things into expansion.  Still using `aliases_to_insert` , but I'm checking it against the new `resolved_aliases`, which survives each time the prior is drained.  I think that should suffice, given the test I added has two instances.  I'll have to look into the broken tests though.

alexcrichton (commented):
Sorry but I don't have a ton of time to help write this change myself. The advice that I can offer is:

* Internally this should be using a `HashMap` instead of a vector search ideally
* This should be implemented for both core and component aliases
* This list of prior aliases to choose from should be populated whenever an `alias` item is visited in addition to injection of aliases.

PR #652, opened 2022-06-27
wasm-smith: allow generating single function modules

abrown:
This change adds `SingleFunctionModule`, an alternative to `Module` and
`ConfiguredModule` that allows the user to specify the function
signature to generate. It avoids previous issues with `Module` where
`min_funcs` and `max_funcs` would be configured as `1` but the function
would at times not be exported.

### DIFF
--- a/crates/wasm-smith/src/core.rs
+++ b/crates/wasm-smith/src/core.rs
@@ -233,6 +233,68 @@ impl<'a> Arbitrary<'a> for MaybeInvalidModule {
     }
 }
 
+pub struct SingleFunctionModule {
+    signature: FuncType,
+    module: Module,
+}
+
+impl SingleFunctionModule {
+    /// Encode this Wasm module into bytes.
+    pub fn to_bytes(&self) -> Vec<u8> {
+        self.module.to_bytes()
+    }
+
+    /// Build a Wasm module containing a single function with the following
+    /// characteristics:
+    /// - exports the function as `"generated_function"`
+    /// - the function matches the parameters and results of the given
+    ///   `signature` function type
+    /// - the module may contain other Wasm objects (e.g., memories, tables) and
+    ///   these may or may not be exported
+    pub fn build(&mut self, u: &mut Unstructured, allow_invalid: bool) -> Result<()> {
+        let ty = Type::Func(self.signature.into());
+        self.module.valtypes = configured_valtypes(&*self.module.config);
+
+        // Record the type of the single function. Since the module should only
+        // have one function, we do not import additional functions (i.e.,
+        // `arbitrary_imports()`) or other function types (i.e.,
+        // `arbitrary_types()`).
+        self.module.types.push(ty);
+        let ty_idx = self.module.types.len().try_into().unwrap();
+
+        self.module.record_type(&ty);
+        self.module.arbitrary_tags(u)?;
+
+        // Since this module will have a single function, this block replaces
+        // `arbitrary_funcs()`.
+        self.module
+            .funcs
+            .push((ty_idx, self.module.func_type(ty_idx).clone()));
+        self.module.num_defined_funcs += 1;
+
+        self.module.arbitrary_tables(u)?;
+        self.module.arbitrary_memories(u)?;
+        self.module.arbitrary_globals(u)?;
+
+        // The module will always export its single function.
+        // `arbitrary_exports()` may re-export it, but this is acceptable.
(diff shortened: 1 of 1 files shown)


crates/wasm-smith/src/core.rs
```diff
@@ -233,6 +233,68 @@ impl<'a> Arbitrary<'a> for MaybeInvalidModule {
...
+            .exports
+            .push(("generated_function".into(), ExportKind::Func, 0));
+        self.module.arbitrary_exports(u)?;
+
+        self.module.arbitrary_start(u)?;
+        self.module.arbitrary_elems(u)?;
+        self.module.arbitrary_data(u)?;
+
+        // No need to modify the generated code.
+        self.module.arbitrary_code(u, allow_invalid)?;
```
abrown, line 293: This comment is probably incorrect: I think there is a need to modify the generated code to use the function parameters and eventually emit the desired results.

crates/wasm-smith/src/core.rs
```diff
@@ -233,6 +233,68 @@ impl<'a> Arbitrary<'a> for MaybeInvalidModule {
     }
 }
 
+pub struct SingleFunctionModule {
+    signature: FuncType,
```
abrown, line 237: This `FuncType` could be generated with `arbitrary_func_type`.

crates/wasm-smith/src/core.rs
```diff
@@ -233,6 +233,68 @@ impl<'a> Arbitrary<'a> for MaybeInvalidModule {
...
+    pub fn build(&mut self, u: &mut Unstructured, allow_invalid: bool) -> Result<()> {
+        let ty = Type::Func(self.signature.into());
+        self.module.valtypes = configured_valtypes(&*self.module.config);
+
+        // Record the type of the single function. Since the module should only
+        // have one function, we do not import additional functions (i.e.,
+        // `arbitrary_imports()`) or other function types (i.e.,
+        // `arbitrary_types()`).
+        self.module.types.push(ty);
+        let ty_idx = self.module.types.len().try_into().unwrap();
```
abrown, line 263: Obviously this kind of thing messes with the meaning of `Config::min_funcs` and `Config::max_funcs`. This `SingleFunctionModule` could grow its own `Config` object but it sure would be nice to figure out a way to re-use the existing `Config`.

alexcrichton:
I like the idea of reusing `Module` where possible since it's got quite a lot internally already. If we can though I think I'd prefer to continue to lean heavily on `Config`. I'm not too too concerned about trying to rationalize all the configuration options with one another so if something falls through the cracks there that seems reasonable to me. 

I think though the main purpose of this would be to generate function signatures differently though, right? In that we'd generate arbitrary parameters but we wouldn't know the results until after the function was generated (presumably taking all of the results from the function)

abrown:
> I think though the main purpose of this would be to generate function signatures differently though, right? In that we'd generate arbitrary parameters but we wouldn't know the results until after the function was generated (presumably taking all of the results from the function)

You mentioned this in the other thread but I didn't quite understand the "why." Is it because not restricting the result types would be easier to implement or something else? It seems valuable to provide a target for the function to "head towards" but perhaps that is too difficult (I haven't really looked at the code builder changes yet).

fitzgen:
> but the function would at times not be exported.

We could also add a `export_all` method that performs a post-generation pass to export everything (or `export_all_funcs` and just export the functions)

fitzgen:
> One of my worries about the current generation of wasm-smith is that we generate a bunch of code which has all these nice values on the wasm stack, only to have them all get dropped at the end of the function because the function has a different signature. That means that differential fuzzing is pretty unlikely to actually pick up any differences in execution since the values going in/out are always canned.
> 
> I don't know how best to fix it but picking the result types later was a strawman at doing so. It's still not great though I think.

Another thing we could do is, when we have a memory, write the extra values to memory instead of dropping them.

abrown:
> That's sort of another thing which makes me think we should continue to add heuristics and config options but probably no need to make a whole new generator just yet.

Yeah, thinking about this more, a combination of [`export_everything`](https://github.com/bytecodealliance/wasm-tools/pull/657), [`allowed_instructions`](https://docs.rs/wasm-smith/0.11.1/wasm_smith/trait.Config.html#method.allowed_instructions), and whatever we come up with for https://github.com/bytecodealliance/wasm-tools/issues/655 is probably enough to generate small functions with observable results.

PR #656, opened 2022-06-28
Add a small `strip` subcommand to `wasm-tools`

alexcrichton:
I often find myself wanting this, namely to strip most sections but not
the `name` section. This differs from the behavior of the `wasm-strip`
tool from wabt so I figured I'd add it to our suite.


fitzgen (approved):
Add to the README too?

PR #657, opened 2022-06-28
wasm-smith: add configuration option to export everything

### LINKED ISSUE #654: `wasm-smith`: add method to export everything
Add a `export_everything` method to `Module` that does a post-generation pass to export everything in the module.

Various Wasmtime fuzz targets will call functions *if they are exported* or hash memories *if they are exported*.

An `export_everything` method would give us an easy way to export everything.

abrown:
Closes #654.

### DIFF
--- a/crates/wasm-smith/src/config.rs
+++ b/crates/wasm-smith/src/config.rs
@@ -129,6 +129,12 @@ pub trait Config: 'static + std::fmt::Debug {
         100
     }
 
+    /// Export all WebAssembly objects in the module. This overrides
+    /// [`Config::min_exports`] and [`Config::max_exports`]. Defaults to false.
+    fn export_everything(&self) -> bool {
+        false
+    }
+
     /// The minimum number of element segments to generate. Defaults to 0.
     fn min_element_segments(&self) -> usize {
         0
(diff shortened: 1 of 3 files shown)


abrown:
The issue mentions a post-processing pass but here it is implemented as a `Config` option; it will likely be easier to find this way.

crates/wasm-smith/src/core.rs
```diff
@@ -918,36 +918,48 @@ impl Module {
...
+        if self.config.export_everything() {
+            for choices_by_kind in choices {
+                for (kind, idx) in choices_by_kind {
+                    let name = unique_string(1_000, &mut export_names, u)?;
+                    let ty = self.type_of(kind, idx);
+                    self.type_size += 1 + ty.size();
+                    self.exports.push((name, kind, idx));
                 }
+            }
+            Ok(())
```
fitzgen, line 930: Can this be `return Ok(())` and then we can remove the `else` and the extra level of indentation below?

fitzgen (approved):
r=me with nitpick below addressed -- thanks!

abrown:
Should be merge-able...

crates/wasm-smith/src/config.rs
```diff
@@ -129,6 +129,12 @@ pub trait Config: 'static + std::fmt::Debug {
         100
     }
 
+    /// Export all WebAssembly objects in the module. This overrides
+    /// [`Config::min_exports`] and [`Config::max_exports`]. Defaults to false.
+    fn export_everything(&self) -> bool {
+        false
+    }
```
alexcrichton, line 136: Could this be added to `SwarmConfig` as well? (probably `false` during the `Arbitrary` impl though)

alexcrichton (approved):

Issue #266, opened 2021-04-22
wasm-smith: add a "no trapping" mode


fitzgen:
I think we could do this with a post-processing pass, similar to what we do with `ensure_termination`.

We'd walk over each instruction and potentially insert some code right before it:

* We would insert a couple instructions to ensure that a division instruction's denominator is never zero
* We would insert a couple instructions to mask heap addresses to ensure they are within the memory's minimum size
* Similar for `table.get` and `table.set`
* Similar for trapping floating point conversion instructions
* Every `unreachable` would be replaced with code to create dummy result values (ie zeroes) and then `br` out of the current control frame

We would also have to make sure that active data/elem segments were always in bounds of their memories/tables.

I think that's everything? I might be missing some trapping cases, but I think the approach would work for everything.

cc @alexcrichton

alexcrichton:
Done in https://github.com/bytecodealliance/wasm-tools/pull/769

PR #659, opened 2022-06-29
🪤 Non-trapping mode for wasm-smith

itsrainy:
Partially addresses https://github.com/bytecodealliance/wasm-tools/issues/266. This inserts instructions around or in place of instructions that have the potential to trap.

Completed:
- Replace `unreachable`s with dummy-value `return`s
- Return dummy results for `call_indirect` calls
- Return dummy value for out of range load operations
- Drop the value from the stack for out of range store operations
- Change the divisor from 0 to 1 for div and rem operations
- Change the divisor to 1 for signed div and rem operations that would result in 2<sup>n-1</sup>
- Handle float -> int conversion
  - Will trap for `NaN`, `Infinity`, and out of range values ([ref](https://webassembly.github.io/spec/core/exec/numerics.html#op-trunc-u))

Not completed, but punting to a later PR:
- Handle table operations
- Handle `memory.{init,copy,fill}` ([ref](https://webassembly.github.io/spec/core/exec/instructions.html#xref-syntax-instructions-syntax-instr-memory-mathsf-memory-fill))
- Vector/SIMD Instructions (i'm not sure what to do with these)


itsrainy:
putting this up for some initial review as I work on the todos

alexcrichton (commented):
This is looking awesome, thanks for this!

One thing I might recommend adding, could a fuzzer (or a runner like `crates/fuzz-stats`) be added to test out that there are no traps? I presume that at this moment in time it'd quickly find the `todo!` panics but it might be useful to have something in-tree which tests the no-trap mode and asserts that traps in fact do not come out.

crates/wasm-smith/src/notrap.rs
```diff
@@ -0,0 +1,651 @@
...
+                    // Replace `unreachable` with an early return of dummy values.
+                    //
+                    // We *could* instead abstractly interpret all these
+                    // instructions and maintain a stack of types (the way the
+                    // validation algorithm does) and insert dummy values
+                    // whenever an instruction expects a value of type `T` but
+                    // there is an `unreachable` on the stack. This would allow
+                    // us to keep executing the rest of the code following the
+                    // `unreachable`, but also is a ton more work, and it isn't
+                    // clear that it would pay for itself.
```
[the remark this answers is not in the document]
fitzgen: I think this is a question we can kick down the road to a different PR. Personally I don't think we should completely 100% disable `unreachable`, so I think this pass would have to handle it somehow anyways.

itsrainy:
> One thing I might recommend adding, could a fuzzer (or a runner like crates/fuzz-stats) be added to test out that there are no traps? I presume that at this moment in time it'd quickly find the todo! panics but it might be useful to have something in-tree which tests the no-trap mode and asserts that traps in fact do not come out.

I think this is also something I will push to a later PR if that's cool (likely once the simd and table instructions are handled, since that's the point when such a fuzzer will be informative).

crates/wasm-smith/src/core/notrap.rs
```diff
@@ -0,0 +1,826 @@
...
+                    | Instruction::V128Store64Lane { memarg: _, lane: _ } => {
+                        return Err(NotSupported { opcode: inst })
+                    }
+
+                    Instruction::MemoryCopy { src: _, dst: _ } => todo!(),
+                    Instruction::MemoryFill(_) => todo!(),
+                    Instruction::MemoryInit { mem: _, data: _ } => todo!(),
+
+                    // /*
+                    // Unsigned integer division and remainder will trap when
```
fitzgen, line 354:
```suggestion
                    // Unsigned integer division and remainder will trap when
```

crates/wasm-smith/src/core/notrap.rs
```diff
@@ -0,0 +1,826 @@
...
+                    // the divisor is 0. To avoid the trap, we will set any 0
+                    // divisors to 1 prior to the operation
+                    // The code below is equivalent to this expression:
+
+                    //     local.set $temp_divisor
+                    //     (select (i32.eqz (local.get $temp_divisor) (i32.const 1) (local.get $temp_divisor))
+                    // */
```
fitzgen, line 361:
```suggestion
                    // divisors to 1 prior to the operation.
                    //
                    // The code below is equivalent to this expression:
                    //
                    //     local.set $temp_divisor
                    //     (select (i32.eqz (local.get $temp_divisor) (i32.const 1) (local.get $temp_divisor))
```

crates/wasm-smith/src/core/notrap.rs
```diff
@@ -0,0 +1,826 @@
...
+                        // [dividend:op_type divisor:op_type]
+                        new_insts.push(inst);
+                        // [result:op_type]
+                    }
+
+                    // /*
+                    // Signed division and remainder will trap in the following instances:
+                    //     - The divisor is 0
+                    //     - The result of the division is 2^(n-1)
+                    // */
```
fitzgen, line 392:
```suggestion
                    // Signed division and remainder will trap in the following instances:
                    //     - The divisor is 0
                    //     - The result of the division is 2^(n-1)
```

crates/wasm-smith/src/core/notrap.rs
```diff
@@ -0,0 +1,826 @@
...
+                        // [dividend:op_type 1:op_type divisor:op_type]
+                        new_insts.push(Instruction::LocalGet(temp_divisor));
+                        // [dividend:op_type 1:op_type divisor:op_type divisor:op_type]
+                        new_insts.push(eqz_inst(op_type));
+                        // [dividend:op_type 1:op_type divisor:op_type is_zero:i32]
+                        new_insts.push(Instruction::Select);
+                        // [dividend:op_type divisor:op_type]
+
+                        // If dividend and divisor are -int.max and -1, replace
+                        // divisor with 1?
```
fitzgen, line 418:
```suggestion
                        // divisor with 1.
```

fitzgen (approved):
LGTM with the little comment nitpicks below, thanks!

Issue #653, opened 2022-06-28
wast: Expose offset of spans


ThePuzzlemaker:
Currently `Span` only supports `Span::new` from an offset, and `Span::linecol_in`, but no way to get the offset (the field is `pub(crate)` and there are no relevant `impl`s). This would be helpful for libraries/applications that want to use a unified span type.

ThePuzzlemaker:
Closed by #661. Forgot to add an annotation for GH to do it automatically :P :sweat_smile:

PR #661, opened 2022-07-01
Add Span::offset

ThePuzzlemaker:
See #653. I used `&self` since `Span::linecol_in`, a similar function, uses `&self` rather than `self`.

### DIFF
--- a/crates/wast/src/token.rs
+++ b/crates/wast/src/token.rs
@@ -38,6 +38,11 @@ impl Span {
         }
         (text.lines().count(), 0)
     }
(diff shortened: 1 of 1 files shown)


ThePuzzlemaker:
It might also be good to call this something like `start_offset` or similar in case anyone wants to add full span tracking to `wast`, that way it won't be backwards-incompatible or confusing. But that's a decision I'll leave to the maintainers.

alexcrichton:
Looks good to me, thanks!

PR #663, opened 2022-07-01
Add allowed_instructions to SwarmConfig  

alexcrichton:
This commit updates `SwarmConfig` to use the input to filter what
kinds/classes of instructions can be generated. This should
theoretically help explore the instruction space a bit more by having
generated modules where entire classes of instructions are disabled or
are the only ones enabled.


abrown (approved):
LGTM (yet another case where I forgot to add the `Config` option to `SwarmConfig`!)

fitzgen:
Note that some kinds of instructions won't ever really be generated without another kind enabled at the same time (e.g. can't do memory loads/stores without also having numeric instructions to generate addresses).

It is probably fine to just let the fuzzer explore the config combos till it finds interesting combinations of allowed kinds, but if for whatever reason we find that it can't do that in practice, something we can try in the future is having dependencies between instruction kinds.

PR #664, opened 2022-07-01
Fix location of recursive type parsing depth check

alexcrichton:
In #646 I accidentally put the check in the wrong function and then
forgot to check it against the original fuzz inputs.


peterhuene:
Jeez, I should have caught this too.  Sorry about that.

peterhuene (approved):

PR #677, opened 2022-07-08
wast: Update GC text format for latest changes

eqrion:
Updates the GC text format for:
  * Removal of RTT value types and fallout
  * New instructions
  * New type system (recursion groups/declared subtyping)


alexcrichton:
Looks great to me, thanks!

eqrion:
@alexcrichton Would this also be able to get published in a new version of wat/wast? I believe I'm an owner of the crate, so I can also publish if there are directions for the current procedure.

alexcrichton:
Sure! We use the `publish.rs` script in this repo nowadays and to bump just one crate it's `rustc publish.rs && ./publish bump wast:major` and it'll take care of other transitive version bumps -- https://github.com/bytecodealliance/wasm-tools/pull/680

PR #680, opened 2022-07-13
Bump wast to 44.0.0

alexcrichton:


eqrion (approved):
I just ran `./publish bump wast:major && ./publish publish` locally and it was able to publish wast/wat, but failed on wasm-tools when verifying the package:

```
   Compiling wasm-tools v1.0.9 (/Users/rhunt/src/wasm-tools/target/package/wasm-tools-1.0.9)
error[E0599]: no method named `index` found for struct `FuncValidator` in the current scope
  --> src/bin/wasm-tools/validate.rs:75:86
   |
75 |                     .with_context(|| format!("func {} failed to validate", validator.index()))
   |                                                                                      ^^^^^ method not found in `FuncValidator<ValidatorResources>`
```

I'm guessing this is #668 and indicates that wasmparser needs a bump too?

PR #683, opened 2022-07-19
`wasmparser`: make `is_subtype_of` machinery in the validator public.

peterhuene:
This PR makes the `is_subtype_of` machinery in the validator public so that
other tools can make use of it without having to reimplement the subtype rules.

### DIFF
--- a/crates/wasmparser/src/readers/component/types.rs
+++ b/crates/wasmparser/src/readers/component/types.rs
@@ -166,7 +166,8 @@ impl PrimitiveValType {
         matches!(self, Self::String)
     }
 
-    pub(crate) fn is_subtype_of(&self, other: &Self) -> bool {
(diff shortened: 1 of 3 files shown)


alexcrichton (approved):
I'm a bit wary of "simply export everything internal" in the sense that none of this was really designed per se, it's just what exists and it may not be the mot suitable interface to export. Otherwise though I don't have a better alternative and otherwise designing an interface explicitly would likely be quite nontrivial.

peterhuene:
Let me take another stab at this that at least keeps `SnapshotList` internal.

peterhuene:
I've reverted making `SnapshotList` public; as `TypesRef` is already public, this PR now only changes the visibility on the `is_subtype_of` methods at the cost of having to delegate to an internal method that uses the private types.

alexcrichton (approved):

PR #687, opened 2022-07-22
wasm-encoder: add a ConstExpr type for offsets and init expressions

nagisa:
`Element` does not support all the various constant expressions that are
valid in the given position, making it infeasible to generate some valid
webassembly modules, for example those whose elements refer to imported
globals.

For the time being I went with a plain `Instruction`, since that is what
is being used in other places that require a constant expression, most
notably offsets. However, if we wanted to keep this more constrained, I
imagine we should add an equivalent of `wasm_parser::InitExpr` and use
that instead for both elements and offsets.

### DIFF
--- a/crates/wasm-encoder/src/core/code.rs
+++ b/crates/wasm-encoder/src/core/code.rs
@@ -2738,59 +2738,61 @@ pub struct ConstExpr {
 
 impl ConstExpr {
     /// Create a new empty constant expression builder.
-    pub fn new() -> Self {
+    pub fn empty() -> Self {
         Self { bytes: Vec::new() }
     }
 
-    /// Add raw bytes to this constant expression.
-    pub fn raw(&mut self, bytes: impl IntoIterator<Item=u8>) -> &mut Self {
-        self.bytes.extend(bytes);
-        self
+    /// Create a constant expression with the specified raw encoding of instructions.
+    pub fn raw(bytes: impl IntoIterator<Item = u8>) -> Self {
+        Self {
+            bytes: bytes.into_iter().collect(),
+        }
     }
 
-    fn insn(&mut self, insn: Instruction) -> &mut Self {
-        insn.encode(&mut self.bytes);
-        self
+    fn new_insn(insn: Instruction) -> Self {
+        let mut bytes = vec![];
+        insn.encode(&mut bytes);
+        Self { bytes }
     }
 
-    /// Append a `global.get` instruction to this constant expression.
-    pub fn global_get(&mut self, index: u32) -> &mut Self {
-        self.insn(Instruction::GlobalGet(index))
+    /// Create a constant expression containing a single `global.get` instruction.
+    pub fn global_get(index: u32) -> Self {
+        Self::new_insn(Instruction::GlobalGet(index))
     }
 
-    /// Append a `ref.null` instruction to this constant expression.
-    pub fn ref_null(&mut self, ty: ValType) -> &mut Self {
-        self.insn(Instruction::RefNull(ty))
+    /// Create a constant expression containing a single `ref.null` instruction.
+    pub fn ref_null(ty: ValType) -> Self {
+        Self::new_insn(Instruction::RefNull(ty))
     }
 
-    /// Append a `ref.func` instruction to this constant expression.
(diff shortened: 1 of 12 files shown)


nagisa:
I see the failure, will fix it shortly.

alexcrichton (commented):
I think this is more general but I'm not sure that the representation here is the right one as well. Previously the possible set of elements was discretely enumerated but now with just a blind list of instructions it assumes that each element is represented by a singular instruction. While that's true today there's already extensions to constant expressions for data/element offsets which will allow multiple expressions.

I think this may want to add another variant to `Element` for imported globals or otherwise perhaps change the API a bit to have something like an `Elements` builder or something that's otherwise opaque around the internal representation to allow for incrementally building up the bytes where there could be some sort of function `Elements::element(&mut self, instr: &Instruction<'_>)` which would more clearly signify that it's one-instruction-per-element currently.

nagisa:
Hm, do you think it would make sense to pursue a `ConstExpr` type? I would be happy to work on it, and I think it would be quite beneficial in some other places too (offsets most notably).

alexcrichton:
Yeah that also seems reasonable to me since `wasm-encoder` will otherwise have to change to support the [extended-const proposal](https://github.com/WebAssembly/extended-const)

nagisa:
I made a push of a work-in-progress change. Notably missing are adjustments to all the places where constant expressions are expected (current approach only adjusts the `elem`s), but the `ConstExpr` type is sort-of complete.

I’m also thinking about some more drastic changes to the `Instruction` which would help to simplify the `ConstExpr` et al. but that might end up becoming a separate PR altogether in order to avoid blocking this PR on potentially contentious but optional changes.

crates/wasm-encoder/src/core/code.rs
```diff
@@ -2728,6 +2728,79 @@ impl Encode for Instruction<'_> {
...
+        Self { bytes: Vec::new() }
+    }
+
+    /// Add raw bytes to this constant expression.
+    pub fn raw(&mut self, bytes: impl IntoIterator<Item=u8>) -> &mut Self {
+        self.bytes.extend(bytes);
+        self
+    }
+
+    fn insn(&mut self, insn: Instruction) -> &mut Self {
```
alexcrichton, line 2751: This seems reasonable to me to also make `pub` in case anyone wants to put custom instructions/etc for one reason or another (although I like the concrete methods as well too)

alexcrichton (commented):
Seems reasonable to me!

crates/wasm-encoder/src/core/code.rs
nagisa: Yeah, my goal here was to ensure the builder only really allows const operations, and `raw` is already there as a backdoor for people to use.

crates/wasm-smith/src/core.rs
```diff
@@ -328,10 +328,16 @@ enum DataSegmentKind {
...
-        offset: Instruction,
+        offset: ConstExpr,
     },
 }
 
+#[derive(Debug)]
+pub(crate) enum GlobalInitExpr {
+    FuncRef(u32),
+    ConstExpr(ConstExpr),
+}
```
nagisa, line 339: I don’t love this, but there was code out there that was matching on `Instruction::FuncRef` before and adding a type like this seemed like the path of least resistance.

nagisa (commented):
I adjusted `ConstExpr`’s methods to be constructors that populate it with a single instruction for the time being, since it led to less convoluted code where `Instruction` was being used already. In the future adding something like `ConstExpr::append(Instruction)` can make sense but I would rather that sort of change be separate from this PR. And either way it seems to me like modifying the API for this wouldn’t be a huge deal down the line either way.

crates/wasm-encoder/src/core/code.rs
nagisa: I left this as private for the time being. It does seem like expanding this type is going to be inevitable, so I feel like finding use-cases first makes sense.

alexcrichton (commented):
Seems reasonable enough to me, thanks! If anyone feels so inspired future cleanups I think are reasonable as well.

PR #689, opened 2022-07-22
wasm-encoder: support already-encoded instructions

nagisa:
wasm-encoder already has a budding support for splating raw bytes into a
section of a module as raw bytes. Right now it seems quite supported on
a fairly ad-hoc basis, but is extremely useful when transformations are
being applied.

One particularly high-value addition to this support that I found is
with `Instructions` – there are a ton of these and, when modifying a
module, most of these will be copied over, with only a couple of
interesting instructions being inspected and/or modified.

Right now this requires effectively a HUGE match statement spanning
hundreds of lines to implement. `Instruction::Raw` would provide
a reasonable escape hatch here, much like e.g. `RawSection` and such do.

### DIFF
--- a/crates/wasm-encoder/src/core/code.rs
+++ b/crates/wasm-encoder/src/core/code.rs
@@ -303,6 +303,9 @@ impl Encode for BlockType {
 #[non_exhaustive]
 #[allow(missing_docs, non_camel_case_types)]
 pub enum Instruction<'a> {
+    // Raw encoding of an instruction.
+    Raw(Cow<'a, [u8]>),
+
     // Control instructions.
     Unreachable,
     Nop,
@@ -783,6 +786,8 @@ pub enum Instruction<'a> {
 impl Encode for Instruction<'_> {
     fn encode(&self, sink: &mut Vec<u8>) {
         match *self {
+            Instruction::Raw(ref encoding) => sink.extend(&**encoding),
+
             // Control instructions.
             Instruction::Unreachable => sink.push(0x00),
             Instruction::Nop => sink.push(0x01),


alexcrichton:
Naively I would expect that this is a bit too granular of a basis to accept raw bytes, but I also can't quite envision what you're describing with a modification pass. Could you perhaps elaborate a bit more?

I do agree that anything which takes instructions today should also be able to take raw bytes, but I would currently assume that would be better done with dedicated methods rather than adding a new variant.

nagisa:
Aha, a wild XY problem appears! Fair.

So, I’m parsing a module with `wasmparse`, modifying certain parts of it and writing it out (through `wasm-encoder`). While there are crates for this (such as `walrus`), I have found that they do not fit my requirements. So I’m rolling my own and a large chunk of that ended up quite similar to [wasm-mutate::mutators::translate::op](https://github.com/bytecodealliance/wasm-tools/blob/ed454cd940dfe045eae1974b0eb858aa5ecb0f15/crates/wasm-mutate/src/mutators/translate.rs#L273-L921). In my case, however, I’m only interested in inspecting and modifying just a few instructions – `Call`, `ReturnCall` and such, leaving all the rest as-is, so a lot of this large match is just largely an identity conversion.

I agree that this solution is not particularly ideal, but it works in instances such as [GlobalSection::global](https://docs.rs/wasm-encoder/latest/wasm_encoder/struct.GlobalSection.html#method.global) without significant codebase-wide changes. Adding another method everywhere where an `Instruction` is accepted sounds like it would be pretty sub-par experience as well. If anything else, a solution similar to `RawSection` would probably make more sense to me API design wise.

Now, I’m happy to explore alternative solutions, though. Would a `RawSection`-like approach with a trait make sense to you?

alexcrichton:
Ah ok thanks for explaining that. I would sort of expect you had a two-finger-lookalike thing in a situation like that where one finger records the last instruction you weren't interested in and the second finger goes up to an instruction you're interested in. When you hit an interesting instruction (`Call`/`ReturnCall` in your case) you'd flush the two fingers in bulk to the output function and reset the offsets. It sounds like though with `Instruction::Raw` you're encoding one instruction at a time?

Otherwise though I think as shown in https://github.com/bytecodealliance/wasm-tools/pull/687 we probably need an actual abstraction for an `InitExpr` rather than using `Instruction` in various places. With something like that I think instructions are only used in `InitExpr` and `Function`?

nagisa:
> It sounds like though with Instruction::Raw you're encoding one instruction at a time?

Yeah, my thinking was to structure the code as such:

```rust
fn op(in: wasmparser::Operator) -> wasm_encoder::Instruction {
    match op {
        wasmparser::Operator::Call { .. } => { wasm_encoder::Instruction::Call(do_things()) }
        wasmparser::Operator::ReturnCall { .. } = > { wasm_encoder::Instruction::ReturnCall(do_things()) }
        op => { wasm_encoder::Instruction::Raw(input[op.range()].to_vec().into())
    }
}
```

and then use this wherever I need to deal with operators and instructions, but I could see a finger-based approach working out quite well too, especially if we end up changing offsets and such to use a dedicated type. I’ll work on that first and maybe return to this later.

Issue #695, opened 2022-07-30
wasmparser: Improve performance by avoiding unpredictable branches


Robbepop:
While working on `wasmi` I found that overhead from parsing and validating Wasm bytecode is quite significant using `wasmparser`.
My suspicion is that a significant part of this overhead is due to numerous unpredictable branches given the `wasmparser` API for Wasm operator parsing. It is of importance that this part of the `wasmparser` crate is really efficient given that every Wasm execution requires to parse and validate all Wasm operators which make up the greatest part of the Wasm file usually.

## Current API

Currently `wasmi` does roughly the following in order to parse, validate and translate Wasm operators to its internal `wasmi` bytecode:
```rust
    /// Parses and validates the Wasm operators and translate them to `wasmi` bytecode.
    fn translate_operators(&mut self) -> Result<()> {
        let mut reader = self.func_body.get_operators_reader()?;
        while !reader.eof() {
            let (operator, offset) = reader.read_with_offset()?;
            self.validator.op(offset, &operator)?; // #1
            self.translate_operator(operator)?; // #2
        }
        reader.ensure_end()?;
        self.validator.finish(reader.original_position())?;
        Ok(())
    }
```
This might be similar for other users of the `wasmparser` crate given its API.

## The Problem

I marked 2 lines with `// #1` and `// #2`. Those two lines are were the unpredictable branches mostly happen due to matching on the `Operator` variant.
With a different interface we could get rid of both unpredictable branches. One solution is to introduce a trait `VisitOperator` that has a visitor-like method for each `Operator` variant with its variants fields as method parameters and an `offset` field given that `read_with_offset` returns a pair of `(offset: u32, op: Operator)`.
This way we could add a new API for the [`OperatorsReader`](https://docs.rs/wasmparser/0.88.0/wasmparser/struct.OperatorsReader.html) to visit the next `Operator` given some visitor. Furthermore we could add another API to also `validate_and_visit` the next `Operator`. Internally the `OperatorsReader` knows which `Operator` to visit next and directly calling the proper method of the new `VisitOperator` trait. Interestingly the `wasmparser::Validator` could just implement the very same trait as users would do in order to be able to visit each `Operator` efficiently.

## Backwards Compat

This proposal should not break backwards compat since we are just adding a new API and do not plan to change existing ones.
I plan to research about the gains in the `wasmparser-nostd` fork but I already conducted similar optimizations elsewhere in `wasmi` and received significant gains.

## Proposal Downsides

[...]

PR #697, opened 2022-08-02
Add VisitOperator trait and impls

Robbepop:
Closes https://github.com/bytecodealliance/wasm-tools/issues/695.

@alexcrichton


Robbepop:
Initial benchmarks indicate that somehow validation improved even though the current benchmarks do not even make use of the new visitor infrastructure. Instead we are still matching on `Operator` but it seems that even this part got slightly faster with this PR.
```
validate benchmark      time:   [11.015 ms 11.055 ms 11.105 ms]                               
                        change: [-7.9227% -7.0061% -6.1734%] (p = 0.00 < 0.05)
                        Performance has improved.
  4 (4.00%) high mild
  5 (5.00%) high severe
```
Using the new visitor based validation we can see additional improvements:
```
validate benchmark      time:   [10.700 ms 10.745 ms 10.792 ms]                               
                        change: [-10.546% -9.6199% -8.7807%] (p = 0.00 < 0.05)
                        Performance has improved.
```

crates/wasmparser/src/validator/func.rs
```diff
@@ -210,3 +208,567 @@ mod tests {
...
+
+#[allow(unused_variables)]
+#[rustfmt::skip]
+impl<T> VisitOperator<usize> for FuncValidator<T>
+where
+    T: WasmModuleResources,
+{
+    type Output = Result<()>;
+
+    fn visit_nop(&mut self, offset: usize) -> Self::Output { forward!(self.visit_nop(offset)) }
```
alexcrichton, line 245: To make this a bit easier on rustc I think it might be better to not use a closure internally where each method dispatches to `visit` and then that dispatches back to a closure. Instead I think it might be better here to instead have the macro expand to what `visit` is already doing (but with fewer closures)

alexcrichton (commented):
This is looking good, thanks! One worry I have is that this is introducing two ways to do the same thing that we need to keep in sync an up-to-date. I think that the overhead here is somewhat low because it's easy to spot mistakes and such but I want to make sure that everything is at least thoroughly tested. For example I don't think there should be two different paths that read the binary encoding of an operator. The style implemented in the operator validator where `Operator` is deleted to visit methods I think makes sense. 


I unfortunately can't open `validator/operators.rs` in a browser to provide more detailed feedback:

* Could the `VisitOperator` trait have a method `visit_operator` which delegates to the appropriate `visit` method internally? This could then be used by the operator validator I believe.
* Passing around `resources: &I` seems pretty onerous. If you'd like I think we could move ownership of the resources into the operator validator instead of having it separately owned in the function validator.
* If possible I'd ideally like to remove the dependence on `Operator` if we can. Given this PR is there any reason for it to stick around?

For benchmarks the in-repository benchmark isn't really that good I think, do you have more representative numbers on a more idiomatic module perhaps? I suspect that they're still wins, but it may be bigger than 10% since the modules in this repository are all test files which have few instructions.

crates/wasmparser/src/validator/func.rs
Robbepop: Good to know that closures are heavy on `rustc`. Will do what you proposed and see if it works equally well.

crates/wasmparser/src/binary_reader.rs
```diff
@@ -1553,6 +1553,726 @@ impl<'a> BinaryReader<'a> {
...
+    /// the `Operator`.
+    #[rustfmt::skip]
+    pub fn visit_operator<T>(&mut self, visitor: &mut T) -> Result<<T as VisitOperator<usize>>::Output>
+    where
+        T: VisitOperator<usize>,
+    {
+        let pos = self.original_position();
+        let code = self.read_u8()? as u8;
+        Ok(match code {
+            0x00 => visitor.visit_unreachable(pos),
```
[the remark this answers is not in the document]
Robbepop: I also thought that we could get rid of the `Operator` `enum` entirely with the new design, however, I didn't want to take it so far with just this PR and wanted to keep things simple. Afterall, there can be follow-ups if we decide on a path forward.

crates/wasmparser/src/binary_reader.rs
```diff
@@ -1553,6 +1553,726 @@ impl<'a> BinaryReader<'a> {
...
     }
 
+    /// Reads the next available `Operator` and calls the respective visit method.
+    ///
+    /// # Errors
+    ///
+    /// If `BinaryReader` has less bytes remaining than required to parse
+    /// the `Operator`.
+    #[rustfmt::skip]
+    pub fn visit_operator<T>(&mut self, visitor: &mut T) -> Result<<T as VisitOperator<usize>>::Output>
```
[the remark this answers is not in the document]
Robbepop: I actually have a personal use case where I would not like to return a `Result` for any of the visitation methods. One idea is to have a `where E: From<BinaryReadyError>` bound in order to flatten the `Result`. However, the current design is more flexible since it allows the user to take control over this behavior. The user can still perform this flattening. In benchmarks I tried to flatten but it actually was bad for the performance so I reverted it.

crates/wasmparser/src/binary_reader.rs
```diff
@@ -1553,6 +1553,726 @@ impl<'a> BinaryReader<'a> {
...
+    where
+        T: VisitOperator<usize>,
```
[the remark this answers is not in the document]
Robbepop: I made `VisitOperator` generic over this `Input` type which usually is `usize` to neatly integrate the existing validation procedures into the new framework. This was mainly done to keep the code changes as minimal and contained as possible. I also think it is a nice design to have an `Input` and `Output` duality in the trait which makes it very versatile. For `wasmi` I have use cases where `Input = ()` and it is nice to have this `Input` to not have to define yet another entire data structure just to capture the common input as a user.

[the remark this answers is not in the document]
Robbepop: One thing I wanted to add is that I personally would have liked for `VisitOperator<usize>` to actually be something more along the lines of a strongly typed `VisitOperator<Offset>` which also more carries over the purpose of the internal `usize` where `Offset` is of type `struct Offset(usize)`.

Robbepop:
> This is looking good, thanks! One worry I have is that this is introducing two ways to do the same thing that we need to keep in sync an up-to-date. I think that the overhead here is somewhat low because it's easy to spot mistakes and such but I want to make sure that everything is at least thoroughly tested. For example I don't think there should be two different paths that read the binary encoding of an operator. The style implemented in the operator validator where `Operator` is deleted to visit methods I think makes sense.
> 
> I unfortunately can't open `validator/operators.rs` in a browser to provide more detailed feedback:
> 
>     * Could the `VisitOperator` trait have a method `visit_operator` which delegates to the appropriate `visit` method internally? This could then be used by the operator validator I believe.

This should totally work but kinda feels like it defeats the purpose of the `VisitOperator` trait. However, it should be fairly straight forward to provide it with a decent default implementation so that users can basically ignore it. Another potential solution is to get rid of the `Operator` `enum` entirely and do everything in terms of the new `VisitOperator` trait.

>     * Passing around `resources: &I` seems pretty onerous. If you'd like I think we could move ownership of the resources into the operator validator instead of having it separately owned in the function validator.

If we move ownership into the `OperatorValidator` then the only thing that distincts `OperatorValidator` from `FuncValidator` is an `index` field. Technically though it should work out when looking at the usage patterns of the `resources` in `FuncValidator`. Then we could make `VisitOperator<usize> impl for OperatorValidator` and put the entire error handling logic in there as well. I fear that the amount of refactoring work involved in this is going to be huge.

>     * If possible I'd ideally like to remove the dependence on `Operator` if we can. Given this PR is there any reason for it to stick around?

What do you mean by this? Remove the `Operator` `enum` entirely from the crate? If so: Yes I think we could potentially do that but then upgrading experience for current `wasmparser` users would be pretty bad. Maybe we could go a smooth transition using deprecation of the "old" techniques first and in the next release remove them? I personally don't mind a break on my end though since I would use the new APIs anyways.

> For benchmarks the in-repository benchmark isn't really that good I think, do you have more representative numbers on a more idiomatic module perhaps? I suspect that they're still wins, but it may be bigger than 10% since the modules in this repository are all test files which have few instructions.

In `wasmi` the biggest Wasm file we use for compilation and validation benchmarks is [this one here](https://github.com/robbepop/wasm_kernel/tree/6abd5b38d9775b3957087fd9b1e8cae9418602a0) compiled to Wasm.

crates/wasmparser/src/binary_reader.rs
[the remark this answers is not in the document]
Robbepop: Generally we could remove the generic parameter of the `VisitOperator` trait together with the `offset: usize` parameter but that would decrease the versatility of the entire trait approach and would cause a huge amount of boilerplate code. By introducing this generic `Input` type we spare us a lot of this boilerplate code and make the trait very versatile for users of the `wasmparser` crate as well as the `wasmparser` crate itself.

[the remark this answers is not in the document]
alexcrichton: I'm not sure I understand what you mean by removing this would create a huge amount of boilerplate. My point is that each method is always invoked with a `usize` offset and it's always available from the binary decoder. I'd ideally prefer to not try to design a trait that's super general for all purposes and rather keep it relatively form-fitted for `wasmparser`'s use case.

I don't mind using an `Offset` type but I think that should be in a separate PR because `usize` is used pervasively throughout `wasmparser` as an offset type, so the PR would be quite large to change everything necessary.

Also to clarify I'm not saying that the `usize` argument should be removed from the methods, just the type parameter on the trait because I'm not sure how it would vary and still be callable by this crate.

crates/wasmparser/src/binary_reader.rs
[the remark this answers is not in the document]
alexcrichton: The base layer is always returning a `Result` though because an operator may not be encoded correctly. Once there's one layer of a `Result` I would assume that each individual method returning a `Result` would be trivially inline-able by LLVM to statically determine whether that arm never returned an error or not.

I could try to help look into regressing benchmarks though if you've got those.

crates/wasmparser/src/binary_reader.rs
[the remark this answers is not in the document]
alexcrichton: I think it's ok to defer work to later, but I personally prefer to not have even a temporary state internally where large amounts of the library are duplicated. If `Operator` isn't removed yet then I think that the `Operator`-reading operation should be defined in terms of this visitor instead.

alexcrichton:
> This should totally work but kinda feels like it defeats the purpose of the VisitOperator trait

Ah this is just a convenience method if `Operator` sticks around, I don't expect high-perf cases to call it (e.g. validation). And yes this would be a default method so only one location would have to worry about defining it.

> Maybe we could go a smooth transition using deprecation of the "old" techniques first and in the next release remove them?

This crate does not provide a strong API stability guarantee, there's a whole lot of major versions and new things get added all the time. This basically isn't a factor when designing this crate. Instead we've been designing this crate as "keep it up to day with abstractions that all fit together", which is why given this trait I'm not sure we still need an `Operator` enum since it doesn't really make sense any more. If you'd prefer to leave it in that's ok because I agree it's work to remove, but I would at least want to try to remove duplication of decoding if not other duplicate pieces of functionality. Everything should still be tested in this crate at the very least.

> There benchmarks show that compiling and validating it using wasmi via wasmparser takes roughly 6ms which is a lot and also the reason why I started this investigation

Is that 6ms before this change or after this change? Are you able to see other bottlenecks in `wasmparser`?

Robbepop:
> Ah this is just a convenience method if Operator sticks around, I don't expect high-perf cases to call it (e.g. validation). And yes this would be a default method so only one location would have to worry about defining it.

Will add the `visit_operator` method then. I don't plan on removing `Operator` just now. Instead I'd like to merge the PR and do that as a follow-up.

> Is that 6ms before this change or after this change? Are you able to see other bottlenecks in `wasmparser`?

In `wasmi` we unfortunately cannot use `wasmparser` crate direct but have to use a fork `wasmparser-nostd` for which I have not yet implemented this feature. Waiting for upstream to have it which is why I am behind this feature. ^^

I do see some bottlenecks with regards to small allocations but nothing I have benchmarked in detail, yet. Your concern about the double parsing is something where I have absolute no idea how we could improve the situation tbh.

crates/wasmparser/src/binary_reader.rs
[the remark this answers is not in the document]
Robbepop: I will see what I can do to remove the generic parameter of the `VisitOperator` trait with non-pervasive changes to `FuncValidator` and `OperatorValidator`.

crates/wasmparser/src/binary_reader.rs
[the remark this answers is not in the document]
Robbepop: In the common `wasmparser` benchmarks I saw the following differences:

- Current Method: roughly 10% improvements compared to `master`
- Flattening `Result`: just roughly 5% improvements compared to `master`

I expected both versions to have equal performance but this measured slow-down kinda defeats the purpose of this optimization. The experimental flattening was implemented on the `VisitOperator` impl for `FuncValidator`.

---

Could we maybe experiment with flattening of `Result`s in a follow-up PR? I feel like this PR is already kinda pumped and would like to chunk up work into smaller pieces.

crates/wasmparser/src/binary_reader.rs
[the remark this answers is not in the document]
Robbepop: I will implement the `Operator`-reading operation in terms of the new `VisitOperator`. Although this is going to require yet another big `VisitOperator` impl.

Robbepop:
@alexcrichton I moved the `WasmModuleResources` into the `OperatorValidator`. This cleaned up a lot of code in there and I am quite happy with the results. Unfortunately, the [`ModuleState::check_init_expr`](https://github.com/Robbepop/wasm-tools/blob/rf-visit-operator/crates/wasmparser/src/validator/core.rs#L321) method is unhappy about these changes and causes a borrow check error. I have not yet been able to fully resolve it since I am missing some information about the underlying behavior there. Do you have an idea how we should resolve this best?

Robbepop:
I have implemented 2 different approaches to resolve the borrow checking issue mentioned [here](https://github.com/bytecodealliance/wasm-tools/pull/697#issuecomment-1205802898).

1. Implemented an `OperatorValidtorInner` which has no `resources` field and allow it to be temporarily transformed into a full `OperatorValidator` given some generic `resources: impl WasmModuleResources`. While this temporary conversion is costly it is only needed for validating const expressions. Therefore common validation of operators is still fast.
2. Implemented `OperatorValidator` with a `VisitOperator<(usize, T)>` pair where `T: WasmModuleResources`. While this makes the fowarding of parameters even worse than before this change it was possible to clean up some error handling parts with it and also it resolves the borrow checking issue.

The big issue with both approaches: Benchmarks show that both refactorings result in validation performance regressions from 10% down to 5% improvements which is significant.

crates/wasmparser/src/validator/operators.rs
```diff
@@ -23,75 +23,184 @@
...
+/// The Wasm control list and type stack.
+pub(crate) struct Stack {
+    /// This is a list of flags for wasm features which are used to gate various
+    /// instructions.
+    features: WasmFeatures,
+    /// The `control` list is the list of blocks that we're currently in.
+    control: Vec<Frame>,
+    /// The `operands` is the current type stack.
+    operands: Vec<Option<ValType>>,
+}
```
alexcrichton, line 56: How come this type was split out from the operators validator?

crates/wasmparser/src/validator/operators.rs
```diff
@@ -310,1885 +380,3521 @@ impl OperatorValidator {
...
+                func_type_at(resources, idx, offset)?;
                 Ok(())
             }
         }
     }
 
     /// Validates a `call` instruction, ensuring that the function index is
     /// in-bounds and the right types are on the stack to call the function.
-    fn check_call(
+    fn check_call<T>(
```
alexcrichton, line 601: How come the usage of `impl WasmModuleResources` changed to `<T>` on many methods?

crates/wasmparser/src/validator/operators.rs
```diff
@@ -310,1885 +380,3521 @@ impl OperatorValidator {
...
+        T: WasmModuleResources,
+    {
+        self.check_simd_enabled(offset)?;
+        let idx = self.check_memarg(memarg, 3, offset, resources)?;
+        self.stack.pop_operand(offset, Some(idx))?;
+        self.stack.push_operand(offset, ValType::V128)?;
+        Ok(())
+    }
+
+    #[rustfmt::skip]
```
alexcrichton, line 915: How come this is here?

crates/wasmparser/src/validator/operators.rs
```diff
@@ -310,1885 +380,3521 @@ impl OperatorValidator {
...
+    pub fn process_operator<T>(
```
alexcrichton, line 916: Is this still necessary vs having callers call `visit_operator`?

crates/wasmparser/src/readers/core/operators.rs
```diff
@@ -1176,3 +1187,1280 @@ impl<'a> Iterator for OperatorsIteratorWithOffsets<'a> {
...
+
+    /// Visits the [`Operator`] `op` using the given `input`.
+    ///
+    /// # Note
+    ///
+    /// This is a convenience method that is intended for non-performance
+    /// critical use cases. For performance critical implementations users
+    /// are recommended to directly use the respective `visit` methods or
+    /// implement [`VisitOperator`] on their own.
+    #[inline(always)]
```
alexcrichton, line 1205: This is such a big method I think it's best to leave this to LLVM, not force it here.

crates/wasmparser/src/readers/core/operators.rs
```diff
@@ -1176,3 +1187,1280 @@ impl<'a> Iterator for OperatorsIteratorWithOffsets<'a> {
...
+    #[rustfmt::skip]
```
alexcrichton, line 1206: This PR commonly uses `rustfmt::skip`, but could that be removed? It may make code a bit longer but I think having consistent styling is more important myself.

crates/wasmparser/src/readers/core/operators.rs
```diff
@@ -1176,3 +1187,1280 @@ impl<'a> Iterator for OperatorsIteratorWithOffsets<'a> {
         Some(result)
     }
 }
+
+/// Trait implemented by types that can visit all [`Operator`] variants.
+#[allow(missing_docs)]
+pub trait VisitOperator<'a, Input> {
```
alexcrichton, line 1193: Personally I still don't understand why there's an `Input` type parameter here. The source, `BinaryReader`, can only pass `usize` here. The other usage of `(usize, &T)` in the validator could instead be implemented with something like `impl VisitOperator for SomeTemporary<'a> { ... }` where the `&T` lives in the `SomeTemporary<'_>` or something like that.

I think you mentioned that there may be a perf difference? Can you link to two commits which differ just with this that I could help take a look?

crates/wasmparser/src/readers/core/operators.rs
```diff
@@ -1176,3 +1187,1280 @@ impl<'a> Iterator for OperatorsIteratorWithOffsets<'a> {
...
+            Operator::I32x4LaneSelect => self.visit_i32x4_lane_select(input),
+            Operator::I64x2LaneSelect => self.visit_i64x2_lane_select(input),
+            Operator::F32x4RelaxedMin => self.visit_f32x4_relaxed_min(input),
+            Operator::F32x4RelaxedMax => self.visit_f32x4_relaxed_max(input),
+            Operator::F64x2RelaxedMin => self.visit_f64x2_relaxed_min(input),
+            Operator::F64x2RelaxedMax => self.visit_f64x2_relaxed_max(input),
+        }
+    }
+
+    fn visit_nop(&mut self, input: Input) -> Self::Output;
```
alexcrichton, line 1741: Personally I still think that this should be `Result<Self::Output>` since this primarily intended for the validator right now which already returns results (and otherwise avoids `foo??`). You mentioned there's a perf difference, though, so is there something I can look at commit-wise to help drill down?

crates/wasmparser/src/binary_reader.rs
```diff
@@ -2507,3 +2258,552 @@ impl fmt::Debug for BrTable<'_> {
         f.finish()
     }
 }
+
+/// A factory to construct [`Operator`] instances via the [`VisitOperator`] trait.
+pub struct OperatorFactory<'a> {
```
alexcrichton, line 2263: I think it's ok to avoid this being public.

crates/wasmparser/src/binary_reader.rs
```diff
@@ -2507,3 +2258,552 @@ impl fmt::Debug for BrTable<'_> {
...
+    fn visit_return_call_indirect(&mut self, _offset: usize, index: u32, table_index: u32) -> Self::Output { Operator::ReturnCallIndirect { index, table_index } }
+    fn visit_drop(&mut self, _offset: usize) -> Self::Output { Operator::Drop }
+    fn visit_select(&mut self, _offset: usize) -> Self::Output { Operator::Select }
+    fn visit_typed_select(&mut self, _offset: usize, ty: ValType) -> Self::Output { Operator::TypedSelect { ty } }
+    fn visit_local_get(&mut self, _offset: usize, local_index: u32) -> Self::Output { Operator::LocalGet { local_index } }
+    fn visit_local_set(&mut self, _offset: usize, local_index: u32) -> Self::Output { Operator::LocalSet { local_index } }
+    fn visit_local_tee(&mut self, _offset: usize, local_index: u32) -> Self::Output { Operator::LocalTee { local_index } }
+    fn visit_global_get(&mut self, _offset: usize, global_index: u32) -> Self::Output { Operator::GlobalGet { global_index } }
+    fn visit_global_set(&mut self, _offset: usize, global_index: u32) -> Self::Output { Operator::GlobalSet { global_index } }
+    fn visit_i32_load(&mut self, _offset: usize, memarg: MemoryImmediate) -> Self::Output { Operator::I32Load { memarg } }
```
alexcrichton, line 2309: One perf-related thing I might recommend looking into if you're interested is to pass larger-than-pointer-sized-values, like `MemoryImmediate` and some of the vector-related ones by pointer instead of by value. That may make some operations here more efficient (not this specific impl per se but the other impls)

alexcrichton (commented):
I left some inline comments as well below, but the main thing I'd like to resolve now is the input/output types of this new trait. I'd again prefer to not try to be super general since this is primarily just meant for parsing/validation, but I also realize that the purpose of this is performance so I don't want to compromise on that. Mainly I'd like to help dig into perf differences.

Otherwise I would again reiterate that the in-tree benchmark isn't really representative of anything. It's a data point but not really a strong one. Compile settings and/or changes in that number aren't the most important since this is not a carefully-chosen or written benchmark to be representative of what consumers might run into.

Otherwise it looks like the compile errors have been fixed here. Future refactorings can consider some different designs if necessary perhaps.

> I feel like this PR is already kinda pumped and would like to chunk up work into smaller pieces.

I definitely agree this is already large, but to some degree this is an inherently large change and it's tough to escape that. The `wasmparser` crate is a foundational dependency of many tools and it's one where I want to make sure we try really hard to get the design right. I wish there were a better way to manage large changes like this but I don't think there's an easy way out, dealing with this is going to require a relatively significant investment of work.

crates/wasmparser/src/validator/operators.rs
Robbepop: It was to resolve some borrow checking issues. Also I was under the impression that it cleaned up the code a bit actually. I often see myself splitting out grouped functionality into their own types to resolve borrow checking issues elegantly.

crates/wasmparser/src/readers/core/operators.rs
Robbepop: As weirdly as it sounds but I was able to resolve a performance regression using this `#[inline(always)]` annotation. Remove it?

crates/wasmparser/src/readers/core/operators.rs
Robbepop: The `#[rustfmt::skip]` are only used for parts of bulky code to allow for that part to be easily bulk-transformed. I commonly used this throughout the lifetime of this PR and would only want to remove those annotations once we are absolutely sure to merge it.

crates/wasmparser/src/validator/operators.rs
Robbepop: The #[rustfmt::skip] greatly help to easily bulk transforme the code. I commonly used this throughout the lifetime of this PR and would only want to remove those annotations once we are absolutely sure to merge it.

crates/wasmparser/src/validator/operators.rs
Robbepop: Good question: No. We can definitely remove it.

crates/wasmparser/src/validator/operators.rs
Robbepop: Haha, interesting catch! This is an artifact of the refactorings that went through a phase where `OperatorValidator` had ownership of `resources` and then lost it again due to aforementioned problems and then due to my own personal preferences for the `<T>` syntax this syntax evolved. We can change it back if you dislike the new syntax.

crates/wasmparser/src/readers/core/operators.rs
Robbepop: I implemented another potential solution that does not require an `Input` parameter [in this branch](https://github.com/Robbepop/wasm-tools/tree/rf-visit-operator-inner). However, I disliked the solution for 2 reasons.

1. It performed worse than the other solution and the original design.
2. It kinda felt like a hacky attempt to me whereas the other solution for which I went was just clean imo.

crates/wasmparser/src/readers/core/operators.rs
Robbepop: Here is were our misunderstanding comes from: This trait and API is not just designed for the validator. If we want to replace both unnecessary unpredictable branches we have the need to design this trait for user side as well since the 2nd unnecessary unpredictable branch is in user code. So the `VisitOperator` trait is expected to be implemented by users of the `wasmparser` trait.

For example: In `wasmi` we want to implement `VisitOperator` by the component that translates from Wasm bytecode to our internal `wasmi` bytecode. The very same component will also drive the validation via this trait.

crates/wasmparser/src/binary_reader.rs
Robbepop: True!

crates/wasmparser/src/binary_reader.rs
Robbepop: Cool! Thanks for the info. Gonna try that out and post results.

Robbepop: Cool! Thanks for the info. Gonna try that out and post results.

crates/wasmparser/src/validator/operators.rs
Robbepop: I reverted these syntactical changes.

crates/wasmparser/src/binary_reader.rs
Robbepop: I implemented it and hopefully did not forget any of the variants but benchmarks indicated no real performance difference.

Robbepop:
@alexcrichton I was able to remove the generic `Input` parameter of the `VisitOperator` trait using a new design while also keeping the implementation as efficient as before.

crates/wasmparser/src/readers/core/operators.rs
Robbepop: Resolved, since I just implemented a design that got rid of the unwanted `Input` parameter.

alexcrichton:
I'm ok deferring dealing with `Self::Output` as a `Result` to later myself. Having to do `??` isn't the end of the world. Ideally though I'd like to move to something that looks like:

```
type Ok;
type Err: From<BinaryReaderError>;
```

which I think should solve your use case?

Otherwise I have posted https://github.com/Robbepop/wasm-tools/pull/2 to simplify the operator validator (in my mind at least). The main goal of the operator validator is to be as absolutely simple and close to the spec as we can manage. This is critical for correctness so having extra layers of abstractions and distinctions like `push_ctrl` vs `push_control` can be confusing. I also took the opportunity to stop passing around `&resources` everywhere and all "meaty" methods are now just on `OperatorValidatorTemp` which has struct field access to the resources.

alexcrichton:
Er sorry and to clarify with https://github.com/Robbepop/wasm-tools/pull/2 I personally feel this is good to go and iteration can continue in-tree.

Robbepop:
> which I think should solve your use case?

Yes that would be fine for my use case.

Thank you for the review and for creating the issue which allows us to merge this PR and defer the rest of the work. I understand your concern with keeping the operator validator as simple as possible.

Robbepop:
I just benchmarked the PR with your simplifications and have to say that benchmarks drastically went down from 18% improvements to just 12% improvements. Even if you are not convinced by the `wasmparser` benchmarks this difference is quite significant unfortunately.

crates/wasmparser/src/binary_reader.rs
alexcrichton: I tried various tricks for this and was unable to get more than a 1% perf improvement.

My rough thinking is that passing `MemoryImmediate` and `V128Const` and such by value require allocating extra stack space in the visit function or otherwise more memory copies since things are passed by value. My thinking was that visiting an operator would pass in `&mut OperatorImmediates` which is allocated in some caller-specified location which means that, theoretically, `visit_operator` can work with a smaller stack frame. 

In practice it doesn't seem to help much though so I don't think it's worth pursuing at this time.

alexcrichton:
No I simply assumed that performance wouldn't change, let me investigate why it changed locally and see if I can't figure it out. I'm with you though in that this is supposed to be performant, I am not trying to compromise on the efficiency here.

Robbepop:
> No I simply assumed that performance wouldn't change, let me investigate why it changed locally and see if I can't figure it out. I'm with you though in that this is supposed to be performant, I am not trying to compromise on the efficiency here.

For best results try benchmarking using the following profile since it makes a lot of difference using this new API:
```
[profile.bench]
lto = "fat"
codegen-units = 1
```

alexcrichton:
I checked out the commit prior to mine and I added a benchmark that additionally validates the entirety of [spidermonkey.wasm](https://github.com/bytecodealliance/sightglass/blob/main/benchmarks/spidermonkey/benchmark.wasm). Before after my PR to this PR I get:

| bench | before | after |
|------|-------|----------|
| validate | 11.9ms | 12.0ms |
| spidermonkey | 81.8ms | 79.2ms |

Taking a look at the profile I wrote this trivial patch:

```diff
diff --git a/crates/wasmparser/src/binary_reader.rs b/crates/wasmparser/src/binary_reader.rs
index f54a1414..983c0057 100644
--- a/crates/wasmparser/src/binary_reader.rs
+++ b/crates/wasmparser/src/binary_reader.rs
@@ -994,12 +994,19 @@ impl<'a> BinaryReader<'a> {
     ///
     /// If `BinaryReader` has no bytes remaining.
     pub fn read_u8(&mut self) -> Result<u8> {
-        self.ensure_has_byte()?;
-        let b = self.buffer[self.position];
+        let b = match self.buffer.get(self.position) {
+            Some(b) => *b,
+            None => return self.eof_err(),
+        };
         self.position += 1;
         Ok(b)
     }

+    #[cold]
+    fn eof_err<T>(&self) -> Result<T> {
+        Err(BinaryReaderError::eof(self.original_position(), 1))
+    }
+
     /// Advances the `BinaryReader` up to four bytes to parse a variable
     /// length integer as a `u32`.
     ///
```

and got this speedup:

| bench | before | after | with patch |
|------|-------|----------|-------|
| validate | 11.9ms | 12.0ms | 11.4ms |
| spidermonkey | 81.8ms | 79.2ms | 69.0ms |

Basically I can't really reproduce the numbers you're seeing. For performance work I think it would be best to settle on a set of benchmarks to validate (maybe `spidermonkey.wasm` is all that's needed to be added?) and go from there. Additionally I do not think that benchmarking should be done with LTO enabled. While I understand that it improves performance (and I measured it to improve perf locally) it doesn't reflect what consumers of the crate always use so it can paper over "easily missing" `#[inline]` annotations for example.

Otherwise I suspect there's quite a lot of low hanging fruit on `wasmparser` to improve performance, we've never really focused all that much on perf yet. Most of the perf of Wasmtime has been runtime execution since Cranelift takes awhile to compile things. We of course appreciate perf work on `wasmparser` though!

Are you able to quantify where the slowdown you're seeing is coming from? Did I miss something perhaps? I'm also on an aarch64 machine instead of an x86_64 machine which could make things different.

Robbepop:
Let me check out your patch and play around with it.
Your previous changes looked unsuspicious to me. I have only wild guesses atm as to why the performance difference is significant.

Robbepop:
@alexcrichton I conducted benchmarks using the `wasmparser` benchmark suite, the `spidermonkey.wasm` and the 780kB `.wasm` from `wasmi` with both default and optimized `profile.bench`. The following tables summarize the results.

### `main` Branch

| Benchmark | Default Profile | Opt. Profile | Diff% |
|---|---|---|---|
| `parse` | 47.228 ms | 38.095 ms | -19.602% |
| `validate` | 88.523 ms | 87.496 ms | |

### `rf-visit-operator` Branch

| Benchmark | Default Profile | Opt. Profile | Diff% | Diff% to `main` (Default) | Diff% to `main` (Opt.) |
|---|---|---|---|---|---|
| `parse` | 51.140 ms | 40.412 ms | -20.977% | +8.2832% | +6.0806% |
| `validate` | 79.736 ms | 70.894 ms | -11.089% | -9.9262% | -18.974% |

### `rf-visit-operator` Branch (Cleanup Reverted)

| Benchmark | Default Profile | Opt. Profile | Diff% | Diff% to `main` (Default) | Diff% to `main` (Opt.) |
|---|---|---|---|---|---|
| `parse` | 51.335 ms | 40.490 ms | -21.125% | +9.3440% | +7.0290% |
| `validate` | 78.600 ms | 68.066 ms | -13.402% | -10.849% | -21.926% |

Where Opt. Profile is
```toml
[profile.bench]
lto = "fat"
codegen-units = 1
```

### Conclusions

1. The `parsing` benchmark using the default `profile.bench` regressed compared to `main` branch by roughly 8-9%. This is not okay and needs investigation. I honestly have no clue why that is.
2. There is a small but significant performance difference between the `rf-visit-operator` branch and its version with the `OperatorValidator` cleanup reverted for both profiles.
3. Using the optimized `profile.bench` has major performance improvements for the `rf-visit-operator` branches but only minor for the `main` branch. Therefore `rustc` is capable to optimize the `VisitOperator` API better.

### Remarks

While I can understand profiling for the default `bench` profile is important for most users I think it is at least as important to also profile for the profile with maximum amounts of optimizations turned on for the cases were that matters. In `wasmi` we would never compile a release without this particular profile since it provides up to 400% better performance in some cases with an average performance boost of 100%. Of course `wasmi` is probably an outlier as an interpreter but certainly not the only performance sensitive project out there.

alexcrichton:
Using an x86_64 machine I can roughly reproduce these results. In taking a deeper look all I can conclude though is that there's just so much low-hanging fruit I don't think it makes sense at this time to draw broad conclusions. For example the lto build before my patch is indeed ~8% faster than the lto build after my patch, but all the changes in functions seem to just be little layout things here and there where all the hot functions have so many better ways to be optimized than the changes I made.

I'm unable to reproduce your slowdown relative to `main`, but that probably depends a lot on what precise benchmark you're using. I don't know what `parse` corresponds to and if it's using things like `read_operator` that seems fine to regress since that's not intended to be optimized anyway.

My read on this is that this PR should go ahead and get merged and further iteration can happen in-tree. There's lots of little things I see which can be improved in the validator such as outlining error paths to cold functions to enable more inlining and such.

How do you feel about that? Basically taking the current state of this PR as a baseline and continuing to optimize from there.

Robbepop:
Let's merge and progress in-tree! :)

PR #698, opened 2022-08-03
wasmparser: rename init_expr to offset_expr

nagisa:
Some of the expressions representing an offset into a memory or table
entity are named `init_expr`. This has caused me a significant amount of
confusion multiple times now. Each time I ended up having to stop
working on my code and going to re-read the spec and the implementation
of wasmparser.

Naming these expressions `offset_expr` would make it much clearer what
they represent and put them closer to both the spec wording and wat
representation (`(offset ...)`) as well.

This is a breaking change though.

### DIFF
--- a/crates/dump/src/lib.rs
+++ b/crates/dump/src/lib.rs
@@ -161,11 +161,11 @@ impl<'a> Dump<'a> {
                         }
                         ElementKind::Active {
                             table_index,
-                            init_expr,
+                            offset_expr,
                         } => {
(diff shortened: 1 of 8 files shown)


alexcrichton:
Thanks!

If you're interested renaming `InitExpr` to something like `ConstExpr` may be worth it. I think this was all originally done for globals where it is indeed an initialization expression but it's used in different contexts as you've seen for data/element segments where "init expr" doesn't make sense (and I like the term "offset expr" there as well)

PR #699, opened 2022-08-04
wasm-tools: rename InitExpr to ConstExpr

nagisa:
This type is used in more contexts than just for the initializer of the
globals. Furthermore, the specification calls these a constant
expression, so lets harmonize the implementation in wasm-tools to use
the more generic name.

I believe the only remaining occurrences of `[Ii]nit.?[Ee]xpr` are in 
relation to globals, which is intended.


alexcrichton:
Thanks!

PR #701, opened 2022-08-05
Function references

dhil:
This patch implements (most parts of) the frontend for the function
references proposal (currently we have not implemented initialised
locals). Moreover, the crates shrink, smith, and mutate have yet to be
made compatible with the new features --- once the changeset has been
approved we can adjust those crates.

Boardly, the implementation of function references follows the
proposed spec closely, and as such, this patch introduces some
important changes to both the type structure and the validation
algorithm.

The structure of `ValType` is no longer a flat, instead, it contains a
new constructor `Ref(RefType)` which is parameterised by `RefType`,
which is a new struct type. This type is itself parameterised by a
`HeapType`, which is a new enumeration type. In addition, `ValType` is
extended with the constructor `Bot` to represent the special value
bottom type. It is special in the sense that it can only appear during
validation as there is not surface syntax for it (this is in
accordance with the spec). Concretely, the `ValType`, `RefType`, and
`HeapType` are now defined as follows.

```rust
pub enum ValType {
    /// The value type is i32.
    I32,
    /// The value type is i64.
    I64,
    /// The value type is f32.
    F32,
    /// The value type is f64.
    F64,
    /// The value type is v128.
    V128,
    /// The value type is a reference. Which type of reference is decided by
    /// RefType. This is a change in syntax from the function references proposal,
    /// which now provides FuncRef and ExternRef as sugar for the generic ref
    /// construct.
    Ref(RefType),
    /// Special bottom type.
    Bot,
}

pub struct RefType {
    /// Whether it's nullable
    pub nullable: bool,
    /// The relevant heap type
    pub heap_type: HeapType,
}

pub enum HeapType {
    /// It seems by example that u32s are directly used for arbitrary indexes,
    /// but maybe a higher-level structure like TypeRef is relevant here?
    Index(u32),
    /// From reference types
    Func,
    /// From reference types
    Extern,
    /// Special bottom heap type
    Bot,
}
```

The validation algorithm has been adapted to follow that of the
pseudocode algorithm in the appendix of the spec. We have made a few
administrative changes such as plumbing an instance of
`WasmModuleResources` around, because whenever a reference type has
heap type `Index(i)` we need to look up the defined type at index
`i`. The interface of `WasmModuleResources` has been extended with the
following new methods.

```rust
fn type_index_of_function(&self, func_idx: u32) -> Option<u32>;
fn matches(&self, t1: ValType, t2: ValType) -> bool;
fn check_value_type(
    &self,
    t: ValType,
    features: &WasmFeatures,
    offset: usize
) -> Result<(), BinaryReaderError>;
```

These methods are used during validation:

 * `type_index_of_function` is necessary to obtain the result type of `RefFunc`
 * `matches` implements the subtyping relation on types; it is used in
   both operand validation and in element validation. It may need to
   peek into context to retrieve function types.
 * `check_value_type` is also used by both operand and module validation.

The validation procedure has obviously been extended with the new
instructions; the validation of a few previous instructions has been
updated:

* `Select` computes the least upper bound of the two operand
   types.
* `RefFunc` has been changed to match its new definition.
* `TableCopy` uses subtyping (matches relation) rather than
  (structural) equality in accordance with the updated specification.
* `RefIsNull` has been adjusted to use the new `pop_ref` method (see
  below).

The validation helper methods have changed too.

```rust
fn pop_operand(
    &mut self,
    expected: Option<ValType>,
    resources: &impl WasmModuleResources,
) -> OperatorValidatorResult<ValType>

fn pop_ref(
    &mut self,
    resources: &impl WasmModuleResources,
) -> OperatorValidatorResult<RefType>
```

Note that the interface of `pop_operand` has changed: it no longer
returns an `Option<ValType>`, instead it returns simply a `ValType`
(wrapped in an `OperatorValidatorResult` as before); the new `Bot`
type subsumes all previous instances of `Unknown` which were
represented by `None`. This change makes the validation logic more
uniform. The new method `pop_ref` pops a reference type off the
operand stack; it errors if the stack is empty or there is a type
mismatch.

Comments and feedback are most welcome --- we have tried to follow the
structure of the code base to the best of our abilities, however, we
are not completely sure we have put the shared logic
(e.g. `type_index_of_function`, `matches`, and `check_value_type`) in
the right places. Also, help with getting the crates shrink, smith,
and mutate up to date would be much appreciated.

Additionally, we have another patch that makes wasmtime compile and
pass the testsuite (without function references) with our modified
wasm-tools (c.f. https://github.com/effect-handlers/wasmtime/pull/3).


alexcrichton:
Thanks for this! This'll take some time to digest but I hope to get at least an initial review of this in the next day or two. In the meantime though it looks like there's some CI failures which may need fixing up. (looks like other crates in the repo which need updates)

crates/wasmparser/src/validator/operators.rs
```diff
@@ -1409,160 +1550,175 @@ impl OperatorValidator {
...
+                        return Err(OperatorValidatorError::new(format!(
+                            "unknown function {}: function index out of bounds",
+                            function_index,
+                        )));
+                    };
                 if !resources.is_function_referenced(function_index) {
                     return Err(OperatorValidatorError::new("undeclared function reference"));
                 }
-                self.push_operand(ValType::FuncRef)?;
+                if self.features.function_references {
```
alexcrichton, line 1580: Given that the type-checking here is purely internal to this crate, I think it should be ok to remove this `if` and always push the non-nullable type, right?

crates/wasmparser/src/validator/operators.rs
```diff
@@ -777,6 +876,47 @@ impl OperatorValidator {
...
+                    _ => bail_op_err!("type mismatch: instruction requires function reference type but stack has {}", ty_to_str(ValType::Ref(rt)))
+                }
+            }
+            Operator::ReturnCallRef => {
+                self.check_function_references_enabled()?;
+                if !self.features.tail_call {
+                    return Err(OperatorValidatorError::new(
+                        "tail calls support is not enabled",
+                    ));
+                }
```
alexcrichton, line 903: Could this be extracted to a `check_tail_call_enabled()` helper function in the same manenr as `check_function_references_enabled`?

alexcrichton (commented):
Thanks for this! This is looking pretty good to me, but I'd like to push a bit more on the representation of `ValType` if possible. I'm worried about the inflation here from a 1-byte type to a 16-byte type and I'm also worried about users of this library having to work with the `Bot` type when it's otherwise just an implementation detail of the validator. I left some comments below but let me know if anything doesn't make sense.

crates/wasmparser/src/validator/operators.rs
```diff
@@ -542,20 +587,20 @@ impl OperatorValidator {
                 ));
             }
             Some(tab) => {
-                if tab.element_type != ValType::FuncRef {
+                if tab.element_type != FUNC_REF {
```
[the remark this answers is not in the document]
CosineP: Great catch, thank you.   This should be matches, not equality, according to the spec.

crates/wasmparser/src/validator/operators.rs
```diff
@@ -744,16 +789,70 @@ impl OperatorValidator {
...
                 }
                 for ty in default_types.rev() {
-                    self.pop_operand(Some(ty))?;
+                    self.pop_operand(Some(ty), resources)?;
                 }
                 self.unreachable();
             }
+            Operator::BrOnNull { relative_depth } => {
+                self.check_function_references_enabled()?;
+                let rt = self.pop_ref(resources)?;
```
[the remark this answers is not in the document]
CosineP: Good point.    The pseudo-code from the spec doesn't seem to match the inference rule to me.    Let me look into it.   I'll also check about br_on_non_null, where the prose and inference rule may be a bit at odds on a similar issue.

crates/wasmparser/src/validator/operators.rs
CosineP: `push_operand` will validate that the type is valid with respect to the enabled features, which will fail when the feature isn't enabled and the reference is non-nullable.   This could be changed, it just matches the existing pattern I noticed

crates/wasmparser/src/readers/core/types.rs
```diff
@@ -29,12 +29,54 @@ pub enum ValType {
...
-    /// The value type is an extern reference.
-    ExternRef,
+    /// The value type is a reference. Which type of reference is decided by
+    /// RefType. This is a change in syntax from the function references proposal,
+    /// which now provides FuncRef and ExternRef as sugar for the generic ref
+    /// construct.
+    Ref(RefType),
+    /// Special bottom type.
+    Bot,
 }
```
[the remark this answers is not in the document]
eqrion: If it helps, SpiderMonkey represents all of its 'type codes', ValType, FieldType, RefType, and StackType, using this packing scheme [1]. Right now it's uintptr_t, but could be uint32_t.

One subtlety is around storing a RefType inside a ValType with this packing scheme. We currently return ValType::kind() == Ref in this case, as nearly every consumer only cares that the ValType is a ref and not whether it's a funcref or externref, etc. But we also allow getting the full type information, if desired, from ValType. So the packing scheme stores the full RefType code, but is abstracted away by ValType [2].

Our design for ValType/RefType will likely change as we implement type canonicalization, but that might not be something you want to implement in wasmparser.

[1] https://searchfox.org/mozilla-central/rev/4f2984be127d2e7c788cf1848d63dca63022beec/js/src/wasm/WasmValType.h#37
[2] https://searchfox.org/mozilla-central/rev/4f2984be127d2e7c788cf1848d63dca63022beec/js/src/wasm/WasmValType.h#131

crates/wasmparser/src/validator/operators.rs
alexcrichton: Ah ok that makes sense. In general we try to avoid control flow based on enabled features though. Ideally unconditionally enabling a feature in `wasmparser` would simply be "remove all the `if !enabled { return err }` branches" as otherwise we run the possibility of having divergent behavior in the validator whether or not the feature is enabled or not. I don't think that this individual block would result in any issues, but if possible I think it would be good to allow `RefType` in the validator and only validate that user-provided types (e.g. function types, block types) don't contain `RefType` if the function-references proposal is disabled.

crates/wasmparser/src/readers/core/types.rs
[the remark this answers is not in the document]
alexcrichton: Thanks for the links and explanation! That all sounds great to me, and it makes me think that given the enhancements to the type information here we should probably go with a 32-bit packed integer for now for our type representation, and then pack on helper methods and such onto that as necessary and needed within wasmparser

alexcrichton:
As a heads up I just merged https://github.com/bytecodealliance/wasm-tools/pull/697 which is a large restructuring of the validator. I'm happy to help rebase this on top of that if y'all have issues with it.

crates/wasmparser/src/readers/core/types.rs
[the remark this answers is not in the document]
dhil: I think this is a legitimate concern, though, I'd be interested in the actual performance overhead (there's obviously a space overhead, but I don't have a good feel how much this representation affects the execution time). We're certainly open to explore alternative designs such as the scheme outlined above.

On the note of `Bot`, we also can eliminate it from the current representation by a slight refactor of the validator, where we lift all utility functions (e.g. `matches`, `check_eq_*`, etc) point-wise to work over `Option<ValType>`, where `None` is used to represent `Bot`. My current understanding of the optimisations performed by the Rust compiler is that the cost of using `Option<T>` is negligible for most cases. Also, this representation should make it simple to preserve backwards compatibility for any publicly exposed APIs too.

In the first instance, we have implemented something that follows the spec closely to provide a correct foundation for the work that we (Luna&I) are doing in wasmtime. I am happy to iterate and optimise this implementation into an acceptable shape.

crates/wasmparser/src/readers/core/types.rs
```diff
@@ -29,12 +29,54 @@ pub enum ValType {
...
-    FuncRef,
-    /// The value type is an extern reference.
-    ExternRef,
+    /// The value type is a reference. Which type of reference is decided by
+    /// RefType. This is a change in syntax from the function references proposal,
+    /// which now provides FuncRef and ExternRef as sugar for the generic ref
+    /// construct.
+    Ref(RefType),
+    /// Special bottom type.
+    Bot,
```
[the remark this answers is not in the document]
dhil: Correct. Currently, `Bot` is an artefact of validation. However, it is not entirely clear to me that `Bot` is not observable externally, as dead code will be ascribed type `Bot` -- I don't know whether such code can cause this type can leak out.

crates/wasmparser/src/validator/operators.rs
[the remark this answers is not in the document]
dhil: Yes, that is correct. We have `matches(non-nullable, nullable) = true`, meaning a non-nullable reference may be used whenever nullable reference is expected.

crates/wasmparser/src/readers/core/types.rs
```diff
@@ -29,12 +29,54 @@ pub enum ValType {
...
+    pub heap_type: HeapType,
+}
+
+/// A heap type from function references. When the proposal is disabled, Index
+/// is an invalid type.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub enum HeapType {
+    /// It seems by example that u32s are directly used for arbitrary indexes,
+    /// but maybe a higher-level structure like TypeRef is relevant here?
+    Index(u32),
```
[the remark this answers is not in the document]
dhil: Sure. I think this is an instance where we simply copied the constructor used by the wast parser. Happy to change this constructor name to `TypedFunc`.

dhil:
Thanks @alexcrichton! Regarding [your comment about #697](https://github.com/bytecodealliance/wasm-tools/pull/701#issuecomment-1211101891), we would very much appreciate some help rebasing. I don't know what the best course of action is here: do we refactor our implementation first and then rebase, or vice versa?

crates/wasmparser/src/readers/core/types.rs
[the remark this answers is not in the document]
alexcrichton: There are some relatively primitive benchmarks in this repository now which will get executed with `cargo bench` in the `crates/wasmparser` directory. Mind running those locally and seeing what the results are before/after this PR?

crates/wasmparser/src/readers/core/types.rs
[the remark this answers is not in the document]
alexcrichton: Currently the validator doesn't provide fine-grained type information like this about specific opcodes to actually see the bottom type. I think though with a `ValType(u32)` representation we could stick to not exposing it for now with the option of doing so more easily in the future if necessary (publicly exposing it that is)

alexcrichton:
I think it's up to y'all whether you'd like to refactor first or rebase. I mostly just wouldn't recommend doing them in parallel. If you'd like I can try to do the rebase later today.

crates/wasmparser/src/binary_reader.rs
```diff
@@ -1325,19 +1332,68 @@ impl<'a> BinaryReader<'a> {
...
+                Ok(HeapType::Func)
+            }
+            0x6F => {
+                self.position += 1;
+                Ok(HeapType::Extern)
+            }
+            x if 0x80 & x == 0 => {
+                self.position += 1;
+                Ok(HeapType::Index(x as u32))
+            }
```
alexcrichton, line 1348: I was reading over this again and I don't think that this binary decoding is quite right. It's actually unfortunately subtle I think what needs to happen here, but I think what you can do is call `read_var_s33` and then validate that the result is in-bounds aftewards (e.g. not negative as the binary encoding specifies).

alexcrichton: I think this basically needs to be similar in structure to `read_block_type` if you're looking to pattern-match

alexcrichton (commented):
I sent https://github.com/effect-handlers/wasm-tools/pull/14 just now as well with a recent merge. In Zulip y'all had a packed version of `ValType` which I think would be good to add here as well since I think it's quite likely we'll want that in the long run.

The changes in https://github.com/bytecodealliance/wasm-tools/pull/753 I think should help a little bit with perf since one of the issues I saw was that `Result<ValType>` was pretty big and the inlining that PR enables should help remove some of that (I'm not certain about this though)

crates/wasmparser/src/validator/core.rs
```diff
@@ -760,6 +769,102 @@ impl Module {
...
+                            let n2 = self.func_type_at(n2.into(), types, 0).unwrap();
+                            self.eq_fns(n1, n2, types)
+                        }
+                        (_, _) => false,
+                    }
+            }
+            _ => ty1 == ty2,
+        }
+    }
+    fn eq_fns(&self, f1: &impl WasmFuncType, f2: &impl WasmFuncType, types: &TypeList) -> bool {
```
alexcrichton, line 825: This isn't necessarily a concern for this PR itself but it's something I wanted to note lest I forget it. This is a recursive function which means that a number of bad things can happen:

* The stack depth here is proportional to the number of types in a program which I think can lead to stack overflow in a host.
* The "width" or "size" of a type is proportional to how long it takes to calculate `eq_valtypes`. Given the linear list of types specified in wasm it's relatively easy to create an exponentially sized type -- for example a function takes a parameter and a return value both of which are `(ref $prev_func)`, do that N times and the size is 2^N. I think that means that executing `eq_valtypes` as written can take an exponential amount of time.

These are more concerns for stabilization or the proposal itself perhaps, but it's something we won't be able to forget before declaring the function-references feature as completely done.

crates/wasmparser/src/binary_reader.rs
```diff
@@ -1329,19 +1336,74 @@ impl<'a> BinaryReader<'a> {
...
+                Some(ValType::Ref(EXTERN_REF))
+            }
+            x @ (0x6B | 0x6C) => {
+                self.position += 1;
+                Some(ValType::Ref(RefType {
+                    nullable: x == 0x6C,
+                    heap_type: self.read_heap_type()?,
+                }))
+            }
+            _ => None,
```
alexcrichton, line 1403: I think that this, and transitively `read_heap_type` aren't quite right because they need to have the fallback case of any `s33`-encoded value. To test out the exponential behavior I pointed out below I created a module with 100 types but the module failed to parse I think due to what I think is an issue in `read_heap_type` (although I'm not 100% sure).

I think it might be good to add some test cases with a type index of 256 and place that into various places (e.g. block types, params, results, `(ref $foo)`, etc) and make sure that the binary encoding/decoding works everywhere.

crates/wasmparser/src/validator/operators.rs
```diff
@@ -560,6 +592,20 @@ impl<'resources, R: WasmModuleResources> OperatorValidatorTemp<'_, 'resources, R
...
+        }
+        Ok(())
+    }
+
+    fn check_tail_call_enabled(&self, offset: usize) -> Result<()> {
+        if !self.features.tail_call {
+            bail!(offset, "tail call support is not enabled",);
+        }
+        Ok(())
+    }
```
alexcrichton, line 607: With recent refactorings it's ok to remove these  since it's handled by the `validate_proposal` macro

crates/wasmparser/src/validator/core.rs
alexcrichton: To put this in code, this input module:

```wasm
(module
  (type $t1 (func))
  (type $t2 (func
    (param (ref $t1) (ref $t1) (ref $t1) (ref $t1) (ref $t1) (ref $t1) (ref $t1) (ref $t1) (ref $t1) (ref $t1))
  ))
  (type $t3 (func
    (param (ref $t2) (ref $t2) (ref $t2) (ref $t2) (ref $t2) (ref $t2) (ref $t2) (ref $t2) (ref $t2) (ref $t2))
  ))
  (type $t4 (func
    (param (ref $t3) (ref $t3) (ref $t3) (ref $t3) (ref $t3) (ref $t3) (ref $t3) (ref $t3) (ref $t3) (ref $t3))
  ))
  (type $t5 (func
    (param (ref $t4) (ref $t4) (ref $t4) (ref $t4) (ref $t4) (ref $t4) (ref $t4) (ref $t4) (ref $t4) (ref $t4))
  ))
  (type $t6 (func
    (param (ref $t5) (ref $t5) (ref $t5) (ref $t5) (ref $t5) (ref $t5) (ref $t5) (ref $t5) (ref $t5) (ref $t5))
  ))
  (type $t7 (func
    (param (ref $t6) (ref $t6) (ref $t6) (ref $t6) (ref $t6) (ref $t6) (ref $t6) (ref $t6) (ref $t6) (ref $t6))
  ))
  (type $t8 (func
    (param (ref $t7) (ref $t7) (ref $t7) (ref $t7) (ref $t7) (ref $t7) (ref $t7) (ref $t7) (ref $t7) (ref $t7))
  ))
  (type $t9 (func
    (param (ref $t8) (ref $t8) (ref $t8) (ref $t8) (ref $t8) (ref $t8) (ref $t8) (ref $t8) (ref $t8) (ref $t8))
  ))
  (type $t10 (func
    (param (ref $t9) (ref $t9) (ref $t9) (ref $t9) (ref $t9) (ref $t9) (ref $t9) (ref $t9) (ref $t9) (ref $t9))
  ))
  (type $t11 (func
    (param (ref $t9) (ref $t9) (ref $t9) (ref $t9) (ref $t9) (ref $t9) (ref $t9) (ref $t9) (ref $t9) (ref $t9))
  ))

  (table 1 1 (ref null $t10))

  (func (param (ref null $t11))
      i32.const 0
      local.get 0
      table.set)


)
```

currently takes nigh infinite time to validate right now

dhil: For this particular example, it should be enough to short-circuit the function equivalence checking when the two indices are nominally the same, e.g.
```rust
(HeapType::TypedFunc(n1), HeapType::TypedFunc(n2)) => {
  n1 == n2 || {
  let n1 = self.func_type_at(n1.into(), types, 0).unwrap();
  let n2 = self.func_type_at(n2.into(), types, 0).unwrap();
  self.eq_fns(n1, n2, types) }
}
```
Naturally, this does not solve the general problem when `n1` and `n2` point to function types that are structurally the same. To that avail, I reckon we have to canonicalise the types, e.g. using hash-consing or similar technique.

crates/wasmparser/src/validator/core.rs
```diff
@@ -217,10 +246,19 @@ impl ModuleState {
...
+                            "a non-nullable element must come with an initialization expression",
+                            offset,
+                        ));
+                    }
+                }
+                validate_count(count)?;
                 for f in reader.into_iter_with_offsets() {
                     let (offset, f) = f?;
-                    if e.ty != ValType::FuncRef {
+                    if e.ty != RefType::FUNCREF {
```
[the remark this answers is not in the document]
dhil: The testsuite still passes after removal of those two branches.

alexcrichton:
Ok here's a summary of the remaining outstanding issues with this PR that I'm aware of:

1. The [representation of `ValType`](https://github.com/bytecodealliance/wasm-tools/pull/701#discussion_r940519444) uses some `#[repr(packed)]` tricks to be 4-bytes but this means that you can't have more than 65k function types in a module. Currently we allow up to one million types so this probably needs to be extended in the future. This notably affects the `ref.func` instruction currently where `ref.func` will not validate, even if the `function-references` proposal is disabled, if the type index is too large.
2. The performance of the validator has regressed ~15-20% locally in benchmarks. My guess is that this is largely due to the 4x size increase of `ValType` from one byte to four. I have looked and I don't know how to recover this performance easily. Some local benchmarking shows that wasmparser, as-is on `main`, is already ~2x slower than v8's validator (compared against node.js), so this isn't necessarily a showstopper.
3. It's possible to [construct a module](https://github.com/bytecodealliance/wasm-tools/pull/701#discussion_r992617443) that is valid but takes ~infinite time to validate.
4. The `HeapType::Bot` variant [is exposed in the public API](https://github.com/bytecodealliance/wasm-tools/pull/701#discussion_r940527015) despite all consumers basically only going to use `unreachable!` in that case. The current "bottom type" is encapsulated in the validator and this "heap bottom type" is not.

Personally I would consider the threshold for merging this PR to be along the lines of "if you're not using the function-references proposal you're not affected". In that sense (3) and (4) seem ok. They can always be fixed as a follow-up before turning on this proposal by default. I think (1) can be solved by having a conditional implementation of `visit_ref_func` depending on whether the `function_references` proposal is enabled or not. (I know I originally pushed for not having checks for "is function-references enabled" but the implementation limit is causing me to backtrack on that).

For (2) I don't really entirely know what to do about that. I'm not aware of any current embedding of `wasmparser` where validation performance is of the utmost importance. For example while we care about it in Wasmtime Cranelift compilation time always dwarfs the validation time by a huge amount. Along those lines, in a macro sense, the slowdown may be inevitable until someone puts in more elbow grease to optimize the validator. Our validator is certainly by no means the highest-of-performance already in that there's a lot of "we prefer simpler code" idioms throughout it.

Overall though I'd like to get a second opinion on these points and the state of the PR itself. @fitzgen would you be up for being a second pair of eyes here?

alexcrichton:
I fixed (4) in https://github.com/effect-handlers/wasm-tools/pull/38

crates/wasmparser/src/validator/core.rs
alexcrichton: Right yeah a short circuit could be added, but doubling the size of the above module, which is still quite small, could cause the short-circuit to never be hit. In that sense while a short-circuit is still good to have (as most idiomatic modules would likely hit it) we'll still need something to handle the pathological case.

Canonicalization is one idea but one we've so far avoided with a similar implementation in the component model. This problem comes up in the component model as it also has recursive types. I think that this also comes up with the gc proposal for other engines as well. It might be worthwhile to see how other engines avoid this and mirror that. The component model solution is to track the "size" of each type where it's the sum of all the component sizes, meaning that `$t10` above would be be ~10^10 which would exceed the threshold for the allowed size. The "size" is a rough proxy for "how many things must be iterated over to check compatibility".

fitzgen:
> I'm not aware of any current embedding of `wasmparser` where validation performance is of the utmost importance.

+cc @Robbepop. You've sent PRs improving `wasmparser` performance in the past, is `wasmparser` performance critical for `wasmi`?

fitzgen:
> 1. The [representation of `ValType`](https://github.com/bytecodealliance/wasm-tools/pull/701#discussion_r940519444) uses some `#[repr(packed)]` tricks to be 4-bytes but this means that you can't have more than 65k function types in a module. Currently we allow up to one million types so this probably needs to be extended in the future. This notably affects the `ref.func` instruction currently where `ref.func` will not validate, even if the `function-references` proposal is disabled, if the type index is too large.

Have we checked what implementation limits others have? If V8/SM/JSC have the same limits we have than it isn't really an issue. But we should probably at least have our limit as high as the lowest of them.

Robbepop:
> > I'm not aware of any current embedding of `wasmparser` where validation performance is of the utmost importance.
> 
> +cc @Robbepop. You've sent PRs improving `wasmparser` performance in the past, is `wasmparser` performance critical for `wasmi`?

@fitzgen In `wasmi` the parsing and validation performance is kind of critical since it makes up for a good chunk of the total runtime especially in our use case at Parity were we run Wasm smart contracts with it. Those smart contracts are usually rather small Wasm blobs (5-50 kilobytes) and execute pretty quickly due to the fact that users have to pay for every cycle of computation.

alexcrichton:
To be clear the (1) limitation is not the end of the world, if it only affects validation when `function-references` is enabled. In that case we can leave it as a "fix this before turning this on" feature. I think all that's necessary is what I [outlined above](https://github.com/bytecodealliance/wasm-tools/pull/701#issuecomment-1419334925), conditionally implementing the `ref.func` validation. Otherwise though I think it's clear that 65k isn't enough and we'll need to play more representation tricks with the type information.

@Robbepop would you happen to have benchmarks you'd be able to execute on-hand? It'd be great to see the impact of this specific PR on your embedding to see if the 20% slowdown in just validation that I measured locally is both reproducible and also to see what portion of your benchmark is slowed down as well. If not though, no worries!

Robbepop:
@alexcrichton I ran a benchmark on our CI. Of interest are the `translate_` benchmarks which show to have regressed by roughly 5% all in all with the `wasmparser` version of this PR.
https://github.com/paritytech/wasmi/pull/656#issuecomment-1421503362

fitzgen:
Looking at the representation of `ValType` (and `RefType` and `HeapType`) I think we could do a bit better.

We need at least 20 bits to represent at least 1 million type indices.

If we represent `ValType` as a `u32`, use the first byte as a discriminant, and left the other three bytes (24 bits) available to each variant as they see fit, that is more than enough for our 1 million types.

Thoughts?

alexcrichton:
I do think a representation like that would work, yeah, but I also want to avoid going too far down this rabbit hole just yet. I think @dhil and @CosineP have subsequent work based on this PR they're interested in making progress on so in the spirit of that I don't think that the 65k-limit is a blocker for this PR. I believe there's a fair amount of API design work to get a bit-packing API working since it won't be as straightforward as today's `match` used extensively throughout `wasmparser` and consumers, so separating out that change to a standalone PR I think is worthwhile.

fitzgen:
Alex and I just talked offline about this PR and its performance regression. We agreed that ultimately the regression is acceptable. It is pretty much a fact of life that `ValType` is going to have to grow in size to accommodate future Wasm proposals, not just the func refs proposal but the GC proposal as well. There certainly is room for performance improvement in the crate and its validator, but it doesn't need to hold up this PR from landing, since it isn't something we can just change about this PR; it is going to need to be wide spread and touch a lot of wasmparser to clean up idioms all over the place and shrink sizes of various things.

Robbepop:
I would very much appreciate to have a `wasmparser` relase without this PR merged and the fix for the `floats` feature so that `wasmi` can sit on it for a while until the performance regressions are hashed out eventually. (gives us time)

alexcrichton:
Ok yeah if that works for y'all @Robbepop I'm happy to do that. I'm going to try to publish new versions today and I'll do that before merging this.

For the layout of `ValType`, I should clarify that I don't know if it's 100% responsible for the 15% performance regression. I know that it's inflated from the prior 1 bytes to 4 bytes in this PR and I think that's contributing to the regression, but I don't think it completely explains it. I have been having difficulty digging in and figure out what else to optimize here.

alexcrichton (approved):
Ok now that a version of wasmparser was just published I'm going to go ahead and merge this. Thanks again for your work in implementing this @dhil and @CosineP!

PR #703, opened 2022-08-10
Update wasm-tools for the latest component model proposal changes

peterhuene:
This PR updates the various tools for the [latest changes in the component model proposal](https://github.com/WebAssembly/component-model/pull/69).

The changes include:

* The removal of the `unit` type.
* Variant case types are now optional.
* The `expected` type has been renamed to `result` and its types made optional.
* Function type encoding now supports multiple return values; function types now have either a singular unnamed parameter/result or a zero-or-more list of named parameters/results.


crates/wast/src/component/types.rs
```diff
@@ -271,6 +268,31 @@ impl<'a> Parse<'a> for PrimitiveValType {
     }
 }
 
+impl Peek for PrimitiveValType {
```
alexcrichton, line 271: Is this and `Peek for ComponentDefinedType` necessary to define? I find it difficult to keep `Peek` and `Parse` in sync so try to leave out `Peek` unless it's necessary

alexcrichton (approved):
Looks great to me! Just one minor nit but otherwise seems good to land to me, thanks for this!

This is probably not going to be too difficult to integrate into Wasmtime, but I suspect this will be significantly more effort to integrate into `wit-bindgen`.

crates/wast/src/component/types.rs
peterhuene: The parsing of `ResultType` has an optional `ComponentValType`, so this and the implementation for `ComponentDefinedType` were needed to prevent the implementation of `ResultType` from being very complex.

I realize that `Peek` has a bit of a maintenance overhead since 1) it deals with strings rather than `kw` and 2) it's easy to drift from the parse implementation.

I'm definitely open to alternative approaches, though.

peterhuene:
@fitzgen Do you want to also review the changes? There were some questionable ones I made to `wasm-smith` (namely removing `NamedType` and `OptionalNamedType`) that I'd love your input on.

crates/wast/src/component/types.rs
alexcrichton: Oh right yeah that makes sense. An nah as long as it's necessary I think it's fine to have!

fitzgen:
> @fitzgen Do you want to also review the changes?

looking now

fitzgen (approved):
`wasm-smith` changes look great!

PR #704, opened 2022-08-10
Soup up the wasmparser benchmarking slightly

alexcrichton:
Add in some modules from the sightglass benchmark and also update the
"parse" benchmark to use the new `visit_operator` method of reading
operators, the most optimal one, instead of the old `read_operator`.


crates/wasmparser/benches/benchmark.rs
```diff
@@ -230,7 +235,7 @@ fn it_works_benchmark(c: &mut Criterion) {
     let mut inputs = collect_benchmark_inputs();
     // Filter out all benchmark inputs that fail to parse via `wasmparser`.
     inputs.retain(|input| read_all_wasm(input.wasm.as_slice()).is_ok());
-    c.bench_function("it works benchmark", move |b| {
+    c.bench_function("parse all test modules", move |b| {
```
Robbepop, line 238: I think this identifier is supposed to be without white space usually so that we can filter benchmarks easily. In `wasmi` we namespace benchmarks using `/` for example: `"parse/visit"` and `"parse/read"` or so. With whitespace it becomes daunting to filter benchmark test cases via CLI.

crates/wasmparser/benches/benchmark.rs
```diff
@@ -263,14 +268,625 @@ fn validate_benchmark(c: &mut Criterion) {
...
         })
     });
+
+    for file in std::fs::read_dir("benches").unwrap() {
+        let file = file.unwrap();
+        let path = file.path();
+        if path.extension().and_then(|s| s.to_str()) != Some("wasm") {
+            continue;
+        }
+        let name = path.file_stem().unwrap().to_str().unwrap();
```
Robbepop, line 285: You may want to pack these into a helper function and move the call into the `c.bench_function` call since otherwise these are always called independent of the benchmark filters. Putting all the setup code into the `bench_function` closure help with profiling single benchmarks, for example when using `flamegraph`.

crates/wasmparser/compare-with-main.sh
```diff
@@ -1,10 +0,0 @@
-#!/usr/bin/env bash
```
Robbepop, line 1: Happy that this file is gone. I lost work multiple times since apparently it clears the git workspace.

crates/wasmparser/benches/benchmark.rs
alexcrichton: I think you can pass regexes to criterion to get it to work regardless, but I prefer `/`-separated tests anyway.

crates/wasmparser/benches/benchmark.rs
alexcrichton: I don't think that the `PathBuf` manipulation here is worth factoring out, but I used `once_cell` to factor out some other bits and pieces into more lazy computations to ensure they happen at most once and only if necessary.

Robbepop: That's a cool application of `once_cell`. Gotta do the same in `wasmi`. :)

fitzgen (approved):

PR #706, opened 2022-08-10
Optimize `local` lookup in the validator

alexcrichton:
Previously `wasmparser`'s validation would avoid creating a
`Vec<ValType>` for locals to prevent to avoid allocating too much
memory, instead creating a binary-searchable-array for local indices.
Most wasm functions, however, have very few locals and frequently access
them. This commit instead creates a linear flat array for the first N
locals and when the size of locals exceeds that count it falls back to
the old binary search method.

Locally this improve the validation time for the benchmarks in #704 by
up to 10%.

### DIFF
--- a/crates/wasmparser/src/validator/operators.rs
+++ b/crates/wasmparser/src/validator/operators.rs
@@ -44,19 +44,7 @@ macro_rules! bail_op_err {
 }
 
 pub(crate) struct OperatorValidator {
-    // The total number of locals that this function contains
-    num_locals: u32,
-    // This is a "compressed" list of locals for this function. The list of
-    // locals are represented as a list of tuples. The second element is the
-    // type of the local, and the first element is monotonically increasing as
-    // you visit elements of this list. The first element is the maximum index
-    // of the local, after the previous index, of the type specified.
-    //
-    // This allows us to do a binary search on the list for a local's index for
-    // `local.{get,set,tee}`. We do a binary search for the index desired, and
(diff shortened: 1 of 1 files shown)


crates/wasmparser/src/validator/operators.rs
```diff
@@ -71,6 +59,34 @@ pub(crate) struct OperatorValidator {
...
+#[derive(Default)]
+struct Locals {
+    // Total number of locals in the function.
+    num_locals: u32,
+
+    // The first MAX_LOCALS_TO_TRACK locals in a function. This is used to
+    // optimize the theoretically common case where most functions don't have
+    // many locals and don't need a full binary search in the entire local space
+    // below.
+    first: Vec<ValType>,
```
Robbepop, line 75: Technically, since `ValType` is a C-like enum with 1 byte in size we could stuff 24 `ValType` in an array that results in the same size as a single `Vec<ValType>` and avoid an additional indirection. Maybe worth benchmarking if this is primarily about an optimization anyways.

alexcrichton: The [`ValType` type is about to become at least 4 bytes](https://github.com/bytecodealliance/wasm-tools/pull/701) I think and from there I think a `SmallVec`-style thing might still be useful but probably less so in the context of https://github.com/bytecodealliance/wasm-tools/issues/714. In any case I mostly wanted to just land a few easy wins without going too overboard on the complexity of the validator.

peterhuene (approved):

PR #709, opened 2022-08-11
s/notrap/no_traps and publicly export NotSupported error

itsrainy:


alexcrichton (approved):
Seems reasonable to me! I think that this may need a `rustfmt` though

PR #716, opened 2022-08-15
Fix a typo in parameter orderings

alexcrichton:
This fixes a bug when using the `Operator`-based interface to the
`Validator` where the `mem` and `segment` indices were accidentally
swapped for the `memory.init` instruction. This commit goes a bit
further though and reorders method parameters to match their order of
appearance in the binary encoding, swapping the parameters in both the
`memory.init` and `memory.copy` visit functions.

### DIFF
--- a/crates/dump/src/lib.rs
+++ b/crates/dump/src/lib.rs
@@ -1371,9 +1371,9 @@ impl<'a> VisitOperator<'a> for Dump<'_> {
             memarg: MemoryImmediate,
             lane: SIMDLaneIndex,
         )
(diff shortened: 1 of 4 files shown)


fitzgen (approved):

Robbepop:
@alexcrichton What is your stance on using new type patterns for parameters with ambiguous types? This could have avoided such problems at large scale refactorings.

alexcrichton:
Yes I think it'd be good to use that if we can. I personally like where `wast` ended up where all instructions either have a payload or they don't, and instructions like the ones updated here all get uniquely typed payloads used just by that instruction.

PR #720, opened 2022-08-15
Bump wasm-smith version

itsrainy:


fitzgen (approved):

alexcrichton:
I think unfortunately this may not be sufficient since the current version of `wasm-smith` requires new publication of its `wasmparser` and `wasm-encoder` dependencies. I'm doing the whole shebang of version bumps in https://github.com/bytecodealliance/wasm-tools/pull/719 since I need to pull in the validator and such changes into Wasmtime anyway.

alexcrichton:
Ok I'm currently publishing everything with https://github.com/bytecodealliance/wasm-tools/pull/719 so this'll get to crates.io momentarily.

PR #729, opened 2022-08-17
Add a feature to `wasmprinter` to print binary offsets

alexcrichton:
I often find myself trying to debug error messages here and there
related to low-level validation. These error messages have an offset
within the original file of the offending opcode, but finding that
opcode within the text format is typically not trivial. This was the
original motivation for `wasm-tools dump` but that output is also quite
verbose and not always easy to parse. Consequently this problem
motivates this commit.

This commit adds a `-p` flag to the `wasm-tools print` subcommand which
will print, in comments at the start of each line, the binary offset of
the corresponding item in the binary. This notably prints the binary
offset of all instructions and should ideally make it much easier to use
just the text format to find the offending instruction given a
particular offset.

### DIFF
--- a/crates/wasmparser/src/readers.rs
+++ b/crates/wasmparser/src/readers.rs
@@ -55,9 +55,20 @@ pub trait SectionReader {
 }
 
 /// Implemented by sections with a limited number of items.
-pub trait SectionWithLimitedItems {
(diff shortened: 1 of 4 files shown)


fitzgen (approved):
Nice, this will be super useful

PR #733, opened 2022-08-18
wasmparser: Add a macro to define all instructions

alexcrichton:
This commit adds a new "iterator macro" to the `wasmparser` crate to
iterate over all the instructions defined by `wasmparser`. This is used
to define both the `Operator` enum as well as the `VisitOperator` trait.
Many of the various utilities in `wasmparser` as consequently greatly
simplified using this macro:

* The `Operator` enum is defined with the macro
* The default `visit_operator` method is defined with this macro
* The `VisitOperator` trait itself is defined with the macro
* The "dump" support is now defined with this macro
* The `VisitOperator for FuncValidator` trait implementation is now
  defined with this macro.
* The `VisitOperator for OperatorFactory` trait implementation is now
  defined with this macro.
* The `VisitOperator for NopVisitor` in benchmarking now uses this macro.

Overall this should help provide a convenient way to uniformly handle
all instructions or otherwise handle one or two instructions as a
special case. In thinking how to integrate the `VisitOperator` trait
into `wasmtime` one thing I want is to have a "validate then call
visit"-style of iteration which I'm planning to define with this macro
as well. Furthermore there's some more usage of `Operator` within
`wasm-tools` which I think can more easily be removed with this macro.


Robbepop (approved):
This is really cool and will remove the burden of synchronizing everything. And having it publicly exported is also super useful to user crates.

fitzgen (approved):

PR #735, opened 2022-08-18
wasmparser: Remove unnecessary `unsafe` block

fitzgen:
No need to check `self.owned` when we can just check whether the result of
`Arc::get_mut` is some or none.

We keep the `MaybeOwned<T>` type instead of replacing it with `Arc<T>`, however,
to help document invariants for when it is used.

### DIFF
--- a/crates/wasmparser/src/validator/core.rs
+++ b/crates/wasmparser/src/validator/core.rs
@@ -972,36 +972,66 @@ mod arc {
     use std::ops::Deref;
     use std::sync::Arc;
 
+    enum Inner<T> {
+        Owned(T),
+        Shared(Arc<T>),
+
+        Empty, // Only used for swapping from owned to shared.
+    }
+
     pub struct MaybeOwned<T> {
-        owned: bool,
-        arc: Arc<T>,
+        inner: Inner<T>,
     }
 
     impl<T> MaybeOwned<T> {
-        #[allow(clippy::cast_ref_to_mut)]
+        #[inline]
         fn as_mut(&mut self) -> Option<&mut T> {
-            if !self.owned {
-                return None;
+            match &mut self.inner {
+                Inner::Owned(x) => Some(x),
+                Inner::Shared(_) => None,
+                Inner::Empty => Self::unreachable(),
             }
-            debug_assert!(Arc::get_mut(&mut self.arc).is_some());
-            Some(unsafe { &mut *(Arc::as_ptr(&self.arc) as *mut T) })
         }
 
+        #[inline]
         pub fn assert_mut(&mut self) -> &mut T {
             self.as_mut().unwrap()
         }
 
         pub fn arc(&mut self) -> &Arc<T> {
-            self.owned = false;
-            &self.arc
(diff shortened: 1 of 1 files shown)


crates/wasmparser/src/validator/core.rs
```diff
@@ -973,34 +973,26 @@ mod arc {
...
 
     impl<T> MaybeOwned<T> {
-        #[allow(clippy::cast_ref_to_mut)]
         fn as_mut(&mut self) -> Option<&mut T> {
-            if !self.owned {
-                return None;
-            }
-            debug_assert!(Arc::get_mut(&mut self.arc).is_some());
-            Some(unsafe { &mut *(Arc::as_ptr(&self.arc) as *mut T) })
+            Arc::get_mut(&mut self.arc)
```
alexcrichton, line 981: I mentioned this on Zulip as well, but I think it's worth measuring the performance impact of this since the original purpose of this was to avoid atomics in release mode for `assert_mut` here. If the performance impact is negligible though it seems reasonable to get rid of the `unsafe` block.

fitzgen:
Okay, using `Arc::get_mut` did result in a ~5% hit to benchmarks. Rewrote a little bit and it seems like things are within the noise on our real world benchmarks now:

```
validate/tests          time:   [8.8355 ms 8.8622 ms 8.8910 ms]
                        change: [+2.5521% +3.0319% +3.5134%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 5 outliers among 100 measurements (5.00%)
  5 (5.00%) high mild

Benchmarking validate/spidermonkey: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 5.9s, or reduce sample count to 80.
validate/spidermonkey   time:   [58.544 ms 58.753 ms 58.974 ms]
                        change: [-1.2258% -0.6252% -0.0524%] (p = 0.03 < 0.05)
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild

validate/pulldown-cmark time:   [2.3431 ms 2.3598 ms 2.3811 ms]
                        change: [+0.2819% +1.0555% +2.0408%] (p = 0.01 < 0.05)
Found 17 outliers among 100 measurements (17.00%)
  8 (8.00%) high mild
  9 (9.00%) high severe

validate/bz2            time:   [940.90 us 942.53 us 944.31 us]
                        change: [+0.2020% +1.0266% +1.8724%] (p = 0.01 < 0.05)
Found 10 outliers among 100 measurements (10.00%)
  1 (1.00%) low mild
  2 (2.00%) high mild
  7 (7.00%) high severe

validate/intgemm-simd   time:   [2.1326 ms 2.1497 ms 2.1721 ms]
                        change: [-0.1844% +0.7127% +1.9124%] (p = 0.17 > 0.05)
                        No change in performance detected.
Found 11 outliers among 100 measurements (11.00%)
  9 (9.00%) high mild
  2 (2.00%) high severe
```

alexcrichton (approved):

Issue #732, opened 2022-08-18
`wasmparser`: Add API to access FuncValidator state


Robbepop:
## The Problem

When using the `wasmparser` crate for Wasm parsing and validation currently users have to mirror certain structures in some use cases.
Therefore this work is duplicated and could easily be deduplicated by making the `wasmparser` crate offer an API to access its internal state.
For example in `wasmi` we mirror the value stack and the way we register local variables in order to translate Wasm to `wasmi` bytecode.

The new `VisitOperator` API allows to interweave the validator into the user code very flexibly.

## Example API

- `fn len_locals(&self) -> u32;`: Returns the number of registered local variables.
- `fn len_params(&self) -> u32;`: Returns the number of paramters of the function.
- `fn len_results(&self) -> u32;`: Returns the number of results of the function.
- `fn get_local(&self, index: u32) -> Option<ValType>;`: Returns the type of the local variable at the given index.
- `fn get_value_nth_back(&self, index: u32) -> Option<ValType>;`: Returns the value type of the n-th value from the value stack from the back where an index of 0 is the last value.
- `fn len_values(&self) -> u32;`: Returns the current height of the value stack.

Combined with the new `VisitOperator` API this allows user code to rely on the `wasmparser` validator without duplicating the logic to sync the state properly.

## Alternative: Return Result

Usually accesses to the state need to be performed before validation since validation would cause the state to change. Therefore it might be useful to not return `Option<ValType>` for `get_local` and `get_value_nth_back` but instead return a validation `Result`. This means that validation was partly driven by the user code though.

## Alternative: Peek Validation

An alternative is to have a `peek_validate` API that validates the Wasm operator while not changing the validator state but returning a `ValidatorStateTransfer` "object" instead that applies the required state transitions in the validator later via another API called by the user.

## Future Work

We could do the same for the control stack but that is a bit trickier since it usually carries information specialized to its task, e.g. different information requirements for validation and translation. So for now I would not include the control stack into the initial API but we may introduce such an API at later stages.

alexcrichton:
I think this would be a great idea to add! I like the idea of adding accessors for the information already there as opposed to adding new functionality like `peek_validate`.

Robbepop:
Cool, I will prepare a PR for this then. :)

PR #738, opened 2022-08-19
`wasmparser`: Add getters for useful internal state to the FuncValidator

Robbepop:
Closes https://github.com/bytecodealliance/wasm-tools/issues/732.

This adds the following APIs:

- `fn len_locals() -> u32`
- `fn get_local_type(index) -> ValType`
- `fn operand_stack_height() -> u32` (already existed)
- `fn get_operand_type(index) -> Option<Option<ValType>>`
- `fn control_stack_height() -> u32`
- `fn get_control_frame(depth) -> Option<&Frame>`

### DIFF
--- a/crates/wasmparser/src/validator/operators.rs
+++ b/crates/wasmparser/src/validator/operators.rs
@@ -228,7 +228,9 @@ impl OperatorValidator {
     /// Returns the optional value type of the value operand at the given
     /// `index` from the top of the operand stack.
     ///
-    /// Returns `None` if the `index` is out of bounds.
+    /// - Returns `None` if the `index` is out of bounds.
+    /// - Returns `Some(None)` if there is a value with unknown type
+    /// at the given `index`.
     ///
     /// # Note
     ///
--- a/crates/wasmparser/src/validator/operators.rs
+++ b/crates/wasmparser/src/validator/operators.rs
@@ -43,7 +43,7 @@ macro_rules! bail_op_err {
 }
 
 pub(crate) struct OperatorValidator {
-    locals: Locals,
+    pub(super) locals: Locals,
 
     // This is a list of flags for wasm features which are used to gate various
     // instructions.
@@ -67,7 +67,7 @@ pub(crate) struct OperatorValidator {
 const MAX_LOCALS_TO_TRACK: usize = 50;
 
 #[derive(Default)]
-struct Locals {
+pub(super) struct Locals {
     // Total number of locals in the function.
     num_locals: u32,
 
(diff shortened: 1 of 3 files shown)


crates/wasmparser/src/validator/operators.rs
```diff
@@ -90,29 +90,53 @@ struct Locals {
...
-    // The type signature of this frame, represented as a singular return type
-    // or a type index pointing into the module's types.
-    block_type: BlockType,
-    // The index, below which, this frame cannot modify the operand stack.
-    height: usize,
-    // Whether this frame is unreachable so far.
-    unreachable: bool,
+pub struct Frame {
+    /// Indicator for what kind of instruction pushed this frame.
+    pub kind: FrameKind,
```
Robbepop, line 102: I was not sure if we want to have public fields or getter methods instead. The style of the codebase uses public fields mostly so that's what I did.

crates/wasmparser/src/validator/operators.rs
```diff
@@ -196,10 +220,40 @@ impl OperatorValidator {
...
 
+    /// Returns the optional value type of the value operand at the given
+    /// `index` from the top of the operand stack.
+    ///
+    /// Returns `None` if the `index` is out of bounds.
+    ///
+    /// # Note
+    ///
+    /// An `index` of 0 will refer to the last operand on the stack.
+    pub fn peek_operand_at(&self, index: usize) -> Option<Option<ValType>> {
```
Robbepop, line 236: I really dislike the `Option<Option<ValType>>` return type but also did not want to erase potentially important information by flattening the `Option`. What would you recommend to do here? What cases are there that push a `None` `ValType` btw?

crates/wasmparser/src/validator/operators.rs
```diff
@@ -90,29 +90,53 @@ struct Locals {
     all: Vec<(u32, ValType)>,
 }
 
+/// A Wasm control flow block on the control flow stack during Wasm validation.
+//
+// # Dev. Note
+//
 // This structure corresponds to `ctrl_frame` as specified at in the validation
 // appendix of the wasm spec
 #[derive(Debug)]
```
Robbepop, line 99: Technically I do not see a reason why `Frame` could not also derive `Copy` and `Clone`.

alexcrichton: Seems reasonable to me to add yeah

crates/wasmparser/src/validator/operators.rs
alexcrichton: Yeah I think this is fine

crates/wasmparser/src/validator/operators.rs
alexcrichton: I think this is ok for now, this isn't something we need to set in stone from day 1. It's ok to iterate on this over time and improve later. Could the documentation explain the meaning of `Some(None)` though?

crates/wasmparser/src/validator/operators.rs
```diff
@@ -196,10 +220,40 @@ impl OperatorValidator {
...
+    }
+
+    /// Returns a shared reference to the control flow [`Frame`] of the
+    /// control flow stack at the given `depth` if any.
+    ///
+    /// Returns `None` if the `depth` is out of bounds.
+    ///
+    /// # Note
+    ///
+    /// A `depth` of 0 will refer to the last frame on the stack.
```
alexcrichton, line 252: I think it's ok to skip the docs here on these functions since this `OperatorsValidator` isn't exposed publicly.

alexcrichton: In that these docs are all great to have but it's best I think to leave them on the main public methods for explanation and visibility

alexcrichton (commented):
Seems reasonable to me!

crates/wasmparser/src/validator/operators.rs
Robbepop: Hmmm we are a bit flaky here since some of those internal methods do have quite some docs and others do not. I can remove the docs since you asked for that but now `control_stack_height` method has docs while `get_frame` has no docs ... :upside_down_face:

Robbepop:
@alexcrichton applied all your review suggestions

crates/wasmparser/src/validator/operators.rs
alexcrichton: Sorry I don't mean to say that we shouldn't have documentation, even for internals that don't show up in `rustdoc`. Moreso I prefer to avoid duplicate documentation where possible

alexcrichton (approved):

PR #739, opened 2022-08-19
`wasmparser`: Refactor `FuncType`

Robbepop:
This refactors the `FuncType` type in the `wasmparser` crate for the following benefits:

- Instead of accessing the bare fields `FuncType` now has a proper API which helped in cleaning up parts of the codebase. Also this may help in future refactors of the crate.
- `FuncType` memory footprint (`size_of`) slightly decreased.
- Halfs the amount of heap memory allocations performed during parsing.
- Provides better space locality between parameter and result types of the `FuncType`.
- Some of the `parse` benchmarks show improvements above noise treshold using the `lto="fat",codegen-units=1` profile. No significant difference for the default profile. (x86)

The main point of this PR is not optimizations but to reduce unnecessary pressure on the heap memory allocator. This especially is useful for embedded platforms that cannot afford a high-perf allocator. Besides that the refactoring introduces the new API which helps to clean up other parts of the codebase.

### DIFF
--- a/crates/wasmparser/src/readers/core/types.rs
+++ b/crates/wasmparser/src/readers/core/types.rs
@@ -66,7 +66,7 @@ impl Debug for FuncType {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.debug_struct("FuncType")
             .field("params", &self.params())
-            .field("results", &self.results())
+            .field("returns", &self.results())
             .finish()
     }
 }
--- a/crates/wasmparser/src/readers/core/types.rs
+++ b/crates/wasmparser/src/readers/core/types.rs
@@ -103,17 +103,13 @@ impl FuncType {
     /// Returns a shared slice to the parameter types of the [`FuncType`].
     #[inline]
     pub fn params(&self) -> &[ValType] {
-        // Safety: `len_params` is asserted to be less than or equal to
-        //         `params_results.len()` by construction.
-        unsafe { self.params_results.get_unchecked(..self.len_params) }
+        &self.params_results[..self.len_params]
     }
 
     /// Returns a shared slice to the result types of the [`FuncType`].
     #[inline]
     pub fn results(&self) -> &[ValType] {
-        // Safety: `len_params` is asserted to be less than or equal to
-        //         `params_results.len()` by construction.
-        unsafe { self.params_results.get_unchecked(self.len_params..) }
+        &self.params_results[self.len_params..]
     }
 }
 
--- a/crates/wasmparser/src/readers/core/types.rs
+++ b/crates/wasmparser/src/readers/core/types.rs
@@ -55,10 +55,54 @@ pub enum Type {
 /// Represents a type of a function in a WebAssembly module.
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
 pub struct FuncType {
-    /// The function parameter types.
-    pub params: Box<[ValType]>,
-    /// The function result types.
-    pub returns: Box<[ValType]>,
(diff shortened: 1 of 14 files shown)


crates/wasmparser/src/readers/core/types.rs
```diff
@@ -55,10 +55,56 @@ pub enum Type {
...
+            len_params,
+        }
+    }
+
+    /// Returns a shared slice to the parameter types of the [`FuncType`].
+    #[inline]
+    pub fn params(&self) -> &[ValType] {
+        // Safety: `len_params` is asserted to be less than or equal to
+        //         `params_results.len()` by construction.
+        unsafe { self.params_results.get_unchecked(..self.len_params) }
```
Robbepop, line 98: Initially I did not use `unsafe` for `params` and `results` methods but given the new API and its invariants the usage of `unsafe` methods yields slight performance boost for the optimized profile.
Still I wonder if we want to introduce `unsafe` here.

alexcrichton: Ideally we wouldn't have any `unsafe` code in `wasmparser`, so could you benchmark and see the effect with/without this?

alexcrichton:
Looks reasonable to me, although I think it would be best to remove the `unsafe` if possible

crates/wasmparser/src/readers/core/types.rs
Robbepop: Yeah I benchmarked this while working on the PR and the `unsafe` is not critical but just nice to have since it improves some `parse` benchmarks by roughly 2.5%. There is no significant difference between `main` branch and this PR when using safe versions of these methods. So the gain is not big and we can definitely remove the usage of `unsafe` here.

Robbepop: **edit:** I just benchmarked it again (since there have been some changes in between) and the performance is actually affected negatively under the default profile without those `unsafe` operations ... It kind of makes sense since the previous implementation naturally also had no bounds checking when both `params` and `returns` was still split.

alexcrichton: At least for me locally the difference in benchmarks all looks to be noise, I'm measuring +/- 3% with this change. I don't think that these bounds checks are so common that it would actually have any significant impact on performance.

Robbepop: Yeah wins and losses are definitely <3% overall. I more or less copied Criterion's meaning about noise or significance. :sweat_smile: 
In summary: With `unsafe` we win 0-3%, without we seems to be losing 0-3% on `parse` benchmarks.

alexcrichton: At least for me successive runs of `cargo bench` even with the same source code produces +/- 3% variance so anything under 5 I basically don't trust as being accurate personally.

Robbepop: That's very interesting since there is a lot less variance on my machine. But yeah, I definitely agree with you in that case.

alexcrichton: Ok I'm gonna go ahead and merge this since I think the memory wins are still good, but we can keep an eye on this over time. It's not 100% crucial that this crate has 0 unsafe blocks but given the context of `wasmparser` the bar is set pretty high for adding unsafe blocks

alexcrichton (approved):

crates/wasmparser/src/readers/core/types.rs
Robbepop: sounds good to me!
maybe we can add some more different benchmarks in future that can assert certain performance properties better than the currenr set.

PR #740, opened 2022-08-20
Fix a few minor issues in `wasmprinter`

alexcrichton:
I was investigating a fuzz-generated issue discovered on oss-fuzz where
`wasmprinter` would OOM with a very small input and found a number of
minor issues.

* The main problem which resulted in OOM was that `wasmprinter` was
  printing the `(value ...)` results for a component start function and
  the number of results weren't limited in any way. This commit updates
  the binary parsing of this field to place a limit, when it's parsed,
  on the maximum value.

* Another issue I noticed was that `wasmparser` accepting an invalid
  component start section where data remained after the start section
  was successfully parsed. I added a case to validation such that after
  the start function is read the reader is asserted to be at eof.

* Finally I noticed a similar issue when parsing the `module` subsection
  of the `name` section where the module's name discarded trailing data
  at the end of the module name subsection instead of error'ing out on
  the presence of this data.

I've fixed all these issues locally and added some tests for the
component model `start`-related issues.

### DIFF
--- a/crates/wasmparser/src/binary_reader.rs
+++ b/crates/wasmparser/src/binary_reader.rs
@@ -208,7 +208,7 @@ impl<'a> BinaryReader<'a> {
             arguments: (0..size)
                 .map(|_| self.read_var_u32())
                 .collect::<Result<_>>()?,
-            results: self.read_var_u32()?,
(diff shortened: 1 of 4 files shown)


peterhuene:
Merging since Alex is out today and these are small, minor fixes.

PR #744, opened 2022-08-25
wasm-compose: improve the tool to make configuration files optional

peterhuene:
This PR improves `wasm-compose` such that it no longer requires a configuration file to be present to operate.

To support this change, the tool is now explicitly told which component is the root component to compose.

To resolve dependencies, a set of search paths are searched automatically for components satisfying the dependencies of the root (and transitively their dependencies as well).

Dependencies that are not found are kept as instance imports in the composed component (i.e. can now import host interfaces from the composed component).

The tool now only exports everything from the root component in the composed output.

A configuration may still be specified to have the same flexibility in specifying the complete or partial instantiation graph and where to explicitly locate dependencies.

Additionally, exported instances from an instance may now be easily aliased such that those exported instances may participate in the instantiation graph too.


fitzgen:
> To resolve dependencies, a set of search paths are searched automatically for components satisfying the dependencies of the root (and transitively their dependencies as well).

(Haven't looked at the code yet, but) are these paths passed explicitly or is there some set of default/implicit paths? The latter is reminding me of all the times I've been frustrated with `PYTHONPATH` and pulling my hair out trying to figure out why some dep isn't being found or something like that... :grimacing:

fitzgen:
> Dependencies that cannot be located will remain as instance imports in the
composed component

I could imagine this being frustrating if you expect something to be bundled in but it wasn't and instead of getting an error about that, the tool silently made it an instance import, and you didn't find out till later when instantiation fails because of that missing instance import.

Not saying that the "strict" opposite (make everyone explicitly list their instance imports and anything that isn't found and isn't an explicitly listed instance import results in an error) is necessarily better (I could imagine this being very tedious, and I'm not sure whether that is worth it for avoiding these errors that take a while to bubble up).

But this is something to think about and maybe file follow up issues for new CLI flags to enable/disable "strict" mode or just to have a place to really hash out what we want the developer experience to be here.

fitzgen:
An idea: I haven't got to the actual code that searches through paths and all that yet, but I think having logging something like

```
trying to resolve dependency: `backend`
  no explicit dependency for `backend` in configuration
  looking in directory `.`
    did not find dependency
  looking in direcotry `./dependencies`
    did not find dependency
  looking in directory `/usr/lib/wasm-components`
    did not find dependency
  failed to find `backend` dependency, leaving `backend` as an import
```

And that is hopefully easier to enable than `RUST_LOG`, which many folks (expecially non-Rust devs) won't know to try or think of. Probably a `--verbose` CLI flag  or something.

crates/wasm-compose/tests/compositions/complex/root.wat
```diff
@@ -2,4 +2,8 @@
   (type (instance (export "m" (func (param "x" string) (result string)))))
   (import "b1" (instance (type 0)))
   (import "b2" (instance (type 0)))
+  (alias export 0 "m" (func))
+  (alias export 1 "m" (func))
+  (export "m1" (func 0))
+  (export "m2" (func 1))
 )
```
fitzgen, line 9: Pre-existing, but this file is missing a final newline.

fitzgen (approved):
Code itself looks great! Just the above comments about developer experience, which aren't things that this PR needs to tackle, of course.

Thanks @peterhuene!

peterhuene:
> (Haven't looked at the code yet, but) are these paths passed explicitly or is there some set of default/implicit paths? The latter is reminding me of all the times I've been frustrated with PYTHONPATH and pulling my hair out trying to figure out why some dep isn't being found or something like that... 😬

The paths are explicitly passed in via CLI options or via a CLI option to point at the config file which itself explicitly lists the search paths; currently no environment variable is respected.

> I could imagine this being frustrating if you expect something to be bundled in but it wasn't and instead of getting an error about that, the tool silently made it an instance import, and you didn't find out till later when instantiation fails because of that missing instance import.

It currently logs a warning whenever it fails to find a dependency and treats it as an instance import.

> But this is something to think about and maybe file follow up issues for new CLI flags to enable/disable "strict" mode or just to have a place to really hash out what we want the developer experience to be here.

I think that's a great idea for a follow-up feature to error for unresolved dependencies instead of what it does now. That would play really well once the tool ingests world files so it knows exactly what the expected host environment provides.

> An idea: I haven't got to the actual code that searches through paths and all that yet, but I think having logging something like...

The tool does currently log every path it probed for a component at debug level (although relying on `RUST_LOG` to enable).

> And that is hopefully easier to enable than RUST_LOG, which many folks (expecially non-Rust devs) won't know to try or think of. Probably a --verbose CLI flag or something.

Definitely another great follow-up feature to add.

PR #746, opened 2022-08-29
Update `wasm-tools` to latest component model proposal

peterhuene:
This PR updates the various tools in the repo to the latest component model proposal changes, which include:

* The removal of the core alias section (and reordering of subsequent sections).
* Aliases now have a variant for core exports.
* Module types now only encode an outer alias to a core type.

The text format has also changed to remove core aliases in favor of `(alias core export ...)`.

See webassembly/component-model#95 for the proposal changes.


fitzgen (approved):
LGTM!

PR #753, opened 2022-08-31
Implement another round of validator optimizations

alexcrichton:
This commit implements another batch of optimizations for the operator
validator in wasmparser. This results in a 15-20% improvement in
validation time for the benchmarked modules in this repository. The
optimizations implemented here are:

* The `pop_operand` method has been split into two with a "fast path"
  that can be inlined into callers and a slower path that is not
  inlined. The fast path is expected to be always taken for valid
  modules which is where the expected type is indeed on the stack. The
  slow path handles the case of mismatched types, unreachable code, etc.
  This is by far the largest optimization implemented in this commit.

* Creation of a `BinaryReaderError` has been updated to uniformly use a
  new `format_err!` macro and `bail!` macro, both modeled after the
  preexisting `format_op_err` and `bail_op_err` macros. These
  additionally use a new `BinaryReaderError::fmt` constructor to perform
  the string allocation within the constructor rather than within the
  caller, ideally reducing code size and the footprint of these
  constructors.

* The `br_if` validation now only calls `self.label_types(...)` once
  instead of twice.

* The `Locals::get` method has been split manually into two functions
  with the fast-path it's-in-the-vector being inlineable and the binary
  search is outlined to a separate function.

### DIFF
--- a/crates/wasmparser/src/binary_reader.rs
+++ b/crates/wasmparser/src/binary_reader.rs
@@ -998,6 +998,7 @@ impl<'a> BinaryReader<'a> {
     /// # Errors
     ///
     /// If `BinaryReader` has no bytes remaining.
+    #[inline]
     pub fn read_u8(&mut self) -> Result<u8> {
         let b = match self.buffer.get(self.position) {
             Some(b) => *b,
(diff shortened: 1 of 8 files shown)


Robbepop (approved):
looking forward to use those new optimizations :)
glad to see that the resulting code looks even cleaner than before generally.

peterhuene (approved):

PR #763, opened 2022-09-07
Update the spec test suite submodule

alexcrichton:
This commit updates the `tests/testsuite` submodule which is a copy of the spec tests from upstream and the various proposals repositories. The major change included here is the update to the lexing of the text format to allow "reserved" tokens to have string tokens embedded within them (in a sense). This doesn't affect any parsing in the crate, only the tokenization of the input.

Additionally there are a few minor updates to error messages plus an new exceptional case where the spec interpreter is printing "unknown global" for defined globals which I'm not really sure what's going on.

### DIFF
--- a/crates/wast/src/core/wast.rs
+++ b/crates/wast/src/core/wast.rs
@@ -76,6 +76,8 @@ pub enum WastRetCore<'a> {
     RefExtern(u32),
     /// A non-null funcref is expected.
     RefFunc(Option<Index<'a>>),
+
+    Either(Vec<WastRetCore<'a>>),
 }
 
 static RETS: &[(&str, fn(Parser<'_>) -> Result<WastRetCore<'_>>)] = {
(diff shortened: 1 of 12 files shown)


alexcrichton:
I've now additionally updated this to include the inclusion of `relaxed-simd` into the testsuite repo as well.

fitzgen (approved):

dkegel-fastly:
This seems to bump the minimum supported rust compiler; I'm now getting
```
#25 101.6   --> /root/.cargo/registry/src/github.com-1ecc6299db9ec823/wast-47.0.0/src/names.rs:84:26
#25 101.6    |
#25 101.6 84 |         format!("unknown {ns}: failed to find name `${}`", id.name()),
#25 101.6    |                          ^^^^
```
in a dockerfile that had been working, so my little downstream project's unhappy now.

I guess I get to update it from alpine 2.15; I was about to anyway for other reasons.

PR #764, opened 2022-09-08
wasmparser: self-contain for_each_operator!

nagisa:
This adjusts the `for_each_operator!` macro to avoid it requiring ambient context (imports) with specific names. Instead it’ll use correctly-scoped fully qualified names, making it somewhat easier to use.


Robbepop (approved):
nice usability improvement!

alexcrichton (approved):
Looks great to me, thanks! I think that this may need a rebase though.

PR #767, opened 2022-09-16
OperatorVisitor: qualify index param names

nagisa:
This is intended to make it easier for implementers of this trait to figure what is what (admittedly `index` doesn’t say much when there are all kinds of indexable lists, and some entities deal with indices into multiple lists), especially if they’re using something like RLS’ auto-implement all methods or just reading the documentation.

With `type_index` and `tag_index` these parameters become self-documenting.


nagisa:
Oh, it is going to affect the definitions of the `Operand` type too…

Well, I think it is still worth trying to make this change tbh, so I’ll adjust.

Robbepop (approved):
Thanks a lot for these improvements. In the past I often was confused about this as well.

alexcrichton (approved):

alexcrichton:
Agreed, thank you for these improvements!

PR #768, opened 2022-09-19
Fix wasm-encoder encodings for new relaxed instructions

alexcrichton:
Looks like a few copy/paste errors


yurydelendik (approved):
Yep, copy'n'paste typos. Thank you.

PR #769, opened 2022-09-20
🏗️ Move wasm-smith's non-trapping mode logic to CodeBuilder

itsrainy:
Prior to this change, the non-trapping logic in wasm-smith was implemented as a post-processing step. Random wasm would get generated and then the `no_traps()` logic would attempt to modify the generated instructions to ensure that the resulting wasm wouldn't trap. This led to a couple challenges: there are some instructions (such as `unreachable` and `call_indirect`) that we don't want to emit at all in non-trapping mode, and there are some instructions that we haven't yet implemented a non-trapping solution for (such as SIMD Lane loads/stores). This meant that running wasm-smith in non-trapping mode could still result in a wasm program that would trap. While discussing how best to communicate and work around this nuance, @fitzgen and I realized that these were probably better dealt with while the code is being generated, rather than after the fact.

This change moves much of the non-trapping logic from `src/core/no_traps.rs` to a code_builder module (`src/core/code_builder/no_traps.rs`). It then calls into those functions from CodeBuilder, while entirely avoiding `unreachable`, `call_indirect`, and possibly trapping instructions that we haven't yet dealt with.

TODO:
- [ ] I'm currently working on adding a fuzz target to verify that `disallow_traps` works and to find the cases I inevitably missed


crates/wasm-smith/src/core.rs
```diff
@@ -199,6 +198,16 @@ impl Module {
...
+
+    fn import_count(&self) -> usize {
+        self.imports
+            .iter()
+            .filter(|imp| match imp.entity_type {
+                EntityType::Func(_, _) => true,
+                _ => false,
+            })
+            .count()
+    }
```
fitzgen: Is this actually used anywhere? I don't see anything in this commit.

Either way, it should be named `imported_func_count` or something like that since it is only counting function imports and not all imports.

crates/wasm-smith/src/core/code_builder.rs
```diff
@@ -2414,6 +2428,12 @@ fn memory_copy_valid(module: &Module, builder: &mut CodeBuilder) -> bool {
         return false;
     }
 
+    // The non-trapping case for memory copy has not yet been implemented,
+    // so we are excluding them for now
+    if !module.config.disallow_traps() {
+        return false;
+    }
```
fitzgen, line 2435: Shouldn't this be 

```rust
if module.config.disallow_traps() {
    return false;
}
```

(That is, no `!`)

?

fitzgen (approved):
LGTM, although yeah we should have that fuzz target

crates/wasm-smith/src/core/code_builder.rs
itsrainy: Ah good catch!

crates/wasm-smith/src/core/code_builder.rs
```diff
@@ -2049,6 +2428,12 @@ fn memory_copy_valid(module: &Module, builder: &mut CodeBuilder) -> bool {
         return false;
     }
 
+    // The non-trapping case for memory copy has not yet been implemented,
+    // so we are excluding them for now
+    if !module.config.disallow_traps() {
```
itsrainy, line 2433:
```suggestion
    if module.config.disallow_traps() {
```

crates/wasm-smith/src/core/no_traps.rs
```diff
@@ -1,818 +0,0 @@
...
-            }
-
-            code.instructions = Instructions::Generated(new_insts);
-        }
-        Ok(())
-    }
-
-    /// Mask data and element segments' offsets to be in bounds of their
-    /// associated tables and memories.
-    fn no_trapping_segments(&mut self) {
```
itsrainy, line 547: i just realized I never moved this over

crates/wasm-smith/src/core/code_builder.rs
```diff
@@ -3589,10 +4785,21 @@ fn v128_store(
...
+    if module.config.disallow_traps() {
+        no_traps::store(
+            Instruction::V128Store { memarg },
+            module,
+            builder,
+            instructions,
+        );
+    } else {
+        instructions.push(Instruction::V128Store { memarg });
+    }
```
nagisa, line 4801: I’m wondering if this pattern would be better served by making a choice between trapping or non-trapping variants and then generating an instruction that is guaranteed to be either way? Are there any instructions that would fall in the “unknown” territory somewhere?

This sort of split model would allow to easily add/maintain other kinds of generation “modes” as well (e.g. deterministic vs non-deterministic), and I wonder if it would result in better fuzz coverage (the fuzzer would be able to flip between the two broad categories of instructions at runtime on a per-instruction basis)

itsrainy: @nagisa that's interesting! I'm wondering how useful the "guaranteed to trap" case would actually be. If we were to create a version of this code that does the opposite (ie. makes every division have a denominator of 0, makes every load/store have an out of bounds address, etc), then I would think any generated programs would trap relatively early and not improve coverage all that much except in testing that traps are handled correctly by the runtime. Maybe a wasm-smith flag to generate programs that are guaranteed to _eventually_ trap would be helpful for improving coverage, since that would allow us to explore whether different program state affects the trap-handling behavior of the runtime.

Either way, maybe we discuss this further on https://github.com/bytecodealliance/wasm-tools/issues/266 (or a new issue)? I think work in that direction could be an improvement on this work, but not immediately necessary.

As far as ease of adding/maintaining other modes, I agree that having this turn into something like the below would be unideal:
```rust
if option_1 {
  generate_instruction();
} else if option_2 {
  generate_instruction_but_slightly_differently();
} else {
  generate_instruction_more_differently();
}
```
That being said, I'm inclined to leave that kind of refactoring for when/if we actually add more modes like this rather than trying to think of a more general way of doing this that may never be needed.

nagisa: > That being said, I think I'm inclined to leave that kind of refactoring for when/if we actually add more modes like this

Of course, that was just me thinking out loud, not necessarily suggesting that this must be implemented here and now ^^ Should have made it clear in my original comment.

itsrainy:
@fitzgen I've been finagling with the fuzzer for a little bit and I haven't been able to get any functions to run yet, but I think I might be close(🤞).

The fuzzer currently errs on line 49 with
```
Insufficient resources: memory minimum size of 281474976710656 pages exceeds memory limits
```

I'm not sure where that size is coming from, but I think it's in the Memory being generated by `dummy_imports`. I haven't found any immediately obvious config options for `Store`, `Engine`, or `Module` that seem like they'd let me set that to something reasonable.

fuzz/fuzz_targets/no-traps.rs
```diff
@@ -0,0 +1,68 @@
...
+// `Module`s as input.
+fuzz_target!(|data: &[u8]| {
+    let mut u = Unstructured::new(data);
+    let (wasm_bytes, config) = match wasm_tools_fuzz::generate_valid_module(&mut u, |config, _| {
+        config.disallow_traps = true;
+        config.threads_enabled = false;
+        config.allow_start_export = false;
+        config.max_memories = config.max_memories.min(1);
+        config.memory64_enabled = true;
+        config.exceptions_enabled = false;
```
itsrainy, line 21: These config options were mostly an attempt to get around errors that were preventing the Module or Store from getting instantiated. I would not be surprised if some of these aren't actually necessary.

fuzz/Cargo.toml
```diff
@@ -62,3 +62,10 @@ name = "mutate"
 path = "fuzz_targets/mutate.rs"
 test = false
 doc = false
+
+[[bin]]
+name = "no-traps"
+path = "fuzz_targets/no-traps.rs"
+test = false
+doc = false
+required-features = ["wasmtime"]
```
fitzgen, line 71: Nitpick: no newline at the end of the file.

fuzz/fuzz_targets/no-traps.rs
```diff
@@ -0,0 +1,68 @@
...
+pub mod dummy;
+
+// Define a fuzz target that accepts arbitrary
+// `Module`s as input.
+fuzz_target!(|data: &[u8]| {
+    let mut u = Unstructured::new(data);
+    let (wasm_bytes, config) = match wasm_tools_fuzz::generate_valid_module(&mut u, |config, _| {
+        config.disallow_traps = true;
+        config.threads_enabled = false;
+        config.allow_start_export = false;
```
fitzgen, line 18: I don't think this should be necessary.

fuzz/fuzz_targets/no-traps.rs
```diff
@@ -0,0 +1,68 @@
...
+        config.max_memories = config.max_memories.min(1);
+        config.memory64_enabled = true;
```
fitzgen, line 20: I don't think this should be necessary either.

fuzz/fuzz_targets/no-traps.rs
```diff
@@ -0,0 +1,68 @@
...
+
+// Define a fuzz target that accepts arbitrary
+// `Module`s as input.
+fuzz_target!(|data: &[u8]| {
+    let mut u = Unstructured::new(data);
+    let (wasm_bytes, config) = match wasm_tools_fuzz::generate_valid_module(&mut u, |config, _| {
+        config.disallow_traps = true;
+        config.threads_enabled = false;
+        config.allow_start_export = false;
+        config.max_memories = config.max_memories.min(1);
```
fitzgen, line 19: Not this one actually.

fuzz/fuzz_targets/no-traps.rs
```diff
@@ -0,0 +1,68 @@
...
+        match export.ty() {
+            ExternType::Func(func_ty) => {
+                let args = dummy::dummy_values(func_ty.params());
+                let mut results = dummy::dummy_values(func_ty.results());
+                let func = instance.get_func(&mut store, export.name()).unwrap();
+                func_ty.results();
+                store.add_fuel(1_000).unwrap();
+                match func.call(&mut store, &args, &mut results) {
+                    Ok(_) => continue,
+                    Err(_) => panic!("wasm trapped"),
```
fitzgen, line 62: The trap might be because we ran out of fuel, in which case it is fine and we can continue.

fuzz/fuzz_targets/no-traps.rs
```diff
@@ -0,0 +1,68 @@
...
+        .and_then(|imports| Instance::new(&mut store, &module, &imports))
+        .unwrap();
+    for export in module.exports() {
+        match export.ty() {
+            ExternType::Func(func_ty) => {
+                let args = dummy::dummy_values(func_ty.params());
+                let mut results = dummy::dummy_values(func_ty.results());
+                let func = instance.get_func(&mut store, export.name()).unwrap();
+                func_ty.results();
+                store.add_fuel(1_000).unwrap();
```
fitzgen, line 59: You probably want to reset to exactly 1000 fuel each time, rather than keep adding 1000 in the case where there is left over fuel from last iteration.

https://cs.github.com/bytecodealliance/wasmtime/blob/bb6a8a717ab9dcf0594a577893fb31a6528ada1b/crates/fuzzing/src/oracles.rs?q=set_fuel#L790-L801

fuzz/fuzz_targets/no-traps.rs
```diff
@@ -0,0 +1,68 @@
...
+        simd: config.simd_enabled,
+        relaxed_simd: config.relaxed_simd_enabled,
+        memory64: config.memory64_enabled,
+        threads: config.threads_enabled,
+        exceptions: config.exceptions_enabled,
+        ..wasmparser::WasmFeatures::default()
+    });
+    if let Err(e) = validator.validate_all(&wasm_bytes) {
+        panic!("Invalid module: {}", e);
+    }
```
fitzgen, line 43: Might be a little bit easier to read this fuzz target if validation was pulled out into a `validate` function and then calling exports was also pulled out into its own `call_function_exports` function or something like that. As is, it is a bit of a wall of code to read.

fitzgen (commented):
Looks pretty good, a few suggestions below. Is it running and passing?

fitzgen:
> The fuzzer currently errs on line 49 with

We can configure a maximum to the size of memories generated to avoid this issue.

itsrainy:
> Looks pretty good, a few suggestions below. Is it running and passing?

No not yet! As-is it bails at line 49 every time so it isn't even making it to the function calls. I'll remove the config options that aren't needed but I might need some help to figure out how to properly instantiate the wasm! I roughly followed the pattern in the failed-instantiations fuzzer, but I think I'm missing something

itsrainy:
The only traps this is now throwing for me are StackOverflows on modules with recursive functions that look something like this:
```
  (func (;0;) (type 0) (result f64)
    call 0
     ...
   )
```

These can't easily be avoided without somehow keeping track of the call-stack size, which we'd have to do in the emitted wasm, and then somehow handling the "we're about to exceed our stack depth" case. It takes the fuzzer a few hours to run into this case on my machine.

From chatting with @fitzgen it sounds like ignoring this for now and documenting the behavior in the config docs should be good for the time being.

If I add a case to ignore StackOverflow traps, like this:
```diff, name=no-traps.rs
- Err(err) if err.to_string().contains("all fuel consumed") => continue,
+ Err(err)
+     if err.to_string().contains("all fuel consumed")
+         || err.to_string().contains("call stack exhausted") =>
+ {
+     continue
+ }                  
```

**then the fuzzer runs for 20+ hours on my machine!** I think this means that we've found all the cases that we would expect to trap 🎉 

@fitzgen any other changes you'd like to see here?

fitzgen (approved):
Looks great, thanks @itsrainy! Just one nitpick about documentation.

PR #770, opened 2022-09-21
wasm-tools: update for latest component model changes.

peterhuene:
This PR updates the tools for the latest component model spec changes.

Namely, it makes component function parameters always named.

See WebAssembly/component-model#108 for the spec change.


alexcrichton (approved):
Thanks!

PR #774, opened 2022-09-21
wasm-componentize

pchickey:


pchickey:
We're taking a totally different tack on this, morally this work will continue in wit-componentize over in wit-bindgen

PR #788, opened 2022-10-04
Automatically validate required wasm proposals

alexcrichton:
This commit updates the operator validator within `wasmparser` to automatically validate that the appropriate wasm proposal is enabled for each instruction instead of manually listing the instructions. This is done by using the `@proposal` tag in the `for_each_operator!` macro. A wrapper visitor is used that's macro generated which validates the required proposal before delegating to the actual validator which does per-instruction validation.

This should also fix a few discrepancies where some instructions were listed as `@reference_types` but validated as bulk-memory. Additionally this aligned the `@`-names with the names in `WasmFeatures`.

### DIFF
--- a/crates/wasmparser/src/lib.rs
+++ b/crates/wasmparser/src/lib.rs
@@ -41,10 +41,10 @@
 ///
 /// - `@mvp`: Denoting a Wasm operator from the initial Wasm MVP version.
 /// - `@exceptions`: [Wasm `expection-handling` proposal]
-/// - `@tail_calls`: [Wasm `tail-calls` proposal]
+/// - `@tail_call`: [Wasm `tail-calls` proposal]
 /// - `@reference_types`: [Wasm `reference-types` proposal]
-/// - `@sign_ext_ops`: [Wasm `sign-extension-ops` proposal]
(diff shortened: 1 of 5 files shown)


alexcrichton:
Not currently, although I'm not sure how dynamic configuration within `WasmFeatures` could lead to elision of the checks entirely. The only way I can think of to do that is to have a compile-time feature to remove the ability to disable features in `WasmFeatures`. If something is always-enabled then the check can naturally be elided, but otherwise with the runtime-configurable structure I don't think it's possible to statically elide these checks.

pchickey (approved):

Robbepop:
> Not currently, although I'm not sure how dynamic configuration within `WasmFeatures` could lead to elision of the checks entirely. The only way I can think of to do that is to have a compile-time feature to remove the ability to disable features in `WasmFeatures`. If something is always-enabled then the check can naturally be elided, but otherwise with the runtime-configurable structure I don't think it's possible to statically elide these checks.

Ahh yes I probably was a bit overzealous since this was only possible if we used `dyn` bindings which have other performance issues if used. Anyhow, I still like the change lot since it properly implements separation of concerns and cleans up code in that regard. :)

PR #789, opened 2022-10-04
Remove the `Cell` for fuel in `wasm-mutate`

alexcrichton:
This commit removes the `Cell` used in fuel by taking `&mut self` instead of `&self`. Two locations needed to be updated to handle this mutability and weren't too hard to refactor.


alexcrichton:
The motivation for this change is that clap in 4.0.0 [does not support non-`Sync` arguments](https://github.com/clap-rs/clap/issues/4347) and looks like it won't plan on doing so any time soon, so in the interest of still being upgrade this performs the refactor here.

pchickey (approved):

PR #792, opened 2022-10-05
Add a library function to view `Payload` as a section

alexcrichton:
This makes the `wasm-tools strip` function easier to write and additionally enables this sort of functionality externally as `wit-bindgen` wants to strip custom sections as well.


pchickey (approved):
Yep this is exactly what I was wishing for.

PR #795, opened 2022-10-11
Refactor and simplify wasm opcode printing

alexcrichton:
While looking through `wasmprinter` for other issues I was struck with inspiration to simplify the printing of operators with the `for_each_operator!` macro to avoid the verbosity of the full trait implementation. Along the way I made a few other adjustments as well:

* `*RoundingAverage*` instructions now use `Avgr` in the name to match the instruction name.
* `MemArg` now specifies its `max_align` in a field to uniformly handle this without needing per-instruction treatment.
* Alignment of atomic operators now properly validates that the alignment must be specified at the maximum.
* The text representation of a relaxed simd instruction was updated to have `*_sat_*` in the name to match the in-Rust name.

### DIFF
--- a/crates/wasm-encoder/src/core/code.rs
+++ b/crates/wasm-encoder/src/core/code.rs
@@ -647,7 +647,7 @@ pub enum Instruction<'a> {
     I8x16MinU,
     I8x16MaxS,
     I8x16MaxU,
-    I8x16RoundingAverageU,
+    I8x16AvgrU,
     I16x8ExtAddPairwiseI8x16S,
     I16x8ExtAddPairwiseI8x16U,
     I16x8Abs,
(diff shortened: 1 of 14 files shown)


fitzgen (approved):
Very nice.

Robbepop:
Sorry for being very late. While reading the changes introduced in this PR I noticed that this removed a check for maximum alignment that existed in the removed `read_memarg_of_align`: https://github.com/bytecodealliance/wasm-tools/pull/795/files#diff-154ee93e092342f2f38473cbc114e5d864ebf78c4a302e5988805f4f659e45f3

This reminded me of this `wasmi` validation error (note that `wasmi` uses `wasmparser` for validation): https://github.com/paritytech/wasmi/issues/570

However, `wasmi` uses the older version in which this check was still existing. 🤔
I would love to get some clarification. 😅

alexcrichton:
I removed that because it was duplicating logic already in the validator, and generally where possible we try to have the parser be relatively flexible about what it reads and the validator is what makes sure it's all lined up. With the validator in use there shouldn't have been any change in modules accepted or rejected with this change.

Robbepop:
@alexcrichton Thanks a lot for the clarification!

PR #806, opened 2022-11-02
Add Types{,Ref}::{imports,exports} getters

coolreader18:
I'm not sure if there was a reason this wasn't a thing, but afaict `validator::core::exports::Module::exports/imports` is never accessed to get() at one of their entries, only ever to insert stuff into them. This seems like a missing piece of the API, sorta.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ arbitrary = "1.1.0"
 clap = { version = "4.0.0", features = ["derive"] }
 criterion = "0.3.3"
 env_logger = "0.9"
+hashbrown = { version = "0.12.3", default-features = false }
 indexmap = "1.9.1"
 leb128 = "0.2.4"
 libfuzzer-sys = "0.4.0"
--- a/crates/wasmparser/Cargo.toml
+++ b/crates/wasmparser/Cargo.toml
@@ -14,6 +14,7 @@ exclude = ["benches/*.wasm"]
 
 [dependencies]
 indexmap = { workspace = true }
+hashbrown = { workspace = true }
 
 [dev-dependencies]
 anyhow = { workspace = true }
--- a/crates/wasmparser/src/validator.rs
+++ b/crates/wasmparser/src/validator.rs
@@ -48,6 +48,7 @@ fn test_validate() {
 mod component;
 mod core;
 mod func;
+mod importmap;
 mod operators;
 pub mod types;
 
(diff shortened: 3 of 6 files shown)


alexcrichton:
Seems reasonable to me to add! Could the return values here though be an `impl Iterator` rather than the internal representation? The `imports` representation in particular is a bit wonky to work with.

coolreader18:
Oh, I was hoping to use it as an IndexMap; ModuleType/ComponentType also have `imports`/`exports` public as a IndexMap. Maybe if the `Vec<EntityType>` was wrapped in a `ImportType` struct or something?

coolreader18:
Or an `ImportsMap` that works similar to a multimap or something, since that seems to be what it's functioning as.

coolreader18:
Or actually it's not even really an IndexMap at all, is it? Internal representation should be more something like `{ indices: Vec<(String, String, usize)>, entries: HashMap<(String, String), Vec<EntityType>> }`, since the imports are actually referred to by their index in import index space. Oh maybe like [`OrderedMultiMap`](https://docs.rs/ordered-multimap/latest/ordered_multimap/list_ordered_multimap/struct.ListOrderedMultimap.html)?

coolreader18:
Ok I know that's like, a whole new datastructure I just added, but it's really just a simple wrapper around a hashtable (or, simple if you know how hashmaps work, at least)

alexcrichton (commented):
Thanks for this, but sorry to clarify I dont think that this necessarily needs a whole new data structure. I was instead thinking that an `-> impl Iterator<...>` could be returned which internally did something like a `.flat_map(..)` to get hte flat iterator.

coolreader18:
Yeah, the RawTable probably was a little bit overkill... but I was thinking that since the full IndexMaps were exposed for the other methods, it'd make sense to have at least something with `get()` available here (that was my use case at least, to use it as a map rather than a vec). Also, the `ImportMap` was to fix a (theoretical?) correctness issue where if you flatten `imports` as an IndexMap, the iterator would return import entries in a different order from what they were in the module which now looking at the code that does the typechecking is not actually an issue at all.... hm. How about keeping ImportMap, but just as like a wrapper around a `IndexMap<(String, String), Vec<EntityType>>`, with a `get(module: &str, name: &str) -> Option<EntityType>` and a disclaimer that its iterator is not in the same order as the imports in the module? Or just an `impl Iterator` if that's still preferred, I've realized I'm sorta making a big deal out of this for nothing lol.

alexcrichton:
I do think that iteration order here will be important and should match what was specified in the original module. I forgot that `IndexMap` was already exposed so exposing such a map for exports seems reasonable. For imports perhaps the `impl Iterator` could be exposed for iterating over everything and a specific method could be added for looking up a specific import?

coolreader18:
I don't remember what I needed this for, and if I run into it again, I guess I'd rework this.

PR #811, opened 2022-11-09
wasm-encoder: Change the I64xLeS instruction encoding

adambratschikaye:
Change the `u32` field of the `I64x2LeS` instruction from `0xDD` to `0xDA` (decimal 218) to match the [spec](https://webassembly.github.io/spec/core/binary/instructions.html#vector-instructions).

### DIFF
--- a/crates/wasm-encoder/src/core/code.rs
+++ b/crates/wasm-encoder/src/core/code.rs
@@ -2331,7 +2331,7 @@ impl Encode for Instruction<'_> {
             }
             Instruction::I64x2LeS => {
                 sink.push(0xFD);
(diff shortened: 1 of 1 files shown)


adambratschikaye:
This should fix https://github.com/bytecodealliance/wasm-tools/issues/812.

alexcrichton (approved):
Thanks for catching this!

adambratschikaye:
Thanks for taking a look. By the way, any idea when the next release would be?

alexcrichton:
I'm preparing that in https://github.com/bytecodealliance/wasm-tools/pull/815

PR #813, opened 2022-11-09
FIX: Nop operation instead of unreachable

### LINKED ISSUE #809: wasm-mutate if complement mutator
If the`if` construction has no alternative body, it is replaced by an `unreachable` instruction.
This does not necessarily preserve semantics. 
The solution can be to replace the unreachable instruction by a nop opration.

https://github.com/bytecodealliance/wasm-tools/blob/d86563392253cb88cc9f101a48f1afa965a4469b/crates/wasm-mutate/src/mutators/codemotion/if_complement.rs#L55

cc @fitzgen @alexcrichton

Jacarte:
This fixes #809.
Basically, unreachable does not preserve semantics of the original binary.


fitzgen (approved):
Thanks!

PR #823, opened 2022-11-17
Implement parsing of tables with initializer expressions

yurydelendik:
The function references proposal adds initializer for tables when its type is non-nullable reference. The text format is extended with constant expression (which can have `ref.func`). The binary format adds `0x40 0x00` prefix to the table section entry to indicate expression presence.

https://github.com/WebAssembly/function-references/blob/main/proposals/function-references/Overview.md

### DIFF
--- a/crates/wast/src/core/binary.rs
+++ b/crates/wast/src/core/binary.rs
@@ -427,7 +427,13 @@ impl Encode for Table<'_> {
     fn encode(&self, e: &mut Vec<u8>) {
         assert!(self.exports.names.is_empty());
         match &self.kind {
-            TableKind::Normal(t) => t.encode(e),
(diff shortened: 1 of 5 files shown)


yurydelendik:
See also https://bugzilla.mozilla.org/show_bug.cgi?id=1784499

alexcrichton:
I don't think that it's type level information, though, but rather runtime information? I could be misunderstanding the spec, though, but for example when a table is imported is the initializer element specified? If so I would expect that the spec would grow more logic around the "matching" semantics to test whether two table types are equivalent, and if not I think it's runtime information that is stateful but not part of the type information.

yurydelendik:
Simplified patch by moving `init_expr` out of `TableType`.

PR #826, opened 2022-11-18
Implement latest component model spec.

peterhuene:
This PR implements the latest changes to the component model spec:

* Adds a `URL` field to imports and exports.
* Adds validation of the `URL` field.
* Adds validation of various "names" such that they are kebab-cased.

Consequently, import and export names in components are no longer allowed to be empty,
which required fixes in many component model tests.

### DIFF
--- a/crates/wasmparser/Cargo.toml
+++ b/crates/wasmparser/Cargo.toml
@@ -14,6 +14,7 @@ exclude = ["benches/*.wasm"]
 
 [dependencies]
 indexmap = { workspace = true }
+url = "2.3.1"
 
 [dev-dependencies]
 anyhow = { workspace = true }
--- a/crates/wit-component/src/validation.rs
+++ b/crates/wit-component/src/validation.rs
@@ -168,7 +168,7 @@ pub fn validate_module<'a>(
                 assert!(prev.is_none());
             }
             None if adapters.contains(name) => {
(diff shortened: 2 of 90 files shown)


crates/wit-component/src/validation.rs
```diff
@@ -425,7 +425,7 @@ fn validate_exported_interface(
                     wasm_sig_to_func_type(interface.wasm_signature(AbiVariant::GuestExport, f));
                 let ty = types.function_at(*func_index).unwrap();
                 if ty == &expected_ty {
-                    return Ok(());
+                    continue;
```
alexcrichton, line 428: Oops thanks for catching this!

crates/wasmparser/Cargo.toml
```diff
@@ -14,6 +14,7 @@ exclude = ["benches/*.wasm"]
 
 [dependencies]
 indexmap = { workspace = true }
+url = "2.3.1"
```
alexcrichton, line 17: I feel like we may need to one day add a `component-model` feature to `wasmparser` given the relative weight of the implementation internally and the added dependencies, but that's an issue for another day.

alexcrichton (approved):
All looks great to me, thanks!

crates/wit-component/src/validation.rs
peterhuene: Thanks goes to `clippy` as my rust-analyzer check command :)

PR #828, opened 2022-11-18
Fix the no-traps fuzzer when memory runs out

alexcrichton:
Allow another instance of a trap when an instance runs out of memory due to wanting to create too many tables or too many memories or similar.


pchickey (approved):

pchickey:
We should probably put non-string ways to recognize those errors into Wasmtime as concrete error types now that Trap is anyhow

alexcrichton:
Agreed!

PR #833, opened 2022-11-21
Remove the `wit_component::ComponentInterfaces` type

alexcrichton:
This was a temporary holdover until a `World` existed in `wit-parser`, which it now does, so remove the old type.

### DIFF
--- a/crates/wit-component/src/decoding.rs
+++ b/crates/wit-component/src/decoding.rs
@@ -88,12 +88,11 @@ struct InterfaceDecoder<'a> {
 /// Decode the world described by the given component bytes.
 ///
 /// This function takes a binary component as input and will infer the
-/// `Interface` representation of its imports and exports. More-or-less this
-/// will infer the "world" from a binary component. The binary component at this
-/// time is either a "types only" component produced by `wit-component` or an
-/// actual output of `wit-component`.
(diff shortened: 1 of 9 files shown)


crates/wit-component/src/encoding.rs
```diff
@@ -2169,13 +2169,8 @@ impl ComponentEncoder {
     ///
     /// The string encoding of the specified interface set is supplied here as
```
peterhuene, line 2170:
```suggestion
    /// The string encoding of the specified word is supplied here as
```

crates/wit-component/src/decoding.rs
```diff
@@ -85,54 +85,6 @@ struct InterfaceDecoder<'a> {
...
-    fn from(world: wit_parser::World) -> ComponentInterfaces {
-        ComponentInterfaces {
-            exports: world.exports,
-            imports: world.imports,
-            default: world.default,
-        }
-    }
-}
-
 /// Decode the interfaces imported and exported by a component.
```
peterhuene, line 88:
```suggestion
/// Decode the world described by the given component bytes.
```

peterhuene: Plus an update to the expanded-upon comment below this.

peterhuene (approved):
Just doc comment nits.

PR #848, opened 2022-11-28
wit-parser: re-enable use

fibonacci1729:
**WIP**

This PR temporarily re-enables `use` for the single-file `*.wit` scenario.
The implementation orders interfaces topologically and then processes
each interface's use dependencies in order.

As a concrete example, consider:

```wit
// service.wit

interface service-handler {
    use { request, response } from service-types
    use { error } from service-common

    f: func(req: request) -> result<response, error>
}

interface service-types {
    use { message, request, response } from service-common
}

interface service-common {
    record message {
        text: string
    }

    record request {
        mesg: message
    }

    record response {
        mesg: message,
        code: s32
    }

    enum error {
        ok,
        bad,
    }
}

world service {
    export handler: service-handler
}
```

Running `wasm-tools component new --types-only --wit service.wit -t -o service.wat` produces:

```
(component
  (type (;0;) (record (field "text" string)))
  (type (;1;) (record (field "mesg" 0)))
  (type (;2;) (record (field "mesg" 0) (field "code" s32)))
  (type (;3;) (enum "ok" "bad"))
  (type (;4;) (result 2 (error 3)))
  (type (;5;) (func (param "req" 1) (result 4)))
  (type (;6;)
    (instance
      (alias outer 1 0 (type (;0;)))
      (export "message" (type (eq 0)))
      (alias outer 1 1 (type (;1;)))
      (export "request" (type (eq 1)))
      (alias outer 1 2 (type (;2;)))
      (export "response" (type (eq 2)))
      (alias outer 1 3 (type (;3;)))
      (export "error" (type (eq 3)))
      (alias outer 1 5 (type (;4;)))
      (export "f" (func (type 4)))
    )
  )
  (export "handler" (type 6))
)
```

Signed-off-by: Brian H <brian.hardock@fermyon.com>

### DIFF
--- a/crates/wit-parser/Cargo.toml
+++ b/crates/wit-parser/Cargo.toml
@@ -17,6 +17,7 @@ id-arena = "2"
 anyhow = { workspace = true }
 indexmap = { workspace = true }
 pulldown-cmark = { version = "0.8", default-features = false }
+topo_sort = "0.4"
 unicode-xid = "0.2.2"
 
 [dev-dependencies]
--- a/crates/wit-parser/src/ast.rs
+++ b/crates/wit-parser/src/ast.rs
@@ -172,6 +172,14 @@ pub struct Interface<'a> {
 }
 
 impl<'a> Interface<'a> {
+    pub fn uses(&self) -> impl Iterator<Item = &Use<'a>> {
+        self.items.iter().filter_map(|item| match item {
+            InterfaceItem::Use(item) => Some(item),
+            InterfaceItem::TypeDef(_) => None,
+            InterfaceItem::Value(_) => None,
+        })
+    }
+
     pub fn span(&self) -> Span {
         self.name.span
     }
@@ -200,6 +208,7 @@ impl<'a> Interface<'a> {
 pub enum InterfaceItem<'a> {
     TypeDef(TypeDef<'a>),
     Value(Value<'a>),
+    Use(Use<'a>),
 }
 
 pub struct Id<'a> {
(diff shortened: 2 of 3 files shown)


fibonacci1729:
NOTE: This PR effectively reverts previously removed processing of `use`.

alexcrichton:
Thanks! Personally though I'm a bit hesitant to go down this route before `use` is figured out in terms of its final state. For example one idea we talked about historically was only `use` at the top level and not having to deal with it within `interface` blocks. I do understand that there are pains with `*.wit` today without `use`, but I think it would be best for churn to focus energies on designing a final state for `use`.

npmccallum:
@alexcrichton I want to push back on that. There isn't just pain. There is a lot of pain as there's no way to share types across interfaces. This means that for any system of even moderate complexity, everything has to go in one interface to share types.

What `wit` is missing from my perspective is a way to **evolve**. That is, we need something usable yesterday that we can grow up with. I'm fine even if we break compatibility. But we need to find a way to produce a new wasi snapshot yesterday.

What I want to accomplish is this:

1. `wit` remains unstable.
2. Every 6 months, we have a new version of `wit`; possibly breaking compatibility.
3. Every 6 months, we have a new version of `wasi`; possibly breaking compatibility.

This is infinitely better than doing nothing in `wasi` for years until some glorious future where `wit` arrives.

npmccallum:
@alexcrichton The problem here is that `use` was removed without any replacement and without any real discussion. This invalidated a bunch of my work. And there's no workaround (besides combining all interfaces into a single mega-interface; which defeats the whole purpose of worlds). And re-adding `use` is somewhat trivial. And, after we built consensus in Zulip on re-adding it, you single-handedly rejected it.

I'm asking that we:
1. re-add the small feature as unstable to allow continued progress
2. that we don't remove features without designing their replacements

I'm not asking:
1. that `.wit` be stable
2. that `use` be stable

alexcrichton:
I'm sorry that the removal of `use` invalidated your work, the [README warning](https://github.com/bytecodealliance/wit-bindgen/#about) was intended to indicate the unstable state of `*.wit` but that's been there for a long time.

I understand what you're asking for and what you're not asking for. I again am saying that I personally do not want to go back to an interim state only to break things immediately again. In my opinion energy is ideally spent towards designing `use`.

> And, after we built consensus in Zulip on re-adding it, you single-handedly rejected it.

I think this is somewhat disingenuous. Almost everything so far in `wit-bindgen` and `*.wit` was developed with no consensus whatsoever at any point along the way. That is something I would personally like to change now and is what the `wit-bindgen` repository has been moving towards. I've also discussed this in historical wit-bindgen meetings about the planned path forward here and everything.

There is no syntax, documentation, or semantics specified for `use` upstream in the component-model's `*.wit` explainer. There is additionally no PR for such syntax, documentation, or semantics. The direction of `wit-bindgen` is now to follow the component model rather than trying to jump ahead and prototype things - the time for that is over now. This is another reason I am personally reluctant to merge interim changes that do not have any upstream consensus or design-on-the-records.

If you're worried about me single-handedly rejecting things I would encourage you to ask for other reviewers on this. Everything here is just my own personal opinion. There is no formal governing body for most of the components in play here which means everything is ad-hoc. I'm certainly not going to revert this if someone else merges it.

fibonacci1729:
Hey @npmccallum and @alexcrichton, sorry for the delay weighing in here! I've been iterating to re-enable `use` for multi-file `*.wit` this week. I think i'll have a PR up early next week that supersedes this along with the proposed wit changes that need to be accepted by the component model `WIT.md` to reflect the implementation.

I'm going to close this PR for now.

PR #849, opened 2022-11-29
wasmparser: split parsers for different elems

nagisa:
Given an `elem`, whether it is going to contain function indices or constexprs is determined early on, and once determined, this won’t change.

The way wasmparser’s API was structured suggested that such change within a single element is in fact possible, which confused me every time I encountered this. By introducing two distinct readers for each kind of element items, this property is encoded in the API, much like it is in wasm_encoder.

Although unverified and not at all the focus of this PR, this change may also improve the performance of parsing items, as the branch checking for whether the item is an expression of a function is now outside of the loop by construction.

### DIFF
--- a/crates/wasm-mutate/src/mutators/translate.rs
+++ b/crates/wasm-mutate/src/mutators/translate.rs
@@ -1,6 +1,9 @@
 use crate::{Error, Result};
 use wasm_encoder::*;
-use wasmparser::{DataKind, ElementItem, ElementKind, FunctionBody, Global, Operator, Type};
+use wasmparser::{
+    DataKind, ElementKind, FunctionBody, Global, Operator, SectionReader, SectionWithLimitedItems,
+    Type,
+};
 
 #[derive(Debug, Hash, Eq, PartialEq, Copy, Clone)]
 pub enum Item {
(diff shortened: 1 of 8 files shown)


crates/wasmparser/src/readers/core/elements.rs
```diff
@@ -283,10 +320,16 @@ impl<'a> ElementSectionReader<'a> {
...
+        let items = if exprs {
+            ElementItems::Expressions(ElementExpressionItemsReader::new(
+                &self.reader.buffer[data_start..data_end],
+                self.reader.original_offset + data_start,
+            )?)
+        } else {
+            ElementItems::Functions(ElementFunctionItemsReader::new(
+                &self.reader.buffer[data_start..data_end],
+                self.reader.original_offset + data_start,
+            )?)
```
nagisa, line 332: NB: this changes where the failure to parse the count of element items is reported. I’m happy to implement something that would allow detecting this particular error case separately as was possible before.

crates/wasm-mutate/src/mutators/translate.rs
```diff
@@ -246,31 +246,32 @@ pub fn element(
...
+                    t.translate_const_expr(
+                        &reader.read()?,
+                        &element.ty,
+                        ConstExprKind::ElementFunction,
+                    )
+                })
+                .collect::<Result<Vec<_>, _>>()?;
+            Elements::Expressions(&exprs)
+        }
+    };
```
nagisa, line 270: This particular example is where I think the increase in clarity is definitely worth the additional verbosity in other parts of the code base.

crates/wasmparser/src/readers/core/elements.rs
```diff
@@ -49,49 +49,96 @@ pub enum ElementKind<'a> {
...
+    }
+
+    /// Reads an element function item from the segment.
+    pub fn read(&mut self) -> Result<u32> {
+        let idx = self.reader.read_var_u32()?;
+        Ok(idx)
+    }
+}
+
+impl<'a> IntoIterator for ElementFunctionItemsReader<'a> {
```
alexcrichton, line 92: Could this perhaps use `SectionReader`, `SectionWithLimitedItems`, and `SectionIterator` to avoid a new iterator type? Albeit the name isn't great but I think could help cut down on the boilerplate possibly.

alexcrichton (approved):
Looks great to me, thanks! One minor question about reusing some other iterator abstractions but otherwise I'm happy to merge either way.

crates/wasmparser/src/readers/core/elements.rs
nagisa: Addressed!

PR #857, opened 2022-12-02
Remove a large amount of boilerplate parsing from `wasmparser`

alexcrichton:
Recently I was curious to try out a new `linking` section and `reloc` section parser, but I was struck again with how much boilerplate is required to create a section parser in `wasmparser`. I thought about this for a bit and realized that a huge amount of boilerplate could be removed by shifting the traits slightly in `wasmparser`. Prior to this change wasmparser had traits such as:

* `SectionReader` - implemented by anything that could be incrementally read as a section
* `SectionReaderLimited` - implemented for limited sections which had a length prefix for a fixed number of items - also gave an iterator.

Using these however required a large amount of boilerplate. Most sections had inherent methods duplicated with the trait methods, and there were quite a number of trait impls with lots of methods to implement. Furthermore almost all trait impls were exactly the same, just copy/pasted from elsewhere. 

How better to copy/paste code than to force rustc to do it through monomorphization! This commit removes the above two traits and replaces them with:

* `SectionLimited<'a, T>` - a new concrete `struct` for a section which is a list of `T`. This concrete struct is the deduplication of all the common methods/impls amongst all sections. All the bells and whistles can be added here and automatically shared with everything else.
* `FromReader<'a>` - this is a new trait that `SectionLimited` is based on. The `T` implements this trait and this represents the individual ability to parse a single `T` from a `BinaryReader`. In introducing this trait I opted to remove most of the inherent methods to `BinaryReader` and instead make them trait implementations. This makes them more readily accessible to users (as a trait impl) without having all the parsing centralized in one location.
* `Subsections<'a, T>` and `Subsection<'a>` - these are replacements for sections that are not length-delimited, namely the `name` and `component-name` custom sections.

The diff here removes about 2k lines from `wasmparser` and I think overall modernizes the feel of `wasmparser`'s API. There's still a lot to take in from the crate but I at least personally feel that it's a lot less boilerplate than before and easier to wrap my head around and figure out what needs implementing where (without copy/pasting lots of code from other locations)


alexcrichton:
I'll also note that while this is a very large PR the main guts are:

* The first commit, and the introduction of `SectionLimited`
* One of the commits in the middle, showing the pattern of what it looks like to use the new struct/trait
* The final commit, which removes the prior traits

I also forgot to write in the OP but while this is a breaking change for `wasmparser` it's not expected to be a large breaking change. All the same idioms are still possible (or at least should be or it's a bug). I've got further ideas for improving things and boilerplate here but I wanted to land this first as it's code-wise a large change.

Robbepop:
Does this PR also has performance implications for `wasmparser`?

fitzgen:
> Not intentionally, no, but I haven't verified. This should largely "just" be code motion.

Would be nice to double check this.

fitzgen (approved):
Very nice!

alexcrichton:
I've confirmed any perf difference is just noise locally.

PR #861, opened 2022-12-06
Cleanup skipping methods in `BinaryReader`

alexcrichton:
This commit consolidates and simplifies some methods around skipping data in a `BinaryReader`, typically to get a new reader to parse it later. Notably a field of `BinaryReader` is now private and there's less open-coding around the offsets and internal state management.


pchickey (approved):
Another nice refactor!

PR #864, opened 2022-12-12
WIP: implement use for single file .wit

fibonacci1729:
This PR implements resolution of `use` for single file `.wit`. Interfaces are defined and resolved topologically.

Signed-off-by: Brian H <brian.hardock@fermyon.com>

### DIFF
--- a/crates/wit-parser/src/ast.rs
+++ b/crates/wit-parser/src/ast.rs
@@ -168,6 +168,14 @@ pub struct Interface<'a> {
 }
 
 impl<'a> Interface<'a> {
+    pub fn uses(&self) -> impl Iterator<Item = &Use<'a>> {
+        self.items.iter().filter_map(|item| match item {
+            InterfaceItem::Use(item) => Some(item),
+            InterfaceItem::TypeDef(_) => None,
+            InterfaceItem::Value(_) => None,
+        })
+    }
+
     fn parse(tokens: &mut Tokenizer<'a>, docs: Docs<'a>) -> Result<Self> {
         tokens.expect(Token::Interface)?;
         let name = parse_id(tokens)?;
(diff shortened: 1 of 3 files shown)


crates/wit-parser/src/ast.rs
```diff
@@ -192,6 +200,56 @@ impl<'a> Interface<'a> {
 pub enum InterfaceItem<'a> {
     TypeDef(TypeDef<'a>),
     Value(Value<'a>),
+    Use(Use<'a>),
+}
+
+pub struct Use<'a> {
+    pub names: Option<Vec<UseName<'a>>>,
```
alexcrichton, line 207: I think it's ok to remove the `*` syntax since I don't think we really need that much any more and can always add that in later. For now I'm leaving that out of https://github.com/WebAssembly/component-model/pull/141

crates/wit-parser/src/ast/resolve.rs
```diff
@@ -31,22 +35,65 @@ enum Key {
...
+                names
+                    .into_iter()
+                    .map(|name| ast
+                        .interfaces()
+                        .find(|i| &*i.name.name == name)
+                        .unwrap())
+            },
+            SortResults::Partial(_) => {
+                bail!("interface `use` cycle detected");
+            },
```
alexcrichton, line 82: I think it's definitely ok to punt "really good error messages" to later, but I do think this should at least mention one interface (with a span) in the chain of cyclic interfaces

crates/wit-parser/src/ast/resolve.rs
```diff
@@ -141,14 +184,78 @@ impl Resolver {
...
+            let interface_name = &u.from.name;
+            let from_span = u.from.span;
+            match &u.names {
+                Some(names) => {
+                    for name in names {
+                        let (my_name, span) = match &name.as_ {
+                            Some(id) => (&id.name, id.span),
+                            None => (&name.name.name, name.name.span),
+                        };
+                        let mut found = false;
```
alexcrichton, line 203: Instead of a `found` bool here I think the `None` case can be matched on the `qualified_type_names.get` below I think?

crates/wit-parser/src/ast/resolve.rs
```diff
@@ -4,14 +4,18 @@ use anyhow::Result;
 use indexmap::IndexMap;
 use std::collections::{HashMap, HashSet};
 use std::mem;
+use topo_sort::{TopoSort, SortResults};
 
 #[derive(Default)]
 pub struct Resolver {
+    // Mapping of (interface-name, type-name) to type-id for resolving use.
+    qualified_type_names: HashMap<(String, String), TypeId>,
```
alexcrichton, line 12: Instead of a two-level hashmap like this could this perhaps be a `Vec<HashMap<String, TypeId>>` where the position within the vector corresponds to the `InterfaceId` index and within that is the namespace of names within that interface?

alexcrichton (commented):
Looking good to me!

crates/wit-parser/src/ast/resolve.rs
fibonacci1729: How does this look? The following `wit` ...

```
interface x {
    type m = string
    use { m } from y
}

interface y {
    use { m } from x
}
```

... produces the error:
```
Error: interface y contains a `use` cycle
     --> test.wit:6:11
      |
    6 | interface y {
      |
```

crates/wit-parser/src/ast.rs
fibonacci1729: Should i also make a pass through the tests and delete any `use * ...` related ones.

alexcrichton:
I've folded this into https://github.com/bytecodealliance/wasm-tools/pull/867, although moreso conceptually than literally as I'm not doing a great job of persisting history and such. I wanted to note here though the progress on that!

fibonacci1729:
Excellent. I close this for now! Thanks @alexcrichton

Issue #841, opened 2022-11-26
Block names for `br` are not preserved


d3lm:
When parsing WAT into bytes and re-printing it then I have noticed that for blocks the name for the block itself preserved but not for `br` instructions for instance, e.g.
```
block $myBlock
  br $myBlock
end
```
When parsing and printing it again it produces:
```
block $myBlock ;; label = @1 
  br 0 (;@1;)
end
```
Is this a bug or on purpose, or can we change this so that the blocks use the name if it has one?

alexcrichton:
Just an oversight, would be great to have this implemented!

PR #866, opened 2022-12-13
wasmparser: simplify printing of shadowed labels

nagisa:
The webassembly spec test suite exercises label shadowing in the wat format. `wasmprinter` would then utilize the annotations proposal to maintain equivalent behaviour, but the reference interpreter does not support it, which means that simply passing the spec tests through `wasmparser` and `wasmprinter` would make tests unparseable by the reference interpreter.

To fix this avoid the annotations extension for labels which support shadowing.

### DIFF
--- a/crates/wasmprinter/src/lib.rs
+++ b/crates/wasmprinter/src/lib.rs
@@ -481,12 +481,17 @@ impl Printer {
         ) -> Result<()> {
             for indirect in names {
                 let indirect = indirect?;
-                let mut used = HashSet::new();
+                let mut used = match name {
+                    // labels can be shadowed, so maintaining the used names is not useful.
+                    "label" => None,
+                    "local" => Some(HashSet::new()),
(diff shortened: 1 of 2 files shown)


nagisa:
FWIW I believe that this potentially makes fixing https://github.com/bytecodealliance/wasm-tools/issues/841 much more difficult, as the implementation would need to change to maintain a stack of names, rather than a hashmap like it does now. I went ahead and added a test for it ahead of time to make sure we don’t end up producing wrong results in those cases.

alexcrichton (approved):
Looks reasonable to me, thanks! It's ok I dont think that issue will become too noticeably more difficult with this, as your tests show it'll just need a case to fall back to an integer depth if the selected depth can't be jumped to by name.

nagisa:
I made the code slightly neater – dropped the boolean argument and refactored how the `used` hashset is passed around. rustfmt is now passing too.

alexcrichton (approved):

PR #868, opened 2022-12-19
wasmprinter: add snapshot tests for the outputs

nagisa:
This PR saves the output of every wasmprinter test into the `tests/snapshot` directory. Once these are in place, every time wasmprinter's output is changed, these changes will be reflected neatly in these snapshots. This can aid with the detection of issues and also as a mechanism to facilitate discussion of subjective calls on how formatting should be done (i.e. previous formatting was better, or the new formatting is better in this specific example.)

The downside is that this is 2000+ new files and 70k new lines in the repo.


alexcrichton:
This looks great to me, thanks! Would you be up for removing the `snapshots` component of the path though to place the `.print` next to all the tests? That way in diffs things should be naturally sorted where if a test was changed the `*.print` should show up either just before or just after to easily compare in something like the github diff UI.

Otherwise though looks good to go 👍

nagisa:
There are some issues with implementing this. In particular the `tests/testsuite` is a submodule, so adding files to that directory is going to cause problems. Others are fixable, like for example current code uses files within directories to represent all the modules defined in the `wast` files, but this isn’t too big a deal and could be replaced with `foo.wast@{index}.print` without much of an issue.

alexcrichton (approved):
Ah the point about the `testsuite` module is a good one. In that case let's go with this and fix it up later if truly necessary, which it may likely not be required.

PR #867, opened 2022-12-15
Sync WIT with upstream specification, `use`, packages, and all

alexcrichton:
I'm opening this up now to share some early work on this. This is unfortunately going to again be a very large PR due to the sheer amount of infrastructure required to pull in all the WIT updates, namely to implement https://github.com/WebAssembly/component-model/pull/141. By-and-large the headline features of this PR will be:

* [x] The `use` keyword will be fully implemented. This means that WIT now has the concept of packages, dependencies between packages, documents, etc. Packages can be split between multiple files and packages can depend on other packages as well. All of the `use` statements are preserved in the internal representation to enable code generators to see connections between types and generate appropriate code, e.g. one `struct` in Rust.
* [x] A `wit-parser`-specific convention is invented here for a set of packages residing on the filesystem, intended to probably be in a `wit/` folder at the top of a project. The "root package" is located at `wit/*.wit` and dependencies are located at `wit/deps/$dep/*.wit`. Dependencies can use amongst themselves and this should provide a means by which to depend on packages (like WASI) before registry tooling is in place. Submodules may also help here too perhaps.
* [x] Full support in `wit-component` for all of these features. This means that a component can be created from a world which is itself assembled from many packages/files/interfaces/etc.
* [x] The `wasm-tools component new --types-only` interface is removed in favor of `wasm-tools component wit --binary`. More-or-less the binary format for WIT is now supported under `wasm-tools component wit` which can be used for printing the WIT of a preexisting component, converting a WIT binary to WIT text, or converting WIT text to WIT binary.

I'm opening this up as a work-in-progress at this time. I'm almost done with `wit-parser` and the text syntax bits. A good way to explore this is through the tests updated and added in this PR. Procedually I'll be taking next week and the week after off so this is not going to merge and get published until at least January 2023.

---

[...]

PR #871, opened 2023-01-04
Implement exports-introduce-indice semantics

alexcrichton:
This commit implements part of WebAssembly/component-model#129, specifically the portion where an `export` item now introduces a fresh item into the relevant index space. This is going to be used by #867 to implement encoding/decoding of wit documents, specifically for `use` statements.

This commit mostly implements the infrastructure necessary for this change rather than any other major changes internally. Functionally `export` statements now basically behave the same as `alias` statements within the validator. Eventually more validation will be required to ensure that functions use transitively-named types in all locations. Additionally the name that a type index is exported at will be relevant and will need exposing, but that'll come at a later date as well. It's hoped that exposing this information, as required for WIT encoding/decoding, will be a relatively small layer on the validator from this commit.


crates/wasmparser/src/validator/component.rs
```diff
@@ -808,26 +818,7 @@ impl ComponentState {
...
-                                }
-                                _ => unreachable!(),
-                            }
-                        }
-                        _ => return Err(BinaryReaderError::new(
-                            "only outer type aliases are allowed in component type declarations",
-                            offset,
-                        )),
-                    }
+                    Self::add_alias(components, alias, types, offset)?;
```
peterhuene, line 821: This removes the better error message of "only outer type aliases are allowed..." for component/instance type declarations in favor of a more confusing error message of "unknown instance X: instance index out of bounds".

The latter makes it seem like the encoding was legal except the index was wrong when instead instance export aliasing is not legal in type declarations.

I'm of the opinion that if we're going to allow illegal encoding of aliases in type definitions in `wasm-encoder`, we should do the best validation we can against said encodings in `wasmparser`.

crates/wasmparser/src/validator/component.rs
```diff
@@ -863,28 +854,9 @@ impl ComponentState {
...
-                    }
-                    _ => {
-                        return Err(BinaryReaderError::new(
-                            "only outer type aliases are allowed in instance type declarations",
-                            offset,
-                        ))
-                    }
-                },
+                crate::InstanceTypeDeclaration::Alias(alias) => {
+                    Self::add_alias(components, alias, types, offset)?;
```
peterhuene, line 858: Same comment as above re: validation error messages around illegal encodings of aliases in type declarations.

crates/wast/src/component/types.rs
```diff
@@ -800,7 +807,7 @@ impl<'a> Parse<'a> for ComponentTypeDecl<'a> {
         } else if l.peek::<kw::r#type>() {
             Ok(Self::Type(parser.parse()?))
         } else if l.peek::<kw::alias>() {
-            Ok(Self::Alias(Alias::parse_outer_type_alias(parser, false)?))
+            Ok(Self::Alias(parser.parse()?))
```
peterhuene, line 810: More justification for improving the error message in the validator if we're also relaxing the parsing in `wat` in component/instance type declarations.

tests/snapshots/local/component-model/a.wast/0.print
```diff
@@ -21,5 +21,5 @@
     )
   )
   (import "three" (instance (;0;) (type 1)))
-  (export "one" (value $v))
+  (export (;1;) "one" (value $v))
```
peterhuene, line 24: Perhaps we should be printing `(export "one" (value $v (;1;)))`?

To me it's easier to line up the index comment with the index space when it's near the sort, like we do for definitions and imports (i.e. it's not `(import (;x;) ...)`).

peterhuene (commented):
Looks good; I just have a few feedback comments to go over.

peterhuene:
It also looks like some more test baselines need updates.

tests/snapshots/local/component-model/a.wast/0.print
peterhuene: Now that I say that, I guess that could be confused with saying that `$v` has index `1` when it does not.

Fine with keeping as-is, although I think this format does make it harder to grep for "where is item X with index Y declared".

peterhuene: In favor of just leaving this be as the index comment is being printed where the identifier (if it had one) would be parsed in `wat` anyway.

I'll get used to it.

alexcrichton:
Ah so @peterhuene your comments are all spot on, and reminds me of the other major change this implements. This not only introduces `export`-creates-an-alias but it also implements the behavior where all forms of aliases are now supported in instance and component types. Specifically this is enabled by a new test along the lines of:

```wasm
(component
  (type (component
    (type $t u8)
    (import "x" (instance $i (export "t" (type (eq $t)))))
    (alias export $i "t" (type $my-t))
  ))
)
```

This'll be required to get the `use` statement modeled correctly and is required regardless for resources. This was a pretty subtle change from https://github.com/WebAssembly/component-model/pull/129, specifically [here](https://github.com/WebAssembly/component-model/pull/129/files#diff-59f6b20b0bd073593bb31d6338e2186f876564ffe0bb67b8740b37dabf8e5636L219-L220) where the replacement text allows more than just the `outer` alias.

More-or-less the same goal is still achieved, we're just schlepping around type information. Internally though it means that a few more index spaces are maintained.

This introduction of more support for more forms of `alias` in the type context is what led me to the refactorings you pointed out, namely using an `Alias` in `wasm-encoder` to avoid duplicating the encoding/methods across the alias section and the type sections.

tests/snapshots/local/component-model/a.wast/0.print
alexcrichton: Nah I definitely agree that `(export $name ...` isn't as nice as all the other textual forms. I'm hoping we can shave that yak at a later date, however, and figure out a better bikeshed.

peterhuene:
Ah, that makes so much more sense now. At first glance it seemed like a nice, but unrelated, refactoring that was a part of the changes to introduce an index for exported items.

I didn't realize the spec changed to legalize instance export aliases in type definitions. Sorry for the unnecessary feedback then!

alexcrichton:
No worries of course! I forgot to call it out anyway in the OP.

Mind taking a look at the latest commit as well? I'm hoping to head off confused users asking what obscure error messages are ideally. There's still going to be an unfortunate period of "oh well what versions of things do I need to line everything up?" but hopefully that's better than "why am I getting this weird error message"

peterhuene:
I assume we'll now bump it in the spec as well?

crates/wasmparser/src/validator.rs
```diff
@@ -492,28 +492,32 @@ impl Validator {
...
             }
-            (Encoding::Component, WASM_COMPONENT_VERSION) => {
+            Encoding::Component => {
                 if !self.features.component_model {
-                    return Err(BinaryReaderError::new(
-                        "WebAssembly component model feature not enabled",
+                    bail!(
                         range.start,
-                    ));
+                        "unknown binary version: {num:#x}, \
```
peterhuene, line 509: To match the error messages below:

```suggestion
                        "unknown component version: {num:#x}, \
```

peterhuene (approved):

crates/wasmparser/src/validator.rs
```diff
@@ -492,28 +492,32 @@ impl Validator {
...
+                         note the WebAssembly component model feature not enabled?",
```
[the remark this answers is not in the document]
alexcrichton: Whoops that `?` was there from a previous iteration, I like your message better

crates/wasmparser/src/validator.rs
alexcrichton: Heh I wanted to do that but it runs afoul of [this spec test](https://github.com/WebAssembly/testsuite/blob/4f77306bb63151631d84f58dedf67958eb9911b9/binary.wast#L44) which currently somewhat forces our hand unless specific exemptions for specific errors are added. My laziness won out in this case to acquiesce to the test suite's expectation.

crates/wasmparser/src/validator/component.rs
```diff
@@ -215,8 +215,41 @@ impl ComponentState {
         offset: usize,
     ) -> Result<()> {
         let entity = self.check_type_ref(&import.ty, types, offset)?;
+        self.add_entity(entity, false, offset)?;
+        let name = to_kebab_str(import.name, "import", offset)?;
+
+        match self.imports.entry(name.to_owned()) {
```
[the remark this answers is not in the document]
alexcrichton: Agreed would be good, but I think I'll leave this for a future refactoring for now

PR #876, opened 2023-01-06
wasm-encoder: Return Indices from builder functions where applicable

Kylebrown9:
This refactor changes the returns of builder methods which create indexed entries (e.g. types) so that they return the index of the newly created entry. This means users don't need to otherwise calculate the value or know what it will be.

For example, the README no longer needs the line `let type_index = 0;` since it can be obtained when `types.function(...)` is called.
```rust
// Encode the type section.
let mut types = TypeSection::new();
let params = vec![ValType::I32, ValType::I32];
let results = vec![ValType::I32];
let type_index = types.function(params, results);
module.section(&types);

// Encode the function section.
let mut functions = FunctionSection::new();
functions.function(type_index);
module.section(&functions);
```

### DIFF
--- a/crates/wasm-encoder/src/component/aliases.rs
+++ b/crates/wasm-encoder/src/component/aliases.rs
@@ -106,10 +106,11 @@ impl ComponentAliasSection {
     }
 
     /// Define an alias to a component instance's export.
-    pub fn alias(&mut self, alias: Alias<'_>) -> &mut Self {
+    pub fn alias(&mut self, alias: Alias<'_>) -> u32 {
         alias.encode(&mut self.bytes);
+        let index = self.num_added;
         self.num_added += 1;
-        self
+        index
     }
 }
 
--- a/crates/wasm-encoder/README.md
+++ b/crates/wasm-encoder/README.md
@@ -36,12 +36,11 @@ let mut module = Module::new();
 let mut types = TypeSection::new();
 let params = vec![ValType::I32, ValType::I32];
 let results = vec![ValType::I32];
-types.function(params, results);
+let type_index = types.function(params, results);
 module.section(&types);
 
 // Encode the function section.
 let mut functions = FunctionSection::new();
-let type_index = 0;
 functions.function(type_index);
 module.section(&functions);
 
--- a/crates/wasm-encoder/src/component/canonicals.rs
+++ b/crates/wasm-encoder/src/component/canonicals.rs
@@ -83,7 +83,7 @@ impl CanonicalFunctionSection {
     }
 
     /// Define a function that will lift a core WebAssembly function to the canonical ABI.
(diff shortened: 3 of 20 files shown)


crates/wasm-encoder/src/component/aliases.rs
```diff
@@ -106,10 +106,11 @@ impl ComponentAliasSection {
     }
 
     /// Define an alias to a component instance's export.
-    pub fn alias(&mut self, alias: Alias<'_>) -> &mut Self {
+    pub fn alias(&mut self, alias: Alias<'_>) -> u32 {
         alias.encode(&mut self.bytes);
+        let index = self.num_added;
         self.num_added += 1;
-        self
+        index
```
peterhuene, line 113: This is the index within this particular alias section, which is not the same thing as an item index as there can be multiple alias sections.

Additionally, alias indexes don't make sense as the new index depends on what is being aliased.

peterhuene (changes requested):
I don't think this change makes sense for the component model sections because there isn't a 1:1 with section and index space like there is in core modules.

Kylebrown9:
That makes sense, I've reverted all the component model indexing logic for now and we can revisit it at a future point.
@alexcrichton mentioned that for core wasm it's also sometimes incorrect since imports exist at the beginning of the index space for some things.

From the spec:
> The index space for [functions](https://webassembly.github.io/spec/core/syntax/modules.html#syntax-func), [tables](https://webassembly.github.io/spec/core/syntax/modules.html#syntax-table), [memories](https://webassembly.github.io/spec/core/syntax/modules.html#syntax-mem) and [globals](https://webassembly.github.io/spec/core/syntax/modules.html#syntax-global) includes respective [imports](https://webassembly.github.io/spec/core/syntax/modules.html#syntax-import) declared in the same module. The indices of these imports precede the indices of other definitions in the same index space

Reviewing the index spaces, only a few make sense to include:
* typeidx
* ~~funcidx~~ (includes imports)
* ~~tableidx~~ (includes imports)
* ~~memidx~~ (includes imports)
* ~~globalidx~~ (includes imports)
* elemidx
* dataidx
* ~~localidx~~ (not created by builder)
* ~~labelidx~~ (not created by builder)

So, I'm going to further pare this down to types, elements, and data.

alexcrichton:
I personally think it's ok to remove the `-> &mut Self` builder-style pattern since I don't think it's really used all that often with wasm-encoder anyway. In that sesnse I think returning `-> u32` is fine, but while I don't think the `FuncSection` could return raw indices it could still have something like `FuncSection::count(&self) -> u32` which returns the count of items added there. The documentation for the function could indicate that the count must be added to the number of imports to find the actual index.

For consistency though it might make the most sense to have `len`-style methods on most section builders rather than some returning indices and some returning `&mut Self` perhaps?

PR #878, opened 2023-01-11
wit-component: expose dummy component generation

guybedford:
This exposes the dummy core Wasm component generation, which is useful for binding generation code independent of implementation.

Specifically, I'd like to expose this in order to enable JS binding generation via a similar approach to the wit-bindgen demo.


alexcrichton:
This seems reasonable enough to me, I know I've needed this in quite a few places. 

The precise way it's exposed here I don't think is necessarily the best, but I'm going to merge anyway and I'll handle the merge conflicts in https://github.com/bytecodealliance/wasm-tools/pull/867. Thanks @guybedford!

PR #880, opened 2023-01-18
VisitOperator: implement delegate from &mut VisitOperator

nagisa:
This allows generic code to work with both VisitOperators passed by value and by reference, without client code needing to implement a wrapping type that would act as a container for the reference.


alexcrichton (approved):
Thanks!

PR #882, opened 2023-01-20
Add experimental memory.discard instruction

bvisness:
Adding support for a `memory.discard` instruction proposed in the [memory control proposal](https://github.com/WebAssembly/design/issues/1439), which I am prototyping within SpiderMonkey.

@eqrion

### DIFF
--- a/crates/wasmparser/src/lib.rs
+++ b/crates/wasmparser/src/lib.rs
@@ -338,6 +338,11 @@ macro_rules! for_each_operator {
             @reference_types TableGrow { table: u32 } => visit_table_grow
             @reference_types TableSize { table: u32 } => visit_table_size
 
+            // OxFC prefixed operators
+            // memory control (experimental)
+            // https://github.com/WebAssembly/design/issues/1439
+            @memory_control MemoryDiscard { mem: u32 } => visit_memory_discard
+
             // 0xFE prefixed operators
             // threads
             // https://github.com/WebAssembly/threads
--- a/crates/wasmprinter/src/operator.rs
+++ b/crates/wasmprinter/src/operator.rs
@@ -279,6 +279,12 @@ macro_rules! define_visit {
             $self.memory_index($mem)?;
         }
     );
+    (payload $self:ident MemoryDiscard $mem:ident) => (
+        if $mem != 0 {
+            $self.push_str(" ");
+            $self.memory_index($mem)?;
+        }
+    );
     (payload $self:ident I32Const $val:ident) => (write!($self.result(), " {}", $val)?);
     (payload $self:ident I64Const $val:ident) => (write!($self.result(), " {}", $val)?);
     (payload $self:ident F32Const $val:ident) => (
@@ -362,6 +368,7 @@ macro_rules! define_visit {
     (name MemoryInit) => ("memory.init");
     (name MemoryCopy) => ("memory.copy");
     (name MemoryFill) => ("memory.fill");
+    (name MemoryDiscard) => ("memory.discard");
     (name DataDrop) => ("data.drop");
     (name ElemDrop) => ("elem.drop");
     (name TableInit) => ("table.init");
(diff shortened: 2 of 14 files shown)


tests/local/memory-discard.wat
```diff
@@ -0,0 +1,7 @@
+(module
+  (memory i32 16)
+
+  (func
+    (memory.discard (i32.const 0) (i32.const 16))
```
bvisness, line 5: This is actually not great since my prototype uses bytes instead of pages, but when I update this test + snapshots again (`BLESS=1 cargo test`) it obliterates most of the snapshots in the repo. Trying to sort that out.

crates/wasmparser/src/lib.rs
```diff
@@ -325,6 +325,7 @@ macro_rules! for_each_operator {
             @bulk_memory DataDrop { data_index: u32 } => visit_data_drop
             @bulk_memory MemoryCopy { dst_mem: u32, src_mem: u32 } => visit_memory_copy
             @bulk_memory MemoryFill { mem: u32 } => visit_memory_fill
+            @bulk_memory MemoryDiscard { mem: u32 } => visit_memory_discard
```
alexcrichton, line 328: The `@bulk_memory` here is used as a proposal to gate access to this instruction, so this'll want to be something other than `bulk_memory` since otherwise `memory.discard` will be enabled by default.

In lieu of an official proposal, could this perhaps be something like `@experimental` or `@memory_control`? That'll require adding a field to `WasmFeatures` which is ok and would be necessary for this anyway.

crates/wasmprinter/src/operator.rs
```diff
@@ -279,6 +279,12 @@ macro_rules! define_visit {
             $self.memory_index($mem)?;
         }
     );
+    (payload $self:ident MemoryDiscard $mem:ident) => (
+        if $mem != 0 {
+            $self.push_str(" ");
+            $self.memory_index($mem)?;
+        }
+    );
```
alexcrichton, line 287: If you'd like I think it's ok to remove this special case since not printing `memory.grow 0` is mostly just a compat with MVP wasm at this point. I think instructionsl ike `table.grow` always print their table index as they're newer, so always printing the memory index with `memory.discard` sounds ok to me at least.

alexcrichton (approved):
Looks good to me, thanks! With a new fresh proposal gating this instruction happy to merge

crates/wasmparser/src/lib.rs
bvisness: I'll add `@memory_control` I think!

bvisness: Added a `@memory_control` feature. I hope I set true/false correctly in the various places where it was required.

crates/wasmprinter/src/operator.rs
bvisness: I tried removing this `MemoryDiscard` case entirely but I get the following error:

```
error[E0599]: no method named `mem` found for mutable reference `&mut PrintOperator<'_, '_>` in the current scope
   --> crates/wasmprinter/src/operator.rs:851:5
    |
851 |     wasmparser::for_each_operator!(define_visit);
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: there is a method with a similar name: `memarg`
    |
    = note: this error originates in the macro `wasmparser::for_each_operator` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Not sure how to fix that, but just let me know and I'll update it.

alexcrichton: Ah ok there's a possible fix for this but I'll handle it later, so this is fine as-is. Thanks for testing though!

alexcrichton:
All looks good to me, thanks again! I think CI is showing there's at least one more place to update `WasmFeatures` though

bvisness:
Just patching up those last couple places. Didn't realize I had to run `cargo test --all`.

alexcrichton:
I stepped over myself a bit during publication but these crates are now all published on crates.io

PR #883, opened 2023-01-20
VisitOperator: implement delegates for Box<VisitOperator>

nagisa:
This is also fine because `Box` implements `DerefMut`. Pretty useful for people (like me) who have a `Box<dyn VisitOperator>`, but can't use `&dyn` because of the borrowck limitations around how it handles HRTBs.


nagisa:
It would be awesome to have a wasmparser release once this lands.

alexcrichton (approved):

alexcrichton:
Sure yeah I can publish today. I think https://github.com/bytecodealliance/wasm-tools/pull/882 will want a publish shortly after merging as well so I'll get that too.

alexcrichton:
I stepped over myself a bit during publication but these crates are now all published on crates.io

PR #885, opened 2023-01-23
Fix references to imported types in the text format

alexcrichton:
During the topological sorting of types there wasn't an entry in the
dependency array for imported types which ended up leading to errors
about the types not being defined when, in fact, they are defined!


pchickey (approved):

alexcrichton:
After thinking about this a bit more I was wondering why fuzzing hadn't caught this, and I found a bug in the fuzzer after investigating which is turning up more fuzz bugs, so I'm going to back out the version bump here and fix the fuzz-related issues before bumping.

PR #886, opened 2023-01-23
Fix a slew of fuzz bugs in wit-{parser,component}

alexcrichton:
This PR fixes the `roundtrip-wit` fuzz target to actually be able to generate `use` statements. An accidental bug meant that the fuzzer previously didn't actually create `use` statements, despite that being one of the main goals of the fuzzer. Alas! 

In any case this turned up a number of bugs and additionally found https://github.com/bytecodealliance/wasm-tools/pull/885 in no time at all.


alexcrichton:
Note that this is a draft while the fuzzer is still running as I've got more bugs to fix, and I'll keep pushing them here as I discover them.

alexcrichton:
Ok there ended up being quite a few more issues than I thought there were going to be. @peterhuene I think this is ready to go as I've got 70 cores of fuzzing being clean for ~30 minutes now. I've tried to split everything up where each commit fixes an individual problem and a single fuzz test case. As always though I'm happy to dive into more details with you or expand comments and such.

fuzz/fuzz_targets/roundtrip-wit.rs
```diff
@@ -462,7 +502,6 @@ mod generate {
                 ret.push_str("\n\n");
             }
 
-            self.types_in_interface.clear();
```
peterhuene, line 465: Whoops, sorry for not catching that either!

peterhuene (approved):
Looks 👍. Just minor nits.

fuzz/fuzz_targets/roundtrip-wit.rs
alexcrichton: Heh I'm the one who ran the fuzzer for hours without noticing this 🤡

PR #887, opened 2023-01-23
`wasm-smith`: Add support for the Wasm tail calls proposal

fitzgen:


alexcrichton (approved):
Seems reasonable to me! Also FWIW the on-by-default features of wasm-smith typically try to match the on-by-default features of wasmparser, so mind turning on the tail-calls feature by default in wasmparser now that it's a stage 4? We'll turn it off explicitly in Wasmtime still, however.

fitzgen:
Both `wasmparser` and `wasm-smith` disable the proposal by default right now. Are you suggesting enabling it by default?

alexcrichton:
Oh indeed, let me clarify. For off-by-default proposals we currently generally leave them as `false` in `SwarmConfig` rather than leaving it as `u.arbitrary()` which may mean it's turned on. If you'd like to leave it as off-by-default I think that should be hardcoded as `false` (and remove the need for the latest commit).

That all being said I think it's reasonable to have as on-by-default because that's what we generally do for stage 4 proposals.

fitzgen:
Ahhh gotcha yes of course. I'll flip the defaults since it is stage 4.

alexcrichton (approved):

PR #890, opened 2023-01-24
wasm-mutate: Add an "AddCustomSection" mutator

itsrainy:
#415

### DIFF
--- a/crates/wasm-mutate/src/lib.rs
+++ b/crates/wasm-mutate/src/lib.rs
@@ -17,7 +17,8 @@ mod mutators;
 pub use error::*;
 
 use crate::mutators::{
-    add_function::AddFunctionMutator, add_type::AddTypeMutator, codemotion::CodemotionMutator,
+    add_custom::AddCustomSectionMutator, add_function::AddFunctionMutator,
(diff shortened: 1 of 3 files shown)


fitzgen (approved):

davidmdm:
How do I use this in the context of wasm-tools to add a custom-section to a wasm binary? The help text is not very informative and my investigations have led me here... Which indicates this should be possible via `wasm-tools mutate`... Any help would be great.

itsrainy:
Hey @davidmdm, I'm not involved with this project any more, so you may want to file an issue to get your question answered (people also tend to pretty responsive in the BA zulip: https://bytecodealliance.zulipchat.com/). From what I can remember, this was primarily for use in fuzzing and is meant for adding a ~random custom section to a wasm module, so I don't think this is the tool you are looking for.

fitzgen:
> this was primarily for use in fuzzing and is meant for adding a ~random custom section to a wasm module, so I don't think this is the tool you are looking for.

This is correct. @davidmdm if you want to add your own custom section to an existing Wasm module, you should use the `wasmparser` and `wasm-encoder` crates instead.

PR #893, opened 2023-01-26
Change the canonical ABI name of for top level functions in worlds

alexcrichton:
For a WIT world of the form:

    world foo {
        import foo: func()
    }

previously this would require a core wasm module along the lines of:

    (module
        (import "" "foo" (func))
    )

where notably the module name of the import was the empty string, indicating that it was corresponding to a root-level function. Unfortunately, though, LLVM doesn't easily support this from guest languages. For example this Rust program:

    #[link(wasm_import_module = "")]
    extern "C" {
        fn foo();
    }

    #[no_mangle]
    pub unsafe extern "C" fn e() {
        foo();
    }

generates a wasm module that looks like:

    (module
      (type (;0;) (func))
      (import "env" "foo" (func $_ZN3foo3foo17hc95a7373bc5389c2E (;0;) (type 0)))
      ;; ..
    )

Note here how the `""` empty string was replaced by `env`, the default if nothing else is specified. While this is probably an LLVM bug that should be fixed the timescale of that is long enough that it's not really worth persisting this idiom, so this PR switches the expected module name of root-level function imports to be `$root` now instead of the empty string. This gets LLVM to work and preserves the property that the module name can't otherwise clash with a valid interface name.

### DIFF
--- a/crates/wit-component/src/dummy.rs
+++ b/crates/wit-component/src/dummy.rs
@@ -11,7 +11,7 @@ pub fn dummy_module(resolve: &Resolve, world: WorldId) -> Vec<u8> {
             WorldItem::Function(func) => {
                 let sig = resolve.wasm_signature(AbiVariant::GuestImport, func);
 
-                wat.push_str(&format!("(import \"\" \"{}\" (func", func.name));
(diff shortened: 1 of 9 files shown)


pchickey (approved):
pretty reasonable workaround, but bummer about the LLVM issue

PR #894, opened 2023-01-26
Fix a bug in wit-component when a function is only used within an ada…

alexcrichton:
…pter

This commit fixes a bug in `wit-component` when used with adapters when an imported interface or function is used exclusively within an adapter that otherwise isn't needed by the main module. The fix here was to alter the order of iteration when calculating shims to iterate over all adapters instead of only iterating over the adapters required by the main module.


crates/wit-component/src/encoding.rs
```diff
@@ -753,6 +752,10 @@ impl<'a> EncodingState<'a> {
                     &mut signatures,
                 );
             }
+            let funcs = match self.info.info.adapters_required.get(adapter) {
```
peterhuene, line 755: Nit:

```suggestion
            let funcs = match info.adapters_required.get(adapter) {
```

peterhuene (approved):

PR #897, opened 2023-01-27
Add CI automation for binary releases

alexcrichton:
This commit adds CI that happens on pushes to `main` in addition to tags which produce GitHub Releases and publish the `wasm-tools` binary artifact to the release. I figure that as this project is relied on more for component-based tooling it might be good to have at least a basic version of binary releases to avoid having to always build from source.

Much of the tooling here is copied over from Wasmtime, notably the bits used to produce more compatible binary releases with older versions of glibc.


.github/workflows/build.yml
```diff
@@ -0,0 +1,64 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: write
+
+jobs:
+  build:
+    name: Build wasmtime
```
fitzgen, line 25:
```suggestion
    name: Build wasm-tools
```

fitzgen (approved):

.github/workflows/build.yml
alexcrichton: Oh dear my copy/paste is showing

PR #900, opened 2023-02-01
use main module's realloc if available when applying an adapter

dicej:
Originally, adapters were expected to allocate their memory using `memory.grow`, and `wit-component` would also use `memory.grow` to allocate stack space for the adapter if needed.  However, it turned out this didn't work reliably with `wasi-sdk`'s libc, which assumed it owned the entire memory region, leading to chaos when both the adapter and the main module thought they owned the same pages.

Alex has since patched `wasi-sdk` to address the issue, but it still affects older toolchains, and particularly modules which were built prior to the patch. However, many of those older modules export `cabi_realloc` (or it's older equivalent, `canonical_abi_realloc`), so we might as well use that if it's available.  That's what this commit does.

In cases where the main module exports a realloc function and we have a use for it (i.e. either the adapter imports `cabi_realloc` or it needs a shadow stack), we'll use it.  In cases where we have a use for it but the main module does _not_ export it, we fall back to the old behavior of using `memory.grow` and hope the module was built with a recent-enough `wasi-sdk`.

This also addresses alternative toolchains which are not based on `wasi-sdk` and may expect to own the module's entire memory.  In that case, the toolchain can simply export `cabi_realloc` and `wit-component` will use it.

Signed-off-by: Joel Dice <joel.dice@fermyon.com>

### DIFF
--- a/crates/wit-component/src/gc.rs
+++ b/crates/wit-component/src/gc.rs
@@ -567,6 +567,8 @@ impl<'a> Module<'a> {
 
         let sp = self.find_stack_pointer()?;
 
+        let mut func_names = Vec::new();
+
         if let (Some(realloc), Some(_), None) = (main_module_realloc, sp, realloc_index) {
             // The main module exports a realloc function, and although the adapter doesn't import it, we're going
             // to add a function which calls it to allocate some stack space, so let's add an import now.
@@ -580,6 +582,7 @@ impl<'a> Module<'a> {
                 realloc,
                 EntityType::Function(add_realloc_type(&mut types)),
             );
+            func_names.push((num_func_imports, realloc));
             num_func_imports += 1;
         }
 
@@ -590,7 +593,7 @@ impl<'a> Module<'a> {
                 Definition::Import(_, _) => {
                     // The adapter is importing `cabi_realloc` from the main module, but the main module isn't
                     // exporting it.  In this case, we need to define a local function it can call instead.
-                    realloc_index = Some(funcs.len());
+                    realloc_index = Some(num_func_imports + funcs.len());
                     funcs.function(ty);
                     code.function(&realloc_via_memory_grow());
                 }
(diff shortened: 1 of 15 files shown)


crates/wit-component/src/gc.rs
```diff
@@ -489,11 +617,31 @@ impl<'a> Module<'a> {
...
+
+        // Now that we know how many function imports there are, we can compute the final index of our realloc
+        // function.
+        let realloc_index = realloc_index.map(|index| {
+            index
+                + if main_module_realloc.is_some() {
+                    0
+                } else {
+                    num_func_imports
+                }
```
alexcrichton, line 636: I was having a tough time wrapping my head around this, but I think this only affects the calculation of the `realloc_index` on lines 593 and 623, right? (the "it's imported but not provided" and "it's needed for the stack but not otherwise" cases) I think on those lines the `num_func_imports` variable here is ready to go and can get added in.

crates/wit-component/tests/components/adapt-inject-stack-with-reallocing-adapter/component.wat
```diff
@@ -0,0 +1,158 @@
...
+    (export "cabi_realloc" (func $cabi_realloc))
+  )
+  (core module (;1;)
+    (type (;0;) (func (param i32)))
+    (type (;1;) (func (result i32)))
+    (type (;2;) (func (param i32 i32 i32 i32) (result i32)))
+    (type (;3;) (func))
+    (import "env" "memory" (memory (;0;) 0))
+    (import "new" "get-two" (func $get_two (;0;) (type 0)))
+    (import "__main_module__" "cabi_realloc" (func (;1;) (type 2)))
```
alexcrichton, line 56: As a very minor nit, could this import, which is synthesized by the gc pass, get a name `cabi_realloc`?

alexcrichton (approved):
Looks great to me, thanks! I've got two minor comments but otherwise happy to merge.

crates/wit-component/src/gc.rs
dicej: You're right -- I moved the code around so many times, I lost track.

PR #904, opened 2023-02-02
Fix Windows CI

alexcrichton:
Also bump the version of the `wasm-tools` umbrella crate to trigger CI again.


alexcrichton:
The release ended up working so I backed out the bump here.

pchickey (approved):

PR #910, opened 2023-02-04
feat: direct signature api

guybedford:
This makes a minor change to `wasm_signature` to support a form that provides the complete ABI signature for a function without the import / export call signature truncations when providing `None` in place of `AbiVariant`.

I found I needed this information in bindgen work and there was no other API available to just get the full untruncated core type, without a bunch of copying or reimplementation.

While I understand this change is slightly noisy, it is also in theory a negative diff (by character count at least!). I'm very much open to API suggestions as to how to better make this available, but as such a fundamental part of the component model it would be great to have easy access to this information.

### DIFF
--- a/crates/wit-parser/src/abi.rs
+++ b/crates/wit-parser/src/abi.rs
@@ -700,19 +700,23 @@ pub trait Bindgen {
 impl Resolve {
     /// Get the WebAssembly type signature for this interface function
     ///
+    /// The ABI variant determines the use of indirect params and a retptr.
+    /// If passing no variant, the full signature will be provided without
+    /// the use of indirect params or a retptr.
+    ///
     /// The first entry returned is the list of parameters and the second entry
     /// is the list of results for the wasm function signature.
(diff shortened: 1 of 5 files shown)


crates/wit-parser/src/abi.rs
```diff
@@ -750,7 +754,7 @@ impl Resolve {
         }
     }
 
-    fn push_wasm(&self, variant: AbiVariant, ty: &Type, result: &mut Vec<WasmType>) {
```
guybedford, line 753: It turns out `variant` isn't used in `push_wasm` or its callees at all.

alexcrichton:
I would prefer to not expose the functionality this way since there's no real correspondance to anything in the component model with a "directionless" API per se. Could this be done instead, though, by exposing the `push_wasm` method?

guybedford:
I may be able to write a custom conversion to handle the use case, I'll report back how it goes.

guybedford:
I was able to successfully work around the issue - the only remaining problem was retptr sizing, which can be handled by passing `resolve` to `SizeAlign.fill`.

PR #912, opened 2023-02-06
Disable float types with `floats = false`

### LINKED ISSUE #909: `WasmFeatures::floats = false` allows using float types in parameters and locals
Examples of code that is successfully validated even if `floats: false`:

```wat
(module
	(func $foo (local f32)
		local.get 0
		drop
	)
)

(module
	(func $foo (param f32)
		local.get 0
		drop
	)
)
```

I think wasmparser only scans for specific instructions and doesn't check types.

alexcrichton:
This commit updates the validator to reject the `f32` and `f64` types if the `WasmFeatures::floats` field is disabled.

Closes #909

### DIFF
--- a/crates/wasmparser/src/validator.rs
+++ b/crates/wasmparser/src/validator.rs
@@ -222,8 +222,7 @@ pub struct WasmFeatures {
     /// Whether or not floating-point instructions are enabled.
     ///
     /// This is enabled by default can be used to disallow floating-point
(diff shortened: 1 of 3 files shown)


pchickey (approved):

Robbepop:
Will this also disallow floating point instructions and types in `ConstExpr` such as global initializers?

PR #913, opened 2023-02-07
Implement export type ascription

alexcrichton:
This commit implements the change described in
WebAssembly/component-model#155 and accounts for the update in WebAssembly/component-model#161. The internal plumbing for this change was relatively easy to implement and shouldn't cause too much of a burden even on Wasmtime as a consumer (it may not have to change at all). I'll be working on a follow-up for `wit-component` to use this new construct to fix the original underlying issue.

### DIFF
--- a/crates/wasm-compose/src/encoding.rs
+++ b/crates/wasm-compose/src/encoding.rs
@@ -1225,7 +1225,7 @@ impl<'a> CompositionGraphEncoder<'a> {
                 }
             };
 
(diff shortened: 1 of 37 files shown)


alexcrichton:
I should also mention that this changes the binary encoding of components and subsequently bumps the version in the binary format header.

pchickey (approved):

peterhuene (approved):

PR #917, opened 2023-02-07
Reduce the code size of some wasm-mutate functions

alexcrichton:
I tried compiling the `wasm-tools` CLI itself to wasm and in debug mode it wouldn't validate in Wasmtime or `wasm-tools` itself because a few functions used more than 50k locals. These functions made heavy use of macros internally so I've refactored them to lean less on macros.

A nice side benefit of these commits is that the compile time of the `wasm-mutate` crate, in debug mode, on my machine went from 23s to 6s. Yay!

### DIFF
--- a/crates/wasm-mutate/src/lib.rs
+++ b/crates/wasm-mutate/src/lib.rs
@@ -32,81 +32,6 @@ use std::sync::Arc;
 #[cfg(feature = "clap")]
 use clap::Parser;
 
-macro_rules! define_mutators {
-    (@count) => {0};
-    (@count $first: expr , $( $tail: expr ,) *) => { 1 + define_mutators!(@count $($tail , )*) };
-    (@expand $self:ident, $discriminator: ident, $start: expr, . , $( $rest: expr ,)*) => {};
-    // The dot (.) is the mark to avoid infinite recursion , something like and
-    // LR parser
-    (@expand $self:ident, $discriminator: ident, $start: expr, $first: expr , $( $head: expr ,)*  . , $( $rest: expr ,)*) => {
-        if $discriminator == $start {
-            // Start by the current node
-            let m = $first;
-
-            let can_mutate = m.can_mutate($self);
-            log::trace!("Can `{}` mutate? {}", m.name(), can_mutate);
-            if can_mutate {
-                log::debug!("attempting to mutate with `{}`", m.name());
-                match m.clone().mutate($self) {
-                    Ok(iter) => {
-                        log::debug!("mutator `{}` succeeded", m.name());
-                        return Ok(Box::new(iter.into_iter().map(|r| r.map(|m| m.finish()))))
-                    }
-                    Err(e) => {
-                        log::debug!("mutator `{}` failed: {}", m.name(), e);
-                        return Err(e);
-                    }
-                }
-            }
-            // Follow the tail
-            $(
-                let m = $rest;
-
-                let can_mutate = m.can_mutate($self);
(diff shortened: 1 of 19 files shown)


crates/wasm-mutate/src/lib.rs
```diff
@@ -280,46 +205,57 @@ impl<'wasm> WasmMutate<'wasm> {
...
+            let can_mutate = m.can_mutate(self);
+            log::trace!("Can `{}` mutate? {}", m.name(), can_mutate);
+            if !can_mutate {
+                continue;
+            }
+            log::debug!("attempting to mutate with `{}`", m.name());
+            match m.mutate(self) {
+                Ok(iter) => {
+                    log::debug!("mutator `{}` succeeded", m.name());
+                    return Ok(Box::new(iter.into_iter().map(|r| r.map(|m| m.finish()))));
```
fitzgen, line 251: Interesting! I don't know if you remember, but this was what we originally attempted to write this as, but it was failing the borrow checker in like one of the last things that actually require polonious to fix. But seems not to require that anymore? Not sure what changed! Anyways, super happy to be rid of that nasty macro!!

crates/wasm-mutate/src/mutators.rs
```diff
@@ -92,7 +92,7 @@ pub trait Mutator {
     /// `std::iter::once`, to give the fuzzer a chance to choose a new kind of
     /// mutation.
     fn mutate<'a>(
-        self,
+        &self,
```
fitzgen, line 95: I guess this is what made a difference?

crates/wasm-mutate/src/mutators/peephole/rules.rs
```diff
@@ -19,656 +19,687 @@ impl PeepholeMutator {
...
+                    applier: egg::ConditionalApplier {
+                        condition: cond[1].clone(),
+                        applier: rhs,
+                    },
+                },
+            )
+            .unwrap(),
+            _ => unimplemented!(),
+        });
+    }
```
fitzgen, line 667: Rather than forking `egg` code, what if we just split the rule set into a bunch of different functions based on how they are grouped together? Not required, but it would be nice to avoid copying code out if we can...

fitzgen (approved):
Nice!!

crates/wasm-mutate/src/lib.rs
alexcrichton: Hm good point! I suspect it is related to the by-val-`self` change which I actually just did as a refactoring so it would work with the dynamic dispatch array I had (and I noticed nothing actually needed the by-value). But hey if it works it works!

crates/wasm-mutate/src/mutators/peephole/rules.rs
alexcrichton: While possible I'm honestly not too gung-ho about redoing this in the sense that it isn't really changing that much and we don't have a ton of eyes on it at this time. We're pretty far behind `egg` at this point now as well as well so sticking close to `egg` is sort of a sailed ship at this point. 

All that being said though I think that this variant is concretely less code since it's moving all the rule parsing/compliation into one function instead of having it inlined and monomorphized in lots of locations. The current `egg` API isn't too different from this so updating in theory wouldn't be too bad.

I suppose another way to put all that: how strongly do you feel that this should continue to use the macro as-is from `egg`?

fitzgen: Not strong enough to really argue the point. Let's merge!

PR #918, opened 2023-02-08
Encode exported instances with nested components

alexcrichton:
This commit leverages the type ascription added to the component model in #913 to reimplement how exported instances work within `wit-component`. Instead of bag-of-exports they're now exported with instantiations of nested components, where each nested component imports functions and then simply reexports them with the appropriate type ascribed.


crates/wit-component/src/encoding/types.rs
```diff
@@ -178,6 +179,22 @@ pub trait ValtypeEncoder<'a> {
         })
     }
 
+    /// Optionally imports `id` from a different interface, returning the index
+    /// of the imported type into this index space.
+    ///
+    /// Returns `None` if `id can't be imported.
```
peterhuene, line 185:
```suggestion
    /// Returns `None` if `id` can't be imported.
```

peterhuene (approved):

PR #919, opened 2023-02-09
wit-component: lazily allocate adapter stack when appropriate

dicej:
Previously, I modified wit-component to use the main module `cabi_realloc` export to allocate the adapter stack when present. However,
https://github.com/bytecodealliance/preview2-prototyping/issues/78 pointed out that this can be problematic when `cabi_realloc` calls back into the adapter, e.g. as part of `__wasi_call_ctors`.  Such reentrance can either fail immediately because the function aliases have not yet been initialized or else recurse infinitely if the reentered function tries to allocate.

This commit introduces a new `allocation_state` global variable which, if found in an adapter, is used to facilitate lazy stack allocation. If this variable is available, wit-component instruments every function exported by the adapter with a call to a synthesized `allocate_stack` function which checks `allocation_state` and initiates allocation if it is zero, setting it immediately to one prior to calling `cabi_realloc`, thereby short-circuiting any recursion.  Once allocation is finished, `allocation_state` is set to two.

The adapter's responsibility is to also check `allocation_state` in any function it exports which is likely to be called from the main module's `cabi_realloc` function.  If it is zero or one, the function should return immediately (with whatever dummy return value is appropriate) without trying to use the stack or do an allocation.

Note that if the adapter does _not_ have an `allocation_state` global, we fall back to allocating the stack using memory.grow even if the main module has a `cabi_realloc` function.  This is because we can't be sure `cabi_realloc` won't try to reenter the adapter, and without `allocation_state` we have no way to notice and handle that case correctly.  Technically, we could add a synthetic global, but the adapter wouldn't know about it, so it would not know when to do the immediate dummy return discussed above.

Needless to say, this wit-component <-> adapter relationship has become pretty complicated, but that seems necessary in order to support existing, in-the-wild modules.

I'll follow this PR up with a corresponding one two the preview2Prototyping repo.

Signed-off-by: Joel Dice <joel.dice@fermyon.com>

### DIFF
--- a/crates/wit-component/src/encoding.rs
+++ b/crates/wit-component/src/encoding.rs
@@ -40,11 +40,30 @@
 //! * They cannot define any `elem` or `data` segments since otherwise there's
 //!   no knowledge ahead-of-time of where their data or element segments could
 //!   go. This means things like no panics, no indirect calls, etc.
-//! * Only one mutable global is allowed and it's assumed to be the stack
-//!   pointer. This stack pointer is automatically configured with an injected
-//!   `start` function that is allocated with `memory.grow (i32.const 1)`,
-//!   meaning that the shim module has 64k of stack space and no protection if
-//!   that overflows.
+//! * If the adapter uses a shadow stack, the global that points to it must be a
+//!   mutable `i32` named `__stack_pointer`. This stack is automatically
+//!   allocated with an injected `allocate_stack` function that will either use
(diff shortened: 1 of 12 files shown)


dicej:
Will fix tests -- sorry for the noise.

alexcrichton (commented):
Seems reasonable to me! Could you add a test as well for this to review the output in `*.wat` too?

alexcrichton (approved):

PR #926, opened 2023-02-11
Fix validation panic with unreachable code

alexcrichton:
This fixes a fuzz-found regression from #701 where unreachable code could end up triggering a panic during validation.


Robbepop:
Would it be possible (not inconvenient) if we could backport this fix to version `0.100.0`? I can understand if this causes too much friction.

alexcrichton:
#701 wasn't included in 0.100.0, so I don't believe it's affected.

Robbepop:
> #701 wasn't included in 0.100.0, so I don't believe it's affected.

Ah perfect! Sorry then for my confusion.

peterhuene (approved):

PR #928, opened 2023-02-13
Add a function to extract a world from a document

alexcrichton:
This pulls in bytecodealliance/wit-bindgen#494 as an "official" function to the `wit-parser` crate. I've written an analogue for that function in many locations so the hope is that all bindings generators can standardize on the necessary idioms by using this function. For example the Wasmtime bindings generator will want to update to use this as well.


alexcrichton:
cc @sunfishcode

PR #932, opened 2023-02-14
Fix metadata/strip on components with core modules

alexcrichton:
This fixes the `metadata add` and `strip` loops in the `wasm-tools` CLI for components to correctly handle nested components and nested modules. Previously the output contents would get corrupted due to modules showing up multiple times by accident.


pchickey (approved):
Hmm that is subtle! Thanks for finding this

PR #933, opened 2023-02-16
feat: add binary ranges to metadata

guybedford:
This extends the metadata to include the range of the modules being associated with metadata. This enables some simple usage scenarios like being able to calculate byte sizes for nested modules and even extracting core modules from their offsets.

### DIFF
--- a/crates/wasm-metadata/src/lib.rs
+++ b/crates/wasm-metadata/src/lib.rs
@@ -344,7 +344,7 @@ pub enum Metadata {
         /// The module's producers section, if any.
         producers: Option<Producers>,
         /// Byte range of the module in the parent binary
-        range: Range<usize>
+        range: Range<usize>,
     },
 }
 
@@ -359,7 +359,9 @@ impl Metadata {
                 Version { encoding, .. } => {
                     if metadata.is_empty() {
                         match encoding {
-                            wasmparser::Encoding::Module => metadata.push(Metadata::empty_module(0..input.len())),
+                            wasmparser::Encoding::Module => {
+                                metadata.push(Metadata::empty_module(0..input.len()))
+                            }
                             wasmparser::Encoding::Component => {
                                 metadata.push(Metadata::empty_component(0..input.len()))
                             }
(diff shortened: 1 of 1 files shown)


crates/wasm-metadata/src/lib.rs
```diff
@@ -444,7 +453,11 @@ impl Metadata {
     fn display(&self, f: &mut fmt::Formatter, indent: usize) -> fmt::Result {
         let spaces = std::iter::repeat(" ").take(indent).collect::<String>();
         match self {
-            Metadata::Module { name, producers } => {
+            Metadata::Module {
+                name,
+                producers,
+                range: _,
```
pchickey, line 459: A better idiom for unused fields would be `Module { name, producer, .. }` which allows all unnamed fields to be unbound
```suggestion
                ..
```

crates/wasm-metadata/src/lib.rs
```diff
@@ -459,6 +472,7 @@ impl Metadata {
                 name,
                 producers,
                 children,
+                range: _,
```
pchickey, line 475:
```suggestion
                ..
```

crates/wasm-metadata/src/lib.rs
```diff
@@ -797,7 +827,11 @@ mod test {
 
         let metadata = Metadata::from_binary(&module).unwrap();
         match metadata {
-            Metadata::Module { name, producers } => {
+            Metadata::Module {
+                name,
+                producers,
+                range: _,
```
pchickey, line 833:
```suggestion
                ..
```

crates/wasm-metadata/src/lib.rs
```diff
@@ -825,7 +859,11 @@ mod test {
...
```
pchickey, line 865:
```suggestion
                ..
```

pchickey (approved):
Just an idiom feedback. You'll probably have to run rustfmt after accepting the suggestions so feel free to reject them and do it locally if you prefer

PR #935, opened 2023-02-17
Fix component model bugs in `wasmparser` and `wasm-compose`.

peterhuene:
* `wasmparser`: fix component validation for type instantiation arguments.

  This PR fixes validation when a component is instantiated that imports a
  type and the type is provided as an instantiation argument.

  Previously, the type was being ignored and not populated in the list of
  instantiation arguments, resulting in an incorrect validation message of
  "instantiation argument not provided".

  The fix is to add type arguments to the arguments list.

* `wasm-compose`: fix incorrect type indexes caused by type imports.

  Currently `wasm-compose` writes all core and component types out in
  contiguous sections.

  The problem with this approach is that imports of types occur *after* these
  sections, so an import of a type will be assigned an incorrect type index
  during encoding.

  The fix is to simply encode core types, component types, and imports when
  encountered rather than trying to group them all into the same sections.


crates/wasmparser/src/validator/component.rs
```diff
@@ -1149,22 +1158,21 @@ impl ComponentState {
...
                             )
                         }
                     };
 
                     if !ComponentEntityType::internal_is_subtype_of(arg, types, expected, types) {
                         bail!(
                             offset,
-                            "{} type mismatch for component instantiation argument `{name}`",
-                            expected.desc(),
+                            "type mismatch for component instantiation argument `{name}`"
```
peterhuene, line 1175: FYI, the reason these two messages changed was that they didn't read so well when the description was "type".

alexcrichton (approved):
Good catches!

PR #939, opened 2023-02-21
Improve validation error when component model feature not enabled

rylev:
This tripped me up as the previous error message made it sound like the version encoded in the binary was not a supported version, but it is. The only issue was that the validator did not have the right feature enabled.

### DIFF
--- a/crates/wasmparser/src/validator.rs
+++ b/crates/wasmparser/src/validator.rs
@@ -534,8 +534,7 @@ impl Validator {
                 if !self.features.component_model {
                     bail!(
                         range.start,
-                        "unknown binary version: {num:#x}, \
-                         note: the WebAssembly component model feature is not enabled",
+                        "encoded as a component but the WebAssembly component model feature is not enabled",
                     );
                 }
                 if num == WASM_COMPONENT_VERSION {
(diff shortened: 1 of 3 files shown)


alexcrichton:
Alas I tried to add this and ran into the test failures you're seeing as well! I think though this error message is clearer and worthwhile to keep, so could you edit some of the error message exceptions in `tests/roundtrip.rs` which should get the tests passing again?

rylev:
@alexcrichton so this is a bit more complicated... The error messages are really coming from the fact that some of the tests don't run with component model feature set to true. If they were set to true, the test would pass. However, I think it's not really possible to run these tests with that feature set because the binaries contain a mix of component and module encodings.

So to fix this we would need to update the error messages in https://github.com/WebAssembly/testsuite. I'm not sure if you want me to send a separate PR to that repository changing the error messages there and then in this PR we just change the sha in the submodule.

rylev:
I was also having a hard time reproducing the error locally until I realized the tests were in a submodule. I pushed a fix for how we check that the submodule is initialized - currently it checks that the directory is there (which should always be true), but now we check that the submodule directory is also not empty.

alexcrichton:
Seems reasonable!

rylev:
@alexcrichton the error messages in the testsuite looked correct and changing them in a way that makes sense to the test suite and to wasm-tools would be difficult. I adjusted the error message so that it still passes those tests but provides much more detailed information users will find useful. I think this is good to merge.

alexcrichton (approved):

PR #953, opened 2023-03-07
WIP: support expansion of WIT templates / wildcards

fibonacci1729:
Although still very WIP, this PR adds support for parsing and expanding wildcards (i.e. variable `*`-{imports,exports}) in WIT. 

As a concrete example, consider the WIT:
```fsharp
default world wildcards {
    import imports: self.foo
    export exports: self.foo
}

interface foo {
    *: func()
}
```

Expansion with substitutions:
```toml
[wildcards]
imports = ["a", "b"]
exports = ["x", "y"]
```

Should bindgen to:
```rust
// Generated by `wit-bindgen` 0.3.0. DO NOT EDIT!

#[allow(clippy::all)]
pub mod imports{
  #[allow(clippy::all)]
  pub fn b(){
    
    #[allow(unused_imports)]
    use wit_bindgen::rt::{{alloc, vec::Vec, string::String}};
    unsafe {
      
      #[link(wasm_import_module = "imports")]
      extern "C" {
        #[cfg_attr(target_arch = "wasm32", link_name = "b")]
        #[cfg_attr(not(target_arch = "wasm32"), link_name = "imports_b")]
        fn wit_import(
        );
      }
      wit_import();
    }
  }
  #[allow(clippy::all)]
  pub fn a(){
    
    #[allow(unused_imports)]
    use wit_bindgen::rt::{{alloc, vec::Vec, string::String}};
    unsafe {
      
      #[link(wasm_import_module = "imports")]
      extern "C" {
        #[cfg_attr(target_arch = "wasm32", link_name = "a")]
        #[cfg_attr(not(target_arch = "wasm32"), link_name = "imports_a")]
        fn wit_import(
        );
      }
      wit_import();
    }
  }
  
}


#[allow(clippy::all)]
pub mod exports{
  pub trait Exports {
    fn x();
    fn y();
  }
  
  #[doc(hidden)]
  pub unsafe fn call_x<T: Exports>() {
    
    #[allow(unused_imports)]
    use wit_bindgen::rt::{alloc, vec::Vec, string::String};
    T::x();
  }
  
  #[doc(hidden)]
  pub unsafe fn call_y<T: Exports>() {
    
    #[allow(unused_imports)]
    use wit_bindgen::rt::{alloc, vec::Vec, string::String};
    T::y();
  }
  
}


/// Declares the export of the component's world for the
/// given type.

macro_rules! export_wildcards(($t:ident) => {
  const _: () = {
    
    #[doc(hidden)]
    #[export_name = "exports#x"]
    #[allow(non_snake_case)]
    unsafe extern "C" fn __export_exports_x() {
      exports::call_x::<$t>()
    }
    
    #[doc(hidden)]
    #[export_name = "exports#y"]
    #[allow(non_snake_case)]
    unsafe extern "C" fn __export_exports_y() {
      exports::call_y::<$t>()
    }
    
  };
  
  #[used]
  #[doc(hidden)]
  #[cfg(target_arch = "wasm32")]
  static __FORCE_SECTION_REF: fn() = __force_section_ref;
  #[doc(hidden)]
  #[cfg(target_arch = "wasm32")]
  fn __force_section_ref() {
    __link_section()
  }
});

#[cfg(target_arch = "wasm32")]
#[link_section = "component-type:wildcards"]
#[doc(hidden)]pub static __WIT_BINDGEN_COMPONENT_TYPE: [u8; 287] = [2, 0, 3, 119, 105, 116, 4, 100, 101, 109, 111, 9, 119, 105, 108, 100, 99, 97, 114, 100, 115, 0, 97, 115, 109, 12, 0, 1, 0, 7, 164, 1, 1, 65, 4, 1, 66, 2, 1, 64, 0, 1, 0, 4, 1, 42, 0, 1, 0, 4, 3, 102, 111, 111, 13, 112, 107, 103, 58, 47, 100, 101, 109, 111, 47, 102, 111, 111, 5, 0, 1, 65, 4, 1, 66, 3, 1, 64, 0, 1, 0, 4, 1, 98, 0, 1, 0, 4, 1, 97, 0, 1, 0, 3, 7, 105, 109, 112, 111, 114, 116, 115, 13, 112, 107, 103, 58, 47, 100, 101, 109, 111, 47, 102, 111, 111, 5, 0, 1, 66, 3, 1, 64, 0, 1, 0, 4, 1, 120, 0, 1, 0, 4, 1, 121, 0, 1, 0, 4, 7, 101, 120, 112, 111, 114, 116, 115, 13, 112, 107, 103, 58, 47, 100, 101, 109, 111, 47, 102, 111, 111, 5, 1, 4, 9, 119, 105, 108, 100, 99, 97, 114, 100, 115, 19, 112, 107, 103, 58, 47, 100, 101, 109, 111, 47, 119, 105, 108, 100, 99, 97, 114, 100, 115, 4, 1, 0, 68, 9, 112, 114, 111, 100, 117, 99, 101, 114, 115, 1, 12, 112, 114, 111, 99, 101, 115, 115, 101, 100, 45, 98, 121, 2, 13, 119, 105, 116, 45, 99, 111, 109, 112, 111, 110, 101, 110, 116, 5, 48, 46, 55, 46, 49, 16, 119, 105, 116, 45, 98, 105, 110, 100, 103, 101, 110, 45, 114, 117, 115, 116, 5, 48, 46, 51, 46, 48, 11, 19, 1, 4, 100, 101, 109, 111, 9, 112, 107, 103, 58, 47, 100, 101, 109, 111, 3, 0, 0];

#[inline(never)]
#[doc(hidden)]
#[cfg(target_arch = "wasm32")]
pub fn __link_section() {}

```

### DIFF
--- a/crates/wit-parser/src/ast.rs
+++ b/crates/wit-parser/src/ast.rs
@@ -473,20 +473,20 @@ impl<'a> Value<'a> {
     pub(crate) fn name(&self) -> &'a str {
         match &self.name {
             VariableId::Id(id) => id.name,
-            VariableId::Var(_) => "*",
+            VariableId::Wildcard(_) => "*",
         }
     }
 
     pub(crate) fn span(&self) -> Span {
         match self.name {
             VariableId::Id(Id { span, .. }) => span,
-            VariableId::Var(span) => span,
+            VariableId::Wildcard(span) => span,
         }
     }
 }
 
 pub enum VariableId<'a> {
-    Var(Span), // `*`
+    Wildcard(Span), // `*`
     Id(Id<'a>),
 }
 
@@ -704,7 +704,7 @@ fn parse_variable_id<'a>(tokens: &mut Tokenizer<'a>) -> Result<VariableId<'a>> {
         Some((_span, Token::Id | Token::ExplicitId)) => parse_id(tokens).map(VariableId::Id),
         Some((span, Token::Star)) => {
             tokens.expect(Token::Star)?;
-            Ok(VariableId::Var(span))
+            Ok(VariableId::Wildcard(span))
         }
         other => Err(err_expected(tokens, "an identifier or `*`", other).into()),
     }
--- a/crates/wit-parser/src/ast.rs
+++ b/crates/wit-parser/src/ast.rs
@@ -462,8 +462,29 @@ struct Stream<'a> {
 
 pub struct Value<'a> {
     docs: Docs<'a>,
-    name: Id<'a>,
+    name: VariableId<'a>,
     kind: ValueKind<'a>,
+}
+
+impl<'a> Value<'a> {
+    pub(crate) fn name(&self) -> &'a str {
+        match &self.name {
+            VariableId::Id(id) => id.name,
+            VariableId::Var(_) => "*",
+        }
+    }
+
+    pub(crate) fn span(&self) -> Span {
+        match self.name {
+            VariableId::Id(Id { span, .. }) => span,
+            VariableId::Var(span) => span,
+        }
+    }
+}
+
+pub enum VariableId<'a> {
+    Var(Span), // `*`
+    Id(Id<'a>),
 }
 
 struct Union<'a> {
@@ -550,7 +571,7 @@ impl<'a> InterfaceItem<'a> {
             Some((_span, Token::Union)) => {
                 TypeDef::parse_union(tokens, docs).map(InterfaceItem::TypeDef)
             }
-            Some((_span, Token::Id)) | Some((_span, Token::ExplicitId)) => {
+            Some((_span, Token::Star | Token::Id | Token::ExplicitId)) => {
                 Value::parse(tokens, docs).map(InterfaceItem::Value)
             }
             Some((_span, Token::Use)) => Use::parse(tokens).map(InterfaceItem::Use),
@@ -668,10 +689,21 @@ impl<'a> TypeDef<'a> {
 
 impl<'a> Value<'a> {
     fn parse(tokens: &mut Tokenizer<'a>, docs: Docs<'a>) -> Result<Self> {
-        let name = parse_id(tokens)?;
+        let name = parse_variable_id(tokens)?;
         tokens.expect(Token::Colon)?;
         let kind = ValueKind::parse(tokens)?;
         Ok(Value { docs, name, kind })
+    }
+}
+
+fn parse_variable_id<'a>(tokens: &mut Tokenizer<'a>) -> Result<VariableId<'a>> {
+    match tokens.clone().next()? {
+        Some((_span, Token::Id | Token::ExplicitId)) => parse_id(tokens).map(VariableId::Id),
+        Some((span, Token::Star)) => {
+            tokens.expect(Token::Star)?;
+            Ok(VariableId::Var(span))
+        }
+        other => Err(err_expected(tokens, "an identifier or `*`", other).into()),
     }
 }
 
--- a/crates/wit-parser/src/ast.rs
+++ b/crates/wit-parser/src/ast.rs
@@ -8,6 +8,9 @@ use std::path::{Path, PathBuf};
 
 pub mod lex;
 
+pub use expand::expand;
+mod expand;
+
 pub use resolve::Resolver;
 mod resolve;
 pub mod toposort;
--- a/crates/wit-parser/src/ast/expand.rs
+++ b/crates/wit-parser/src/ast/expand.rs
@@ -1,60 +1,100 @@
-use anyhow::Result;
+use anyhow::{anyhow, Result};
+use indexmap::IndexMap;
 use std::collections::{HashMap, HashSet};
-use crate::{Resolve, World, WorldId, PackageId, WorldItem};
+use crate::{Resolve, World, WorldId, PackageId, WorldItem, InterfaceId};
 
 pub type Substitutions = HashMap<String, HashMap<String, HashSet<String>>>;
 
-pub fn expand(resolve: &mut Resolve, package_id: PackageId, world: Option<&str>, substitutions: Substitutions) -> Result<()> {
-    let world_id = resolve.select_world(package_id, world)?;
-    expand_world(resolve, world_id, substitutions)
-}
-
-fn expand_world(resolve: &mut Resolve, world_id: WorldId, substitutions: Substitutions) -> Result<()> {
-    let World { name, imports, exports, .. } = resolve
-        .worlds
-        .get(world_id)
-        .cloned()
-        .unwrap();
-
-    match substitutions.get(&name) {
-        Some(substitutions) => {
-            expand_world_items(resolve, imports, substitutions, "import")?;
-            expand_world_items(resolve, exports, substitutions, "export")?;
-            Ok(())
-        },
-        None => {
-            // TODO: return error if world contains templated 
-            // items yet no substitutions were provided
-            Ok(())
-        },
-    }
-}
-
-fn expand_world_items(
-    resolve: &mut Resolve, 
-    items: impl IntoIterator<Item = (String, WorldItem)>, 
-    substitutions: &HashMap<String, HashSet<String>>,
-    _direction: &str,
+pub fn expand(
+    resolve: &mut Resolve,
+    package_id: PackageId,
+    world: Option<&str>,
+    substitutions: Substitutions
 ) -> Result<()> {
-    for (name, item) in items {
-        if let WorldItem::Interface(interface_id) = item {
-            let interface = resolve.interfaces.get_mut(interface_id).unwrap();
+    let world_id = resolve.select_world(package_id, world)?;
+    resolve.expand_wildcards(world_id, substitutions)
+}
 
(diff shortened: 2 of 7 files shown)


crates/wit-parser/src/ast.rs
```diff
@@ -462,10 +465,31 @@ struct Stream<'a> {
...
     kind: ValueKind<'a>,
 }
 
+impl<'a> Value<'a> {
+    pub(crate) fn name(&self) -> &'a str {
+        match &self.name {
+            VariableId::Id(id) => id.name,
+            VariableId::Wildcard(_) => "*",
+        }
+    }
```
alexcrichton, line 478: I think it might be best to leave this out as an accessor since it's probably best for most consumers to do a `match` on this and have them determine how to handle it (unless it's super frequently called, but I'm not sure that it is)

crates/wit-parser/src/ast/resolve.rs
```diff
@@ -560,13 +563,22 @@ impl<'a> Resolver<'a> {
...
                     let docs = self.docs(&value.docs);
+                    let name = value.name();
+                    let span = value.span();
                     match &value.kind {
+                        ValueKind::Func(func) if matches!(value.name, VariableId::Wildcard(_))=> {
+                            let func = self.resolve_function(docs, name, func)?;
+                            let prev = self.interfaces[interface_id]
+                                .wildcard
+                                .replace(func);
+                            assert!(prev.is_none());
```
alexcrichton, line 574: I think this'll need to be a first-class error since I think `*: func()` can show up twice textually?

crates/wit-parser/src/ast/expand.rs
```diff
@@ -0,0 +1,100 @@
...
+                            "world {world_name} {desc} item {name} contains wildcards but no substitutions were provided",
+                        ))?;
+                    *interface_id = self.expand_interface(*interface_id, iface_subs); 
+                }
+            }
+        }
+        Ok(())
+    }
+
+    fn expand_interface(&mut self, interface_id: InterfaceId, subnames: HashSet<String>) -> InterfaceId {
```
alexcrichton, line 87: FWIW one minor thing to consider here is that `HashSet` has random iteration order but we likely want this expansion process to be deterministic so it might be good to take a `Vec` or list of some form here instead with a defined order.

crates/wit-parser/src/ast/expand.rs
```diff
@@ -0,0 +1,100 @@
...
+        let mut new_interface = self.interfaces[interface_id].clone();
```
alexcrichton, line 88: Ah unfortunately cloning this won't be so simple I think due to the fact that there's lots of "back references" to the owner of an item. For example types have an `InterfaceId` as their owner and that would also need to be updated here. There's additionally references from the document to its interfaces that it contains.

Overall I'm not sure though that cloning and expanding like this will work out. Would it be possible to fill-out in-place? For example basically converting a wildcard-containing interface to one without wildcards?

crates/wit-parser/src/ast/expand.rs
```diff
@@ -0,0 +1,100 @@
...
+    package_id: PackageId,
+    world: Option<&str>,
+    substitutions: Substitutions
+) -> Result<()> {
+    let world_id = resolve.select_world(package_id, world)?;
+    resolve.expand_wildcards(world_id, substitutions)
+}
+
+impl Resolve {
+    pub fn expand_wildcards(&mut self, world_id: WorldId, substitutions: Substitutions) -> Result<()> {
```
alexcrichton, line 19: The top level of this map is the name of the `world_id` specified, so could one layer of the map be peeled off perhaps and the argument here is "just" `HashMap<String, T<String>>`?

alexcrichton (commented):
Thanks! I've got two high-level questions from this:

* How are y'all planning on representing `*: func()` in the wasm-encoded version of a document?
* How are y'all expecting the `expand_wildcards` method will be invoked and used elsewhere? It sounds like for guests you're expecting a config file of sorts to get expanded in the macro, but in the past there's also been discussion that a wildcard in Rust, for example, would turn into a macro which when invoked would generate the bindings for a new accessor (or export). I'm not sure how either will look like when done, though, so I'm curious if y'all have thoughts.

fibonacci1729:
@alexcrichton Thanks for the quick feedback! Answering your questions:

1. We haven't discussed yet how `*: func()` gets encoded into a `.wasm`, I'm hoping to get that convo going once everyone is back from Germany this week.
2. This is a good question. We approached this originally from the perspective of being pretty agnostic to the langauge toolchain. I haven't given too much more though about using macros for this in `rust` but i think that would be worthwhile to explore and supporting in addition to the config format. FWIW the config file here is largely demonstrative for prototyping.

crates/wit-parser/src/ast/resolve.rs
fibonacci1729: Ahh great catch!

crates/wit-parser/src/ast/expand.rs
fibonacci1729: Ahh you're absolutely correct. I'll have to give this more thought.

Funny enough, my original approach was expanding in-place...However, templated interfaces need to be expanded more-than-once which is the motivation for alloc'ing new interfaces.

I'm guessing it would be quite painful to patch the back references to expanded interfaces in types and documents.

fibonacci1729: After chatting with @dicej, a viable approach might be to reconstruct `Resolve` as we are expanding instead of trying to mutably update as is done now. I anticipate the tricky part of this will be how to reconcile the topological order of things while inserting new interfaces.

alexcrichton: That might work, yeah, but personally I don't really know a good way of handling all this. You can take a look at the support in `resolve.rs` for merging two `Resolve`s to see what the order-of-magnitude is for ids that need remapping. 

One other tricky part is that if a wildcard interface is both exported and imported then the import/export refer to the same ID, but if you can specify different substitutions for both then that ID mapping needs to become different. Then that introduces two interfaces with possibly clashing names which also further needs resolution.

alexcrichton:
In some sense the `*`-import system is intended to "compete" to a degree with "just make a function that takes a string parameter". I realize there are different properties of each of these choices with respect to static analyzability but from a developer's perspective they achieve the same goal of getting a named thing. I'm wary of the `*`-import system feeling bespoke and confusing to end-users when all is said-and-done relative to "just call the function". 

This isn't really a concrete concern per se but it's one where I'm not sure that this is going to actually become a reality. The `*`-handling so far doesn't feel necessarily well-integrated with the rest of the component model and/or doesn't necessarily have a great path to being integrated into each language. It's still very early-days for `*`-imports, though, and it's not like there's a defacto sense for what each language in the component model "feels like" as it's still very early-days for the component model itself as well. It may be worth considering as development goes along, though, because I would fear that if the end-state of `*`-imports/exports feels unnatural enough in various languages it may not end up getting used and the purpose of the system, static analyzability, would be lost.

dicej:
@alexcrichton I think the concerns you raised are valid, and we definitely need to start thinking about how to make this feature ergonomic and idiomatic to use in each language.  I do expect that this feature will end up being used in various standard WASI worlds (e.g. for config, metrics, key-value stores, etc.), which would mean developers will _have_ to use it in order to target those worlds.  I don't know if that will make you feel better (because the feature _will_ be used) or worse (because it will be resented and complained about if we mess up the ergonomics).

The current situation in Rust (per https://github.com/bytecodealliance/wit-bindgen/compare/main...dicej:wit-templates) doesn't seem terribly awkward to me:
```rust
wit_bindgen::generate!({
    path: "../wit/config.wit",
    substitutions_inline: r#"
        [config]
        imports = ["db-url", "max-connections", "favorite-color"]
    "#
})
// ...
// Use the bindings as usual, e.g.:
let conn = connect(imports::db_url());
```
No doubt there is room for improvement, though!  For instance, a more Rust-y syntax for inline substitutions instead of TOML-in-a-string-literal.  Or looking for them in Cargo.toml by default, etc.

alexcrichton:
Nah yeah I understand to a degree that there will be "forced" usage which isn't really problematic per-se but I think it's also an important consideration for other host-based or similar APIs where I don't think we'd want to end up in a world where WASI is the only one using this and everyone else avoids it due to it being unergonomic.

I'm personally wary of adding TOML into the mix here myself, mostly as it feels like "yet another thing to read up on and understand". That's sort of inherent to the feature though so I don't know of a great way to handle that. When learning the component model you basically already need a huge complexity budget to spend. Ideally wildcards are a small addition to that budget where you only have to consider it for a moment and go "oh yeah that makes sense" as opposed to having to spend a significant amount of "budget" to figure everything out.

To be clear though I'm sort of describing the properties we ideally want in an end state, I'm not reviewing what's here or commenting on complexity because much of the whole picture is still missing. Basically I don't mean this as a critique but rather moreso early thoughts on how this might shape up. I see risks in having things like separate configuration files and possible contortions to feed in substitution information to all language generators, but it's tough to see how everything plays out without the full end-to-end story in my head.

PR #954, opened 2023-03-07
Add a ReorderCustomSectionMutator

itsrainy:
#415 
Also did the following cleanup:
- Move AddCustomSectionMutator to custom.rs to keep all the custom section mutators together
- Add CustomSectionMutator to our list of mutators in lib.rs so it starts getting used


crates/wasm-mutate/src/info.rs
```diff
@@ -309,6 +309,31 @@ impl<'a> ModuleInfo<'a> {
...
+        self.raw_sections.iter().enumerate().for_each(|(i, s)| {
+            if src_idx < dest_idx && i == dest_idx {
+                module.section(&self.raw_sections[src_idx]);
+            }
+            if i != src_idx {
+                module.section(s);
+            }
+            if dest_idx < src_idx && i == dest_idx {
+                module.section(&self.raw_sections[src_idx]);
+            }
```
alexcrichton, line 332: I paused for a second here to try to review this carefully to make sure something wasn't encoded twice since the `if` statements all execute on each iteration of the loop. This looks good, but would this simplification also work perhaps?

```rust
let refs = self.raw_sections.iter().collect::<Vec<_>>();
refs.swap(src_idx, dest_idx);
let mut module = wasm_encoder::Module::new();
for section in refs {
    module.section(section);
}
module
```

alexcrichton (approved):
Nice!

crates/wasm-mutate/src/info.rs
itsrainy: That would also reorder the sections, but it would do so slightly differently. Rather than moving `src` to `dest` and keeping the order of all the others the same, it would also affect the placement of where the section at dest_idx is if I'm understanding correctly. So rather than a reorder like this:
```
// [a, b, c]
move_section(0, 2)
// [b, c, a]
```
it would reorder like this:
```
// [a, b, c]
move_section(0, 2)
// [c, b, a]
```

I think that's probably fine? I'll defer to you and @fitzgen since y'all have a bit more context here than I do.

alexcrichton: Oh true yeah, could `.remove` and `.insert` on vectors be used for that then?

fitzgen: I think that would work, but at this point I don't think it is any nicer than what is currently in the PR. But I don't feel super strongly.

alexcrichton: True yeah, in that case I'll merge!

PR #955, opened 2023-03-08
wasmprinter: Add `Printer::print_offsets_and_lines`

fitzgen:
This new method allows users to get the disassembly line by line along with the binary offset associated with each line (in a programmatic way and without actually printing the binary offset in the disassembly).


alexcrichton:
This seems reasonable at a high-level, but at a lower level I'd be a little worried that when printing a large module this would have a significant performance impact due to the list-of-lines representation rather than one big string.

Instead though could a vector-of-offsets be maintained where the `i`th elemet of the vector is the byte offset of the start of the `i`th line?

fitzgen:
That is a much better approach, d'oh.

fitzgen:
Okay, much happier with this new version. @alexcrichton PTAL.

alexcrichton (approved):

PR #956, opened 2023-03-08
feat(wit-parser): add parser for `include`

Mossaka:
This PR adds wit parsing for `include` and `include ... with { }` syntax as proposed by the [union of worlds proposal](https://github.com/WebAssembly/component-model/issues/169) and followed the WIT lexical structured proposed by this [PR](https://github.com/WebAssembly/component-model/pull/174). 

This PR, however, does not implement the `include` path resolver. Since the path to include to is a world id path instead of a interfaceid, making the code for `resolve_path` un-reusuable for resolving include path. A follow up PR or continuation of this PR will resolve this by introducing a `resolve_include_path` function to assist.


crates/wit-parser/src/ast/resolve.rs
```diff
@@ -434,7 +435,7 @@ impl<'a> Resolver<'a> {
         for item in world.items.iter() {
             let (name, kind, desc, spans, interfaces) = match item {
                 // handled in `resolve_types`
-                ast::WorldItem::Use(_) | ast::WorldItem::Type(_) => continue,
+                ast::WorldItem::Use(_) | ast::WorldItem::Type(_) | ast::WorldItem::Include(_) => continue,
```
Mossaka, line 438: This is temporary and needs to be implemented for the resolver to work.

crates/wit-parser/src/ast.rs
```diff
@@ -226,6 +229,42 @@ impl<'a> ExternKind<'a> {
...
+                tokens.expect(Token::With)?;
+                tokens.expect(Token::LeftBrace)?;
+                while !tokens.eat(Token::RightBrace)? {
+                    let mut name = UseName {
+                        name: parse_id(tokens)?,
+                        as_: None,
+                    };
+                    if tokens.eat(Token::As)? {
+                        name.as_ = Some(parse_id(tokens)?);
+                    }
```
alexcrichton, line 255: I realize this is how `use` is parsed but I think the `as` keyword should be required for `with`?

crates/wit-parser/src/ast.rs
```diff
@@ -47,6 +47,7 @@ impl<'a> Ast<'a> {
                             WorldItem::Type(_) => {}
                             WorldItem::Import(Import { kind, .. }) => imports.push(kind),
                             WorldItem::Export(Export { kind, .. }) => exports.push(kind),
+                            WorldItem::Include(i) => f(None, &i.from, Some(&i.names))?,
```
alexcrichton, line 50: I don't think that `Some(&i.names)` is correct to pass along here because that's intended for "names used from this interface" whereas this list of names is instead "renamed imports/exports"

alexcrichton:
Oh I should also clarify, I would personally prefer to get more of `include` implemented before landing incremental support. The patch here is pretty small and I don't mind reviewing something larger, and I think it's better to land a whole feature if possible.

crates/wit-parser/src/ast.rs
Mossaka: Got it, does it mean I should just pass a `None`?

Mossaka:
I've implemented the basic shape for `include` worlds to form a bigger world and added a few 
examples to make `self`, `pkg` and `deps` working. This PR is not ready to be merged yet because I still need to implement the de-duplicate logic and add lots of tests. But if you could give an early quick review, it would be very helpful for me to iterate through this. @alexcrichton

Mossaka:
Oh, should clarfiy that you can ignore all the log debug that's just temporary for me to debug something and will be cleaned up later.

crates/wit-parser/src/ast.rs
```diff
@@ -228,6 +255,45 @@ impl<'a> ExternKind<'a> {
     }
 }
 
+#[derive(Debug)]
+pub struct Include<'a> {
+    /// The path to the world to include
+    from: UsePath<'a>,
+
+    /// The names of the imports and exports to be replaced with
+    names: Vec<UseName<'a>>,
```
alexcrichton, line 264: I think that this isn't quite exactly `UseName` since the `as` keyword is required? For example `with { a, b, c }` doesn't mean much and probably shouldn't parse

alexcrichton (commented):
Looking good to me!

Mossaka:
Closed by #1054

PR #957, opened 2023-03-08
`wasmparser`: turn table index in the elements section into an `Option`

koute:
The `wasmparser` crate currently doesn't disambiguate between the two possible encodings of the table index in the elements section while `wasm-encoder` does. I'd like to be able to generate bit-exact `.wasm` blobs by parsing with `wasmparser` and piping everything to `wasm-encoder`, and this makes it impossible. So this PR makes `wasmparser` do what `wasm-encoder` does and store the index inside of an `Option`.

### DIFF
--- a/crates/wasm-mutate/src/mutators/translate.rs
+++ b/crates/wasm-mutate/src/mutators/translate.rs
@@ -263,7 +263,7 @@ pub fn element(
                 ConstExprKind::ElementOffset,
             )?;
             ElementMode::Active {
-                table: Some(t.remap(Item::Table, *table_index)?),
+                table: table_index.map(|i| t.remap(Item::Table, i)).transpose()?,
                 offset: &offset,
             }
         }
--- a/crates/wasmparser/src/readers/core/elements.rs
+++ b/crates/wasmparser/src/readers/core/elements.rs
@@ -40,7 +40,7 @@ pub enum ElementKind<'a> {
     /// The element segment is active.
     Active {
         /// The index of the table being initialized.
-        table_index: u32,
+        table_index: Option<u32>,
         /// The initial expression of the element segment.
         offset_expr: ConstExpr<'a>,
     },
@@ -91,9 +91,9 @@ impl<'a> FromReader<'a> for Element<'a> {
             }
         } else {
             let table_index = if flags & 0b010 == 0 {
-                0
+                None
             } else {
-                reader.read_var_u32()?
+                Some(reader.read_var_u32()?)
             };
             let offset_expr = reader.read()?;
             ElementKind::Active {
--- a/crates/wasmparser/src/validator/core.rs
+++ b/crates/wasmparser/src/validator/core.rs
@@ -207,7 +207,7 @@ impl ModuleState {
                 table_index,
                 offset_expr,
             } => {
-                let table = self.module.table_at(table_index, offset)?;
+                let table = self.module.table_at(table_index.unwrap_or(0), offset)?;
                 if !self
                     .module
                     .matches(ValType::Ref(e.ty), ValType::Ref(table.element_type), types)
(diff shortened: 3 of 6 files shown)


alexcrichton:
In isolation I think this is ok, but I don't think that `wasmparser` nor `wasm-encoder` is suitable for your goal of bit-exact wasm blobs. One major aspect neither handle is overlong encodings of leb128 values. For example `wasm-ld` will produce 5-byte encodings of all targets for `call` instructions to have the linker efficiently fill them in. The `wasmparser` crate has no means of conveying this information and similarly the `wasm-encoder` crate has no means of creating an overlong encoding.

Additionally this trick is not isolated to just table indices in element sections, but it's been pervasively used throughout `wasmparser`'s development. For example [memory indices](https://github.com/bytecodealliance/wasm-tools/blob/861bfaa8bf48fd3afb4d01595cfc75c08fc69d34/crates/wasmparser/src/binary_reader.rs#L252-L257) is another case where lack-of-integer is silently translated to "you specified 0". [Data segments also do something similar](https://github.com/bytecodealliance/wasm-tools/blob/861bfaa8bf48fd3afb4d01595cfc75c08fc69d34/crates/wasmparser/src/readers/core/data.rs#L68-L72), and this is just what I remember off the top of my head so I think there's more.

koute:
> In isolation I think this is ok, but I don't think that `wasmparser` nor `wasm-encoder` is suitable for your goal of bit-exact wasm blobs.

While this is a little unfortunate I don't really need this in *all cases* in general, just for a particular set of `.wasm` blobs I'm dealing with.

We're currently using `parity-wasm` for manipulating `.wasm` blobs and we'd like to migrate to `wasmparser` + `wasm-encoder`, and obviously having the new codepath be able to give out exactly the same `.wasm` as the old one makes this process easier.

This is somewhat offtopic, but may I ask - are there any plans to unify the types used by `wasmparser` and `wasm-encoder`? In isolation I think those two crates work pretty well, but when you want to mutate an existing `.wasm` blob translating between the `wasmparser` types and `wasm-encoder` types is a little tedious (especially translating `wasmparser::Operation` to `wasm_encoder::Instruction` which is a ~600 lines `match`), and seems a somewhat unnecessary when they're mostly the same. Couldn't we have e.g. a `wasm-types` crate and unify those types between `wasmparser` and `wasm-encoder` where it makes sense?

alexcrichton:
Ok if this is useful to you I think it's reasonable to include as it's relatively minor anyway, as long as you're ok with the other shortcomings that wasmparser has for that sort of use case.

> are there any plans to unify the types used by wasmparser and wasm-encoder?

Currently not at this time. Historically this was also done by the `walrus` crate for the rustwasm working group, but that crate has sort of fallen out of style (and it's behind the times and lacking maintenance, so I wouldn't recommend it nowadays). Not to say that you're the first person who wants something like this, though, as I've also wanted this. Even in this repository the `wasm-mutate` crate translates between wasms and parts of `wit-component` do the same. I think it would be great for something like this to exist, but at this time I don't think anyone knows how best to construct this. Adding something like a `wasm-types` crate I think may be a bit too simple of a solution since there's a lot of subtelties about representation between wasmparser and wasm-encoder that are different and have repercussions on performance in various situations.

If you're interested, though, [some macro soup](https://github.com/bytecodealliance/wasm-tools/blob/cfb51b3fef2286d77d6dad2fbcb8e0c9849813fc/crates/wit-component/src/gc.rs#L1103-L1202) can ease the translation today.

koute:
> Ok if this is useful to you I think it's reasonable to include as it's relatively minor anyway, as long as you're ok with the other shortcomings that wasmparser has for that sort of use case.

Yes, I still think this should be useful!

alexcrichton (approved):

PR #962, opened 2023-03-17
Implement structural equivalence for interfaces when merging worlds.

sunfishcode:
When merging two worlds together, coalesce interfaces that are structurally equivalent to each other.

The motivation for this is that I want to be able to write library code that uses wit bindings that can be used in any world that contains that interface.

One such library is std, which will want to depend on wasi-filesystem and other interfaces, but remain independent of the specific world that the user will ultimately chose to link everything into.

This doesn't yet include any tests because I'm unsure how to add them for this. I have confirmed that a cargo component build with this is able to build a working component for a crate that uses its own `wit_bindgen::generate!` invocation for interfaces that are also pulled in through Cargo.toml:
```toml
[package.metadata.component]
target = { path = "wit", world = "command" }
```

### DIFF
--- a/crates/wit-parser/src/resolve.rs
+++ b/crates/wit-parser/src/resolve.rs
@@ -1,7 +1,8 @@
 use crate::ast::lex::Span;
 use crate::{
-    Document, DocumentId, Error, Function, Interface, InterfaceId, Results, Type, TypeDef,
-    TypeDefKind, TypeId, TypeOwner, UnresolvedPackage, World, WorldId, WorldItem,
+    Case, Document, DocumentId, Error, Field, Function, Interface, InterfaceId, Params, Record,
+    Result_, Results, Stream, Tuple, Type, TypeDef, TypeDefKind, TypeId, TypeOwner, Union,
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! I am personally to perform this merging purely based on structural equivalence however since that definitely gets more funky with resource types in the mix. Instead though a rough idea for handling this sort of case would be to use the URLs and/or unique IDs of interfaces. This is something that hasn't really been gamed out a ton right now but the thinking would be the when the same interface is imported or exported twice when merging worlds the interfaces are merged together based on their unique ID. If the IDs don't match then merging doesn't happen and an error is emitted, otherwise merging happens.

One important case I think this is necessary for is when one world imports one function from `wasi-filesystem`, for example, and another world imports a different function. When those two worlds are merged together they should produce the union of the two with both functions and both types, so going purely off structural equivalence may not be enough? This sort of partial merging may also be a bit of a pipe dream at this point in time though in which case we may need to structurally match two worlds to double-check that if the IDs are the same then the worlds get merged.

sunfishcode:
This is now superceded by #974.

PR #964, opened 2023-03-20
feat: preliminary support for WIT templates

dicej:
This is a minimum viable implementation of WIT templates per https://github.com/WebAssembly/component-model/issues/172.  It supports interfaces with at most one wildcard function, which may be expanded (i.e. monomorphized) with a set of substitutions provided by the application developer when generating guest bindings.  I will be posting separate PRs to the `wit-bindgen` and `wasmtime` repos to implement and test guest and host binding generation, respectively.


crates/wit-parser/src/ast/resolve.rs
```diff
@@ -563,11 +566,21 @@ impl<'a> Resolver<'a> {
...
+                        if !matches!(value.name, VariableId::Wildcard(_)) {
+                            self.define_interface_name(
+                                value.name(),
+                                value.span(),
+                                TypeOrItem::Item("function"),
+                            )?;
+                        }
+                        let func = self.resolve_function(&value.docs, value.name(), func)?;
+                        let prev = if matches!(value.name, VariableId::Wildcard(_)) {
+                            self.interfaces[interface_id].wildcard.replace(func)
```
fibonacci1729, line 578: I think we'll want to handle `prev.is_some()` (indicating multiple `*` appears textually) here and return a concrete syntax error positioned at the span of the wildcard.

PR #965, opened 2023-03-22
preserve `cabi_post_` exports when appropriate

dicej:
Also, consider `canonical_abi_realloc` import equivalent to `cabi_realloc` when scanning adapter imports.  This helps when adapting older modules.

### DIFF
--- a/crates/wit-component/src/gc.rs
+++ b/crates/wit-component/src/gc.rs
@@ -35,6 +35,13 @@ pub fn run(
     }
     let mut not_required = IndexSet::new();
     for name in module.exports.keys().copied() {
+        // If we need `name` then we also need cabi_post_`name`:
+        let name = if let Some(suffix) = name.strip_prefix("cabi_post_") {
+            suffix
+        } else {
+            name
+        };
+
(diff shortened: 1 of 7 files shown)


alexcrichton (approved):
Makes sense to me, thanks! To confirm, this is for a custom non-wasi-preview1 adapter, right? Would you be up for adding a test for this too?

dicej:
> Makes sense to me, thanks! To confirm, this is for a custom non-wasi-preview1 adapter, right? Would you be up for adding a test for this too?

It's for a custom fork of the wasi-preview1 adapter for Spin apps: https://github.com/fermyon/spin-componentize/tree/main/adapter.  I originally intended for it to be a separate adapter, but it turned out the Spin parts and the WASI parts needed to be in the same adapter so they can share state and control reentrancy (per https://github.com/bytecodealliance/preview2-prototyping/issues/97), and a fork was the easiest way to achieve that.  Anyway, one of the Spin exports requires a post-return function, hence this patch.

And yes, I'll add a test tomorrow.  And I'll try to be more proactive about that so you won't have to remind me in the future :)

Issue #967, opened 2023-03-30
Runtime panic encoding a guest component utilizing `wasi-http`


rvolosatovs:
WebAssembly components currently cannot utilize https://github.com/WebAssembly/wasi-http on commit `53e2d258ae3008d254bda79ec958355ded8751a9` with:

- `wit-bingen 0.4.0` 
- `wit-component 0.7.4` 
- `wasi_snapshot_preview1 47b30fdad11553841bbee4c5632101bbd645854b` 

Encoding a component constructed using this toolset produces a runtime panic in `wit-component` at https://github.com/bytecodealliance/wasm-tools/blob/1e0052974277b3cce6c3703386e4e90291da2b24/crates/wit-component/src/encoding.rs#L518

```
'IndexMap: key not found'
```

The index being looked for as a fallback is missing in the imported instance map

The failure can be reproduced via `cargo test` or `cargo run` (which takes a WASI adapter as an optional parameter and prints the resulting component to stdout on success) in this repository https://github.com/rvolosatovs/wasi-http-guest-repro

Note, that I have added a test case to both https://github.com/bytecodealliance/wit-bindgen and `wit-component` (updating the existing `wasi-http` interface and adding a new one importing it as a dependency). I will open PR(s) shortly after. All the tests pass :thinking:

peterhuene:
The problem is that we're not merging worlds between the component being produced and any adapters, so if you import any name that the adapter also imports, `wit-component` will become confused as it needs to encoding two imports of the same name.

The real fix here is to properly merge worlds between the world of the component and any adapters so that there's a single imported instance of that interface that can be referenced from both the adapter and the component.

If you're using `cargo-component` for this, you can work around it by disabling the automatic adaptation from preview 1 by passing `--target wasm32-unknown-unknown` (obviously you'd then not be able to use the stdlib for WASI).

alexcrichton:
I'm running out of time today to work on this, but I am actively pursuing a fix to this which is exactly what @peterhuene outlined. I think I've got a good grasp on the problem now and I've got small reproductions and what I think is a way forward. I hope to have a PR for this next week, as it's a nontrivial feature to implement.

PR #968, opened 2023-03-30
refactor: provide more descriptive panic message

rvolosatovs:
Refs #967 

From:
> 'IndexMap: key not found'

To:
> interface with index 7 of type with index 54 missing from imported instances

While the message is still pretty cryptic, at least the domain the panic originates from is somewhat clear


peterhuene:
If no one has yet looked at fixing #967, I should have some time soon to dive into it after I get a bunch of feature work on cargo component in.

alexcrichton:
I believe this was addressed in https://github.com/bytecodealliance/wasm-tools/pull/974 so I'm going to close this. Thanks for the PR though!

PR #969, opened 2023-03-30
tests(wit-component): update `wasi-http` interface

rvolosatovs:
Sync to https://github.com/WebAssembly/wasi-http/commit/53e2d258ae3008d254bda79ec958355ded8751a9

This was done as part of debugging #967


alexcrichton:
Sure! FWIW this is only used as a test for the parser itself, so there's no need to keep it in sync in the future and if issues are encountered specific smaller tests can be added to showcase the problem.

alexcrichton (approved):

PR #975, opened 2023-04-09
Return error when demangling component

rylev:
`wasm-tools` will happily demangle a component and in the process convert it to an invalid core module.

I would add support for demangling components, but I wanted to check first whether the right approach is to more or less copy/paste the demangling code that currently handles just modules. I believe it wouldn't be possible to add support in the encoder for generically adding sections to either a component or a core module as sections are different depending on if they are embedded in a component or a core module.


alexcrichton (approved):
Thanks! I thought I fixed this awhile back but I apparently did not!

PR #976, opened 2023-04-10
Update wasm-tools compose example with latest cargo component

rylev:
I could not get the example for `wasm-tools compose` to work with the `cargo component` on crates.io. I installed the latest version from the main branch, and updated the example to work with that. I've noted in the README that this is the version that the example is known to work with. 

> *Note*: there are several whitespace changes to the README since my editor is configured to remove trailing whitespaces. Let me know if you want me to revert this.


alexcrichton (approved):
Thanks!

Issue #977, opened 2023-04-11
wasm-tools compose creates invalid instances within the composed component


rylev:
After I run the `wasm-tools compose` [example](https://github.com/bytecodealliance/wasm-tools/tree/main/crates/wasm-compose/example), I cannot run `wasm-tools component wit` on the composed component. I get the following error:

```
Error: failed to decode WIT document

Caused by:
    0: failed to decode WIT from import `streams`
    1: unexpected unnamed type
```

The issue seems to be that the composed component has incorrectly encoded instances. You can see this when looking at one of the instances before and after composition happen. Note: the instance in question is not involved in the composition beyond being a shared dependency of the two components being merged into one composed component:

**Before composition**

```wat
(component
 (type (;0;)
    (instance
      (type (;0;) u32)
      (export (;1;) "wall-clock" (type (eq 0)))
      (type (;2;) (record (field "seconds" u64) (field "nanoseconds" u32)))
      (export (;3;) "datetime" (type (eq 2)))
      (type (;4;) (func (param "this" 1) (result 3))) ;; NOTE: the result points to the "datetime" export
      (export (;0;) "now" (func (type 4)))
    )
  )
;; ...
```

**After composition**

```wat
(component
  (type (;0;)
    (instance
      (type (;0;) u32)
      (export (;1;) "wall-clock" (type (eq 0)))
      (type (;2;) (record (field "seconds" u64) (field "nanoseconds" u32)))
      (export (;3;) "datetime" (type (eq 2)))
      (type (;4;) (func (param "this" 0) (result 2))) ;; NOTE: the result type points directly to the record
      (export (;0;) "now" (func (type 4)))
    )
  )
;; ...
```

At some point this instance seems to have one of its functions changed so that its result type points directly to a record field (which is illegal) instead of pointing to a type.

@alexcrichton noted the following [on Zulip](https://bytecodealliance.zulipchat.com/#narrow/stream/206238-general/topic/wasm-tools.20compose.20example/near/348479284):

> I suspect that this may be an issue with both the decoder and encoder perhaps? The constructs used here were added after the initial implementation of wasm-compose so it hasn't been updated to use exports in the same way. The decoder probably needs an update to read the structure of use-the-exported-item and the encoder needs to be updated to use items as they're exported. Shouldn't be too hard ideally as wit-component deals with similar problems as well, but if you'd like I think opening an issue for this would be good

PR #978, opened 2023-04-11
Improve wit component errors when failing to decode function

rylev:
I needed this when debugging for #977, and it seems useful to keep. Let me know what you think!


alexcrichton (approved):
Thanks!

PR #980, opened 2023-04-13
Encode instance exports properly

### LINKED ISSUE #977: wasm-tools compose creates invalid instances within the composed component

rylev:
Fixes #977 

Instance exports were not being reflected in the TypeMap meaning that future declarations would refer to the direct type definition instead of the export definition. This updates that handling so that the type map is updated with the index of the export. 

Big thanks to @peterhuene for the helpful hints!

> *Note*: I also updated some error handling to give better errors to help with a new error I'm seeing. Let me know if you'd rather I separate that out into a new PR.


alexcrichton:
Thanks! Would it be easy enough to add a test for this as well?

crates/wasm-compose/src/encoding.rs
```diff
@@ -143,8 +143,11 @@ impl<'a> TypeEncoder<'a> {
         }
 
         for (name, url, ty) in exports {
-            let ty = self.component_entity_type(&mut encoded, &mut types, ty);
-            encoded.export(name, url, ty);
+            let type_definiton = self.0.type_from_id(ty.id().unwrap()).unwrap();
+            let export = self.component_entity_type(&mut encoded, &mut types, ty);
+            // Point the type index to the export instead of the concrete type definition
+            types.insert(PtrKey(type_definiton), encoded.type_count());
+            encoded.export(name, url, export);
```
peterhuene, line 150: The test failure is because we only want to do this map update when the item being exported is a type, so that the type map points at the exported (i.e. "named") index rather than the original type definition index.

To do that, I think something like this would work (here and the other places we needed this change):

```suggestion
            let id = match &ty {
                ComponentEntityType::Type { created: id, .. } => Some(*id),
                _ => None,
            };

            let ty = self.component_entity_type(&mut encoded, &mut types, ty);
            if let Some(id) = id {
                let ty = self.0.type_from_id(id).unwrap();
                types.insert(PtrKey(ty), encoded.type_count());
            }

            encoded.export(name, url, ty);
```

crates/wasmparser/src/validator/types.rs
```diff
@@ -816,6 +816,20 @@ impl ComponentEntityType {
             Self::Value(ty) => ty.type_size(),
         }
     }
+
+    /// The underlying `TypeId` for this entity type if one exists
+    pub fn id(&self) -> Option<TypeId> {
```
peterhuene, line 821: I think we can remove this function if we stick to something from the above suggestion.

crates/wasm-compose/tests/compose.rs
```diff
@@ -75,6 +75,13 @@ fn component_composing() -> Result<()> {
                 )
             })?;
 
+            wit_component::decode("component", &bytes).with_context(|| {
+                format!(
+                    "failed to decode component bytes as a wit component `{}`",
```
peterhuene, line 80: Nit: 

```suggestion
                    "failed to decode WIT from component bytes for test cast `{}`",
```

peterhuene (approved):
Looks great, thanks for fixing this!

Just the one minor nit and I think this is good to merge.

peterhuene:
Thanks @rylev!

PR #989, opened 2023-04-19
feat: cabi_realloc_adapter

guybedford:
This adds support for a core `cabi_realloc_adapter` function export as taking precedence over the `cabi_realloc` function export for the implementation of `cabi_realloc` that is provided back to the adapter itself when importing from the component.

This resolves https://github.com/bytecodealliance/preview2-prototyping/issues/145.

### DIFF
--- a/crates/wit-component/src/encoding.rs
+++ b/crates/wit-component/src/encoding.rs
@@ -1171,11 +1171,19 @@ impl<'a> EncodingState<'a> {
 
             let mut core_exports = Vec::new();
             for export_name in info.needs_core_exports.iter() {
+                let mut core_export_name = export_name.as_str();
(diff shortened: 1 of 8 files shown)


guybedford:
I've integrated this fix into componentize-js where the stack allocation attempts were freeing adapter memory inadvertently and can confirm that it fully resolves the issues.

alexcrichton:
I do still feel like the adapter support is hacks-on-hacks but that's also sort of the purpose for its existence. I also can't think of anything that's substantively better here, so looks good to me 👍

Thanks!

alexcrichton (approved):

pchickey:
Thank you for implementing this Guy!

Issue #986, opened 2023-04-17
`wasm-compose` creating an ill formed wasm component


rylev:
I'm still seeing issues after #977 has been fixed. This is what I'm seeing:

```
$ wasm-tools component wit broken-composed.wasm

Error: failed to decode WIT document

Caused by:
    0: failed to decode WIT from export `inbound-http`
    1: failed to register type export 'request'
    2: failed to convert unaliased type
    3: failed to convert record field 'method'
    4: unexpected unnamed type of kind 'enum'
```

Perhaps the issue comes from the following subcomponent in the composed component:

```wat
 (component (;0;)
      (alias outer 1 22 (type (;0;)))
      (import "import-handle-request" (func (;0;) (type 0)))
      (alias outer 1 23 (type (;1;)))
      (export (;2;) "request" (type 1))
      (alias outer 1 24 (type (;3;)))
      (export (;4;) "response" (type 3))
      (type (;5;) (func (param "req" 2) (result 4)))
      (export (;1;) "handle-request" (func 0) (func (type 5)))
    )
```

You'll notice that that request is being exported based on an outer alias. Could it be that this brings request into scope but the types that request relies on like method are still not in scope? Still trying to wrap my head around outer aliases so I might be getting lost here.

peterhuene:
@rylev Thanks for reporting this issue. `wasm-compose` was written prior to how `wit-component` encodes the binary representation of a WIT package, so there's probably some additional discrepancies with what `wit-component::decode` is expecting.

I'll have to look into it, but I suspect it's similar to the problem in #977.

alexcrichton:
The outer aliases are in theory "simply" artifacts of the `wit-component` creation process (what I'm hoping is an inner component of those in the wasm-compose-created-component). Do you have a reproduction I could help poke around though to see if I can see what's up too?

rylev:
I've posted a reproduction on to [a Zulip thread for this issue](https://bytecodealliance.zulipchat.com/#narrow/stream/206238-general/topic/wasm-tools.23986.20ill-formed.20composed.20component/near/351138144).

PR #990, opened 2023-04-19
Add a `--skeleton` option to `wasm-tools print`

alexcrichton:
Often when debugging larger wasm files I want to skip the data segments, element segments, and often entire function bodies. This commit adds a debugging option `--skeleton` to skip printing all these "big" items to ideally only print the "shape" of a module for easier manual inspection. An example issue I'm using this with is #986.


pchickey (approved):
Thanks, I agree this is useful!

PR #997, opened 2023-04-24
Require 'alias core export' when aliasing core item

### LINKED ISSUE #996: Error validating core export alias
Given the following wat:

```wat
(component
  (core module $numbers
    (func (export "one") (result i32) (i32.const 1))
    (func (export "two") (result i32) (i32.const 2))
  )
  (core module $doSomething
    (func (import "theNumbers" "myNumber") (result i32))
  )
  (core instance $firstInstance (instantiate $numbers))
.
  (core func $two (alias core export $firstInstance "two")) ;; this line will throw a validation error
  (core instance $secondInstance (instantiate $doSomething
    (with "theNumbers" (instance
      (export "myNumber" (func $two))
    ))
  ))
)
```

I would expect that this validates properly, but it does not:

```bash
$ wasm-tools validate example.wat -f component-model

Error: expected keyword `export`
     --> examples/out-alias/out-alias.wat:12:26
      |
   12 |   (core func $two (alias core export $firstInstance "two"))
      |
```

According to the spec explainer, "`core export` [is] of a core module instance". Removing the `core` causes the wat to validate even though it shouldn't.

If we switch the alias definition around to its other form...

```wat
(alias core export $firstInstance "two" (core func $two ))
```

... the validation succeeds, and (as expected) when `core` is removed from before `export`, the validation fails. 

> Note: this causes the example used in the spec explainer to not validate.

rylev:
Fixes #996 

When aliasing a core item "alias core export" is now required as per the spec description.

### DIFF
--- a/crates/wast/src/component/alias.rs
+++ b/crates/wast/src/component/alias.rs
@@ -4,17 +4,22 @@ use crate::parser::{Parse, Parser, Result};
 use crate::token::{Id, Index, NameAnnotation, Span};
 
 /// A inline alias for component exported items.
+///
+/// Handles both `core export` and `export` aliases
 #[derive(Debug)]
-pub struct InlineExportAlias<'a> {
+pub struct InlineExportAlias<'a, const CORE: bool> {
     /// The instance to alias the export from.
     pub instance: Index<'a>,
     /// The name of the export to alias.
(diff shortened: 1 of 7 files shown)


alexcrichton (approved):
Thanks!

PR #1004, opened 2023-04-28
GC: wasmparser: Add (ref null? i31) type

imikushin:
Unboxed integers (`i31`) is the most trivial reference type added by the Wasm GC specification, and the easiest to parse 🙂

This PR adds and enables a roundtrip test to demonstrate support for all flavors of `i31` references.

Also, change RefType internal structure to represent types introduced by the GC spec as described below.

RefType is a bit-packed enum that fits in a `u24` aka `[u8; 3]`. 
Note that its content is opaque (and subject to change), but its API is stable. 
It has the following internal structure:
```
[nullable:u1] [indexed==1:u1] [kind:u2] [index:u20]
[nullable:u1] [indexed==0:u1] [type:u4] [(unused):u18]
```
, where 
- `nullable` determines nullability of the ref
- `indexed` determines if the ref is of a dynamically defined type with an index (encoded in a following bit-packing section) or of a known fixed type
- `kind` determines what kind of indexed type the index is pointing to:
  ```
  10 = struct
  11 = array
  01 = function
  ```
- `index` is the type index
- `type` is an enumeration of known types:
  ```
  1111 = any

  1101 = eq
  1000 = i31
  1001 = struct
  1100 = array

  0101 = func    
  0100 = nofunc  

  0011 = extern  
  0010 = noextern

  0000 = none
  ```
- `(unused)` is unused sequence of bits

### DIFF
--- a/tests/snapshots/local/gc/gc-i31.wat.print
+++ b/tests/snapshots/local/gc/gc-i31.wat.print
@@ -1,0 +1,7 @@
+(module
+  (type (;0;) (func (result i32)))
+  (func $f (;0;) (type 0) (result i32)
+    (local $a (ref i31)) (local $b (ref null i31)) (local $c (ref null i31))
+    unreachable
+  )
+)
--- a/crates/wasm-compose/src/encoding.rs
+++ b/crates/wasm-compose/src/encoding.rs
@@ -280,6 +280,7 @@ impl<'a> TypeEncoder<'a> {
             heap_type: match ty.heap_type() {
                 wasmparser::HeapType::Func => HeapType::Func,
                 wasmparser::HeapType::Extern => HeapType::Extern,
+                wasmparser::HeapType::I31 => HeapType::I31,
                 wasmparser::HeapType::TypedFunc(i) => HeapType::TypedFunc(i.into()),
             },
         }
--- a/crates/wasm-encoder/src/core/types.rs
+++ b/crates/wasm-encoder/src/core/types.rs
@@ -103,6 +103,8 @@ pub enum HeapType {
     Func,
     /// An extern reference. When nullable, equivalent to `externref`
     Extern,
+    /// An i31 reference. When nullable, equivalent to `i31ref`
+    I31,
     /// A reference to a particular index in a table.
     TypedFunc(u32),
 }
@@ -112,6 +114,7 @@ impl Encode for HeapType {
         match self {
             HeapType::Func => sink.push(0x70),
             HeapType::Extern => sink.push(0x6F),
+            HeapType::I31 => sink.push(0x6A),
             // Note that this is encoded as a signed type rather than unsigned
             // as it's decoded as an s33
             HeapType::TypedFunc(i) => i64::from(*i).encode(sink),
--- a/crates/wasm-mutate/src/module.rs
+++ b/crates/wasm-mutate/src/module.rs
@@ -87,6 +87,7 @@ pub fn map_ref_type(ref_ty: wasmparser::RefType) -> Result<RefType> {
         heap_type: match ref_ty.heap_type() {
             wasmparser::HeapType::Func => HeapType::Func,
             wasmparser::HeapType::Extern => HeapType::Extern,
+            wasmparser::HeapType::I31 => HeapType::I31,
             wasmparser::HeapType::TypedFunc(i) => HeapType::TypedFunc(i.into()),
         },
     })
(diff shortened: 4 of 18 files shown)


tests/snapshots/local/gc/gc-i31.wat.print
```diff
@@ -0,0 +1,7 @@
+(module
+  (type (;0;) (func (result i32)))
+  (func $f (;0;) (type 0) (result i32)
+    (local $a (ref i31)) (local $b (ref null i31)) (local $c (ref null i31))
```
alexcrichton, line 4: Could a special-case be added to `wasmprinter` to pring `(ref null i31)` as `i31ref` if that's an abbreviation?

alexcrichton (commented):
Thanks for this! Out of curiosity are you interested in implementing the GC proposal in wasmparser? 

Otherwise though, one thing I'm curious about is whether it's worth getting "extra clever" with the representation here. For example an alternative representation would be:

[nullable:u1] [kind:u3] [index:u20]
[nullable:u1] [kind:u5] [(unused):u18]

where `kind:u3` has the first two bits as "what is being index" and the third bit is always set:

```
001 = func
011 = struct
101 = array
```

where alternatively `kind:u5` always has the 3rd bit unset:

```
00000 = func
01000 = struct
10000 = array
    
11000 = any
11001 = eq
11010 = i31
11011 = nofunc  
11100 = extern  
11101 = noextern
11110 = none
```

or... something along those lines. The idea here being that "is this a function" only needs to look at 2 bits and it doesn't matter whether it's indexed or not. Everything else would fit around the indexed func/struct/array.

I don't know whether this is worth it though and it sort of highly depends on what the perf bottlenecks are in gc validation, if any. As I wrote all this out I suspect that going with a simpler scheme like you've implemented is probably the way to go.

tests/snapshots/local/gc/gc-i31.wat.print
fitzgen: Seconding that having `wasmprinter` use the abbreviation short hand when applicable would be great, but it is fine if that is a follow up PR.

fitzgen (approved):
Looks like a great start for adding Wasm GC support to `wasmparser`!

imikushin:
Thanks for the quick response @alexcrichton!

> Thanks for this! Out of curiosity are you interested in implementing the GC proposal in wasmparser?

This is my intention 🙂

> Otherwise though, one thing I'm curious about is whether it's worth getting "extra clever" with the representation here. For example an alternative representation would be:
> 
> [nullable:u1] [kind:u3] [index:u20] 
> [nullable:u1] [kind:u5] [(unused):u18]
> 
> where `kind:u3` has the first two bits as "what is being index" and the third bit is always set:
> 
> ```
> 001 = func
> 011 = struct
> 101 = array
> ```
> 
> where alternatively `kind:u5` always has the 3rd bit unset:
> 
> ```
> 00000 = func
> 01000 = struct
> 10000 = array
>     
> 11000 = any
> 11001 = eq
> 11010 = i31
> 11011 = nofunc  
> 11100 = extern  
> 11101 = noextern
> 11110 = none
> ```
> 
> or... something along those lines. The idea here being that "is this a function" only needs to look at 2 bits and it doesn't matter whether it's indexed or not. Everything else would fit around the indexed func/struct/array.

Hopefully, it can already be determined if the ref is a function using my layout 😅 
This was a goal. It turns out to also work for structs (but not arrays, apparently, as they have the same 2 bits as `any` and `eq`)

> I don't know whether this is worth it though and it sort of highly depends on what the perf bottlenecks are in gc validation, if any. As I wrote all this out I suspect that going with a simpler scheme like you've implemented is probably the way to go.

Yeah, and we can change it (if needed) when we know more.

tests/snapshots/local/gc/gc-i31.wat.print
imikushin: Is there a good reason for that? (beside it being a historical convention)

`(ref null i31)` looks cleaner to me. It kinda bugs me that `i31ref` and `(ref i31)` are subtly different in this way. 

So, I personally would very much prefer `(ref null i31)` which is explicit and obvious.

Also, having to spend extra characters on `null` can makes one think twice if they really want to make a billion dollar mistake 🙂

imikushin:
> Looks like there's still a compile error in the workspace:
> 
> https://github.com/bytecodealliance/wasm-tools/actions/runs/4826112620/jobs/8597648525?pr=1004#step:4:361
> 
> ```
> error[E0063]: missing field `gc` in initializer of `WasmFeatures`
>    --> crates/wasm-smith/tests/core.rs:285:5
>     |
> 285 |     WasmFeatures {
>     |     ^^^^^^^^^^^^ missing `gc`
> ```
> 
> FYI, you can build all crates in the workspace and run all of their tests with `cargo test --all`.

Thanks! Fixed.

fitzgen (approved):

PR #1006, opened 2023-04-28
wasmparser: `{return_,}call_ref` take `u32` type index immediate

nagisa:
Not 100% sure if my reading of the proposal is correct, but my understanding is that these instructions decode a u32 type index immediate parameter, and not the `heaptype`.


crates/wasmparser/src/validator/operators.rs
```diff
@@ -1270,7 +1270,8 @@ where
         self.check_return()?;
         Ok(())
     }
-    fn visit_call_ref(&mut self, hty: HeapType) -> Self::Output {
+    fn visit_call_ref(&mut self, type_index: u32) -> Self::Output {
+        let hty = HeapType::TypedFunc(type_index);
```
nagisa, line 1274: This smells like something that could be validated more directly…

alexcrichton:
I believe this is correct given my reading of the [current spec as well](https://webassembly.github.io/function-references/core/binary/instructions.html#control-instructions), so thanks!

I think there's some test failures that can be fixed with `BLESS=1`, and additionally it looks like `wasmprinter` may be putting two spaces between `call_ref` and the type argument, so would you be up for fixing that while you're here?

nagisa:
Yeah, I’m planning to fix things up so that its all mergeable. Was just waiting for a confirmation that my understanding is correct, before I sunk too much time into a misunderstanding :)

alexcrichton:
Oh this may also want to update the `wast` crate's parsing of the `CallRef` instruction I think?

nagisa:
This is good to go, I believe.

alexcrichton:
Would you be up for updating [these references as well](https://github.com/bytecodealliance/wasm-tools/blob/3b84bcba18a5e0ceb1e5dd27d7997284177082d5/crates/wast/src/core/expr.rs#L525-L526) along with [these](https://github.com/bytecodealliance/wasm-tools/blob/3b84bcba18a5e0ceb1e5dd27d7997284177082d5/crates/wasm-encoder/src/core/code.rs#L325C23-L327)?

nagisa:
I went ahead and updated `wast`, but `wasm-encoder` was already changed to `u32`s in a previous change. Did you mean something else, perhaps?

alexcrichton:
Oops sorry no that's fine, sorry it's been a bit of a week for me.

alexcrichton (approved):

nagisa:
Would it be possible to get a release with this change in?

alexcrichton:
Sure yeah, I've posted that to https://github.com/bytecodealliance/wasm-tools/pull/1025

PR #1009, opened 2023-04-28
GC: wasmprinter: Print shorthand i31ref type alias

imikushin:
For consistency with funcref and externref.

Follow-up for https://github.com/bytecodealliance/wasm-tools/pull/1004#discussion_r1180675039


fitzgen (approved):
Thanks!

PR #1010, opened 2023-04-28
GC: wasmparser: Add the rest of heap types

imikushin:
This adds all of the heap types and the corresponding ref types from the GC spec. 

Tested by another enabled roundtrip test: `tests/local/gc/gc-heaptypes.wat`


crates/wasm-encoder/src/core/types.rs
```diff
@@ -67,6 +67,8 @@ impl RefType {
         nullable: true,
         heap_type: HeapType::Extern,
     };
+
+    // TODO(imikushin): add all other RefTypes
```
fitzgen, line 71: All of them are added now, no? Can this comment be removed?

fitzgen (commented):
LGTM modulo a question about a `TODO` comment

crates/wasm-encoder/src/core/types.rs
imikushin: Silly me :)

Removed.

PR #1012, opened 2023-05-03
🤓 [wasm-parser] Add readers for core dump custom sections.

itsrainy:
These are implemented following the [spec in the tool-conventions repo](https://github.com/WebAssembly/tool-conventions/blob/main/Coredump.md).

### DIFF
--- a/crates/wasmparser/src/readers/core/coredumps.rs
+++ b/crates/wasmparser/src/readers/core/coredumps.rs
@@ -1,0 +1,150 @@
+use crate::{BinaryReader, FromReader, Result};
+
+/// The data portion of a custom section representing a core dump. Per the
+/// tool-conventions repo, this section just specifies the executable name that
+/// the core dump came from while the rest of the core dump information is
+/// contained in a corestack custom section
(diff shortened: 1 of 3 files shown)


crates/wasmparser/src/readers/core/coredumps.rs
```diff
@@ -0,0 +1,150 @@
...
+/// assert!(frame.funcidx == 42);
+/// assert!(frame.codeoffset == 51);
+/// assert!(frame.locals.len() == 1);
+/// assert!(frame.stack.len() == 1);
+/// ```
+pub struct CoreDumpStackSection<'a> {
+    /// The thread name
+    pub name: &'a str,
+    /// The stack frames for the core dump
+    pub frames: Vec<CoreDumpStackFrame>,
```
fitzgen, line 57: We might eventually want to split these out into a sub-reader and not eagerly parse and allocate a `Vec` for them, but this is fine for now.

fitzgen (approved):
Looks great, thanks!

PR #1014, opened 2023-05-03
GC: wasmparser: Parse i31 ops

imikushin:
Parse GC ops: 
- `i31.new`
- `i31.get_s`
- `i31.get_u`

### DIFF
--- a/crates/wasmparser/src/readers/core/types.rs
+++ b/crates/wasmparser/src/readers/core/types.rs
@@ -244,7 +244,7 @@ impl RefType {
 
     const INDEX_MASK: u32 = (1 << 20) - 1; // 20 bits #19-#0 (if `indexed == 1`)
 
-    /// An nullable untyped function reference aka `(ref null func)` aka
+    /// A nullable untyped function reference aka `(ref null func)` aka
     /// `funcref` aka `anyfunc`.
     pub const FUNCREF: Self = RefType::from_u32(Self::NULLABLE_BIT | Self::FUNC_TYPE);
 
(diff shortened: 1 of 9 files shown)


crates/wasmparser/src/validator/operators.rs
```diff
@@ -3335,6 +3336,24 @@ where
...
+    fn visit_i31_get_s(&mut self) -> Self::Output {
+        match self.pop_ref()? {
+            Some(RefType::REF_I31) | Some(RefType::I31REF) => {}
+            _ => bail!(self.offset, "type mismatch: expected (ref null? i31)"),
+        };
+        self.push_operand(ValType::I32)
+    }
+    fn visit_i31_get_u(&mut self) -> Self::Output {
+        match self.pop_ref()? {
+            Some(RefType::REF_I31) | Some(RefType::I31REF) => {}
```
alexcrichton, line 3352: Could this perhaps be modeled as `ty.heap_type() == HeapType::I31`? That feels a little more clear to me that it's accepting both nullable and non-nullable references.

crates/wasmparser/src/readers/core/types.rs
```diff
@@ -284,6 +284,36 @@ impl RefType {
...
+    pub const REF_EQ: Self = RefType::from_u32(Self::EQ_TYPE);
+
+    /// A non-nullable reference to a struct aka `(ref struct)`.
+    pub const REF_STRUCT: Self = RefType::from_u32(Self::STRUCT_TYPE);
+
+    /// A non-nullable reference to an array aka `(ref array)`.
+    pub const REF_ARRAY: Self = RefType::from_u32(Self::ARRAY_TYPE);
+
+    /// A non-nullable reference to an i31 object aka `(ref i31)`.
+    pub const REF_I31: Self = RefType::from_u32(Self::I31_TYPE);
```
alexcrichton, line 315: While I realize this is following the naming of the text format, `I31REF` vs `REF_I31` feels a little too subtle to me as the only distinguishing factor being the nullability. I'm not sure of better names myself, though, so could this perhaps see how long we can go without the `REF_*` names instead?

alexcrichton (commented):
Looks reasonable to me! I'm curious if you think the `REF_*` names will be common so we should keep them though, and if so it may be good to find a way to insert "non null" into the name somehow.

crates/wasmparser/src/readers/core/types.rs
imikushin: I've added these new constants because it was awkward to use `RefType::new()`, which returns `Option<RefType>` and then pulling the actual RefType out of it. 

I'm more than happy to rename the constants introduced in a prior PR with explicit NULL, e.g. `REF_NULL_ANY` instead of `ANYREF` (modeled by the shorthand names, which IMHO simply introduce keywords without new meaning).

imikushin: Also, more than happy to replace printed shorthand reftypes with `(ref null <heap_type>)` throughout.

crates/wasmparser/src/validator/operators.rs
imikushin: Just thought I'd follow the historical convention with those shorthand types :)

crates/wasmparser/src/readers/core/types.rs
alexcrichton: Bikeshed: how about `RefType::I31` for a non-nullable i31 and `RefType::I31.nullable()` for a nullable i31? This'd remove the `RefType::*REF` types like `RefType::ANYREF` and would be replaced with `RefType::ANY.nullable()` for example.

crates/wasmparser/src/validator/operators.rs
alexcrichton: Oh does this exist elsewhere already? If so we should probably go back and update those too (no need to do that in this PR though).

crates/wasmparser/src/readers/core/types.rs
imikushin: Love this idea 👍

imikushin: 🤔 some of the (older) `*REF`s (like `FUNCREF` and `EXTERNREF`) are already used in pattern matches and other stuff a bit.

Let me see if I can safely remove the newer ones keeping the older ones intact.

imikushin: Turns out, one can't simply use `RefType::ANY.nullable()` as a pattern match, even if `nullable()` is a `const fn`!

imikushin: Fixed!

alexcrichton (approved):
Keeping the old ones makes sense as a temporary holdover to me, I think that's a reasonable compromise 👍

Issue #1013, opened 2023-05-03
wasmparser static `TypeId` assert can trigger with `-Zrandomize-layout`


repi:
The following wasmparser code can fail if compiling with nightly and `-Zrandomize-layout` due to a static assert triggering on the size of `TypeId` when the field order gets reshuffled.

```rust
/// Represents a unique identifier for a type known to a [`crate::Validator`].
#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
pub struct TypeId {
    /// The index into the global list of types.
    pub(crate) index: usize,
    /// The effective type size for the type.
    ///
    /// This is stored as part of the ID to avoid having to recurse through
    /// the global type list when calculating type sizes.
    pub(crate) type_size: u32,
    /// A unique integer assigned to this type.
    ///
    /// The purpose of this field is to ensure that two different `TypeId`
    /// representations can be handed out for two different aliased types within
    /// a component that actually point to the same underlying type (as pointed
    /// to by the `index` field).
    unique_id: u32,
}

// The size of `TypeId` was seen to have a large-ish impact in #844, so this
// assert ensures that it stays relatively small.
const _: () = {
    assert!(std::mem::size_of::<TypeId>() <= 16);
};
```

should we increase the limit her, remove the assert, or use `repr(C)` on `TypeId` to enforce the field order of this type? 

We ran into this when building our application with nightly and `-Zrandomize-layout` which builds all dependent crates including wasmtime and wasmparser with it as well and this failed to build. Another option could be to move the assert to a unit test so it only gets triggered in the development of wasmparser and not in usages of the crate?

alexcrichton:
Oh dear sorry for the breakage! I think it's probably best to use `#[repr(C)]` here which I think should fix the issue for you. If you'd like to send a PR that'd be appreciated!

PR #1015, opened 2023-05-05
Freeze field layout of `TypeId`

repi:
When compiling with `-Zrandomize-layout` the fields of `TypeId` can be rearranged in a way where the type becomes larger than 16 bytes which triggered a static assert.

Resolves: #1013


alexcrichton (approved):
Thanks! Could you also add a small comment as to why it's `repr(C)`?

PR #1016, opened 2023-05-05
Don't generate WIT with the name `memory`

alexcrichton:
This commit fixes a fuzzer-found issue in the `wit-smith` crate where if a function is called `memory` then it's not possible to feed such a WIT document through `wit-component` at this time. This is because linear memory is called `memory` and must be called that. Until this restriction is later lifted perhaps this claims the name `memory` early on to prevent any interface items from being named that.


pchickey (approved):
Should `cabi_realloc` be added as well?

alexcrichton:
Thankfully that one's already covered since it's inherently not a kebab-case name, so the core module won't ever have a conflict with that (e.g. `cabi-realloc: func()` would generate an export of `"cabi-realloc"` which is ok)

PR #1017, opened 2023-05-09
Add a simple `addr2line` subcommand

alexcrichton:
This commit adds a simple `addr2line` subcommand to `wasm-tools` which can be used for parsing a WebAssembly binary's custom sections looking for DWARF and mapping addresses to filenames and line numbers. This is modelled somewhat after the native `addr2line` utility where multiple addresses can be passed and inlined frames and such are printed. Overall this is a pretty thin wrapper around the `addr2line` crate and is similar to Wasmtime's support for symbolicating a backtrace.


pchickey (approved):
Nice and simple

PR #1023, opened 2023-05-11
GC: wasmparser: storage types

imikushin:
This builds on #1022 and adds support for packed storage types for array and struct fields.

A storage type is either a packed storage type (`i8`, `i16`) or a regular value type.


fitzgen (commented):
These changes LGTM, so once the base lands we can rebase and land this.

fitzgen:
This PR needs a rebase now.

PR #1026, opened 2023-05-12
[wasmparser/wasm-encoder] Update coredump encoding and parsing

itsrainy:
This updates the core dump representations to align with https://github.com/WebAssembly/tool-conventions/pull/207, namely:
- Adding parsing/encoding for "coremodules" sections
- Adding parsing/encoding for "coreinstances" sections
- Adding an "instance_index" to each encoded stack frame
- Adding roundtrip tests for the above and updating the existing tests

### DIFF
--- a/crates/wasmparser/src/readers/core/coredumps.rs
+++ b/crates/wasmparser/src/readers/core/coredumps.rs
@@ -125,14 +125,15 @@ impl<'a> FromReader<'a> for CoreDumpInstance {
 /// # Examples
 ///
 /// ```
-/// let data: &[u8] = &[0x00, 0x04, 0x6d, 0x61, 0x69, 0x6e, 0x01, 0x00, 0x2a,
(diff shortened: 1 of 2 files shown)


crates/wasmparser/src/readers/core/coredumps.rs
```diff
@@ -31,6 +31,94 @@ impl<'a> FromReader<'a> for CoreDumpSection<'a> {
...
+    /// section.
+    pub module_index: u32,
+
+    /// Which of the coredump's memories are this instance's memories, via
+    /// indexing into the memory index space.
+    pub memories: Vec<u32>,
+
+    /// Which of the coredump's globals are this instance's globals, via
+    /// indexing into the global index space.
+    pub globals: Vec<u32>,
```
fitzgen, line 94: All of these `Vec`s we will probably want to turn into sub-readers eventually, but we can wait till we have a reason to do that.

fitzgen (approved):
LGTM!

PR #1033, opened 2023-05-16
fix: enable verbose to component wit opts

### LINKED ISSUE #1032: Verbose does not work on `component wit` command
It appears that `wasm-tools component wit <path> --verbose` does not actually print the logs.  I suspect that this is due to the fact that `self.verbose.init_logger()` is not invoked in `run`. Will submit a PR to fix this issue.

Mossaka:
Fixes #1032


alexcrichton:
Ah this just recently changed in https://github.com/bytecodealliance/wasm-tools/pull/1028, so I think this may no longer be necessary? Does the latest `wasm-tools` version work for you?

PR #1040, opened 2023-05-16
Add force-merge option for adapters

guybedford:
I've got a workflow where I'm adapting a preview1 component to preview2, where that component itself has world bindings to preview2 as well. The version APIs exactly match (so there should be type equality), but I'm getting the error:

```
     Error: failed to merge WIT world of adapter `wasi_snapshot_preview1` into main package

Caused by:
    duplicate import found for interface `environment`
```

This change adds a `force_merge` option to the adapter to inform that imports are known to be equivalent.


guybedford:
Turned out I was using bad binary versions - so this is unnecessary after all.

pchickey:
I was about to say, every time I've hit this error it was from stale binaries

PR #1042, opened 2023-05-17
Rename HeapType::TypedFunc(u32) to Indexed(u32)

imikushin:
`HeapType::TypedFunc(u32)` has been (as of [typed references](https://github.com/WebAssembly/function-references) proposal) used to refer function types defined in the module.

It is parsed from `(ref null? {index})`. With introduction of new indexed types for GC, there's unfortunately no way to tell what type the index is pointing to while parsing: they used to be all `func` types, but now can also be `array`, `struct` and more.

Renaming HeapType::TypedFunc(u32) to Indexed(u32) reflects this change.

Summary of the changes:
- Match RefTypes where one or both has HeapType::Indexed.
- `RefType::new()` will now create indexed RefType with unknown kind.
- Use `RefType::indexed_func()` instead of `RefType::new()` for indexed func refs.
- Add `Self::ARRAY_KIND` to the list of matched RefType kinds - when getting HeapType of a RefType.
- Check if two indexed types are equal.

### DIFF
--- a/crates/wasmparser/src/readers/core/types.rs
+++ b/crates/wasmparser/src/readers/core/types.rs
@@ -192,8 +192,8 @@ impl Debug for RefType {
             (false, HeapType::Extern) => write!(f, "(ref extern)"),
             (true, HeapType::Func) => write!(f, "funcref"),
             (false, HeapType::Func) => write!(f, "(ref func)"),
-            (true, HeapType::TypedFunc(idx)) => write!(f, "(ref null {idx})"),
-            (false, HeapType::TypedFunc(idx)) => write!(f, "(ref {idx})"),
+            (true, HeapType::Indexed(idx)) => write!(f, "(ref null {idx})"),
+            (false, HeapType::Indexed(idx)) => write!(f, "(ref {idx})"),
         }
     }
 }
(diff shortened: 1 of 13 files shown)


crates/wasmparser/src/readers/core/types.rs
```diff
@@ -333,15 +333,6 @@ impl RefType {
...
         debug_assert!(x & (0b11111111 << 24) == 0);
 
-        // if indexed, kind must be struct/array/func
-        debug_assert!(
-            x & Self::INDEXED_BIT == 0
-                || matches!(
-                    x & Self::KIND_MASK,
-                    Self::FUNC_KIND | Self::ARRAY_KIND | Self::STRUCT_KIND
-                )
-        );
```
imikushin, line 343: If the RefType refers to an indexed type, the 2-bit area masked by `Self::KIND_MASK` can be anything. `00` means "we don't know the kind".

fitzgen: I think we might actually want to completely remove the kind from indexed `RefType`s, since it is redundant with the information in the actual type section. Maintaining the same information in multiple places just allows the possibility for things to get out of sync. Can you make a follow up to remove the indexed kind and actually index into the type section instead everywhere where we are currently using the kind?

PR #1043, opened 2023-05-17
[wast] Add support for final types.

jpages:
The GC proposal contains a final type: https://github.com/WebAssembly/gc/blob/main/proposals/gc/MVP.md#type-definitions

This commit add support for it in wast. All types are considered final unless the `sub` keyword is used.
A type marked `(sub final)` will be final.

We are currently implementing this at Mozilla (see https://bugzilla.mozilla.org/show_bug.cgi?id=1825088)

### DIFF
--- a/crates/wast/src/core/binary.rs
+++ b/crates/wast/src/core/binary.rs
@@ -173,10 +173,25 @@ impl Encode for RecOrType<'_> {
 
 impl Encode for Type<'_> {
     fn encode(&self, e: &mut Vec<u8>) {
-        if let Some(parent) = &self.parent {
-            e.push(0x50);
-            (1 as usize).encode(e);
-            parent.encode(e);
+        match (&self.parent, self.final_type) {
+            (Some(parent), Some(true)) => {
+                // Type is final with a supertype
+                e.push(0x4e);
+                e.push(0x01);
+                parent.encode(e);
+            }
+            (Some(parent), Some(false) | None) => {
+                // Type is not final and has a declared supertype
+                e.push(0x50);
+                e.push(0x01);
+                parent.encode(e);
+            }
+            (None, Some(false)) => {
+                // Sub was used without any declared supertype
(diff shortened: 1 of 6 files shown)


crates/wast/src/core/binary.rs
```diff
@@ -174,10 +174,22 @@ impl Encode for RecOrType<'_> {
...
             (1 as usize).encode(e);
             parent.encode(e);
         }
+        else {
+            // No supertype but sub was used
+            if self.final_type.is_some() && self.final_type.unwrap() == false {
+                e.push(0x50);
+                (0 as usize).encode(e);
+            }
+        }
```
alexcrichton, line 192: To simplify this a bit, could it perhaps become:

```rust
match (&self.parent, self.final_type) {
    (Some(parent), Some(true)) => {
        e.push(0x4e);
        e.push(0x01); // same as `(1 as usize).encode(e)` I think
        parent.encode(e);
    }
    (Some(parent), Some(false) | None) => {
        e.push(0x50);
        e.push(0x01);
        parent.encode(e);
    }
    (None, Some(false)) => {
        e.push(0x50);
        e.push(0x00);
    }
    (None, _) => {} // comment about why no prefix is necessary
}
```

Or.. well after I type this out I'm less certain. I don't know anything much about the gc spec but I'm not sure I fully understand this encoding scheme, so I'm also happy to defer to you.

alexcrichton (approved):
Thanks! I don't follow the encoding scheme for this much but I also haven't taken the time to read the upstream proposal, so that's probably on me. I left a suggestion but feel free to say "no it's better as-is" too

crates/wast/src/core/binary.rs
jpages: Thanks, it's definitely easier to understand that way! I rebased with your suggestion and some additional comments.

alexcrichton (approved):

PR #1044, opened 2023-05-19
Add text syntax for the producers custom section

alexcrichton:
This commit adds textual syntax for the `producers` custom section to the `wast` parser. This support is added with the [annotations proposal][annot] where the `@producers` field is now reserved for this custom section. The impact of this commit is that the `@producers` field will now show up by default for many binaries such as those produced by LLVM. This will make the textual format unparseable for any text tool which doesn't support the custom annotations proposal. Note, though, that `wast` has supported this proposal for awhile now and the text should work seamlessly with older versions.

Some extra validation is added during parsing that the field names are `language`, `sdk`, and `processed-by` to ensure other names don't leak through.

[annot]: https://github.com/WebAssembly/annotations/


pchickey (approved):
Nice work!

PR #1051, opened 2023-05-23
gc: Gate value types on enabled features

alexcrichton:
This commit adds gates for types in the GC proposal when present in value types which gates presence in locations like function parameters, block types, instruction payloads, etc.


fitzgen (approved):
Thanks!

PR #1053, opened 2023-05-28
added initial parsing for resources and shared handles

silesmo:
Adds the initial pass for the parsing of resources and shared handles.

Do we want to store resources as a `TypeDefKind`? I assume we might want to move this to an indexmap of resources and just store a resource id in handles?

@sunfishcode

### DIFF
--- a/crates/wit-parser/src/ast/resolve.rs
+++ b/crates/wit-parser/src/ast/resolve.rs
@@ -652,7 +652,7 @@ impl<'a> Resolver<'a> {
                         assert!(prev.is_none());
                     }
                     ValueKind::Static(_) => {
-                        unreachable!("static functions are only supported in resources")
+                        bail!("static functions are only supported in resources")
                     }
                 },
                 ast::InterfaceItem::Use(_) | ast::InterfaceItem::TypeDef(_) => {}
--- a/crates/wit-parser/src/ast/resolve.rs
+++ b/crates/wit-parser/src/ast/resolve.rs
@@ -84,7 +84,6 @@ enum Key {
     Variant(Vec<(String, Option<Type>)>),
     Handle(Handle),
     Record(Vec<(String, Type)>),
-    Resource(Vec<(String, Type)>),
     Flags(Vec<String>),
     Tuple(Vec<Type>),
     Enum(Vec<String>),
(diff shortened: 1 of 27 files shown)


crates/wit-parser/src/ast/resolve.rs
```diff
@@ -650,6 +651,9 @@ impl<'a> Resolver<'a> {
                             .insert(value.name.name.to_string(), func);
                         assert!(prev.is_none());
                     }
+                    ValueKind::Static(_) => {
+                        unreachable!("static functions are only supported in resources")
```
sunfishcode, line 655: This should be a `bail!` rather than an `unreachable!` since it indicates an error in the source code.

sunfishcode (commented):
The parsing and syntax looks good to me!

My instinct here is that resources would be better off not being a `TypeDefKind`, because it's my understanding they'll never be first-class. The only place they can ever appear is as the referent of handles, and also, handles don't refer to anything other than resources.

crates/wit-parser/src/ast/resolve.rs
silesmo: Yes agreed, I will make that change.

silesmo:
> The parsing and syntax looks good to me!
> 
> My instinct here is that resources would be better off not being a `TypeDefKind`, because it's my understanding they'll never be first-class. The only place they can ever appear is as the referent of handles, and also, handles don't refer to anything other than resources.

Yes that makes sense. I was unsure if we wanted to support types other than resources in the future. If that's not the case then I will change resources to not be `TypeDefKind`.

sunfishcode:
Looks good!

sunfishcode (approved):

alexcrichton:
Thanks for working on this!

@silesmo are you able to fill work on filling out the remaining `todo!()`? I think there'll be a fair amount of design work to get them all filled out, but we can hopefully help out with them over time too.

Additionally would you be up for updating `WIT.md` upstream with the changes here? I think it's good to keep a documentation record of what this crate supports, even if it's worded along the lines of "this will change in the future" to ensure there's somewhere documenting what's supported.

silesmo:
> Thanks for working on this!
> 
> @silesmo are you able to fill work on filling out the remaining `todo!()`? I think there'll be a fair amount of design work to get them all filled out, but we can hopefully help out with them over time too.
> 
> Additionally would you be up for updating `WIT.md` upstream with the changes here? I think it's good to keep a documentation record of what this crate supports, even if it's worded along the lines of "this will change in the future" to ensure there's somewhere documenting what's supported.

@alexcrichton Yes, my plan is to start working on that this weekend!

Yes I agree, I will make sure to update the documentation as well :)

PR #1059, opened 2023-06-09
GC: wasmparser: Subtypes

imikushin:
Introducing the `SubType` struct:
```rust
pub struct SubType {
    pub is_final: bool,
    pub supertype_idxs: Vec<u32>,
}
```

Here, `StructuralType` is an enum (previously named `Type`).

`supertype_idxs` is a vector of indexes of supertypes, limited to at most one supertype as per GC MVP.

A valid subtype matches its supertypes. Supertypes cannot be final.


imikushin:
x86_64-linux and x86_64-windows CI jobs were canceled for some reason.

crates/wasmparser/src/readers/core/types.rs
```diff
@@ -840,15 +1022,50 @@ pub struct TagType {
...
 
-impl<'a> FromReader<'a> for Type {
+impl<'a> FromReader<'a> for StructuralType {
+    fn from_reader(reader: &mut BinaryReader<'a>) -> Result<Self> {
+        read_structural_type(reader.read_u8()?, reader)
+    }
+}
+
+fn read_structural_type(
+    op_code: u8,
```
fitzgen, line 1034: Nitpick: could we name this something other than `op_code` here and below in the `FromReader` impl for `SubType`? That makes me think of instruction opcodes where as this doesn't have anything to do with instructions, just types.

Perhaps `type_discriminant`? Or `leading_byte`?

crates/wasmparser/src/validator/component.rs
```diff
@@ -952,7 +953,11 @@ impl ComponentState {
 
         self.check_options(None, &info, &options, types, offset)?;
 
-        let lowered_ty = Type::Func(info.into_func_type());
+        let lowered_ty = Type::Sub(SubType {
+            is_final: false,
+            supertype_idxs: vec![],
+            structural_type: StructuralType::Func(info.into_func_type()),
+        });
```
fitzgen, line 960: Looking at all these conversions between `Type`, `SubType`, `StructuralType`, and each structural type variant, it seems like we could have some more `From` implementations to make this less noisy and more automatic.

To really benefit from this, and because `From` is not transitive by default, I think we would want all of these impls:

```rust
impl From<FuncType> for StructuralType { ... }
impl From<ArrayType> for StructuralType { ... }
impl From<StructType> for StructuralType { ... }

impl From<StructuralType> for SubType { ... }
impl From<FuncType> for SubType { ... }
impl From<ArrayType> for SubType { ... }
impl From<StructType> for SubType { ... }

impl From<SubType> for Type { ... }
impl From<StructuralType> for Type { ... }
impl From<FuncType> for Type { ... }
impl From<ArrayType> for Type { ... }
impl From<StructType> for Type { ... }
```

This can be done in a follow up PR if you prefer.

crates/wasmparser/src/validator/core.rs
```diff
@@ -494,14 +495,85 @@ pub(crate) struct Module {
...
+        offset: usize,
+    ) -> Result<Type> {
+        if ty.supertype_idxs.len() > 1 {
+            return Err(BinaryReaderError::new(
+                "subtypes with multiple supertypes not supported",
+                offset,
+            ));
+        }
+        for type_index in ty.supertype_idxs.iter() {
+            // check the supertype exists, is not final and the subtype matches it
```
fitzgen, line 528:
```suggestion
            // Check the supertype exists, is not final, and the subtype matches it.
```

crates/wasmparser/src/validator/types.rs
```diff
@@ -208,23 +205,35 @@ impl Type {
     /// Converts the type to a core function type.
     pub fn as_func_type(&self) -> Option<&FuncType> {
         match self {
-            Self::Func(ty) => Some(ty),
+            Type::Sub(SubType {
+                is_final: _,
+                supertype_idxs: _,
+                structural_type: StructuralType::Func(ft),
+            }) => Some(ft),
```
fitzgen, line 212: You can use `..` to ignore multiple fields at a time, instead of writing them all out here and below:

```suggestion
            Type::Sub(SubType {
                structural_type: StructuralType::Func(ft),
                ..
            }) => Some(ft),
```

tests/local/gc/gc-subtypes.wat
```diff
@@ -0,0 +1,7 @@
+;; --enable-gc
+
+(module
+  (type $a (func))
+  (type $b (sub $a (func)))
+  (type (sub final $a (func)))
+)
```
fitzgen, line 7: I think we need more tests for subtyping, basically every branch of all the `Matches` implementations should be covered by our tests:

* [ ] val types that are and are not the same
* [ ] ref types and each case of same/different nullability
* [ ] heap types that are the same, not the same but sub types, and not subtypes
* [ ] each of the "known"/abstract heap types (`struct`, `noextern`, `eq`, etc)
* [ ] basically, each case in `Matches for HeapType`
* [ ] final vs not final sub types in all combinations
* [ ] super types that the sub type does vs doesn't match
* [ ] too many super types

there are a couple of these partially covered in the `gc-subtypes-invalid.wast` test, but we need to be very thorough and deliberate around exercising all these corners and combinations.

fitzgen (commented):
Thanks!

This is looking good, but I have a few nitpicks and also a request for more complete testing.

Once those things are addressed we can merge this.

crates/wasmparser/src/readers/core/types.rs
imikushin: Would `opcode` (without the `_`) work? Seems to be the term used for this purpose in the spec (for example, in https://webassembly.github.io/spec/core/appendix/index-types.html)

Alternatively, how about `type_code`?

crates/wasmparser/src/validator/component.rs
imikushin: Added an issue to follow this up: #1071

crates/wasmparser/src/readers/core/types.rs
fitzgen: Ah, I didn't realize the spec used `opcode` for this. In that case, yes, let's go with that!

crates/wasm-encoder/src/core/types.rs
```diff
@@ -250,11 +251,29 @@ impl TypeSection {
...
-            self.field(f.ty, f.mutable);
+            self.field(&f.ty, f.mutable);
         }
         self.num_added += 1;
         self
     }
+
+    /// Define an explicit subtype in this type section.
+    /// Must be followed by the definition of the structural type (func, array, struct).
+    pub fn subtype(&mut self, is_final: bool, supertype_idxs: &Vec<u32>) -> &mut Self {
```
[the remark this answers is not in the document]
imikushin: Fixed.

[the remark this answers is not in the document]
imikushin: Took approach (1).

tests/local/gc/gc-subtypes.wat
imikushin: Added more tests.

fitzgen (approved):
LGTM with a couple minor nitpicks. I'll be on vacation for a week starting tomorrow, so maybe @alexcrichton can merge this once you've addressed the nitpicks.

Thanks!

crates/wasm-encoder/src/core/types.rs
```diff
@@ -250,11 +322,45 @@ impl TypeSection {
...
+        } else if ty.supertype_idx.is_some() {
+            self.bytes.push(0x50);
+        }
+        if ty.supertype_idx.is_some() {
+            // for both final and non-final
+            1.encode(&mut self.bytes);
+            if let Some(idx) = ty.supertype_idx {
+                idx.encode(&mut self.bytes);
+            }
+        }
```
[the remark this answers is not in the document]
imikushin: Reverting this suggestion for now: this seems to have broken the tests locally 🤔

tests/local/gc/gc-subtypes.wat
```diff
@@ -0,0 +1,112 @@
+;; --enable-gc
+
+(module
+  (type $a (func))
+  (type $b (sub $a (func)))
+  (type $c (sub $b (func)))
+  (type $b1 (sub final $a (func)))
+  ;; (type (sub final $b1 (func))) ;; invalid
+  ;; (type (sub $a $b (func))) ;; invalid: WAT parser error: "expected `(`"
```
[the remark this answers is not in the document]
imikushin: Already testing these in `gc-subtypes-invalid.wast`. Removing.

crates/wasm-encoder/src/core/types.rs
```diff
@@ -250,11 +322,45 @@ impl TypeSection {
...
         for f in fields.iter() {
-            self.field(f.ty, f.mutable);
+            self.field(&f.element_type, f.mutable);
         }
         self.num_added += 1;
         self
     }
+
+    /// Define an explicit subtype in this type section.
+    /// Must be followed by the definition of the structural type (func, array, struct).
```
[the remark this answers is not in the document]
imikushin: Thanks! Great catch.

crates/wit-component/src/gc.rs
```diff
@@ -574,34 +574,14 @@ impl<'a> Module<'a> {
...
-                        .collect();
-                    types.struct_(fs);
-                }
+            let mut is_empty_func_type = false;
+            types.subtype(&map.subtype(&ty), &mut is_empty_func_type);
+
+            // Keep track of the "empty type" to see if we can reuse an
+            // existing one or one needs to be injected if a `start`
+            // function is calculated at the end.
+            if is_empty_func_type {
```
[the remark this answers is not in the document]
imikushin: I agree, this _is_ kinda ugly. Thanks for pointing it out!

crates/wasm-encoder/src/core/types.rs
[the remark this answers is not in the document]
imikushin: Using a different approach: turn the option into a vector (and encoding it).

crates/wit-component/src/gc.rs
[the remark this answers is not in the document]
imikushin: Fixed.

imikushin:
@alexcrichton I believe I've addressed all @fitzgen's feedback (thank you!).

crates/wast/src/names.rs
```diff
@@ -51,11 +51,13 @@ impl<'a> Namespace<'a> {
     }
 
     pub fn register_specific(&mut self, name: Id<'a>, index: u32, desc: &str) -> Result<(), Error> {
-        if let Some(_prev) = self.names.insert(name, index) {
-            return Err(Error::new(
-                name.span(),
-                format!("duplicate identifier for {}", desc),
-            ));
+        if let Some(prev) = self.names.insert(name, index) {
```
alexcrichton, line 54: I think it would be best to leave out this change since, as you commented on the original issue, this isn't a complete fix. I find that this makes the state of things more confusing since it appears to sometimes work but still sometimes not work.

crates/wasmparser/src/validator/core.rs
```diff
@@ -494,14 +495,71 @@ pub(crate) struct Module {
...
+    }
+
+    fn check_subtype(
+        &mut self,
+        ty: SubType,
+        features: &WasmFeatures,
+        types: &mut TypeAlloc,
+        offset: usize,
+    ) -> Result<Type> {
+        if let Some(type_index) = ty.supertype_idx {
```
alexcrichton, line 521: I think this may need a feature check for `gc`?

crates/wasmparser/src/readers/core/types.rs
```diff
@@ -50,6 +53,12 @@ const _: () = {
     assert!(std::mem::size_of::<ValType>() == 4);
 };
 
+pub(crate) trait Matches {
+    fn matches<'a, F>(&self, other: &Self, type_at: &F) -> Result<bool>
+    where
+        F: Fn(u32) -> Result<&'a SubType>;
```
alexcrichton, line 59: One thing I might recommend is to remove the `Result` here since indices at this point should always be valid because the definitions of types are validated to all be correct. Additionally it might be better to pass a concrete type here rather than a generic `Fn(u32) -> &SubType` but instead have something like `cx: &Validator` or similar.

crates/wasmparser/src/readers/core/types.rs
```diff
@@ -510,6 +531,17 @@ impl RefType {
...
 }
 
+impl Matches for RefType {
+    fn matches<'a, F>(&self, other: &Self, type_at: &F) -> Result<bool>
+    where
+        F: Fn(u32) -> Result<&'a SubType>,
+    {
+        Ok((*self == *other)
+            || (other.is_nullable() || !self.is_nullable())
+                && self.heap_type().matches(&other.heap_type(), type_at)?)
```
alexcrichton, line 541: Could parentheses be used to clarify the order of operations here? I don't know the binding precedence of `||` vs `&&` myself

crates/wasmparser/src/readers/core/types.rs
```diff
@@ -595,6 +627,56 @@ pub enum HeapType {
...
+                HeapType::None,
+                HeapType::I31 | HeapType::Eq | HeapType::Any | HeapType::Array | HeapType::Struct,
+            ) => Ok(true),
+            (HeapType::NoExtern, HeapType::Extern) => Ok(true),
+            (HeapType::NoFunc, HeapType::Func) => Ok(true),
+            (HeapType::NoFunc, HeapType::Indexed(a)) => match (type_at(*a)?).structural_type {
+                StructuralType::Func(_) => Ok(true),
+                _ => Ok(false),
+            },
+            (a, b) => Ok(a == b),
```
alexcrichton, line 675: I've mentioned this before, but personally I'm not a fan of catch-all clauses like this. This is very easy to forget to update if a new variant is added because of `#[derive(PartialEq)]` which can lead to further bugs in behavior. Where possible I'd recommend exhaustively matching remaining cases, for example something like:


```
(a @ (HeapType::T1 | HeapType::T2), b) => Ok(a == b),
```

which explicitly documents the list of "catch-all" types being caught here and it's obvious there's no payload so equality can only be identity-based. That way if new types are added in the future this'll generate a compiler error with a non-exhaustive check.

crates/wasmparser/src/readers/core/types.rs
```diff
@@ -662,6 +744,29 @@ pub enum Type {
...
+    /// The structural type of the subtype.
+    pub structural_type: StructuralType,
+}
+
+impl Matches for SubType {
+    fn matches<'a, F>(&self, other: &Self, type_at: &F) -> Result<bool>
+    where
+        F: Fn(u32) -> Result<&'a SubType>,
+    {
+        Ok(!other.is_final
```
alexcrichton, line 763: I'm not familiar with the GC proposal myself, but a lot of these boolean checks feel iffy to me. For example if the other type is `final`, shouldn't an exact structural match be performed? This seems like it rejects all matching entirely at that point. That seems to me like two types which have the same internal indices may just happen to pass some equality test prior to this but it seems like `(type $t1 (func))` and `(type $t2 (func))` should return "true" for a matches predicate if they're both final. 

I could very well be missing something about the gc proposal, though, since I am not up to speed on it.

crates/wasmparser/src/readers/core/types.rs
```diff
@@ -691,6 +796,20 @@ pub struct StructType {
...
+impl Matches for StructuralType {
+    fn matches<'a, F>(&self, other: &Self, type_at: &F) -> Result<bool>
+    where
+        F: Fn(u32) -> Result<&'a SubType>,
+    {
+        match (self, other) {
+            (StructuralType::Func(a), StructuralType::Func(b)) => a.matches(b, type_at),
+            (StructuralType::Array(a), StructuralType::Array(b)) => a.matches(b, type_at),
+            (StructuralType::Struct(a), StructuralType::Struct(b)) => a.matches(b, type_at),
+            _ => Ok(false),
```
alexcrichton, line 808: Like above I'd recommend removing catch-all cases like this, and instead here you can do something like:

```
(StructuralType::Func(a), StructuralType::Func(b)) => a.matches(b, type_at),
(StructuralType::Func(_), _) => false,
```

crates/wasmparser/src/readers/core/types.rs
```diff
@@ -762,6 +881,75 @@ impl FuncType {
...
 }
 
+impl Matches for FuncType {
+    fn matches<'a, F>(&self, other: &Self, type_at: &F) -> Result<bool>
+    where
+        F: Fn(u32) -> Result<&'a SubType>,
+    {
+        let r = self.params().len() == other.params().len()
+            && self.results().len() == other.results().len()
+            // Note: matching functions are contravariant in their parameter types.
```
alexcrichton, line 891: Could you expand this comment? For example this switches the order of the `matches` predicate which seems important to call out here, along with its implications.

crates/wasmparser/src/readers/core/types.rs
```diff
@@ -762,6 +881,75 @@ impl FuncType {
...
+}
+
+impl Matches for StorageType {
+    fn matches<'a, F>(&self, other: &Self, type_at: &F) -> Result<bool>
+    where
+        F: Fn(u32) -> Result<&'a SubType>,
+    {
+        match (self, other) {
+            (Self::Val(a), Self::Val(b)) => a.matches(b, type_at),
+            (a, b) => Ok(*a == *b),
```
alexcrichton, line 932: Another catch-all I think would be good to remove

crates/wasmparser/src/readers/core/types.rs
```diff
@@ -840,15 +1028,65 @@ pub struct TagType {
...
+                    is_final: true,
+                    supertype_idx: idxs.first().copied(),
+                    structural_type: read_structural_type(reader.read_u8()?, reader)?,
+                }
+            }
+            opcode => SubType {
+                is_final: false,
+                supertype_idx: None,
+                structural_type: read_structural_type(opcode, reader)?,
+            },
```
alexcrichton, line 1089: There's a fair amount of duplication in this method, but could that perhaps be removed with something like:

```rust
let is_final = match reader.peek()? {
    0x50 | 0x4e => { /* advance reader */ }
    _ => None,
};
let supertype_idx = if is_final.is_some() {
    // read the supertype
} else {
    None 
};
let structural_type = /* preexisting structural type reading */;

SubType { is_final, supertype_idx, structural_type }
```

crates/wasm-mutate/src/info.rs
```diff
@@ -93,7 +93,7 @@ impl<'a> ModuleInfo<'a> {
 
                     // Save function types
                     for ty in reader {
-                        let typeinfo = TypeInfo::try_from(ty?).unwrap();
+                        let typeinfo = TypeInfo::try_from(ty?.structural_type).unwrap();
```
alexcrichton, line 96: I think it would be best to update `wasm-mutate` a bit differently in this commit. Currently `wasm-mutate` doesn't have support for GC, and that's ok, but if given a binary with gc with these changes it runs a risk of "accepting" the binary and making it invalid. For example this is ignoring all fields other than `structural_type`. 

I think it would be better to keep wasm-mutate as-is where it explicitly rejects gc-using binaries. For wasm-mutate that would look like handling the full `SubType` structure but returning errors if any fields are set that look like gc.

crates/wast/src/names.rs
imikushin: Reverting. Will start a new PR with a proper fix.

crates/wasmparser/src/readers/core/types.rs
imikushin: > One thing I might recommend is to remove the Result here since indices at this point should always be valid because the definitions of types are validated to all be correct.

Actually, the primary use case for `matches()` is to validate if the type definitions are correct.

crates/wasmparser/src/readers/core/types.rs
imikushin: Sure thing!

And I thought people didn't like parentheses. Lisp is the best.

crates/wasmparser/src/readers/core/types.rs
alexcrichton: I don't think that's the case? Whenever a type definition is reached as part of validation there are no type-matching queries performed at that time and instead the structure of the defined type is validated. This means that the indices are all valid for the type section after it's read, so in the future when the matching predicate comes into play during the code section for example there are no invalid indices.

I believe this is `Module::add_type` which is where indices are validated and afterwards indices don't need validation.

imikushin: > I believe this is Module::add_type which is where indices are validated and afterwards indices don't need validation.

Yeah, that's exactly where `matches()` gets called:

```
<wasmparser::readers::core::types::SubType as wasmparser::readers::core::types::Matches>::matches types.rs:763
wasmparser::validator::core::Module::check_subtype core.rs:525
wasmparser::validator::core::Module::add_type core.rs:507
```

crates/wasmparser/src/readers/core/types.rs
imikushin: > because of `#[derive(PartialEq)]`

Not sure I understand what bad things can happen because of it. Everything else seems clear.

crates/wasmparser/src/readers/core/types.rs
imikushin: Would it be more clear if we rename `matches` to `derives`?

We cannot derive a subtype from a type that is final. I'm also kinda worried that `matches` has broader meaning, and is potentially a bad naming decision.

crates/wasmparser/src/readers/core/types.rs
imikushin: Adding a reference to the GC spec.

crates/wasmparser/src/validator/core.rs
imikushin: Added.

crates/wasmparser/src/readers/core/types.rs
imikushin: Removed.

crates/wasmparser/src/readers/core/types.rs
imikushin: Removed.

crates/wasmparser/src/readers/core/types.rs
imikushin: Renamed.

crates/wasmparser/src/readers/core/types.rs
imikushin: Fixed.

crates/wasm-mutate/src/info.rs
imikushin: Fixed.

crates/wasmparser/src/readers/core/types.rs
imikushin: Also, updated the code according to your comment.

crates/wasmparser/src/readers/core/types.rs
alexcrichton: Ah ok I see that now, but I still think that this `Result` isn't necessary. For example if `check_structural_type` is moved above the subtype check, then everything is guaranteed to be valid while performing a subtype check?

imikushin: Thanks! Updated.

imikushin:
@alexcrichton I believe I've addressed all your comments (so far).

crates/wasmparser/src/readers/core/types.rs
imikushin: Renamed to `inherits` which should reflect the meaning even better (than `derives`).

alexcrichton (approved):

PR #1060, opened 2023-06-10
Adds registry metadata

silesmo:
Adds support for registry metadata as discussed: [https://github.com/bytecodealliance/registry/discussions/116](https://github.com/bytecodealliance/registry/discussions/116)


alexcrichton:
Thanks for this! I'd second all of what @calvinrp said as well. For organization it'd be best to put the new executable under `src/bin/wasm-tools/*.rs`, ideally in the `metadata.rs` file by augmenting that existing comment. This ideally would prevent duplicating the loops to read/write from the existing `wasm-metadata` crate as well. (possibly merging the crates together too)

silesmo:
> Thanks for this! I'd second all of what @calvinrp said as well. For organization it'd be best to put the new executable under `src/bin/wasm-tools/*.rs`, ideally in the `metadata.rs` file by augmenting that existing comment. This ideally would prevent duplicating the loops to read/write from the existing `wasm-metadata` crate as well. (possibly merging the crates together too)

Yes I agree with this. I was hesitant to do this initially since I wasn't sure we even wanted this in the wasm-tools repo. But if we are happy with that when I will make those changes!

silesmo:
> Couple suggestions / questions:
> 
>     * add the command so that it can be used from `wasm-tools` CLI, e.g. `wasm-tools registry-metadata print` (not sure about the naming, maybe `package show` and `package set`?);
> 
>     * there is a `metadata` command already, should we consider putting it under that command? probably not but asking the question
> 
>     * currently need to call with args for the input path, e.g. `registry-metadata-cli -p some.wasm print`; can we also do `registry-metadata-cli print some.wasm`?
> 
>     * add an option to provide a JSON file input or STDIN? (probably minifying the JSON before putting into the custom section)
> 
>     * maybe add test / example cases with `links` being exercised?

Thanks for the suggestions. I will go ahead and make the suggested changes.

alexcrichton:
Oh to be clear I'm not sure if this is the best place for this to live either. I can provide feedback on how to integrate this once it's decided to integrate here, but I'd leave that up to other folks to make the judgement call of whether this belongs here.

silesmo:
I will bring it up in the wasm-registry sig meeting later today and we will see after that.

Kylebrown9:
We discussed this topic yesterday and reached a consensus around putting it in the `wasm-tools` repo so that it can be read from and written to using wasm-encoder/decoder more easily.

One way to think about including this is that it's a not-completely-standard custom section and as long as we call that out, I think it's fine to have support for it in `wasm-tools`. This isn't that different from encoding/decoding DWARF into custom sections for debugging tools. If needed, we can even hide it behind a feature flag to make it only visible/usable by people who want it.

silesmo:
This should be ready to be reviewed now.
@alexcrichton or @Kylebrown9 if you have the time to take a look

crates/wasm-metadata/src/lib.rs
```diff
@@ -291,6 +313,19 @@ fn rewrite_wasm(
                 names.section()?.as_custom().append_to(&mut output);
             }
 
+            CustomSection(c) if c.name() == "registry-metadata" && stack.len() == 0 => {
+                // Pass section through if a new registry metadata isn't provided, otherwise ignore and overwrite with new
```
alexcrichton, line 317: Should this perhaps be an error if metadata is already in the file and metadata is also being added?

alexcrichton (approved):
Seems reasonable to me! One minor question about what to do in the appending case, but otherwise seems reasonable to me to add

crates/wasm-metadata/src/lib.rs
silesmo: I don't know to be honest. It depends on if we want the ability to overwrite it or not I guess? I'm happy to change it to an error if we don't want the ability to overwrite registry-metadata.

I originally had it implemented where you could add new registry metadata by merging it with the existing metadata but it became a bit weird if you changed the license expression, for example if you had a custom license that is referenced in the expression. Then you change the license expression to not include the custom license, then you would also need to remove the old license. What would you do in that case, just set it to an empty array in the input JSON if using the cli (not the best user experience)? Or would we remove it automatically and silently handle it for the user (not what I would want or expect)?

calvinrp: After discussing with @silesmo, I think overwrite makes sense for `registry-metadata`. Feels like the expected behavior for this.

silesmo: Are you happy for me to resolve this comment @alexcrichton ?

crates/wasm-metadata/src/lib.rs
```diff
@@ -672,8 +747,344 @@ fn indirect_name_map(
...
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+pub enum LinkType {
+    Documentation,
+    Homepage,
+    Repository,
+    Funding,
+    #[serde(untagged)]
```
peterhuene, line 1018: Is this a valid serde field attribute? Doesn't it belong on the enum declaration?

peterhuene: Oddly enough, my build was failing locally because of this, but no longer does with a clean build. Perhaps something changed in serde to support it on the variant?

alexcrichton: Oh if you're like me and your `Cargo.lock` was out of date (probably by a year or so for me) it was fixed by a `cargo update`, so I think we may need to update the `serde` dependency version requirement to handle this

peterhuene: Yep, looks like a [change to serde](https://github.com/serde-rs/serde/pull/2403) that the documentation hasn't caught up with that released in a non-semver-compatible way (hence why my build was broken, but a newer patch version of `serde` fixed it).

PR #1061, opened 2023-06-10
Ignore files in deps dir when resolving wit packages

cardoso:
Some operating systems will add files randomly to directories.

Considering files inside `deps` currently serve no purpose and will cause an error on `UnresolvedPackage::parse_dir`, I think ignoring any files inside is the best option for user experience.

If that changes, it would be easy to tweak this filter to allow certain extensions.


crates/wit-parser/src/resolve.rs
```diff
@@ -112,6 +112,12 @@ impl Resolve {
             for dep in path.read_dir().context("failed to read directory")? {
                 let dep = dep.context("failed to read directory iterator")?;
                 let path = dep.path();
+
+                // Files in deps dir are ignored
+                if path.is_file() {
+                    continue;
+                };
```
alexcrichton, line 119: This trailing semicolon can be dropped.

crates/wit-parser/src/resolve.rs
```diff
@@ -112,6 +112,12 @@ impl Resolve {
             for dep in path.read_dir().context("failed to read directory")? {
                 let dep = dep.context("failed to read directory iterator")?;
                 let path = dep.path();
+
+                // Files in deps dir are ignored
```
alexcrichton, line 116: Could you expand this comment to explain why these files are being ignored?

.gitignore
```diff
@@ -6,3 +6,6 @@ publish
 *.wasm
 *.wat
 test.config
+
+# Generated by macOS
+.DS_Store
```
alexcrichton, line 11: FWIW I think this is best stored in your global `~/.gitignore` file rather than having all projects you work on add this entry

alexcrichton (commented):
Thanks! Could you add a test for this as well?

cardoso:
Thanks for the review @alexcrichton . I made the changes and included a test.

alexcrichton (approved):

PR #1066, opened 2023-06-13
Use parse_list to parse include names

fibonacci1729:
Was reading the code and noticed we could use `parse_list` to parse the list of include names.


fibonacci1729:
cc @alexcrichton @Mossaka

alexcrichton (approved):
Nice! 👍

PR #1073, opened 2023-06-15
Adjust whitespace during WIT printing

alexcrichton:
The goal is to put a blank line between functions and types but this was manually done previously which involved meticulously keeping track of when items were printed and how many were remaining. Instead this replaces this logic with a simpler "I'm about to start a new line" request which bumps a counter for how many items have been printed. This will help simplify whitespace management a bit with the upcoming resources implementation.

### DIFF
--- a/crates/wit-component/src/printing.rs
+++ b/crates/wit-component/src/printing.rs
@@ -10,7 +10,7 @@ pub struct WitPrinter {
 
     // Count of how many items in this current block have been printed to print
     // a blank line between each item, but not the first item.
-    items: usize,
+    any_items: bool,
 }
 
 impl WitPrinter {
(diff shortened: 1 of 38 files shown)


crates/wit-component/src/printing.rs
```diff
@@ -1,11 +1,16 @@
...
 use wit_parser::*;
 
 /// A utility for printing WebAssembly interface definitions to a string.
 #[derive(Default)]
 pub struct WitPrinter {
     output: Output,
+
+    // Count of how many items in this current block have been printed to print
+    // a blank line between each item, but not the first item.
+    items: usize,
```
sunfishcode, line 13: Would this be simpler as a `any_items: bool`, which just records whether we've seen zero or more-than-zero? When I saw this was a `usize`, I scanned through the code to see if it was ever decremented or compared with a number other than zero.

sunfishcode (commented):
Looks nice! It's fun seeing all those spurious blank lines in the tests go away :-).

crates/wit-component/src/printing.rs
alexcrichton: Ah yes that's much clearer!

sunfishcode (approved):

PR #1076, opened 2023-06-17
Remove keywords from wit-component that are not in use anymore

omersadika:
In the last few changes a few keywords has been removed from WIT and also from crates/wit-parser, but it still exists in crates/wit-component, causing a wrong print of WIT files when using these keywords. For instance, `self` will be printed as `%self`.


alexcrichton (approved):
Thanks!

PR #1077, opened 2023-06-20
Fix a panic validating resources

alexcrichton:
A remapping lookup was too eagerly assuming that an ID was listed in the remapping when it didn't need to be. This fixes the panic by handling the case where it's not remapped. Additionally a test is added showcasing the problem before.

### DIFF
--- a/tests/local/component-model/resources.wast
+++ b/tests/local/component-model/resources.wast
@@ -1112,3 +1112,18 @@
     ))
   )
   "resource types are not the same")
+
+(component
+  (type (export "x") (component
+    (type $t' (instance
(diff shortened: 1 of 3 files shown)


sunfishcode (approved):

tests/local/component-model/resources.wast
```diff
@@ -1112,3 +1112,18 @@
     ))
   )
   "resource types are not the same")
+
+(component
```
rylev, line 1116: Just a thought - we have git history to tell when these tests were added, but it would probably be helpful to document what these are specifically testing for.

PR #1080, opened 2023-06-20
Require resources use `shared<..>` syntax

alexcrichton:
This is likely to change in the near future since the bare name of a resource will be synonymous with `own<T>`, but for the time being while `shared<T>` is implemented this explicitly disallows using `foo: resource-name`, for example.


alexcrichton:
No longer needed after https://github.com/bytecodealliance/wasm-tools/pull/1082

PR #1081, opened 2023-06-20
Reject worlds with interfaces that can access both imports and exports

alexcrichton:
This commit is a fix for the issue described in
WebAssembly/component-model#208. The process of elaborating a world's exports is no longer as straightforward as imports and additionally is no longer infallible. Instead metadata is tracked to ensure that transitively added imports are always used as imports and never both as imports and exports.


crates/wit-parser/tests/ui/parse-fail/import-and-export5.wit.result
```diff
@@ -0,0 +1,5 @@
+interface transitively depends on both an imported and exported copy of the same interface
+     --> tests/ui/parse-fail/import-and-export5.wit:13:10
+      |
+   13 |   export anon: interface {
+      |          ^---
```
sunfishcode, line 5: At first glance, this error message seems confusing, because the import-and-export5.wit file doesn't contain the `import` keyword, so the user may not know what this is referring to.

alexcrichton: Yes I definitely agree, and that's the subject of [this FIXME](https://github.com/bytecodealliance/wasm-tools/pull/1081/files#diff-fc97c873ffa8b5cf64dd5efcc94b11744589a338eb589f3975bf0f290326b79cR1239-R1255) which I'm hoping to leave for later as an improvement. These sorts of error messages in my experience are possible but pretty difficult to write up and implement so I'd ideally like to make sure it's motivated first as opposed to "only I'll see this for years as I'm writing tests"

sunfishcode: Ah, I hadn't seen the code yet :-). This makes sense. Perhaps a little different wording, perhaps "interface transitively depends on an interface in incompatible ways" or so, which would be a little less informative, but less likely to give people a misleading impression?

alexcrichton: Sounds good to me 👍

sunfishcode (approved):

PR #1082, opened 2023-06-20
Own, borrow handles

silesmo:
Removes `shared` handle in favor for `own` and `borrow` and adds implicit conversion of a resource to an owned resource handle, if it's a handle type is not explicitly defined.

@alexcrichton @sunfishcode


alexcrichton (approved):
Looks great to me, thanks!

PR #1083, opened 2023-06-20
Fix WIT sugar for resource names being `own<T>`

alexcrichton:
This commit fixes an issue where WIT resources should not have to be
wrapped in `own<T>` but instead are allowed to be defined as just `T`.
This is a bit tricky when `T` is defined in a foreign package since the
"inject the `own` wrapper" isn't apparent during parsing but only later
once packages are merged together. This phase has been updated
accordingly to inject `own` wrappers as necessary.

> Note: this builds on https://github.com/bytecodealliance/wasm-tools/pull/1079 as the first commit

### DIFF
--- a/crates/wit-parser/src/resolve.rs
+++ b/crates/wit-parser/src/resolve.rs
@@ -590,6 +590,17 @@ pub struct Remap {
     pub interfaces: Vec<InterfaceId>,
     pub worlds: Vec<WorldId>,
     pub packages: Vec<PackageId>,
+
+    /// A cache of anonymous `own<T>` handles for resource types.
+    ///
+    /// The appending operation of `Remap` is the one responsible for
+    /// translating references to `T` where `T` is a resource into `own<T>`
+    /// instead. This map is used to deduplicate the `own<T>` types generated
+    /// to generate as few as possible.
+    ///
+    /// The key of this map is the resource id `T` in the new resolve, and
+    /// the value is the `own<T>` type pointing to `T`.
+    own_handles: HashMap<TypeId, TypeId>,
 }
 
 impl Remap {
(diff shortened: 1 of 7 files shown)


crates/wit-parser/src/resolve.rs
```diff
@@ -744,6 +744,20 @@ impl Remap {
...
 
+        for (id, span) in unresolved.required_resource_types.iter() {
+            let mut id = self.types[id.index()];
+            loop {
+                match resolve.types[id].kind {
+                    TypeDefKind::Type(Type::Id(i)) => id = i,
+                    TypeDefKind::Resource => break,
+                    _ => bail!(Error {
+                        span: *span,
+                        msg: format!("type is not a resource"),
```
sunfishcode, line 755: Could you make this error message a little more explanatory, like "type used in a handle must be a resource" or so?

sunfishcode (approved):

crates/wit-parser/src/resolve.rs
sunfishcode: Ah, this comment is for #1079.

alexcrichton:
Fuzzing ended up showing a number of other issues with this commit, so I've addressed those now in a follow-up commit in the middle here. I can't add most of the fuzz test cases here yet since they require more support in `wit-component`, but they should all be included in a subsequent PR which fills out `wit-component`. 

@sunfishcode could you double-check the new commits here though? They're new-enough I don't want to carry over the review of the first commit to them too

silesmo:
This looks good! I like how you changed it to handle resolving the own handles in the same way regardless of how it's used, removing the need to do it in the `Name` resolution as well as in other places.

alexcrichton:
@sunfishcode to confirm, would you be ok if I carried over https://github.com/bytecodealliance/wasm-tools/pull/1084#pullrequestreview-1515437865 to here?

sunfishcode:
Yes, that's good.

PR #1084, opened 2023-06-21
Add support for resources to `wit-component`

alexcrichton:
This commit fully integrates resource types in the component model with
the `wit-component` crate, implementing features such as:

* A WIT package using `resource` types can now be round-tripped through
  its WebAssembly encoding. This enables use cases such as `wit-bindgen`
  which embed type information in custom sections of core wasm binaries
  produced by native compilers.

* Core modules can be lifted into components and the component can use
  resources. This provides a target for `wit-bindgen` and other code
  generators to use when generating guest code. Resource intrinsics and
  destructors are all available to the core wasm as desired.

* WIT can be inferred from components using resources, where functions
  are represented as `resource`-related functions in WIT.

* The `roundtrip-wit` fuzzer is extended with resources support meaning
  all of the above support will be fuzzed on OSS-Fuzz.

This required a number of refactorings in `wit-component` especially
around how type information was handled. Previous processing was a bit
fast-and-loose because where exactly a type was defined didn't really
matter since everything was nominal. With resource types, however,
definition locations are significant and this required some fixes to
previous processing. One example of this is that
https://github.com/WebAssembly/component-model/issues/208 was discovered through this work and the
fixes required were implemented previously and further handled here in
`wit-component`.

Overall this PR has been almost exclusively fuzz-driven in its
development. I started out with the bare bones of getting simple
components working with resources being imported and exported, then
added fuzzing support to `wit-smith`, then let the fuzzer go wild. Quite
a few issues were discovered which led to all of the refactorings and
processing here in this PR. I definitely won't claim that this is a
simplification at all to `wit-component` by any measure. Rather it's
taking a complicated codebase and making it more complicated. In my
mind though the "saving grace" is that I'm pretty confident in the
testing/fuzzing story here. It's relatively easy to isolate issues and
add test cases for the various things that can crop up and the fuzzer
has quite good coverage of all the various paths through
`wit-component`. All that's to say that this is surely not the "best" or
easiest to understand implementation of resources, but it's intended to
be sufficient for now.

> Note: this builds on https://github.com/bytecodealliance/wasm-tools/pull/1081 and https://github.com/bytecodealliance/wasm-tools/pull/1083 in the first two commits


alexcrichton:
cc @silesmo as you're likely interested in this as well

crates/wit-smith/src/lib.rs
```diff
@@ -24,7 +24,14 @@ pub fn smith(config: &Config, u: &mut Unstructured<'_>) -> Result<Vec<u8>> {
         let unresolved = pkg.sources.parse().unwrap();
         let id = match resolve.push(unresolved) {
             Ok(id) => id,
-            Err(e) => panic!("bad wit parse: {e:?}"),
+            Err(e) => {
+                if e.to_string()
+                    .contains("both an imported and exported copy of the same interface")
```
sunfishcode, line 29: This error message was [changed](https://github.com/bytecodealliance/wasm-tools/pull/1081/commits/1f136b68880da181d5e2f626d592b684d5264a7b) to "an interface in incompatible ways".

crates/wit-component/src/encoding.rs
```diff
@@ -886,6 +1039,43 @@ impl<'a> EncodingState<'a> {
...
+
+        fn import_func_name(f: &Function) -> String {
+            match f.kind {
+                FunctionKind::Freestanding => {
+                    format!("import-func-{}", f.name)
+                }
+
+                // transform `[method]foo.bar` into `import-method-foo-bar` to
+                // have it be a valid kebab-name which can't conflict with
+                // anything else.
```
sunfishcode, line 1064: I'm probably forgetting something here, but why can't `import-method-foo-bar` conflict with user identifiers?

alexcrichton: The imports into this sub-component are all synthesized by wit-component, and freestanding functions are "mangled" with `import-func-` as a prefix where methods are "mangled" with `import-method-` as a prefix, which is intended to remove the possibility of conflict.

sunfishcode (approved):

PR #1085, opened 2023-06-21
fix(wit-parser): fix a few bugs that found by fuzzing

Mossaka:
This PR fixes two bugs:
1. When there are multiple worlds in a WITpackage, but not every one of them are included in a world. This bug emerges from an use of out-of-bound index to get included world's span.
2. The fully resolved world does not check naming conflict among its imported types, functions, exported functions with import/export interfaces.


alexcrichton (approved):
👍

PR #1086, opened 2023-06-23
Do type remapping through component instances

rylev:
Paired with @alexcrichton on this. This fixes an issue where type substitutions where not happening through instances and thus when the validator was ensuring that all types of an export were themselves exported, types exported through instances were not in the set of exported types. This changes registers types that are exported through instances as well.


alexcrichton (approved):
Thanks!

PR #1088, opened 2023-06-27
(wit-parser): throw error if include renaming used on IDs

Mossaka:
this Commits add extra logic to `resolve_include` function to check if there are un-matched names in the `with` block or if the names are referring to an ID.


alexcrichton (approved):
Thanks!

crates/wit-parser/src/resolve.rs
```diff
@@ -1318,6 +1318,21 @@ impl Remap {
...
+        // remove all imports and exports that match the names we're including
+        for import in include_world.imports.iter() {
+            self.remove_matching_name(import, &mut names_);
+        }
+        for export in include_world.exports.iter() {
+            self.remove_matching_name(export, &mut names_);
+        }
+        if !names_.is_empty() {
+            bail!(Error {
+                msg: format!("no import or export kebab-name `{}`. Note that an ID does not support renaming", names_[0].name),
```
rylev, line 1332: Just a note that there might be some confusion about kebab-name and ID. Would saying "interface id" be clearer?

There is potentially a larger conversation about naming here as well: whether "kebab-name" is an appropriate label considering some ids also use kebab case.

Mossaka: The component model repo is probably a better place for this question. I used the terminology from the WIT description

> In the component model imports to a component either use an ID or a kebab-name, and in WIT this is reflected in the syntax:

https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#wit-worlds

PR #1089, opened 2023-06-27
Implement targets check for worlds

fibonacci1729:
**WIP**

Mostly opening this up for a preliminary review to get some feedback on the approach taken here. I'm not entirely satisfied with the errors returned when the `targets` check fails which will require a deeper dive into `wasmparser` to see what can be improved. Specifically, validation fails fast when conformance fails however i think ideally we'd want to precisely describe the set of differences between what's expected and what was provided as input.


crates/wit-component/src/builder.rs
```diff
@@ -143,6 +143,7 @@ impl ComponentBuilder {
             ComponentTypeRef::Instance(_) => inc(&mut self.instances),
             ComponentTypeRef::Func(_) => inc(&mut self.funcs),
             ComponentTypeRef::Type(..) => inc(&mut self.types),
+            ComponentTypeRef::Component(_) => inc(&mut self.components),
```
fibonacci1729, line 146: I'm not sure of any broader implications of introducing this case. So leaving a note to draw attention to it!

alexcrichton: Makes sense to me 👍, the builder right now isn't really general-purpose in the sense that it's "just enough" to get `wit-component` working, but definitely no issues in filling out its support of course.

src/bin/wasm-tools/component.rs
```diff
@@ -366,6 +369,66 @@ impl WitOpts {
...
+    ///
+    /// The file specified can be a `*.wit` file parsed as a single-document
+    /// package. It can be a directory to be parsed as a WIT package. It can be
+    /// a `*.wat` or `*.wasm` file for either the binary representation of a WIT
+    /// package or a component itself to extract the interface from. The type of
+    /// input is inferred from the contents of the path specified.
+    ///
+    /// If not provided or if this is `-` then stdin is read entirely and
+    /// processed.
+    input: Option<PathBuf>,
```
alexcrichton, line 394: Since the input here is wasm then this can be `io: wasm_tools::InputOutput` which covers the text format, stdin, etc. (there's some examples of this throughout the other subcommands, the `input` here from`WitOpts` above is a bit special because the input to that command isn't necessarily wasm unlike many others)

alexcrichton (commented):
Looking good, thanks!

If possible could some tests be added? This may be possible to test through `tests/cli/*` but the format there isn't super-well baked so it may not easily allow for both a wasm file as input and a WIT file as input which could make that hard. Otherwise having dedicated tests for this in `crates/wit-component/tests/*` I think is fine too

src/bin/wasm-tools/component.rs
fibonacci1729: I had considered this; I avoided using `InputOutput` because it introduced an `-o/--output` option which doesn't seem to make much sense in this case. Nevertheless, happy to use the `InputOutput` type if that's the convention for input-only as well!

alexcrichton: Oh true, forgot about that! If you're up for it it might be good to split `InputOutput` into an `Input` and an `Output` (but still have the fused `InputOutput`). Mostly just trying to keep the logic for parsing input consolidated to ensure consistent behavior across subcommands.

fibonacci1729: Perfect, totally agree! I started down this path already should have an update soon. Thanks!

fibonacci1729: Addressed this in #1099

fibonacci1729:
@alexcrichton Updated with a test to cover a failure scenario and a success test. Happy to add more tests but before I go too deep on this approach could you take a look and let me know if this approach is OK or if there is a better way to approach. Thanks!

alexcrichton:
Looks good to me 👍

fibonacci1729:
@alexcrichton added a few more tests that I think rounds out the coverage; let me know if you can think of any case i missed. Otherwise, should we try to land this as is and in a follow-up PR improve the error messaging?

alexcrichton (approved):
Sounds and looks good to me, thanks again!

PR #1090, opened 2023-06-28
add to wasm-tools validate docs

michelleN:
+ Adds where to find available wasm feature options
+ Adds how to see more verbose output in case folks want to see more details of what is being validated


alexcrichton (approved):
Thanks! 

If you're feeling particularly intrepid the [error message here](https://github.com/bytecodealliance/wasm-tools/blob/ea2fe06e332833d3714b455adc03a3d700641705/src/bin/wasm-tools/validate.rs#L145) could theoretically be improved with something like "didn't recognize `foo` but valid features are `bar`, `baz`, ...". Or something like `--wasm-features help` would print them all out or something like that. In any case no need to happen here-and-now, just if you're interested!

PR #1091, opened 2023-06-28
update validate feature error message

michelleN:
+ Adds list of valid features to error message

### DIFF
--- a/src/bin/wasm-tools/validate.rs
+++ b/src/bin/wasm-tools/validate.rs
@@ -139,10 +139,17 @@ fn parse_features(arg: &str) -> Result<WasmFeatures> {
             }
 
             name => {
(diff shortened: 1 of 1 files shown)


alexcrichton (approved):
Thanks!

I think it might be good to add a test as well for this ideally to see how it changes over time, and I think that can be done by adding `tests/cli/bad-feature.wat` which would contain something like:

```wasm
;; RUN: validate --features bad-name

(module)
```

and then `BLESS=1 cargo test --test cli` should create `tests/cli/bad-feature.wat.stderr`.

Although ok actually after I wrote this up I checked `tests/cli.rs` and it asserts that all command executions succeed, so scratch all that adding tests for this would require deeper refactoring, so no need to happen now.

PR #1094, opened 2023-07-01
Update accessors on `TypesRef`

alexcrichton:
Previously accessors of type information would return the actual information itself, and this commit updates them to instead return the `TypeId`. A `TypeId` can be turned into the concrete type information and otherwise upcoming resources support in Wasmtime needs more aggressive use of `TypeId` so I found it easier to have those returned from these methods since it's not possible to go from `&FuncType` to `TypeId`.

Additionally this commit adds accessors for the size of all the index spaces in the component model and core wasm binary format. This enables witnessing intermediate state while validating a component for example which is also used in the upcoming support for resources in Wasmtime.

### DIFF
--- a/crates/wasmparser/src/validator.rs
+++ b/crates/wasmparser/src/validator.rs
@@ -1484,7 +1484,8 @@ mod tests {
             })
         );
 
-        match types.function_at(0) {
+        let id = types.function_at(0).unwrap();
+        match types.type_from_id(id).unwrap().as_func_type() {
             Some(ty) => {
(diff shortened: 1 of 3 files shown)


alexcrichton:
Nah yeah I definitely agree, and this is something I've wanted to fix for awhile. I'd like to add accessors along the lines of:

```
types[id].unwrap_func_type()
```

since most of the time it's `unwrap()`'d anyway so the indexing communicates panicking along with the `unwrap_` in the name. That should tighten up a lot of the code here I think, but I'll leave that for a future PR.

PR #1096, opened 2023-07-04
feat(wasm-compose): --definitions configuration for components with exports defining imports

guybedford:
This adds a new `definitions` / `--definitions` / `-d` configuration to `wasm-tools compose`, taking as input a list of component binaries and then treating the exports of those components as the defined dependencies for composition.

For example, currently the only way to do WASI virtualization is with a configuration file, eg containing:

```
dependencies:
  'wasi:cli-base/environment': wasi_virt.wasm
```

with this PR, the above can be handled fully with the simpler configuration:

```
definitions: ['wasi_virt.wasm']
```

or via the CLI:

```
wasm-compose component.wasm -d wasi_virt.wasm -o output.wasm
```

While also supporting multi import virtualizations.

If any exports override an existing named dependency, the normal unique dependency naming logic will apply, so it seems to fit the model pretty well while offering a more convenient composition workflow for virtualization.


crates/wasm-compose/src/composer.rs
```diff
@@ -56,6 +56,17 @@ impl<'a> CompositionGraphBuilder<'a> {
...
+        // If definition components are provided, parse them, populating
+        // them into dependencies
+        for def in &config.definitions {
+            let component = Component::from_file("", root_path.parent().unwrap().join(def))?;
+            for (expt_name, _) in &component.exports {
+                let mut dep_component = component.clone();
+                dep_component.name = expt_name.to_string();
+                graph.add_component(dep_component)?;
+            }
+        }
```
peterhuene, line 68: I don't think this is the right way to do this as you're duplicating the component in the graph unnecessarily.

Additionally, this would result in multiple defined components in the composition if the root imports more than one instance from the same definition component. I confirmed this by having the test `root.wat` import more than one interface exported by `defs.wat` and it results in having `defs.wat` defined twice in the composed component.

Let me noodle on the best way to do this.

guybedford: I suppose the thing that works here that surprised me was that we could define:

```
dependencies:
  'wasi:cli-base/environment': wasi_virt.wasm
```

and have that correctly wire up `wasi:cli-base/environment` both as the dependency name AND the export name. Which seems to effectively be a sort of flattening.

Strictly speaking I would have expected something like:

```
dependencies:
  'wasi:cli-base/environment': 'wasi_virt.wasm#wasi:cli-base/environment'
```

with `#` meaning "export name", as the flattening that is actually happening here.

guybedford (commented):
I've tested this locally against full virtualization and can confirm it's working as expected. Thanks for carrying this one over the line!

guybedford:
With the CLI, I just attempted a pathing like the following:

```
wasm-tools compose ../dir/test/output/records.component.wasm -d tests/generated/env-deny.virt.wasm -o out.wasm
```

Which gives the error:

```
error: failed to parse component `../dir/test/output/tests/generated/env-deny.virt.wasm`
```

It seems the paths are always taken relative to the base path of the input? Ideally we could maintain the cwd-relative behaviour for the `-d` flag in the CLI?

peterhuene:
I'll get that fixed and rebase this before doing a review.

peterhuene (approved):

Issue #1095, opened 2023-07-02
wast: Parsing giant wat takes too much memory resources


Xyah3PBeHB:
I have a giant wat (`385,794,195 bytes`) with `127384` functions (568 of which are imported functions) and I used `heaptrack` to analyze the memory usage when converting that wat to wasm using `wasm-tool` and `wat2wasm` (from wabt).

# wasm-tool

`heaptrack target/release/wasm-tools parse test.wat -o test1.wasm`

```
total runtime: 37.60s.
calls to allocation functions: 20932994 (556743/s)
temporary memory allocations: 8617065 (229183/s)
peak heap memory consumption: 6.20G
peak RSS (including heaptrack overhead): 6.06G
total memory leaked: 352B
```

# wat2wasm

`heaptrack wabt-1.0.33/bin/wat2wasm test.wat -o test2.wasm --enable-annotations`

```
total runtime: 31.10s.
calls to allocation functions: 23266595 (748122/s)
temporary memory allocations: 1561856 (50220/s)
peak heap memory consumption: 2.83G
peak RSS (including heaptrack overhead): 3.01G
total memory leaked: 0B
```

To my surprise the memory peak of wasm-tool was more than double that of wat2wasm. This is [wat](https://github.com/bytecodealliance/wasm-tools/files/11929997/sample.wat.gz) for testing.

alexcrichton:
Thanks for the report! I've been digging in a bit and there's quite a bit of fruit to pick in this regard, so I'll try to make some PRs in the future to reduce the overhead here.

If I may be so bold as to ask, I'm curious if this caused issues for you somewhere? Or basically was curious what led to the motivation to make a comparison?

Xyah3PBeHB:
First of all, thank you for optimizing the memory usage. I'd be happy to share the origin of this issue.
In the past I have used `wat2wasm` to analyze wasm. Recently, I plan to use `wasmparser` to automate some operations. First of all, I want to try the `wasm-tools` tool set. But in the behavior that I have been able to complete normally in the past (converting giant wat back to wasm), it ran out of memory (my device has not been upgraded for many years and only has 16G memory). So I want to compare it with the tools I used in the past.

alexcrichton:
Makes sense and thanks for the context!

alexcrichton:
[...]

PR #1103, opened 2023-07-06
Shrink the size of wast text tokens

alexcrichton:
This commit is another improvement towards addressing #1095 where the goal here is to shrink the size of `Token` and reduce the allocated memory that it retains. Currently the entire input string is tokenized and stored as a list of tokens for `Parser` to process. This means that the size of a token has a large affect on the size of this vector for large inputs.

Even before this commit tokens had been slightly optimized for size where some variants were heap-allocated with a `Box`. In profiling with DHAT, however, it appears that a large portion of peak memory was these boxes, namely for integer/float tokens which appear quite a lot in many inputs.

The changes in this commit were to:

* Shrink the size of `Token` to two words. This is done by removing all pointers from `Token` and instead only storing a `TokenKind` which is packed to 32-bits or less. Span information is still stored in a `Token`, however.

* With no more payload tokens which previously had a payload such as integers, strings, and floats are now re-parsed. They're sort of parsed once while lexing, then again when the token is interpreted later on. Some of this is fundamental where the parsing currently happens in a type-specific context but the context isn't known during lexing (e.g. if something is parsed as `u8` then that shouldn't accept `256` as input).

The hypothesis behind this is that tokens are far more often keywords, whitespace, and comments rather than integers, strings, and floats. This means that if these tokens require some extra work then that should hopefully "come out in the wash" after and this representation would otherwise allow for other speedups.

Locally the example in #1095 has a peak memory usage reduced from 5G to 4G from this commit and additionally the parsing time drops from 8.9s to 7.6s.

### DIFF
--- a/crates/wast/src/lexer.rs
+++ b/crates/wast/src/lexer.rs
@@ -43,55 +43,100 @@ pub struct Lexer<'a> {
     allow_confusing_unicode: bool,
 }
 
-/// A fragment of source lex'd from an input string.
+/// A single token parsed from a `Lexer`.
+#[derive(Debug, PartialEq)]
+pub struct Token {
+    /// The kind of token this represents, such as whether it's whitespace, a
+    /// keyword, etc.
+    pub kind: TokenKind,
+    /// The byte offset within the original source for where this token came
+    /// from.
+    pub offset: usize,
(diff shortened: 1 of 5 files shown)


fitzgen (approved):
Nice!

PR #1106, opened 2023-07-07
Reduce verbosity in wasmparser `Types` API

alexcrichton:
This additionally applies to `TypesRef` and this PR applies a few changes:

* Going from a `TypeId` to a `Type` is now done with `&types[id]` or `types.get(id)` instead of `type_from_id(id).unwrap()`.
* All index-based accessors return a `T` instead of `Option<T>`. Fallibility is possible through the `*_count()` methods for embedders to test if necessary, but otherwise almost all accesses were followed by `.unwrap()` anyway.
* Conversions from `Type` to a sub-component which were previously `as_foo_type(&self) -> Option<&Foo>` are now expressed as `unwrap_foo(&self) -> &Foo`. This renames `as` to `unwrap` to highlight the panicking behavior and additionally drops the `_type` suffix since that's implied on everything anyway.
* The `type_at` function and its `core: bool` argument are removed in favor of `core_type_at` and `component_type_at` functions.
* The `func_type_at` function was removed as it can be achieved through combining other functions.
* Some methods which returned `&T` changed to return a `TypeId` which represents `T` for consistency.

This is inspired by comments on #1094 and is something I've wanted to address for awhile and will ideally make the `Types` API a bit more ergonomic to use with fewer conversions and less `.unwrap()` boilerplate without any loss of functionality.


crates/wasmparser/src/validator/types.rs
```diff
@@ -1003,46 +999,35 @@ impl<'a> TypesRef<'a> {
...
-
     /// Gets a type based on its type id.
     ///
     /// Returns `None` if the type id is unknown.
-    pub fn type_from_id(&self, id: TypeId) -> Option<&'a Type> {
+    pub fn get(&self, id: TypeId) -> Option<&'a Type> {
         self.list.get(id.index)
     }
 
     /// Gets a type id from a type index.
```
sunfishcode, line 1009: This comment should now at least say "core" type, and mention the index is a "core type index", to distinguish it from the comment for `component_type_at`.

crates/wit-component/src/validation.rs
```diff
@@ -564,7 +564,7 @@ fn validate_imports_top_level(
     for (name, ty) in funcs {
         match resolve.worlds[world].imports.get(&world_key(resolve, name)) {
             Some(WorldItem::Function(func)) => {
-                let ty = types.func_type_at(*ty).unwrap();
+                let ty = types[types.core_type_at(*ty)].unwrap_func();
```
sunfishcode, line 567: How does this code know it has a core type index, rather than a component type index?

crates/wasmparser/src/validator/types.rs
```diff
@@ -1393,29 +1377,26 @@ impl Types {
...
-    pub fn type_at(&self, index: u32, core: bool) -> Option<&Type> {
-        self.as_ref().type_at(index, core)
+    /// # Panics
+    ///
+    /// Panics if `index` is not a valid function index.
+    pub fn core_type_at(&self, index: u32) -> TypeId {
+        self.as_ref().core_type_at(index)
     }
 
     /// Gets a defined core function type at the given type index.
```
sunfishcode, line 1393: This comment no longer describes what it's attached to.

crates/wasmparser/src/validator/types.rs
```diff
@@ -1393,29 +1377,26 @@ impl Types {
...
-    /// Gets a type id from a type index.
-    ///
-    /// Returns `None` if the type index is out of bounds.
-    pub fn id_from_type_index(&self, index: u32, core: bool) -> Option<TypeId> {
-        self.as_ref().id_from_type_index(index, core)
+    pub fn get(&self, id: TypeId) -> Option<&Type> {
+        self.as_ref().get(id)
     }
 
     /// Gets a type at the given type index.
```
sunfishcode, line 1384: This comment should mention "core" types and "core type index".

crates/wasmparser/src/validator/types.rs
```diff
@@ -1279,12 +1256,13 @@ impl<'a> TypesRef<'a> {
 
     /// Gets the type of an component instance at the given component instance index.
     ///
-    /// Returns `None` if the type index is out of bounds or the type has not
-    /// been parsed yet.
-    pub fn component_instance_at(&self, index: u32) -> Option<TypeId> {
+    /// # Panics
+    ///
+    /// This will panic if the `index` provided is out of bounds.
```
sunfishcode, line 1261: This function and several other similar functions also panic if `self` is a core module.

sunfishcode (commented):
Looks good!

crates/wit-component/src/validation.rs
alexcrichton: That's contextually known since the `funcs` map comes from parsing a core wasm module (the input to `wasm-tools component new` which is also where the `types` type information comes from.

sunfishcode (approved):

PR #1117, opened 2023-07-10
Fix "duplicate identifier for field" for subtype fields

### LINKED ISSUE #1092: wasmparser: "duplicate identifier for field" in subtype
For the following module:

```wast
(module
    (type $A (struct (field $vt (mut i32))))
    (type $B (sub $A (struct (field $vt (mut i32)))))
)
```

I get the following output:

```
$ wasm-tools parse demo.wat                                              

error: duplicate identifier for field
     --> demo.wat:3:37
      |
    3 |     (type $B (sub $A (struct (field $vt (mut i32)))))
      |                                     ^
```

For subtypes, the same field in the subtype has to have the same name as in the supertype, otherwise it's impossible to do any kind of field manipulation unless you know the exact type of the object.

In fact, the spec for the gc proposal says:

> For fields, names need only be unique within a single type.
> -- [*`document/core/text/conventions.rst`*](https://github.com/WebAssembly/gc/blob/942c77f128a689bc09dac72059d392769ee71dde/document/core/text/conventions.rst)

---

This happens both on `main` (0263c2f2e0a33fe97569535aa2784b7989cb1dd0) and version 1.0.35.

imikushin:
Field namespaces are per struct now. Inherited fields' names must be exactly the same as in the super-type.

Fixes #1092.

### DIFF
--- a/crates/wast/src/core/resolve/names.rs
+++ b/crates/wast/src/core/resolve/names.rs
@@ -3,6 +3,7 @@ use crate::core::*;
 use crate::names::{resolve_error, Namespace};
 use crate::token::{Id, Index};
 use crate::Error;
+use std::collections::HashMap;
 
 pub fn resolve<'a>(fields: &mut Vec<ModuleField<'a>>) -> Result<Resolver<'a>, Error> {
     let mut resolver = Resolver::default();
@@ -25,7 +26,7 @@ pub struct Resolver<'a> {
     tags: Namespace<'a>,
     datas: Namespace<'a>,
     elems: Namespace<'a>,
-    fields: Namespace<'a>,
+    fields: HashMap<u32, Namespace<'a>>,
     type_info: Vec<TypeInfo<'a>>,
 }
 
@@ -46,16 +47,41 @@ impl<'a> Resolver<'a> {
     }
 
     fn register_type(&mut self, ty: &Type<'a>) -> Result<(), Error> {
+        let type_index = self.types.register(ty.id, "type")?;
+
         match &ty.def {
             // For GC structure types we need to be sure to populate the
             // field namespace here as well.
             //
-            // The field namespace is global, but the resolved indices
-            // are relative to the struct they are defined in
+            // The field namespace is relative to the struct fields are defined in
             TypeDef::Struct(r#struct) => {
                 for (i, field) in r#struct.fields.iter().enumerate() {
                     if let Some(id) = field.id {
-                        self.fields.register_specific(id, i as u32, "field")?;
+                        if let Some(parent_type_index) = ty
+                            .parent
+                            .map(|index| self.types.get_index(&index))
+                            .flatten()
+                        {
+                            if let Some(parent_field_ns) = self.fields.get(&parent_type_index) {
+                                let field_idx =
+                                    parent_field_ns.resolve(&mut Index::Id(id), "field")?;
+                                if field_idx != i as u32 {
+                                    return Err(Error::new(
+                                        id.span(),
+                                        format!(
+                                            "field index mismatch for `{}`: expected {}, got {}",
+                                            id.name(),
+                                            field_idx,
+                                            i
+                                        ),
+                                    ));
+                                }
+                            }
+                        }
+                        self.fields
+                            .entry(type_index)
+                            .or_insert(Namespace::default())
+                            .register_specific(id, i as u32, "field")?;
                     }
                 }
             }
@@ -75,7 +101,6 @@ impl<'a> Resolver<'a> {
             _ => self.type_info.push(TypeInfo::Other),
         }
 
-        self.types.register(ty.id, "type")?;
         Ok(())
     }
 
@@ -610,8 +635,12 @@ impl<'a, 'b> ExprResolver<'a, 'b> {
             }
 
             StructSet(s) | StructGet(s) | StructGetS(s) | StructGetU(s) => {
-                self.resolver.resolve(&mut s.r#struct, Ns::Type)?;
-                self.resolver.fields.resolve(&mut s.field, "field")?;
+                let type_index = self.resolver.resolve(&mut s.r#struct, Ns::Type)?;
+                self.resolver
+                    .fields
+                    .get(&type_index)
+                    .unwrap()
+                    .resolve(&mut s.field, "field")?;
             }
 
             ArrayNewFixed(a) => {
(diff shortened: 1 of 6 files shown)


crates/wast/src/core/resolve/names.rs
```diff
@@ -46,16 +47,41 @@ impl<'a> Resolver<'a> {
...
-                        self.fields.register_specific(id, i as u32, "field")?;
+                        if let Some(parent_type_index) = ty
+                            .parent
+                            .map(|index| self.types.get_index(&index))
+                            .flatten()
+                        {
+                            if let Some(parent_field_ns) = self.fields.get(&parent_type_index) {
+                                let field_idx =
+                                    parent_field_ns.resolve(&mut Index::Id(id), "field")?;
+                                if field_idx != i as u32 {
```
alexcrichton, line 68: Is this strictly necessary or something that could be deferred to type-checking? Ideally the text format is the bare minimum to have developer conveniences like `$foo` but also not enough to duplicate the validation that happens on the binary format. This check here I think may be ok to remove and rely on the binary format validation to perform?

crates/wast/src/core/resolve/names.rs
```diff
@@ -610,8 +635,12 @@ impl<'a, 'b> ExprResolver<'a, 'b> {
             }
 
             StructSet(s) | StructGet(s) | StructGetS(s) | StructGetU(s) => {
-                self.resolver.resolve(&mut s.r#struct, Ns::Type)?;
-                self.resolver.fields.resolve(&mut s.field, "field")?;
+                let type_index = self.resolver.resolve(&mut s.r#struct, Ns::Type)?;
+                self.resolver
+                    .fields
+                    .get(&type_index)
+                    .unwrap()
```
alexcrichton, line 642: This should not be an `unwrap()` but instead should return some form of error message. Note that it should not always be required to be `Some`, however, because the `field` may not actually need resolution. Only if the field is an `Id` is it required that this `.get(...)` succeeds in which case an error would be emitted. For example `struct.get 1000 1000` should be ok (even if it's invalid), but `struct.get 1000 $foo` would only be valid if type `1000` is defined.

crates/wast/src/names.rs
```diff
@@ -51,11 +51,13 @@ impl<'a> Namespace<'a> {
     }
 
     pub fn register_specific(&mut self, name: Id<'a>, index: u32, desc: &str) -> Result<(), Error> {
-        if let Some(_prev) = self.names.insert(name, index) {
-            return Err(Error::new(
-                name.span(),
-                format!("duplicate identifier for {}", desc),
-            ));
+        if let Some(prev) = self.names.insert(name, index) {
+            if prev != index {
```
alexcrichton, line 55: Is this condition still necessary?

tests/local/gc/gc-subtypes-invalid.wast
```diff
@@ -129,3 +129,17 @@
   )
   "type index out of bounds"
 )
+(assert_invalid
+  (module
+    (type $A (struct (field $vt (mut i32))))
+    (type $C (sub $A (struct (field $tv (mut i32)))))
+  )
+  "unknown field"
+)
```
alexcrichton, line 138: Is this invalid according to the spec interpreter? This relates to my comment above about not checking parent types I think, and it's something I'm not sure about because this is valid from a binary format perspective so I would naively expect that this would be a valid module.

alexcrichton (commented):
Thanks, I think it would be worth adding more tests here too which exercise the implementation as well for example by passing `struct.get` instructions with varying levels of validity in the indices specified.

crates/wast/src/core/resolve/names.rs
imikushin: > Is this strictly necessary or something that could be deferred to type-checking?

I think it is: I didn't see if names were preserved in the binary format (pretty sure they are not). I definitely would want to defer it if they were.

And this check is about names only: a field with the same name should be at the same index as in the parent type. None of the binary format validation is happening here.

This seems appropriate: names are only available in text (to my knowledge), so we check them when parsing text.

crates/wast/src/names.rs
imikushin: Yep, imagine if someone writes 
```
(type (struct (field $vt (mut i32)) (field $vt (mut i64))))
```

Adding a test case.

tests/local/gc/gc-subtypes-invalid.wast
imikushin: This is perhaps a stronger check than what is explicitly prescribed by the GC spec. 

But it makes perfect sense: the spec does insist on field types being the same at the same positions as in the parent struct. So, unless we want to introduce a field renaming "feature", we don't want to allow naming inherited fields differently.

imikushin: We may actually want an even stronger check: if the parent struct field is named, the inherited field has to be named (the same) too.

crates/wast/src/core/resolve/names.rs
imikushin: Fixed.

crates/wast/src/names.rs
alexcrichton: I don't believe that motivates this condition because previously that was considered an error as well?

tests/local/gc/gc-subtypes-invalid.wast
alexcrichton: AFAIK the text format doesn't layer more restrictions over the binary format. In that sense I don't think the above would be an error. That being said I don't know, so if you're also not sure I think it would be good to check with the upstream proposal

crates/wast/src/names.rs
imikushin: You're right. The added test works too.

imikushin: Removing.

tests/local/gc/gc-subtypes-invalid.wast
imikushin: > AFAIK the text format doesn't layer more restrictions over the binary format.

Even with this PR it doesn't, because names of fields are gone in binary. The proposed restriction is on the text format. 

Having said that, you're probably right: this is way out of scope of a bug fix PR.

Removing the restriction for now.

crates/wast/src/core/resolve/names.rs
imikushin: Removed as a result of discussion below: https://github.com/bytecodealliance/wasm-tools/pull/1117#discussion_r1260321905

PR #1125, opened 2023-07-14
fix: add `include` to the world parsing error message

Mossaka:
tiny change - adding the missing `include` to an error message


guybedford (approved):

Mossaka:
Wait I was mistaken. There is a `include` in the error message already!

PR #1127, opened 2023-07-17
Update to new GC cast instruction encoding

bvisness:
Per https://github.com/WebAssembly/gc/pull/359#issuecomment-1531774122, the encoding for GC cast instructions has changed to use two separate opcodes for `br_on_cast` and `br_on_cast_fail`, rather than being disambiguated by a flag bit.

I looked around for other places to update, but it seems like parsing isn't implemented for GC instructions yet? Roundtrip tests are also crashing on the latest main, so I'm unable to properly test this. Would appreciate any guidance.


alexcrichton:
Ah yeah support for GC instructions isn't implemented in the rest of the tooling yet, and we don't otherwise run many other GC tests at this time due to the limited support in this repository.

Could you paste the errors you're seeing with the roundtrip suite? Are they related to this PR or probabaly unrelated? (given the green CI run here happy to merge, but wanted to confirm too)

bvisness:
Here's what I get when I run `cargo test` on my Mac:

```
     Running tests/roundtrip.rs (target/debug/deps/roundtrip-8e565f4fdc7362cd)
running 381 test files


thread '<unknown>' has overflowed its stack
fatal runtime error: stack overflow
error: test failed, to rerun pass `--test roundtrip`

Caused by:
  process didn't exit successfully: `/Users/mozilla/Developer/bytecodealliance/wasm-tools/target/debug/deps/roundtrip-8e565f4fdc7362cd` (signal: 6, SIGABRT: process abort signal)
```

I see this test runs fine in CI, so I have no idea what the issue might be.

alexcrichton:
Oh dear! If you're up for a bit of debugging could you try narrowing down to which test is failing? Or capture a stack trace to see what's eating so much stack? 

Otherwise though this looks good to me so I'm gonna go ahead and merge.

alexcrichton (approved):

alexcrichton:
Sure yeah, I'll post that at https://github.com/bytecodealliance/wasm-tools/pull/1128

PR #1130, opened 2023-07-17
Remove public deps from `wasm-metadata`'s API

alexcrichton:
Currently whenever `wasmparser` or `wasm-encoder` have a major version bump this requires a major revision of `wasm-metadata` to be released due to their appearance in the API of `wasm-metadata`. This in turn forces a major version bump of `wit-component` due to `wit-component`'s usage of `wasm-metadata` in its public API.

The `wasmparser` and `wasm-encoder` crates, however, are relatively "unstable" crates in that the amount of features they support for wasm are shifting quite a lot over time meaning that they have a lot of major version bumps. Coupling this rate of change to `wit-component` which is expected to be embedded in many WIT-related bindings generators I'm predicting won't be the best development story.

To decouple these crates this commit removes the `wasmparser` and `wasm-encoder` types from the API of `wasm-metadata` by making them internal implementation details. After this it's possible to have a major revision of `wasmparser` without updating the major revision of `wasm-metadata` or `wit-component` for example.

### DIFF
--- a/crates/wasm-encoder/src/component.rs
+++ b/crates/wasm-encoder/src/component.rs
@@ -20,7 +20,7 @@ pub use self::names::*;
 pub use self::start::*;
 pub use self::types::*;
 
-use crate::{CustomSection, Encode, ProducersSection};
+use crate::{CustomSection, Encode, ProducersSection, RawCustomSection};
 
 // Core sorts extended by the component model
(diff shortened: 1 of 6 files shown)


pchickey (approved):
Thanks - I hadn't considered this problem but luckily the fix is quite easy!

PR #1132, opened 2023-07-18
Fix a few typos in resolver

fibonacci1729:


alexcrichton (approved):
Thanks! I think a test needs updating but otherwise looks good 👍

PR #1133, opened 2023-07-19
[wit-component] add shared-everything linking support

dicej:
Per https://github.com/WebAssembly/component-model/blob/main/design/mvp/examples/SharedEverythingDynamicLinking.md and https://hackmd.io/IlY4lICRRNy9wQbNLdb2Wg.

This adds a new `component link` subcommand, which resembles `component new` but accepts an arbitrary number of "dynamic library" modules (as defined by https://github.com/WebAssembly/tool-conventions/blob/main/DynamicLinking.md), linking them together to create a component whose type is the union of the component types found in the input modules (via the usual custom section representation).  This process can also be invoked programmatically via a new `Linker` API.

`Linker` analyzes and topologically sorts the input modules, then synthesizes two additional modules:

- `main`: hosts the component's single memory and function table and exports any functions needed to break dependency cycles discovered in the input modules. Those functions use `call.indirect` to invoke the real functions, references to which are placed in the table by the `init` module.

- `init`: populates the function table as described above, initializes global variables per the dynamic linking tool convention, and calls any static constructors and/or link-time fixup functions

Finally, `Linker` generates a directed graph of modules and their instantiations and passes it to `ComponentEncoder` to actually produce the component.

The implementation relies on a few modifications to the existing code in `ComponentEncoder` and friends.  Specifically, I've expanded the meaning of "adapter" to include dynamic libraries as well as traditional adapters.  The former are distiguished from the latter by the presence of a new `LibraryInfo` struct.  This approach minimized the number of code changes in the interest of keeping this PR comprehensible, but we will probably want to refactor `ComponentEncoder` in the future to make the distinction between traditional adapters and dynamic libraries clearer, possibly generalizing how we analyze and instantiate modules of various flavors.

I've created a demo which ties everything together, for reference: https://github.com/dicej/component-linking-demo.  It includes `wasi-sdk` and `wasi-libc` patches which have not yet been upstreamed; I'm planning to work on that next.


dicej:
I need to add some more tests before this will be ready to merge, but I'm opening a draft PR in case anyone wants to comment in the meantime.

dicej:
TIL (most of) the `wasm-tools` tests need to be able to run in a Wasm guest, and I broke that by using `wasmtime` in the linking test.  I guess I have two options to address that: either rewrite the linking test to not use `wasmtime` (i.e. turn it into one of those "matches expected output WAT" tests) or exclude the linking test from the set of tests run inside Wasm.  Anyone have a preference?

crates/wit-component/tests/linking.rs
```diff
@@ -0,0 +1,270 @@
...
+    anyhow::{Context, Result},
+    std::{borrow::Cow, path::Path},
+    wasm_encoder::{CustomSection, Encode as _, Module, RawSection},
+    wasmparser::Parser,
+    wit_component::StringEncoding,
+    wit_parser::{Resolve, UnresolvedPackage},
+};
+
+const FOO: &str = r#"
+(module
```
alexcrichton, line 11: One comment I'd have on this is that I think it'd be best if the test suite here could be structured like the other wit-component tests where there's a folder-of-tests and each test is a set of files configuring what needs to be done. The test-runner is more-or-less "go use the library equivalents of `wasm-tools` CLI commands to execute tests".

I've opened https://github.com/bytecodealliance/wasm-tools/pull/1135 which may be able to assist with that, but I've found this style of tests is beneficial for reading tests, writing tests, and debugging tests. (e.g. basically making it such that it's possible to play with everything on the CLI rather than requiring custom Rust code to drive everything.

dicej: I've merged #1135 into https://github.com/dicej/wasm-tools/pull/new/linking-wat and added linking support to the components.rs test runner.  When 1135 is merged, I 'll rebase this PR and pull that in.  Meanwhile, I'll continue adding tests.

BTW, is it okay if I keep the one end-to-end test in linking.rs in addition to all the WAT tests?  It would be nice to have at least one test that actually runs the component.

dicej: I've finished adding tests to the `linking-wat` branch of my fork, so just waiting on 1135 to land at this point.

crates/wit-component/tests/linking.rs
```diff
@@ -0,0 +1,270 @@
...
+"#;
+
+struct MemInfo {
+    memory_size: u32,
+    memory_alignment: u32,
+    table_size: u32,
+    table_alignment: u32,
+}
+
+fn encode(
```
alexcrichton, line 139: Could this perhaps be replaced with the `@dylink` text format in the above `*.wat` files?

crates/wit-component/src/linking.rs
```diff
@@ -0,0 +1,1383 @@
...
+
+    let memory_size = {
+        let mut add_global_export = |name: &str, value, mutable| {
+            let index = globals.len();
+            globals.global(
+                wasm_encoder::GlobalType {
+                    val_type: ValType::I32,
+                    mutable,
+                },
+                &ConstExpr::i32_const(i32::try_from(value).unwrap()),
```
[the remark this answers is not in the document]
alexcrichton: I see now this is used quite a bit throughout this module, so perhaps something like:

```rust
fn const_u32(a: u32) -> ConstExpr { .. }
```

or something like that?

crates/wit-component/src/linking.rs
```diff
@@ -0,0 +1,1383 @@
...
+    let mut resolved = HashMap::new();
+    let mut missing = Vec::new();
+    let mut duplicates = Vec::new();
+
+    let mut triage = |metadata: &'a Metadata, export: Export<'a>| {
+        if let Some((key, value)) = exporters.get_key_value(&export.key) {
+            match value.as_slice() {
+                [] => unreachable!(),
+                [exporter] => {
+                    resolved.insert(*key, *exporter);
```
alexcrichton, line 751: Could this assert that the insertion didn't replace anything?

crates/wit-component/src/linking.rs
```diff
@@ -0,0 +1,1383 @@
...
+fn topo_add<'a>(
+    sorted: &mut IndexSet<usize>,
+    dependencies: &HashMap<usize, HashSet<usize>>,
+    element: usize,
+) {
+    let empty = &HashSet::new();
+    let deps = dependencies.get(&element).unwrap_or(empty);
+
+    // First, add any dependencies which do not depend on `element`
+    for &dep in deps {
```
alexcrichton, line 837: One thing to be careful of here is that `deps` is a `HashSet` which has non-deterministic iteration order between runs. This might be a good place to use `IndexSet` to ensure the output of the `link` command is deterministic.

crates/wit-component/src/linking.rs
```diff
@@ -0,0 +1,1383 @@
...
+                ));
+            }
+        }
+    }
+
+    (resolved, missing, duplicates)
+}
+
+/// Recursively add a library (represented by its offset) and its dependency to the specified set, maintaining
+/// topological order (modulo cycles).
```
alexcrichton, line 827: I'm a bit curious about how this works in the presence of cycles. I've never worked with topo sorts with cycles before, so naively I'd interpret the checks below as they handle a cycle of A->B->A but wouldn't handle deeper cycles and may recurse infinitely? I haven't worked with this before so I'm not sure though.

Put another way I suppose, what's the meaning of a topo sort in the presence of cycles?

crates/wit-component/src/linking.rs
```diff
@@ -0,0 +1,1383 @@
...
+/// Analyze the specified library metadata, producing a map of transitive dependencies, where each library is
+/// represented by its offset in the original metadata slice.
+fn find_dependencies(
+    metadata: &[Metadata],
+    exporters: &HashMap<&ExportKey, (&str, &Export)>,
+) -> Result<HashMap<usize, HashSet<usize>>> {
+    let mut dependencies = HashMap::<_, HashSet<_>>::new();
+    let mut indexes = HashMap::new();
+    for (index, metadata) in metadata.iter().enumerate() {
+        indexes.insert(metadata.name, index);
```
alexcrichton, line 874: Could this, and insertions below, assert that they all succeed? (e.g. don't replace anything)

crates/wit-component/src/linking.rs
```diff
@@ -0,0 +1,1383 @@
...
+            (
+                indexes[k],
+                v.into_iter().map(|v| indexes[v]).collect::<HashSet<_>>(),
+            )
+        })
+        .collect::<HashMap<_, _>>();
+
+    let empty = &HashSet::new();
+
+    loop {
```
alexcrichton, line 901: Could this have a comment to explain what's going on? It looks like a fixpoint-style loop but I'm not quite sure why that's needed here

alexcrichton: I'll also note that there's more `HashMap` iteration here which may run a risk of leading to non-determinism in the output

crates/wit-component/tests/components/error-link-duplicate-symbols/error.txt
```diff
@@ -0,0 +1,114 @@
+duplicate symbol(s): [
```
alexcrichton, line 1: Could this, and some of the other errors here, perhaps be formatted without `Debug` output?

alexcrichton (approved):
Looks reasonable to me, thanks for putting in the work on this! This definitely turned out pretty gnarly, but that's to be expected in this problem domain.

I don't intimiately understand the details of dynamic linking but everything appeared pretty reasonable to me here so I'm leaning on you mostly to ensure it's handled alright (and sounds like with the size of the demos you're working with that'll be ok). The main two meta-level comments I have which are also interspersed in some review are:

* It may be worth watching out for hash maps and iteration order leading to nondeterministic output. In my opinion maps are fine when it's known they won't affect the output order (e.g. only used for lookup), but maps that are iterated over may want to be `Index{Map,Set}` instead.
* In general I try to assert that insertion into a map or set succeeds without replacing something to catch possible bugs where I got something wrong by accident. It may be worth adding asserts to all the various places to handle that?

crates/wit-component/src/linking.rs
```diff
@@ -0,0 +1,1383 @@
...
+                    );
+                }
+            }
+        }
+
+        memory_offset = align(memory_offset, HEAP_ALIGNMENT_BYTES);
+        add_global_export("__heap_base", memory_offset, true);
+
+        let heap_end = align(memory_offset, PAGE_SIZE_BYTES);
+        add_global_export("__heap_end", heap_end, true);
```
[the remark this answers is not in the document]
dicej: LLVM (i.e. wasm-ld) expects them to be.  For example:

```
 $ wasm-tools print /opt/wasi-sdk/share/wasi-sysroot/lib/wasm32-wasi/libc.so | grep '"__heap'
  (import "GOT.mem" "__heap_base" (global $__heap_base (;3;) (mut i32)))
  (import "GOT.mem" "__heap_end" (global $__heap_end (;4;) (mut i32)))
```

crates/wit-component/src/linking.rs
```diff
@@ -0,0 +1,1383 @@
...
+                .enumerate()
+                .map(|(offset, (name, ..))| (*name, table_offset + u32::try_from(offset).unwrap()))
+                .collect::<HashMap<_, _>>();
+
+            for metadata in metadata {
+                for import in &metadata.table_address_imports {
+                    add_global_export(
+                        &format!("{}:{import}", metadata.name),
+                        *offsets.get(import).unwrap(),
+                        true,
```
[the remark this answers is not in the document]
dicej: From https://github.com/WebAssembly/tool-conventions/blob/main/DynamicLinking.md#imports:
> The imported global must be mutable as the dynamic linker might need to modify the value after instantiation.

crates/wit-component/src/linking.rs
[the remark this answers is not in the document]
alexcrichton: alas!

crates/wit-component/src/linking.rs
dicej: In this case, "topo sort in the presence of cycles" means the algorithm makes no guarantees about the relative order of items which participate in a cycle.  If we wanted to, we could refine that a bit so that if A needs 20 things from B and B only needs 3 things from A, then we'd put B first since it doesn't depend on A "as much" as A depends on B, but that was more than I was ready to tackle.

It handles deeper cycles by expecting that `dependencies` includes transitive dependencies, not just the first level of dependencies.

crates/wit-component/src/linking.rs
```diff
@@ -0,0 +1,1383 @@
...
+                        .map(|export| {
+                            let name_address = memory_base + u32::try_from(buffer.len()).unwrap();
+                            write_bytes_padded(&mut buffer, export.key.name.as_bytes());
+
+                            let address = if let Type::Function(_) = &export.key.ty {
+                                Address::Function(
+                                    table_base + get_and_increment(&mut function_count),
+                                )
+                            } else {
+                                Address::Global(export.key.name)
```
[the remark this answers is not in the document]
dicej: `linking::metadata::Type` only has two variants: `Function` and `Global`, so there's nothing to assert for now.  I'll change this to a `match`, though, which will give us an exhaustiveness check in case we add more variants later.

crates/wit-component/src/linking.rs
```diff
@@ -0,0 +1,1383 @@
...
+    fn new(table_base: u32, memory_base: u32, metadata: &'a [Metadata<'a>]) -> Self {
+        let mut function_count = 0;
+        let mut buffer = Vec::new();
+        let mut global_addresses = Vec::new();
+        let mut libraries = metadata
+            .iter()
+            .filter_map(|metadata| {
+                if metadata.dl_openable {
+                    let name_address = memory_base + u32::try_from(buffer.len()).unwrap();
+                    write_bytes_padded(&mut buffer, metadata.name.as_bytes());
```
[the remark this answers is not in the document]
dicej: https://github.com/dicej/wasi-libc/blob/76c7e1e1cfdad577ecd7f61c67ead7a38d62a7c4/libc-top-half/musl/src/misc/dl.c will read the tables we generate here as C structs, so the fields need to be aligned accordingly.

crates/wit-component/src/linking.rs
```diff
@@ -0,0 +1,1383 @@
...
+
+            memory_offset += metadata.mem_info.memory_size;
+            table_offset += metadata.mem_info.table_size;
+
+            for import in &metadata.memory_address_imports {
+                add_global_export(&format!("{}:{import}", metadata.name), 0, true);
+            }
+        }
+
+        {
```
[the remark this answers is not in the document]
dicej: Not necessary, but I find that limiting the scope of variables helps readability by making it clear the value won't be used elsewhere.

crates/wit-component/src/linking.rs
dicej: In this case, we actually expect it may replace something since multiple libraries may import the same symbol.  I've added a comment to note that.  I've also added assertions to a few places where we don't expect replacement.

crates/wit-component/src/linking.rs
dicej: I've changed this to an `IndexSet`.

crates/wit-component/src/linking.rs
dicej: I've added a comment and changed it to an `IndexMap`.

dicej:
Thanks, as always, for the thorough review, @alexcrichton.  I believe I've addressed all your feedback so far.

alexcrichton:
👍 Thanks again!

PR #1134, opened 2023-07-19
GC: wasmparser: recursive type groups support

imikushin:
Introduce support for recursive types as part of the ongoing WebAssembly GC proposal implementation. 

Matching rules for subtypes had to be updated to accommodate recursive types.


crates/wasm-smith/tests/core.rs
```diff
@@ -130,13 +130,15 @@ fn smoke_test_imports_config() {
...
                     // Gather the signature types to later check function types against.
                     for ty in rdr {
-                        match ty.unwrap().structural_type {
-                            wasmparser::StructuralType::Func(ft) => sig_types.push(ft),
-                            wasmparser::StructuralType::Array(_) => {
-                                unimplemented!("Array types are not supported yet.")
-                            }
-                            wasmparser::StructuralType::Struct(_) => {
-                                unimplemented!("Struct types are not supported yet.")
+                        for ty in ty.unwrap().types {
```
fitzgen, line 133: Here and elsewhere, when we are iterating over rec groups, can we name the variable `rec_group` instead of `ty` since it is not a type? Thanks.

crates/wasmparser/src/validator/core.rs
```diff
@@ -493,56 +496,77 @@ pub(crate) struct Module {
 }
 
 impl Module {
-    pub fn add_type(
+    pub fn add_types(
         &mut self,
-        ty: SubType,
+        types_added: Vec<SubType>,
```
fitzgen, line 501: I think this can be a slice, rather than a `Vec` which forces heap allocations on callers, particularly callers that are just passing in a single type.

```suggestion
        types_added: &[SubType],
```

And then single-type callers can use https://doc.rust-lang.org/nightly/std/slice/fn.from_ref.html

fitzgen (commented):
Thanks! A couple things below before we can merge this.

Additionally, can we have some tests for

* [ ] Empty rec groups
* [ ] Invalid modules that try to be type recursive across different groups
* [ ] Any other edge cases you can think of

crates/wasm-smith/tests/core.rs
imikushin: Done.

crates/wasmprinter/src/lib.rs
```diff
@@ -732,7 +742,11 @@ impl Printer {
         for ty in parser.into_iter_with_offsets() {
             let (offset, ty) = ty?;
             self.newline(offset);
-            self.print_type(state, ty)?;
+            if ty.types.len() == 1 {
```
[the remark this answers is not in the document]
imikushin: Done.

crates/wasmparser/src/validator/core.rs
imikushin: Done.

crates/wasmparser/src/readers/core/types.rs
```diff
@@ -771,15 +816,31 @@ pub struct SubType {
     pub structural_type: StructuralType,
 }
 
+/// Represents a recursive type group in a WebAssembly module.
+#[derive(Debug, Clone)]
+pub struct RecGroup {
+    /// The list of subtypes in the recursive type group.
+    pub types: Vec<SubType>,
+}
+
```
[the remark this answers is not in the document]
imikushin: Just pushed a commit making `RecGroup` internally an enum of either a `Vec<SubType>` or a single `SubType` (Thanks @fitzgen for the suggestion!)

Performance wise it's not as bad as `Vec<SubType>` per every `RecGroup`.

crates/wit-component/src/linking/metadata.rs
```diff
@@ -344,8 +345,16 @@ impl<'a> Metadata<'a> {
...
-                        .filter_map(|r| {
-                            r.map(|ty| match ty.structural_type {
+                        .flat_map(|r| match r {
+                            Err(e) => vec![Err(e)],
+                            Ok(rg) => rg
+                                .types()
+                                .iter()
+                                .map(|st| Ok(st.clone()))
+                                .collect::<Vec<_>>(),
+                        })
```
fitzgen, line 354: We shouldn't force additional `Vec` allocations here. It is fine to write a `for` loop if the iterators are pushing you in this direction.

fitzgen (commented):
Looking good, very close! Just one thing left to address before we can merge this. Thanks for your patience!

crates/wit-component/src/linking/metadata.rs
imikushin: Please don't kill it! I spent entire 3 days trying to come up with it 😅

Interesting though, the current code in this PR is slightly faster than this:
```rust
                    types = Vec::new();
                    for rg in reader {
                        for ty in rg?.types() {
                            if let StructuralType::Func(ty) = ty.clone().structural_type {
                                types.push(ty);
                            }
                        }
                    }
```

If we use this snippet, performance drops:
```
parse/spidermonkey      time:   [15.602 ms 15.622 ms 15.641 ms]                               
                        change: [+1.5950% +1.8649% +2.1264%] (p = 0.00 < 0.05)
                        Performance has regressed.

validate/lots-of-types  time:   [741.42 µs 744.22 µs 747.13 µs]                                   
                        change: [+5.1396% +5.5738% +5.9996%] (p = 0.00 < 0.05)
                        Performance has regressed.

parse/lots-of-types     time:   [164.25 µs 164.57 µs 164.87 µs]                                
                        change: [+1.8325% +2.3509% +2.8039%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

fitzgen: Huh! Okay well I can't argue with profiling results!

fitzgen (approved):

PR #1135, opened 2023-07-21
Add text format and printing support for `dylink.0`

alexcrichton:
This is something I threw together which I think might be helpful in testing and working on #1133. In theory this should also make it easier to inspect this custom section in preexisting binaries.

I've invented a text format of this out of nowhere as I don't believe there's a specification or anything else to follow. I've opted to use the `@dylink.0` annotation format to emphasize that it's a custom section, in the same manner that `@producers` uses that today.

Note that this additionally copies some of the `wasmparser`-related parsing code from #1133.


crates/wasmparser/src/readers/core/dylink0.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// The required alignment of the table area, in elements, encoded as a
+    /// power of 2.
+    pub table_alignment: u32,
+}
+
+#[allow(missing_docs)]
+#[derive(Debug)]
+pub struct ExportInfo<'a> {
+    pub name: &'a str,
+    pub flags: u32,
```
dicej, line 36: Eventually, it would be cool to use `bitflags` to represent the possibilities described in https://github.com/WebAssembly/tool-conventions/blob/main/Linking.md.  Maybe add a TODO comment about that?

crates/wasmprinter/src/lib.rs
```diff
@@ -2577,6 +2584,73 @@ impl Printer {
...
+                        self.print_str(s)?;
+                    }
+                    self.end_group();
+                }
+                Dylink0Subsection::ExportInfo(info) => {
+                    for info in info {
+                        self.newline(start);
+                        self.start_group("export-info ");
+                        self.print_str(info.name)?;
+                        write!(self.result, " {}", info.flags)?;
```
dicej, line 2631: Similar to the above, it would be cool if we could print and parse the flags symbolically (e.g. `(binding-weak binding-local)` meaning `WASM_SYM_BINDING_WEAK | WASM_SYM_BINDING_LOCAL`; not sure what the "right" way to represent flags as s-expressions would be).  Not a blocker, of course.

dicej (approved):

crates/wasmparser/src/readers/core/dylink0.rs
alexcrichton: Agreed!

crates/wasmprinter/src/lib.rs
alexcrichton: nah I was thinking the same thing! I was thinking `(export-info "x" binding-weak binding-local 0x80)` where you'd or-together everything after the string `"x"` to represent this, and it could handle unknown flags in theory too with integer literals

alexcrichton: I went ahead an implemented this, but I'll leave the bitflags bits for later

PR #1140, opened 2023-07-24
Update wast to align with GC spec tests

bvisness:
A few changes required to get the spec tests running from the GC repo. Changes include:

- Folded expressions are now allowed in places where the text format calls for a "single instruction". This was used by a few spec tests and isn't ambiguous.
- `ref.any`, `ref.eq`, `ref.struct`, `ref.array`, and `ref.i31` directives were added as returns. They assert that the returned ref has the given ref type. (FYI to those wishing to use this, it basically implies that you should do a `ref.cast` or `ref.test`.)
- `ref.host <u32>` was added as an arg and a return. It represents an internalized host value (i.e. a host value stored as `anyref`).
- `ref.extern` as a return was changed to make the u32 optional. `(ref.extern)` is now a valid ref type assertion without asserting anything about the value.
- Struct field parsing was updated to disallow the `(struct i32)` shorthand (see https://github.com/WebAssembly/gc/issues/333#issuecomment-1480201810)

In addition, a couple spec-related bugs were fixed:

- The binary encoding for rec groups was incorrect. This was an oversight from #1204.
- Types were incorrectly being marked as nonfinal by default; however, types without `sub` are supposed to be final. This was corrected and local GC subtyping tests were fixed.


alexcrichton:
Looks good to me, thanks!

bvisness:
Sorry for the noise on here while I add more changes. Didn't realize you'd get notified for draft PRs.

alexcrichton:
Oh no worries! I'll archive the notifications until this is ready

osa1:
@bvisness do you need help with this PR? I was planning to add support for GC syntax myself this weekend but just saw your PR.

bvisness:
I'll be back to this on Monday but if you want to jump on it first, go for it. If anything my code can maybe serve as reference. 

The biggest difficulty is a disagreement in how to interpret "single instruction" when omitting e.g. offsets for element segments. The spec tests allow folded instructions while wasmparser currently does not. You can see my attempt at changing this in the last couple commits.

alexcrichton:
Oh are there particular `*.wast` files causing trouble? I've had difficulty figuring out how to parse these historically as depending on which tool you're working against at which point of time historically things have been slightly different (or I've just interpreted things wrong)

Basically the element/offset/table text parsing I've always found tricky, and I'm happy to help if you'd like.

bvisness:
This is now successfully building all the GC spec tests on my end. I'm updating the top comment with a description of what I had to do.

alexcrichton:
Yeah it may be worth clarifying, but to be clear I don't think it's worth blocking this PR for that. With some more tests I'd be happy to merge this. Spec tests are currently updated by updating the `tests/testsuite` submodule which points to https://github.com/WebAssembly/testsuite. That repository in turn has a script which updates its contents with the actual spec repos. I'm not sure what the status is with `gc` and the `testsuite` repo and how far apart they are. If `testsuite` upstream isn't updated to include this new syntax it's ok to make custom `tests/local/*.wast` tests for gc things.

bvisness:
The testsuite repo seems to be up to date with the latest GC tests, but how do I actually run the proposal tests? I ran `cargo test --all` but I'm not seeing any `gc` stuff in the output (or indeed any `testsuite` stuff at all).

alexcrichton:
Sorry back from vacation now, but tests are run as part of `cargo test --all` and are specifically run as `cargo test --test roundtrip`. If those aren't picking up gc tests then it's the `tests/roundtrip.rs` file that needs modifying and/or the submodule (sorry I'm not sure the state of gc and the upstream spec tests)

bvisness:
Ah I see, the GC tests were explicitly ignored. I'll work through that now and make sure they're running as expected.

bvisness:
At this point we need to update the test suite repos in order to pass GC spec tests. However, I'm not sure how to update the test suite for GC stuff without pulling in a bunch of other changes too. After running the update script in the test suite, I'm getting a bunch of errors seemingly related to threading, for example. Strangely this is happening even though my update didn't seem to touch threading at all. I did see that there were some kind of git conflicts while running the update script...

Generally I have to admit I'm a bit lost. Could you give me some guidance on how to proceed?

alexcrichton:
Sure yeah! Wanna sync up on [Zulip](https://bytecodealliance.zulipchat.com/) and/or hop on a zoom?

bvisness:
I will find my way onto Zulip tomorrow for sure, and would be happy to work through stuff then or on Monday. Thank you!

bvisness:
Ok, got the spec tests in and passing locally. They will not be passing here in the PR yet because the spec test suite needs to be updated - I have an open PR on the test suite [here](https://github.com/WebAssembly/testsuite/pull/69).

I attempted to preserve validation for the parts of the GC proposal that have been implemented so far, namely subtype validation. The implementation for this was incorrect; types with no explicit `sub` were nonfinal by default, but they should have been final.

Assuming we don't need to rip out the special folded-instruction parser we discussed earlier, I think this should be good to go as soon as the submodule stuff is taken care of. (If the folded instruction stuff will be an issue, I think we should open an issue in the GC repo to clarify what is "one instruction", and I would prefer that discussion not block this work.)

alexcrichton:
All looks good to me! If you want to pull in the submodule I'll go ahead and merge and get this published (plus the gc encoding changes)

bvisness:
Done!

alexcrichton:
I think a `cargo fmt` at the root of the repository should take care of it, but I never manually run that so I'm only 80% sure

alexcrichton (approved):
I went ahead and ran `rustfmt` and pushed up the minor change, so I'll let CI merge this when ready.

PR #1141, opened 2023-07-25
Fix encoding of type information in wasm-compose

alexcrichton:
This commit is a large refactoring and restructuring of how type
information for imports is encoded with `wasm-compose`. This was all
originally written in an era without resource types and without aliasing
of types between instances being well-defined, which means that composed
components end up not currently reflecting the correct type information
in terms of where types are defined and ensuring everything is defined
only once. The test cases added here in this PR either all fail to
generate a valid component today or generate a component that has an
incorrect structure of type information in that it's different from the
sources.

The new implementation of type encoding here is motivated by taking a
larger demo of composition and effectively fixing issues until the whole
thing is composed together. In that sense there's not really a rhyme or
reason to the changes here other than "this is what's necessary to fix
this one test case". All that to say that there are probably still
latent issues to fix and that this would make a fantastic fuzz target
(although I don't know how really myself). For now though the new tests
require most of the implementation here and new tests can be added over
time as well.

At a high level the main part of this restructuring is to follow the
exact definitions of `TypeId` rather than always dereferencing it and
looking at the underlying definition. This involves following alias
chains of exports and additionally adding support for pulling in types
from other instances and such. This is all very similar to what
`wit-component` is already doing when translating from WIT to wasm, but
it's distinct as well in the sense that `wasm-compose` is weaving
multiple components together, each with different type information,
where `wit-component` is translating a WIT document into wasm. All that
to say that while there's at a high level a good deal of shared code
here at a lower level there's not much acutal possible sharing without
what I would predict is quite odd contortions.


peterhuene (approved):
Looks great 👍 Thanks for fixing this.

Just the one nit.

PR #1142, opened 2023-07-25
Disallow empty types in the component model

alexcrichton:
This updates the component validation phase to account for WebAssembly/component-model#218


pchickey:
https://github.com/WebAssembly/component-model/pull/218 doesn't make the restriction for variant types, but this PR does - where do we reconcile that?

alexcrichton:
Oh oops sorry yes I forgot to mention, this additionally fixes mistakes where enums/unions accidentally allowed zero cases but they're both required to have at least one case like variants currently require.

pchickey (approved):

PR #1146, opened 2023-07-26
Impl imports

macovedj:
This PR can serve as an implementation of what is proposed [here](https://github.com/WebAssembly/component-model/pull/222).


crates/wasmparser/src/validator/names.rs
```diff
@@ -320,12 +345,66 @@ impl KebabName {
...
+                        &mut slash_acc,
+                    );
+                    let last_slash = slash[slash.len() - 1];
+                    let at = s.as_str()[last_slash..].find('@').map(|i| i + last_slash);
+                    ParsedKebabName::RegistryId {
+                        colon: colon.iter().map(|i| *i as u32).collect::<Vec<u32>>(),
+                        slash: colon.iter().map(|i| *i as u32).collect::<Vec<u32>>(),
+                        at: at.map(|i| i as u32),
+                    }
+                }
```
alexcrichton, line 386: This one is unfortunately pretty tricky. I don't think the tooling is really ready to support the generalization of ids landed in https://github.com/WebAssembly/component-model/pull/222. There's a lot of work to do on the WIT side of things to support that which isn't part of this PR and would eventually need to make its way in there. For example the `(interface "...")` production should be using the same validation that's added here for `locked`, but it's not currently.

I think it would be best for now to update `RegistryId` to specify whether the `at` is either "none", a "semver", or a "semver range". Otherwise though the locked/unlocked/interface productions here should all use `RegistryId` in the same format, which I think for now we'll need to limit to `foo:bar/baz` with no more colons or slashes. I realize that's not up-to-date with the spec but there will need to be future work for that generalization which is plumbed through to WIT everywhere too.

crates/wast/src/component/import.rs
```diff
@@ -30,17 +67,175 @@ pub enum ComponentExternName<'a> {
     Kebab(&'a str),
     /// This is an interface import where the string is an ID.
     Interface(&'a str),
+    /// This is an implementation import where the string is an ID.
+    Implementation(ImplementationImport<'a>),
```
alexcrichton, line 71: Similar to wasm-encoder and wasmparser, I think this should have the same "flat" representation where the variants of `ImplementationImport` are inlined above. (and additionally split into `Component{Export,Import}Name`

crates/wast/src/component/import.rs
```diff
@@ -30,17 +67,175 @@ pub enum ComponentExternName<'a> {
...
+  /// External Url
+  Url(ImportMetadata<'a>),
+  /// Relative path
+  Relative(ImportMetadata<'a>),
+  /// Just Integrity
+  Naked(ImportMetadata<'a>),
+  /// Locked Registry Import
+  Locked(ImportMetadata<'a>),
+  /// Unlocked Registry Import
+  Unlocked(ImportMetadata<'a>),
```
alexcrichton, line 86: Additionally, like wasmparser/wasm-encoder, I think each of these variants should have a specific payload just to that one variant.

alexcrichton (commented):
Thanks! I think it would be good to add a suite of tests for this too. Adding files to `tests/local/component-model/*.wast` should be good enough. It'll be good to exercise the text parsing, binary encoding, binary validation, and text printing for all of this.

alexcrichton:
Oh one other comment I'd recommend is to configure your text editor to rustfmt-on-save since there's some un-formatted files here  which will get flagged on CI once https://github.com/bytecodealliance/wasm-tools/pull/1174 merges

crates/wasmparser/src/validator/names.rs
```diff
@@ -260,11 +326,11 @@ pub enum KebabNameKind<'a> {
     },
     /// `wasi:http/types@2.0`
     #[allow(missing_docs)]
-    Id {
+    RegistryId {
         namespace: &'a KebabStr,
         package: &'a KebabStr,
-        interface: &'a KebabStr,
-        version: Option<Version>,
+        interface: Option<&'a KebabStr>,
```
alexcrichton, line 332: To keep things somewhat simpler for now could this be left as `&'a KebabStr` and effectively only `a:b/c` works today?

crates/wasmparser/src/validator/names.rs
```diff
@@ -316,14 +509,144 @@ impl KebabName {
...
+                            }),
+                        }
+                    } else {
+                        ParsedKebabName::RegistryId {
+                            colon: colon as u32,
+                            slash: None,
+                            at: None,
+                        }
+                    }
                 }
```
alexcrichton, line 650: I believe this is the more-or-less the same parsing code as `Locked`, so could those be deduplicated?

crates/wasmparser/src/validator/names.rs
```diff
@@ -222,7 +222,73 @@ pub struct KebabName {
...
+                Semver::Semver(other_version) => version == other_version,
+                Semver::SemverRange(_) => false,
+            },
+            Self::SemverRange(version_range) => match other {
+                Semver::Semver(_) => false,
+                Semver::SemverRange(other_range) => version_range == other_range,
+            },
+        }
+    }
+}
```
alexcrichton, line 253: I think this can be replaced with `#[derive(PartialEq)]` on the type?

alexcrichton:
If you'd like I don't mind taking it from here and landing this as well. Implementing these sorts of changes can be pretty tricky so I don't want to put too much on you by accident

crates/wasmparser/src/validator/names.rs
macovedj: The primary motivator here was that when actually using the new implementation imports, there is no `/` generally.  For example, unlocked imports tend to look something like
```
(import (unlocked-dep "foo:five@{>=2.0.0 <3.0.0}") (instance (;0;) (type 1)))
```
It definitely would be nice to simplify if possible, as I added a lot of checks elsewhere to enable the optionality here, but I'm not sure the best way to go about that while simultaneously supporting this use case.

macovedj:
> If it's ok with you @macovedj I'd prefer to hold off on this until [WebAssembly/component-model#253](https://github.com/WebAssembly/component-model/issues/253) is resolved since that will change the binary and textual syntax of imports again. Additionally it will likely change the internals of representation within `wasm-tools` crates, so I think it would be best to handle that all at once rather than having intermediate states to deal with.

Makes sense to me!

alexcrichton:
I've opened up https://github.com/bytecodealliance/wasm-tools/pull/1262 which is intended to be the successor to this PR which catches up with other component model developments as well

PR #1149, opened 2023-07-26
Delete most of `abi` from `wit-parser`

alexcrichton:
This commit deletes most of the `abi` module from the `wit-parser` crate. This functionality is entirely untested in this repository and was originally intended for `wit-bindgen`-style generators. In thinking about bytecodealliance/wit-bindgen#622 I'm going to want to make some invasive changes to this and I'd rather develop everything close to the source rather than across two repositories.

My plan is to move all this code upstream into the `wit-bindgen-core` crate in the `wit-bindgen` repository. Which is to say this is the first half of moving the code, deleting the source here, and I'll make a follow-up PR to re-add all this code into `wit-bindgen-core` so all current consumers will continue to work and/or can switch to depending on `wit-bindgen-core` to preserve the functionality.


pchickey (approved):
:fire:

PR #1154, opened 2023-08-02
Update wasm-compose example based on latest changes.

peterhuene:
This PR updates the `wasm-compose` example based on the latest `cargo-component` changes.

### DIFF
--- a/crates/wasm-compose/example/service.wit
+++ b/crates/wasm-compose/example/service.wit
@@ -1,4 +1,4 @@
-package example:service
+package example:service@0.1.0
 
 interface handler {
     record request {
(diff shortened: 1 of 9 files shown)


crates/wasm-compose/example/service.wit
```diff
@@ -1,4 +1,4 @@
-package example:service
+package example:service@0.1.0
```
Kylebrown9, line 1: What does this version mean?

peterhuene: Right now `cargo-component` automatically versions the components it builds, even if the `package` isn't explicitly versioned in WIT (if it is, it must match what's in `Cargo.toml`). It does this to prevent two sources of truth for the version of the component you're building.

However, the `wasmtime::component::generate!` macro doesn't support specifying a version independent of the WIT; thus we add it here to ensure that the Wasmtime bindings look for the right export from the component.

alexcrichton (approved):

PR #1159, opened 2023-08-07
Probe for post-return instead of asserting existence

alexcrichton:
This commit updates `wit-component` to probe for the existence of post-return functions rather than asserting their existence. This enables moves some more ABI bits to `wit-bindgen-core` since they otherwise don't quite make sense here. This additionally fixes a longstanding issue with `wit-component`, although it's not one that's actually come up in practice yet.


crates/wit-component/src/encoding.rs
```diff
@@ -1100,12 +1106,11 @@ impl<'a> EncodingState<'a> {
 
         // TODO: This should probe for the existence of
         // `cabi_post_{name}` but not require its existence.
```
sunfishcode, line 1108: Is this comment still relevant?

alexcrichton: Oops, forgot to remove!

crates/wit-component/src/validation.rs
```diff
@@ -673,6 +684,20 @@ fn validate_func(
     )
 }
 
+fn validate_post_return(
+    resolve: &Resolve,
+    ty: &wasmparser::FuncType,
+    func: &Function,
+) -> Result<()> {
+    let mut sig = resolve.wasm_signature(AbiVariant::GuestExport, func);
+    sig.params = mem::take(&mut sig.results);
```
dicej, line 693: A comment here might be helpful -- took me a minute to realize this was because the parameters to a post-return function should match the results of the primary function.

dicej (approved):

PR #1161, opened 2023-08-09
Extract `LocalParser` for parsing `local`. (#1156)

### LINKED ISSUE #1156: `Local` struct do not implement `Parse` trait
Regarding [`Local<'a>` struct](https://github.com/bytecodealliance/wasm-tools/blob/559ede0a436240cb7dfe162c568f398655e9b18c/crates/wast/src/core/func.rs#L85) in `crates/wast`, Is there a reason why this struct do not implement `Parse<'a>`?

The alternative, [`parse_remainder`](https://github.com/bytecodealliance/wasm-tools/blob/559ede0a436240cb7dfe162c568f398655e9b18c/crates/wast/src/core/func.rs#L96) is not visible outside the crate. This is preventing me from parsing `local` outside of a `func`.

I can work around this for the time being, but I am curious if there is a reason to do it this way. Thanks.

anoopelias:
Extract `LocalParser` as a struct and implement `Parse` on it.

The current implementation of `parse_remainder` on `Local` is visible only within the crate. This extraction will help create an external facing API for parsing instructions like `(local $num i32)`.

Resolves #1156

### DIFF
--- a/crates/wast/src/core/func.rs
+++ b/crates/wast/src/core/func.rs
@@ -92,27 +92,42 @@ pub struct Local<'a> {
     pub ty: ValType<'a>,
 }
 
+/// Parser for `local` instruction.
+///
+/// A single `local` instruction can generate multiple locals, hence this parser
+pub struct LocalParser<'a> {
+    /// All the locals associated with this `local` instruction.
+    pub locals: Vec<Local<'a>>,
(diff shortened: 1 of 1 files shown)


alexcrichton (approved):
Thanks!

PR #1167, opened 2023-08-16
wit-component: Add WitPrinter::emit_docs option

lann:
Add a configuration option and implementation to print doc comments with associated items.

### DIFF
--- a/crates/wit-component/src/printing.rs
+++ b/crates/wit-component/src/printing.rs
@@ -5,16 +5,36 @@ use std::mem;
 use wit_parser::*;
 
 /// A utility for printing WebAssembly interface definitions to a string.
-#[derive(Default)]
 pub struct WitPrinter {
     output: Output,
 
     // Count of how many items in this current block have been printed to print
     // a blank line between each item, but not the first item.
     any_items: bool,
+
+    // Whether to print doc comments.
+    emit_docs: bool,
+}
+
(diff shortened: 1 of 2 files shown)


lann:
The existing test harnesses don't support this as they all transcode through binary wasm or wat. I'm planning to work on adding doc comments to the binary encoding next, so this change can either wait for that to gain testing or be merged as-is.

lann:
A couple of questions about the implementation:

- Should `emit_docs` even be optional?
  - Should it default to `true`?
  - Should it support emitting block-doc comments (`/** ... */`)?

sunfishcode:
> A couple of questions about the implementation:
> 
>     * Should `emit_docs` even be optional?

Yes, it seems reasonable to have an option.
      
>       * Should it default to `true`?

Yes!

>       * Should it support emitting block-doc comments (`/** ... */`)? (e.g. `.emit_docs(DocFormat::Block)` or something)

Only if someone really wants it; I wouldn't speculatively add it.

alexcrichton (approved):
This all looks great to me, thanks! Happy to merge when you feel it's ready (and I think it's ok to merge at any time even if the binary format doesn't encode comments yet)

lann:
Based on my last round of manual testing I _think_ I'm printing docs everywhere that `Resolve` is parsing them now.

PR #1169, opened 2023-08-17
WIT doc comments encoding

lann:
re: https://github.com/WebAssembly/component-model/issues/213

Example:

```wit
/// Hello, World
world hello {
  /// Inline interface
  export inline-iface: interface {
    /// Inner function
    inner-func: func()
  }
  
  /// World func import
  import world-func: func()

  /// World record
  record world-record {
    /// World record field
    rec-field: u8
  }
}

/// Top-level interface
interface top-iface {
  /// Interface func
  iface-func: func() 

  /// Human resource
  resource human {
    /// Method actor
    actor: func()
  }

  /// Unions are weird
  union weird {
    /// Cases are indexed
    u8,
  }
}
```

```json
{
  "worlds": {
    "hello": {
      "docs": "Hello, World\n",
      "interfaces": {
        "inline-iface": {
          "docs": "Inline interface\n",
          "funcs": {
            "inner-func": "Inner function\n"
          }
        }
      },
      "funcs": {
        "world-func": "World func import\n"
      },
      "types": {
        "world-record": {
          "docs": "World record\n",
          "items": {
            "rec-field": "World record field\n"
          }
        }
      }
    }
  },
  "interfaces": {
    "top-iface": {
      "docs": "Top-level interface\n",
      "funcs": {
        "iface-func": "Interface func\n",
        "[method]human.actor": "Method actor\n"
      },
      "types": {
        "human": {
          "docs": "Human resource\n"
        },
        "weird": {
          "docs": "Unions are weird\n",
          "items": {
            "0": "Cases are indexed\n"
          }
        }
      }
    }
  }
}
```


alexcrichton:
I [wrote up some thoughts](https://github.com/WebAssembly/component-model/issues/213#issuecomment-1684072335) about the usage of JSON here, but my main comment on the format itself is that I'd ideally like to future-proof this a bit in the face of future updates. For example if we change the format in the future I wouldn't want to break parsers accidentally from the past. In that sense I might propose a prefix byte for the JSON blob which is a version number, currently 0x00 or something like that.

Otherwise though would you be up for gaming out the decoding side of this as well? I think it'd be good to have that mostly fleshed out to prove out the idea here works well and can be reasonably processed not only in the encoding process but additionally in the decoding process.

alexcrichton:
Personally at least I think that so long as encoding/decoding works that the schema doesn't matter too too much at this point in time as I'm assuming we'll plan to change it. In that sense this all looks good to me and I think it's good to go with decoding

crates/wit-component/src/encoding/docs.rs
```diff
@@ -0,0 +1,360 @@
...
+                extract_items(&flags.flags, |item| (&item.name, &item.docs))
+            }
+            wit_parser::TypeDefKind::Variant(variant) => {
+                extract_items(&variant.cases, |item| (&item.name, &item.docs))
+            }
+            wit_parser::TypeDefKind::Enum(enum_) => {
+                extract_items(&enum_.cases, |item| (&item.name, &item.docs))
+            }
+            wit_parser::TypeDefKind::Union(_) => {
+                // FIXME: unions are (hopefully) going away
```
lann, line 292: https://github.com/WebAssembly/component-model/pull/237

lann:
Added decoding and a version byte on the custom section.

crates/wit-component/src/encoding/docs.rs
```diff
@@ -0,0 +1,361 @@
...
+        }
+        .encode(&mut raw_section);
+        Ok(raw_section)
+    }
+
+    /// Decode package docs from package-docs custom section content.
+    pub fn decode(data: &[u8]) -> Result<Self> {
+        let version = data.first();
+        if version != Some(&PACKAGE_DOCS_SECTION_VERSION) {
+            bail!("expected package-docs version 0, got {version:?}");
```
alexcrichton, line 90: `s/0/{PACKAGE_DOCS_SECTION_VERSION}/`

alexcrichton (approved):
This looks fantastic to me, thanks again for this! Is there anything else you'd like to add or is this good to go?

lann:
Should be good to go.

PR #1170, opened 2023-08-18
Support WIT package doc comments

### LINKED ISSUE #836: Attach doc-comments to a WIT file
At the moment, you can add doc-comments to any item in a WIT file and the text will appear in the generated code as doc-comments for your target language, however it doesn't look like you can document the WIT file itself.

The idea is that users could write the following `guest.wit` file:

```
//! Functionality exposed by the guest.

/// A function that is called when the WebAssembly module is loaded.
on-load: function()
```

(I'm borrowing the `//!` syntax from Rust but don't have any particular preference)

Generating import bindings with `wit-bindgen wasmtime --import guest.wit` might then look like this:

```rust
pub mod guest {
  //! Functionality exposed by the guest.

  pub struct Guest<T> { ... }

  impl<T> Guest<T> {
    /// A function that is called when the WebAssembly module is loaded.
    pub fn on_load(&self, ...)-> Result<(), wasmtime::Trap> { ... }
  }
}
```

---

Concretely, supporting doc-comments for the WIT file itself will probably involve:

- Decide on a syntax (e.g. `//! ...` or `/// ...` with a newline separating it from the first item in the file)
- Add a `docs: Docs` field to `wit_parser::Interface` and update the parser to populate it
- Update each of the code generators to emit the appropriate doc-comments

lann:
```wit
/// Documentation for an entire package
package wasi:documented
```

At most one file can have a package doc comment.

Fixes #836

### DIFF
--- a/crates/wit-parser/src/ast.rs
+++ b/crates/wit-parser/src/ast.rs
@@ -961,12 +961,16 @@ fn parse_docs<'a>(tokens: &mut Tokenizer<'a>) -> Result<Docs<'a>> {
             Token::Comment => {
                 let comment = tokens.get_span(span);
                 if comment.starts_with("///") || (comment.starts_with("/**") && comment != "/**/") {
-                    if started {
-                        docs.span.end = span.end;
-                    } else {
-                        docs.span = span;
+                    if !started {
+                        docs.span.start = span.start;
                         started = true;
                     }
+                    let trailing_ws = comment
+                        .bytes()
+                        .rev()
+                        .take_while(|ch| ch.is_ascii_whitespace())
+                        .count();
+                    docs.span.end = span.end - (trailing_ws as u32);
                     docs.docs.push(comment.into());
                 }
             }
--- a/crates/wit-parser/src/ast.rs
+++ b/crates/wit-parser/src/ast.rs
@@ -24,12 +24,15 @@ impl<'a> Ast<'a> {
     pub fn parse(lexer: &mut Tokenizer<'a>) -> Result<Self> {
         let mut items = Vec::new();
         let mut package_id = None;
+        let mut docs = parse_docs(lexer)?;
         if lexer.eat(Token::Package)? {
-            package_id = Some(PackageName::parse(lexer)?);
+            let package_docs = std::mem::take(&mut docs);
+            package_id = Some(PackageName::parse(lexer, package_docs)?);
+            docs = parse_docs(lexer)?;
         }
         while lexer.clone().next()?.is_some() {
-            let docs = parse_docs(lexer)?;
             items.push(AstItem::parse(lexer, docs)?);
+            docs = parse_docs(lexer)?;
         }
         Ok(Self { package_id, items })
     }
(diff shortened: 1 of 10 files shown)


crates/wit-parser/src/ast.rs
```diff
@@ -24,12 +24,14 @@ impl<'a> Ast<'a> {
     pub fn parse(lexer: &mut Tokenizer<'a>) -> Result<Self> {
         let mut items = Vec::new();
         let mut package_id = None;
+        let mut docs = parse_docs(lexer)?;
         if lexer.eat(Token::Package)? {
-            package_id = Some(PackageName::parse(lexer)?);
+            let docs = std::mem::take(&mut docs);
```
alexcrichton, line 29: I think this may mean:

```
/// this works
package foo:bar

/// this gets lost
interface foo {}
```

perhaps?

crates/wit-parser/src/ast/resolve.rs
```diff
@@ -129,6 +132,15 @@ impl<'a> Resolver<'a> {
                 }
             }
             self.package_name = Some(cur_name);
+
+            // At most one 'package' item can have doc comments.
+            let docs = self.docs(&cur.docs);
+            if docs.contents.is_some() {
+                if self.package_docs.contents.is_some() {
+                    bail!("found doc comments on multiple 'package' items");
```
alexcrichton, line 140: Could this use `Error` to provide a `Span` to attach this error to?

alexcrichton (commented):
Nice! Could this add a parse-fail test as well for multiple documented `package` items?

crates/wit-parser/src/ast/resolve.rs
lann: I don't have the comment span here; I guess I could use the package item span?

crates/wit-parser/src/ast.rs
lann: Ugh yes. I rewrote this logic twice already!

crates/wit-parser/src/ast/resolve.rs
alexcrichton: Yeah I think it'd be possible in theory to retain the span of documentation, but otherwise pointing at the `package` item I think is fine

lann: I'll give it a shot; it's just awkward because _all_ preceding comments get stored in `ast::Docs` and doc-comments aren't extracted until here in `Resolver::docs`.

alexcrichton (approved):
Thanks! I think tests are failing on CI though?

lann:
Failing on Windows:

```
Caused by:
    failed test: result is not as expected:Diff < left / right > :
     failed to parse package: tests/ui/parse-fail/multiple-package-docs
     
     Caused by:
         found doc comments on multiple 'package' items
              --> tests/ui/parse-fail/multiple-package-docs/b.wit:1:1
               |
             1 | /// Multiple package docs, B
    <          | ^----------------------------
    >          | ^-----------------------------
```

Something with newlines I guess?

lann:
git autocrlf strikes again

PR #1174, opened 2023-08-21
Really check rustfmt on all files in CI

alexcrichton:
I noticed that currently in #1146 the CI there is green but some files are unformatted which got me wondering why that was the case. Turns out that `cargo fmt` won't find files that are referenced through macro-expanded macros, meaning that most of `wast` isn't actually checked for formatting on CI. This commit fixes that.


pchickey (approved):
:upside_down_face:

PR #1176, opened 2023-08-22
[wit-component] support importing bare funcs in adapters

dicej:
This enables importing bare (i.e. no containing interface) functions in adapters -- not just the main module.  This is useful in particular when performing shared-everything linking.


alexcrichton (approved):
👍 nice and easy fix!

PR #1177, opened 2023-08-23
Add serialize to all the exported structs of wit-parser

Mossaka:
This PR adds `Serialize` implementation to all the exported structs of the `wit-parser` crate so that it could output JSON given a WIT package. 

This follows the discussion from https://github.com/bytecodealliance/wit-bindgen/issues/640

Since `wit-parser` uses `id-arena` for allocating graph objects, which doesn't support `serde`, I had to fork this project and added `Serialize` to `Id` and `Arena`. See here for my changes: https://github.com/fitzgen/id-arena/pull/14

## Usage:

wasm-tools component wit <path-to-wit> --json

## Example Output

Give the following WIT

```wit
package foo:foo

interface disambiguate {
  type t = u32
}

world foo {
  type a = u32

  type b = a

  export c: func(a: a) -> b
}
world bar {
  import disambiguate
  use disambiguate.{t}

  export foo: func() -> t
}
world the-test {
  import foo: func(a: a) -> b

  record a {
    x: u32,
  }

  variant b {
    c(a),
  }

  export bar: func(a: a) -> b
}
```

The corresponding JSON format is 
```Json
{
  "worlds": [
    {
      "name": "foo",
      "docs": {
        "contents": null
      },
      "imports": {
        "a": {
          "Type": "type-1"
        },
        "b": {
          "Type": "type-2"
        }
      },
      "exports": {
        "c": {
          "Function": {
            "docs": {
              "contents": null
            },
            "name": "c",
            "kind": "Freestanding",
            "params": [
              [
                "a",
                "type-1"
              ]
            ],
            "results": {
              "Anon": "type-2"
            }
          }
        }
      },
      "package": {
        "idx": 0
      }
    },
    {
      "name": "bar",
      "docs": {
        "contents": null
      },
      "imports": {
        "interface-0": {
          "Interface": "interface-0"
        },
        "t": {
          "Type": "type-3"
        }
      },
      "exports": {
        "foo": {
          "Function": {
            "docs": {
              "contents": null
            },
            "name": "foo",
            "kind": "Freestanding",
            "params": [],
            "results": {
              "Anon": "type-3"
            }
          }
        }
      },
      "package": {
        "idx": 0
      }
    },
    {
      "name": "the-test",
      "docs": {
        "contents": null
      },
      "imports": {
        "a": {
          "Type": "type-4"
        },
        "b": {
          "Type": "type-5"
        },
        "foo": {
          "Function": {
            "docs": {
              "contents": null
            },
            "name": "foo",
            "kind": "Freestanding",
            "params": [
              [
                "a",
                "type-4"
              ]
            ],
            "results": {
              "Anon": "type-5"
            }
          }
        }
      },
      "exports": {
        "bar": {
          "Function": {
            "docs": {
              "contents": null
            },
            "name": "bar",
            "kind": "Freestanding",
            "params": [
              [
                "a",
                "type-4"
              ]
            ],
            "results": {
              "Anon": "type-5"
            }
          }
        }
      },
      "package": {
        "idx": 0
      }
    }
  ],
  "interfaces": [
    {
      "name": "disambiguate",
      "docs": {
        "contents": null
      },
      "types": {
        "t": {
          "idx": 0
        }
      },
      "functions": {},
      "package": {
        "idx": 0
      }
    }
  ],
  "types": [
    {
      "docs": {
        "contents": null
      },
      "kind": {
        "Type": "U32"
      },
      "name": "t",
      "owner": {
        "Interface": {
          "idx": 0
        }
      }
    },
    {
      "docs": {
        "contents": null
      },
      "kind": {
        "Type": "U32"
      },
      "name": "a",
      "owner": {
        "World": {
          "idx": 0
        }
      }
    },
    {
      "docs": {
        "contents": null
      },
      "kind": {
        "Type": "type-1"
      },
      "name": "b",
      "owner": {
        "World": {
          "idx": 0
        }
      }
    },
    {
      "docs": {
        "contents": null
      },
      "kind": {
        "Type": "type-0"
      },
      "name": "t",
      "owner": {
        "World": {
          "idx": 1
        }
      }
    },
    {
      "docs": {
        "contents": null
      },
      "kind": {
        "Record": {
          "fields": [
            {
              "docs": {
                "contents": null
              },
              "name": "x",
              "ty": "U32"
            }
          ]
        }
      },
      "name": "a",
      "owner": {
        "World": {
          "idx": 2
        }
      }
    },
    {
      "docs": {
        "contents": null
      },
      "kind": {
        "Variant": {
          "cases": [
            {
              "docs": {
                "contents": null
              },
              "name": "c",
              "ty": "type-4"
            }
          ]
        }
      },
      "name": "b",
      "owner": {
        "World": {
          "idx": 2
        }
      }
    }
  ],
  "packages": [
    {
      "name": "foo:foo",
      "docs": {
        "contents": null
      },
      "interfaces": {
        "disambiguate": {
          "idx": 0
        }
      },
      "worlds": {
        "foo": {
          "idx": 0
        },
        "bar": {
          "idx": 1
        },
        "the-test": {
          "idx": 2
        }
      }
    }
  ],
  "package_names": {
    "foo:foo": {
      "idx": 0
    }
  }
}
```


alexcrichton:
Would it be possible to avoid the need for `id-arena` to be `Serialize`/`Deserialize`? That includes the `arena_id` in a number of places which is irrelevant for this representation. On one hand I realize that this PR amounts to "let's just dump the internal representation" but at the same time if we can I think it'd be best to do our best to hide the Rust-specific bits that aren't relevant elsewhere.

Mossaka:
> Would it be possible to avoid the need for id-arena to be Serialize/Deserialize?

I believe `arena_id` could be omitted from the output, but `idx` is a key for looking for types, interfaces, worlds and packages.

Mossaka:
> to hide the Rust-specific bits that aren't relevant elsewhere.

What do you consider are Rust-specific bits?

alexcrichton:
Yes `idx` is required but the arena bits are the Rust-specific bits I'm worried about. Other Rust-specific bits could perhaps be serialized `enum` representations but that depends on the output.

Mossaka:
Just updated the repo. Now the JSON output looks like


```JSON
{
  "worlds": [
    {
      "name": "foo",
      "docs": {
        "contents": null
      },
      "imports": {
        "a": {
          "type": 1
        },
        "b": {
          "type": 2
        }
      },
      "exports": {
        "c": {
          "function": {
            "docs": {
              "contents": null
            },
            "name": "c",
            "kind": "freestanding",
            "params": [
              {
                "name": "a",
                "type": 1
              }
            ],
            "results": [
              {
                "type": 2
              }
            ]
          }
        }
      },
      "package": 0
    },
    {
      "name": "bar",
      "docs": {
        "contents": null
      },
      "imports": {
        "interface-0": {
          "interface": 0
        },
        "t": {
          "type": 3
        }
      },
      "exports": {
        "foo": {
          "function": {
            "docs": {
              "contents": null
            },
            "name": "foo",
            "kind": "freestanding",
            "params": [],
            "results": [
              {
                "type": 3
              }
            ]
          }
        }
      },
      "package": 0
    },
    {
      "name": "the-test",
      "docs": {
        "contents": null
      },
      "imports": {
        "a": {
          "type": 4
        },
        "b": {
          "type": 5
        },
        "foo": {
          "function": {
            "docs": {
              "contents": null
            },
            "name": "foo",
            "kind": "freestanding",
            "params": [
              {
                "name": "a",
                "type": 4
              }
            ],
            "results": [
              {
                "type": 5
              }
            ]
          }
        }
      },
      "exports": {
        "bar": {
          "function": {
            "docs": {
              "contents": null
            },
            "name": "bar",
            "kind": "freestanding",
            "params": [
              {
                "name": "a",
                "type": 4
              }
            ],
            "results": [
              {
                "type": 5
              }
            ]
          }
        }
      },
      "package": 0
    }
  ],
  "interfaces": [
    {
      "name": "disambiguate",
      "docs": {
        "contents": null
      },
      "types": {
        "t": 0
      },
      "functions": {},
      "package": 0
    }
  ],
  "types": [
    {
      "docs": {
        "contents": null
      },
      "kind": {
        "type": "u32"
      },
      "name": "t",
      "owner": {
        "interface": 0
      }
    },
    {
      "docs": {
        "contents": null
      },
      "kind": {
        "type": "u32"
      },
      "name": "a",
      "owner": {
        "world": 0
      }
    },
    {
      "docs": {
        "contents": null
      },
      "kind": {
        "type": 1
      },
      "name": "b",
      "owner": {
        "world": 0
      }
    },
    {
      "docs": {
        "contents": null
      },
      "kind": {
        "type": 0
      },
      "name": "t",
      "owner": {
        "world": 1
      }
    },
    {
      "docs": {
        "contents": null
      },
      "kind": {
        "record": {
          "fields": [
            {
              "docs": {
                "contents": null
              },
              "name": "x",
              "type": "u32"
            }
          ]
        }
      },
      "name": "a",
      "owner": {
        "world": 2
      }
    },
    {
      "docs": {
        "contents": null
      },
      "kind": {
        "variant": {
          "cases": [
            {
              "docs": {
                "contents": null
              },
              "name": "c",
              "type": 4
            }
          ]
        }
      },
      "name": "b",
      "owner": {
        "world": 2
      }
    }
  ],
  "packages": [
    {
      "name": "foo:foo",
      "docs": {
        "contents": null
      },
      "interfaces": {
        "disambiguate": 0
      },
      "worlds": {
        "foo": 0,
        "bar": 1,
        "the-test": 2
      }
    }
  ]
}
```

crates/wit-parser/src/resolve.rs
```diff
@@ -1745,3 +1754,183 @@ impl<'a> MergeMap<'a> {
...
+        let resolve = Resolve {
+            packages,
+            interfaces,
+            types,
+            worlds,
+            package_names: Default::default(),
+        };
+        let s = serde_json::to_string(&resolve).unwrap();
+        let json_resolve = JSON_RESOLVE.replace(" ", "").replace("\n", "");
+        assert_eq!(s, json_resolve);
```
Mossaka, line 1933: @ydnar I've added a dummy `Resolve` to match the JSON output

crates/wit-parser/src/lib.rs
```diff
@@ -266,14 +285,24 @@ pub struct IncludeName {
...
 /// The key to the import/export maps of a world. Either a kebab-name or a
 /// unique interface.
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize)]
+#[serde(into = "String")]
 pub enum WorldKey {
     /// A kebab-name.
     Name(String),
     /// An interface which is assigned no kebab-name.
     Interface(InterfaceId),
```
ydnar, line 294: Shoot. Forgot to make this one a JSON number like the other enums.

cardoso:
This is great! Any chance this could evolve to export span information at some point? 

I had to pause https://github.com/bytecodealliance/vscode-wit/pull/16 , but with this + span info, the implementation of LSP for WIT would be much more straightforward especially while the language isn't stable.

I really like the approach used here: https://github.com/dtolnay/clang-ast

crates/wit-parser/src/resolve.rs
```diff
@@ -1745,3 +1754,183 @@ impl<'a> MergeMap<'a> {
...
+    use super::*;
+
+    const EMPTY_RESOLVE: &str =
+        r#"{ "worlds": [], "interfaces": [], "types": [], "packages": [] }"#;
+
+    const JSON_RESOLVE: &str = r#"{
+    "worlds": [
+        {
+        "name": "world1",
+        "docs": { "contents": null },
```
alexcrichton, line 1772: Could `docs` be skipped if the contents are null?

Cargo.toml
```diff
@@ -134,23 +134,7 @@ harness = false
 
 [features]
 # By default, all subcommands are built
-default = [
```
alexcrichton, line 137: Could the formatting changes in this file be undone?

crates/wit-parser/src/lib.rs
```diff
@@ -112,14 +121,21 @@ pub enum AstItem {
...
+#[derive(Debug, Clone, Hash, Eq, PartialEq, Ord, PartialOrd, Serialize)]
+#[serde(into = "String")]
 pub struct PackageName {
     /// A namespace such as `wasi` in `wasi:foo/bar`
     pub namespace: String,
     /// The kebab-name of this package, which is always specified.
     pub name: String,
     /// Optional major/minor version information.
-    pub version: Option<Version>,
+    pub version: Option<SerializableVersion>,
```
alexcrichton, line 132: Instead of using `SerializableVersion` could this use `serialize_with`?

crates/wit-parser/src/lib.rs
```diff
@@ -246,16 +262,19 @@ pub struct World {
...
+    #[serde(skip)]
     pub includes: Vec<WorldId>,
 
     /// All the included worlds names. Empty if this is fully resolved
+    #[serde(skip)]
     pub include_names: Vec<Vec<IncludeName>>,
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Serialize)]
```
alexcrichton, line 277: Does this need `Serialize` if it's `serde(skip)`'d above?

crates/wit-parser/src/lib.rs
```diff
@@ -266,14 +285,24 @@ pub struct IncludeName {
...
     Name(String),
     /// An interface which is assigned no kebab-name.
     Interface(InterfaceId),
 }
 
+impl From<WorldKey> for String {
+    fn from(key: WorldKey) -> String {
+        match key {
+            WorldKey::Name(name) => name,
+            WorldKey::Interface(id) => format!("interface-{}", id.index()),
```
alexcrichton, line 301: This may also be what @ydnar is referring to just above this, but I think it'd be good to separate these two ariants a it better than with an `interface-` prefix since that looks a lot like a kebab-name

crates/wit-parser/src/lib.rs
```diff
@@ -314,24 +346,27 @@ pub struct Interface {
...
+    #[serde(serialize_with = "serialize_optional_id")]
     pub package: Option<PackageId>,
 }
 
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq, Serialize)]
 pub struct TypeDef {
     pub docs: Docs,
     pub kind: TypeDefKind,
     pub name: Option<String>,
```
alexcrichton, line 364: Could this be `serde(skip)`'d if the name is `None`?

crates/wit-parser/src/lib.rs
```diff
@@ -314,24 +346,27 @@ pub struct Interface {
...
     pub owner: TypeOwner,
 }
 
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq, Serialize)]
+#[serde(rename_all = "lowercase")]
```
alexcrichton, line 369: Instead of `"lowercase"` could `"kebab-case"` be used for this and all other serialization keys?

crates/wit-parser/src/resolve.rs
alexcrichton: Thanks for the tests here! I'm a bit worried about the maintainability of this over time though as updating this test is going to be a bit of a pain. Alternatively though could the `tests/ui/*.wit` tests be all automatically serialized to JSON? Similar to how `*.result` files are generated for the `ui/parse-fail/*.wit` tests could the `*.json` tests be automatically generated for all the WIT successful tests? That way it'd be easy to see changes-at-a-glance and additionally auto-update test expectations when anything changes.

crates/wit-parser/src/resolve.rs
ydnar: Sure. Is there any plans to add additional fields to the `Docs` struct? Could it be an optional string?

crates/wit-parser/src/lib.rs
ydnar: Agreed. Since this is just an internal reference to an `Interface` I think this can be serialized to a JSON number.

crates/wit-parser/src/lib.rs
ydnar: Does BA have a style guide or doc somewhere on how to represent names and identifiers? e.g. when to use kebab-case vs camel vs snake / screaming snake?

(given most target languages don’t support kebab-case for identifiers)

crates/wit-parser/src/resolve.rs
alexcrichton: Hm I don't think this would require a separate PR? I'm hoping to change this output in this PR by dropping the `"docs": ...` if it's empty. I think that can be done with `#[serde(skip_if = "...")]` or a similar field attribute name.

crates/wit-parser/src/lib.rs
alexcrichton: Heh I don't think anything is anywhere near official to have these sort of guidelines at a BA level by any means. 

In any case though I figure kebab-case might be good to have since it matches the component model.

crates/wit-parser/src/resolve.rs
Mossaka: Sounds good. I will work on that (expecting delayed commits from me as I am attending WasmCon this week)

Mossaka: Just updated this PR to commit all JSON files corresponding to valid WIT files. The test runner is able to compare and emitted JSON from `resolve` to commited JSON files similar to how `.result` files work.

alexcrichton (approved):
Thanks! I realize that there's still a follow-up here or there, but if you'd prefer I'd be ok landing this first and can iterate on it in-tree.

ydnar:
@alexcrichton thanks! I have a couple commits to land to address feedback. Mind holding off merging until I land those?

alexcrichton:
No worries, sounds good! Let me know when you're ready

crates/wit-parser/src/lib.rs
ydnar: Looks like this is what we want, because `WorldKey` is used in `World` exports and imports `IndexMap<WorldKey, WorldItem>`.

crates/wit-parser/src/lib.rs
ydnar: Upon further investigation, I think we need to have a unique string-serializable key for the `IndexMap<WorldKey, WorldItem>` maps in `World`, given that interfaces may be unnamed:

```rs
    /// Optionally listed name of this interface.
    ///
    /// This is `None` for inline interfaces in worlds.
    pub name: Option<String>,
```

@alexcrichton are you OK with the kebab-case `interface-$N` form or is there another form you’d prefer?

crates/wit-parser/src/lib.rs
ydnar: This is done in https://github.com/ydnar/wasm-tools/commit/68b52f6ed5cbb7f9bdbe8830426bbabaa0fd6127

ydnar: @alexcrichton would reordering some of the fields in the `Resolve` or child structs be a problem?

I’d like to reorganize a little, moving `name` above `docs`, for instance, so the serialization is a bit more human-readable.

crates/wit-parser/src/lib.rs
ydnar: Turns out all of the fields are single words, so kebab and lowercase are identical. Revisit if/when this changes?

crates/wit-parser/src/resolve.rs
ydnar: Done.

crates/wit-parser/src/lib.rs
ydnar: According to the tests, this works with `semver::Version` without `SerializableVersion` at all. So I removed it in be17ced3d194c72c22c6b158e66dca52ad0d8e4e.

crates/wit-parser/src/lib.rs
ydnar: Done.

ydnar:
@alexcrichton @Mossaka is on holiday, so I can’t push commits to his repo (and therefore this PR).

I could open a new PR with this + my subsequent commits, if you want.

alexcrichton:
Sounds good to me!

ydnar:
@alexcrichton OK, #1203 is up!

alexcrichton:
Merged as part of https://github.com/bytecodealliance/wasm-tools/pull/1203, so closing. (thanks again!)

PR #1181, opened 2023-08-25
Require the `gc` proposal with usage of `(rec ...)`

alexcrichton:
This commit is a follow-up to #1134 to require the `gc` feature flag be activated to use the `(rec ...)` construct from the GC wasm proposal. This additionally updates the text parsing and text printing to consider the structure of what's parsed for encoding/printing rather than the size of the types inside. Even a `rec` group of a single type prints a `rec` group or encodes it now instead of silently switching it back to a single-type encoding.


alexcrichton:
CI is failing due to https://github.com/bytecodealliance/wasm-tools/pull/1178, nothing major though

pchickey (approved):

PR #1192, opened 2023-09-06
Preserve operands more aggressively in wasm-smith

### LINKED ISSUE #655: `wasm-smith`: Prefer writing values to memory instead of `drop`ing them
When we `drop` them, then the Wasm engine can do DCE and ignore lots of code.

If we have a non-empty memory, we can write to the first page. Maybe even do a bump allocation thing where first we write to offset 0, then 1, etc. This way an alias analysis won't trivially see through this stuff and make a bunch of expressions DCE'd again.

Could be a post-pass method on `Module` that rewrites `drop`s or it could be done in `CodeBuilder` directly.

alexcrichton:
Currently in wasm-smith the type of a block or function is selected ahead of time and then the code internally is chosen arbitrarily. This means that frequently the result type of the block/function is not the same as whatever happens to be on the operand stack. In this situation `wasm-smith` currently issues a `drop` instruction for all operands that don't match types and then generates an arbitrary value of the required type.

While this works for validation/compilation this has an unintended side effect where modules generated for differential execution end up dropping a lot of values on the floor meaning that divergent state may not actually be stored anywhere to pick up on. This commit instead introduces an alternative strategy where instead of unconditionally emitting a `drop` instruction generated modules will now attempt to combine together all the would-be-dropped values. This is done for now with a simple `add` instruction of the appropriate type to attempt to ensure that if any state diverges in execution it can be witnessed.

The state of dropped operands is stored in globals which are likely to be exported from a module so differential execution can witness differences between two globals.

Closes #655

### DIFF
--- a/crates/wasm-smith/src/core/code_builder.rs
+++ b/crates/wasm-smith/src/core/code_builder.rs
@@ -2,11 +2,12 @@ use super::{
     Elements, FuncType, GlobalInitExpr, Instruction, InstructionKind::*, InstructionKinds, Module,
     ValType,
 };
+use crate::unique_string;
 use arbitrary::{Result, Unstructured};
 use std::collections::{BTreeMap, BTreeSet};
 use std::convert::TryFrom;
 use std::rc::Rc;
-use wasm_encoder::{BlockType, MemArg, RefType};
+use wasm_encoder::{BlockType, ConstExpr, ExportKind, GlobalType, MemArg, RefType};
 mod no_traps;
 
 macro_rules! instructions {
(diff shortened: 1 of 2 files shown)


crates/wasm-smith/src/core/code_builder.rs
```diff
@@ -1080,23 +1124,128 @@ impl CodeBuilder<'_> {
...
+        // If the type of this operand isn't known, for example if it's relevant
+        // to unreachable code, then it can't be combined, so return `false`.
+        let ty = match ty {
+            Some(ty) => ty,
+            None => return false,
+        };
+
+        // Use the input stream to allow a small chance of dropping the value
+        // without combining it.
+        if u.ratio(1, 100).unwrap_or(false) {
```
fitzgen, line 1184: I think we can make this collection of functions return a result and use `?` for readability. Note that all the `Unstructured` methods do their best to avoid returning errors so the `unwrap_or`s aren't changing behavior here.

crates/wasm-smith/src/core/code_builder.rs
```diff
@@ -1080,23 +1124,128 @@ impl CodeBuilder<'_> {
...
+
+        // Depending on the type lookup or inject a global to place this value
+        // into.
+        let (global, combine) = match ty {
+            ValType::I32 => {
+                let global = *self.allocs.global_dropped_i32.get_or_insert_with(|| {
+                    self.allocs.new_globals.push((ty, ConstExpr::i32_const(0)));
+                    inc(&mut self.allocs.globals_cnt)
+                });
+                (global, Instruction::I32Add)
```
fitzgen, line 1196: I have an intuition that `xor` might be better than `add` for these, because with `add` we start losing bits of information when they overflow and disappear.

fitzgen (approved):
Very nice! Thanks!

PR #1201, opened 2023-09-14
Added the ability to add features when decoding Wasm with wit-component

silesmo:
Added the ability to add features when decoding Wasm with wit-component
This is something I needed to get the WASM Analyzer working with arbitrary Wasm files as some Wasm files had features enabled that wasn't supported with the default features used in wit-component.

### DIFF
--- a/crates/wasmparser/src/validator.rs
+++ b/crates/wasmparser/src/validator.rs
@@ -252,6 +252,32 @@ pub struct WasmFeatures {
 }
 
 impl WasmFeatures {
+    /// Returns [`WasmFeatures`] with all features enabled.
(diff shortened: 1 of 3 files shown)


silesmo:
@peterhuene

alexcrichton:
Thanks! For this use case though I think it'd be ok to unconditionally enable all the various features. This isn't a first-class validation step for the core wasm itself and ideally this wouldn't validate core wasm at all. Given how internal it is though I think it's ok to avoid the need for configuration here.

silesmo:
Updated so all features are always enabled @alexcrichton

alexcrichton:
Thanks!

PR #1207, opened 2023-09-15
Serialize TypeOwner::None to JSON null

ydnar:
This is a minor update to the JSON serialization introduced in #1203. It modifies the output so the `TypeOwner` value of `None` in Rust serializes to `null` in JSON.

### Before

```json
    {
      "kind": {
        "type": "u32"
      },
      "name": "example",
      "owner": "none"
    }
```

### After

```json
    {
      "kind": {
        "type": "u32"
      },
      "name": "example",
      "owner": null
    }
```

### DIFF
--- a/crates/wit-parser/src/lib.rs
+++ b/crates/wit-parser/src/lib.rs
@@ -19,7 +19,8 @@ mod live;
 pub use live::LiveTypes;
 mod serde_;
 use serde_::{
-    serialize_anon_result, serialize_id, serialize_id_map, serialize_none, serialize_optional_id, serialize_params,
+    serialize_anon_result, serialize_id, serialize_id_map, serialize_none, serialize_optional_id,
+    serialize_params,
 };
 
 /// Checks if the given string is a legal identifier in wit.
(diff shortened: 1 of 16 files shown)


alexcrichton (approved):
Thanks! If you'd like I think it'd be reasonble to skip the field entirely if the field is `None` too

ydnar:
A better solution might be to change the `owner` field in `TypeDef` to `Option<TypeOwner>`, but that would change the public API of this crate:

```
#[derive(Debug, Clone, PartialEq, Serialize)]
pub struct TypeDef {
    pub name: Option<String>,
    pub kind: TypeDefKind,
    pub owner: Option<TypeOwner>,
    #[serde(skip_serializing_if = "Docs::is_empty")]
    pub docs: Docs,
}
```

ydnar:
> Thanks! If you'd like I think it'd be reasonble to skip the field entirely if the field is `None` too

That would work too. The output would be more concise… see my previous comment on `Option<TypeOwner>` vs a None variant?

PR #1208, opened 2023-09-15
wit-parser: reorder fields for readability

ydnar:
This modifies the field order of WIT structures in the wit-parser crate:

1. The `name` field is moved to the first position in the handful of structs it wasn’t already the first field.
2. The `docs` field is moved to the last (public, serialized) position, since it is A) often empty/omitted, or B) when present, contains paragraphs of plaintext.

These two changes make the JSON serialized output of `Resolve` more consistent and readable. For example, see [random.wit](https://github.com/bytecodealliance/wasm-tools/pull/1208/files#diff-c51a45dab8b06f3fd18a76a24b5f641108d26fb8204f49e63ed15e147c515e83) and random.wit.json in this PR.

Question: does changing the order of fields in a Rust struct break any compatibility guarantee(s) to dependents?

### DIFF
--- a/crates/wit-parser/src/lib.rs
+++ b/crates/wit-parser/src/lib.rs
@@ -62,9 +62,6 @@ pub struct UnresolvedPackage {
     /// The namespace, name, and version information for this package.
     pub name: PackageName,
 
-    /// Doc comments for this package.
-    pub docs: Docs,
-
     /// All worlds from all documents within this package.
     ///
     /// Each world lists the document that it is from.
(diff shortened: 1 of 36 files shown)


ydnar:
The test failures are interesting. Is something converting backslashes to slashes?

alexcrichton:
Ah yeah the `normalize` is intended for error messages with paths to fix across Windows and Unix. I think though it should be ok to skip `normalize` for JSON tests which I think will help fix the failure here?

ydnar:
Makes sense. Can you see if that last commit is OK?

alexcrichton (approved):

PR #1210, opened 2023-09-19
`wasmparser`: Fix structural type matching

fitzgen:


crates/wasmparser/src/readers/core/types.rs
```diff
@@ -522,53 +509,42 @@ impl RefType {
...
-            (false, HeapType::None) => "(ref none)",
-            (false, HeapType::NoExtern) => "(ref noextern)",
-            (false, HeapType::NoFunc) => "(ref nofunc)",
-            (false, HeapType::Eq) => "(ref eq)",
-            (false, HeapType::Struct) => "(ref struct)",
-            (false, HeapType::Array) => "(ref array)",
-            (false, HeapType::I31) => "(ref i31)",
+            (true, HeapType::Func) => "funcref".into(),
+            (true, HeapType::Extern) => "externref".into(),
+            (true, HeapType::Indexed(i)) => format!("(ref null {i})").into(),
```
alexcrichton, line 516: IIRC this `wat()` representation was used in some perf-sensitive situations in validation, namely around it created bloat around error messages I think. I forget the exact context but that's why it was the way it was before. If you're ok leaving this as it was before I think that'd be best, otherwise though I can go look into what it was and see if it's still important

crates/wasmparser/src/validator/operators.rs
```diff
@@ -475,7 +478,7 @@ impl<'resources, R: WasmModuleResources> OperatorValidatorTemp<'_, 'resources, R
             if self.operands.len() == control.height {
                 let desc = match expected {
                     Some(ty) => ty_to_str(ty),
-                    None => "a type",
+                    None => "a type".into(),
```
alexcrichton, line 481: Ah I think this is where my perf-sensitive thinking is coming from. Given that failure is the slow path though I think it's reasonable to outline this and mark it `#[cold]`. (similar to `self.err_beyond_end` IIRC)

If you wouldn't mind double-checking that this is the only place though that'd be good I think

alexcrichton (approved):
Looks good to me 👍, but the stack overflow on CI may be on to something.

PR #1204, opened 2023-09-15
Switch GC binary enconding

yurydelendik:
Following the https://github.com/WebAssembly/gc/pull/372 lead.

## Quick summary of changes

... that I used to created this PR

### Types

nullref: 0x65->0x71
arrayref: 0x66->0x6a
structref: 0x67->0x6b
nullfuncref: 0x68->0x73
nullexternref: 0x69->0x72
i31ref: 0x6a->0x6c
ref $t: 0x6b->0x64
ref null $t: 0x6c->0x63
i16: 0x79->0x77
i8: 0x7a->0x78

### Instructions

ref.as_non_null: 0xd3->0xd4
br_on_null: 0xd4->0xd5
ref.eq: 0xd5->0xd3

struct.get: 0xfb03->0xfb02
struct.get_s: 0xfb04->0xfb03
struct.get_u: 0xfb05->0xfb04
struct.set: 0xfb06->0xfb05
struct.new: 0xfb07->0xfb00
struct.new_default: 0xfb08->0xfb01
array.fill: 0xfb0f->0xfb10
array.get: 0xfb13->0xfb0b
array.get_s: 0xfb14->0xfb0c
array.get_u: 0xfb15->0xfb0d
array.set: 0xfb16->0xfb0e
array.copy: 0xfb18->0xfb11
array.len: 0xfb19->0xfb0f
array.new_fixed: 0xfb1a->0xfb08
array.new: 0xfb1b->0xfb06
array.new_default: 0xfb1c->0xfb07
array.new_data: 0xfb1d->0xfb09
array.new_elem: 0xfb1f->0xfb0a
i31.new: 0xfb20->0xfb1c
i31.get_s: 0xfb21->0xfb1d
i31.get_u: 0xfb22->0xfb1e
ref.test: 0xfb40->0xfb14
ref.cast: 0xfb41->0xfb16
ref.test null: 0xfb48->0xfb15
ref.cast null: 0xfb49->0xfb17
br_on_cast: 0xfb4e->0xfb18
br_on_cast_fail: 0xfb4f->0xfb19
array.init_data: 0xfb54->0xfb12
array.init_elem: 0xfb55->0xfb13
extern.internalize: 0xfb70->0xfb1a
extern.externalize: 0xfb71->0xfb1b

Rest of the encoding was not affected.

I'm working on https://bugzilla.mozilla.org/show_bug.cgi?id=1845374 to verify these changes against it.

PR #1211, opened 2023-09-19
Fix subtype encoding in wasm-encoder

yurydelendik:
Missing piece in #1204
Screenshot 2023-09-19 at 3 29 25 PM

### DIFF
--- a/crates/wasm-encoder/src/core/types.rs
+++ b/crates/wasm-encoder/src/core/types.rs
@@ -336,7 +336,7 @@ impl TypeSection {
             None => vec![],
         };
         if ty.is_final {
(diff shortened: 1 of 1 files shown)


eqrion (approved):

alexcrichton:
Would you like a publish for this?

yurydelendik:
> Would you like a publish for this?

I'm running tests to see if it is important to us. 

It is not a priority atm. It can be done at next convenient moment.

alexcrichton:
Ok no worries! Publishes are mostly only tricky when there's lots of other changes to release (since everything is released as a unit for better or worse), and there's not much on `main` so this should be easy if needed. Just let me know

PR #1226, opened 2023-09-28
add support for exporting resources from adapters

### LINKED ISSUE #1200: wit-component: support resource exports from adapters
#1084 added support for resources to `wit-component`, but appears to only support exporting resources from the main module -- not from adapters.  Given that `wasm-tools component link` represents all dynamic libraries as "adapters", and `componentize-py` uses that to create components, I'd like to add support for exporting resources from adapters.

I'm planning to start working on a PR tomorrow unless there are objections.

dicej:
Previously, you could only export resources from the main module -- not from adapters. Given that wasm-tools component link represents all dynamic libraries as "adapters", and `componentize-py` uses that to create components, we'd like to support both.

Fixes #1200


alexcrichton (approved):
Thanks for this!

PR #1229, opened 2023-10-04
GC: wasmparser: re-enable some skipped GC test cases

imikushin:
Also, add an `assert_invalid` case for trying to subtype a standalone type defined without `(sub )`


fitzgen (approved):
Thanks!

PR #1232, opened 2023-10-04
feat(wit-component): expose wit parsing and wasm detection functions

vados-cosmonic:
WIT parsing and wasm detection functions were used inside wasm-tools but not available to downstream consumers.

This commit refactors the code slightly and puts it in wit-component, and makes two functions available (wit_component::parse_wit_from_path, wit_component::is_wasm_binary_or_wat) for downstream consumers of wit-component to use, along with some basic tests.

## Worth discussing

Some things that are likely issues which I'd like to note a head of time for discussion

### renaming/exposing `is_wasm_binary_or_wat`

Formerly named `is_wasm` (I'm also OK going back to this naming!), I chose to move & expose this function too which wasn't *strictly* necessary -- woud like thoughts on this.

I think this fucntionality would be good as a reusable chunk as well, even if the test is not absolute.


alexcrichton (approved):
Thanks for this! I'm happy to stick with the names you chose as well.

For the test failure I believe this is stemming from the usage of `fs::canonicalize` in the tests. If you remove that though I think the tests should pass on wasm.

vados-cosmonic:
@alexcrichton thanks for taking a look, and thanks for the hint -- I just removed `fs::canonicalize` so hopefully all green CI from here.

PR #1233, opened 2023-10-04
refactor: expose fn for embedding component metadata

### LINKED ISSUE #1231: wit-component: Refactor component metadata embedding to enable downstream re-use
## Context

In wanting to use some component metadata embedding, I found that there wasn't an easy way to perform it at the code level, yet. Ideally [the code that powers `wasm-tools component embed`](https://github.com/bytecodealliance/wasm-tools/blob/03d21b8d9f0dd29441c5de4d6a2fc1505a9fd0d5/src/bin/wasm-tools/component.rs#L207) would be callable from downstream consumers of `wit-component`.

After a bit of [discussion in Zulip](https://bytecodealliance.zulipchat.com/#narrow/stream/217126-wasmtime/topic/parse_wit.20as.20an.20public.20fn.20in.20wit-parser/near/394493625), it seems like this is best as a small refactor of the code in `wit-component`.

## Proposal

Refactoring the code that powers `wasm-tools component embed` to be externally callable *and* used from the CLI.

## Alternatives

We could leave this out, and have people write their own usages of system tools.

## Additional links

vados-cosmonic:
The logic for embedding component metadata as a custom section was used only in the course of `wasm-tools component embed`, which made it difficult to use from outside wasm-tools.

This commit refeactors the code that actually writes out the metadata to a function that can be re-used and is exposed by wit-component.

Resolves #1231

### DIFF
--- a/crates/wit-component/src/lib.rs
+++ b/crates/wit-component/src/lib.rs
@@ -2,11 +2,12 @@
 
 #![deny(missing_docs)]
 
-use anyhow::{bail, Context, Result};
 use std::str::FromStr;
-use std::{fmt::Display, path::Path};
-use wasm_encoder::CanonicalOption;
-use wit_parser::{PackageId, Resolve, UnresolvedPackage};
+use std::{borrow::Cow, fmt::Display, path::Path};
+
+use anyhow::{bail, Context, Result};
+use wasm_encoder::{CanonicalOption, Encode, Section};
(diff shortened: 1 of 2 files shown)


alexcrichton (approved):

vados-cosmonic:
Thanks for the speedy review @alexcrichton :bow:

PR #1235, opened 2023-10-06
Add options to control the functions used to allocate memory

yowl:
This PR adds 2 options to control how realloc and the stack allocation is implemented.  The options if supplied will make these functions use `memory.grow` rather than the `cabi_realloc` (if supplied) of the component.

My first PR here, and pretty new to Rust, so feedback welcomed!

Discussed over in zulip.

Thanks,

### DIFF
--- a/crates/wit-component/src/gc.rs
+++ b/crates/wit-component/src/gc.rs
@@ -707,7 +707,6 @@ impl<'a> Module<'a> {
             num_func_imports += 1;
         }
 
-        let mut realloc_via_memory_grow_index = None;
         for (i, func) in local {
             map.funcs.push(i);
             let ty = map.types.remap(func.ty);
@@ -716,7 +715,6 @@ impl<'a> Module<'a> {
                     // The adapter is importing `cabi_realloc` from the main module, but the main module isn't
                     // exporting it.  In this case, we need to define a local function it can call instead.
                     realloc_index = Some(num_func_imports + funcs.len());
-                    realloc_via_memory_grow_index = realloc_index;
                     funcs.function(ty);
                     code.function(&realloc_via_memory_grow());
                 }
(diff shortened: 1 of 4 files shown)


src/bin/wasm-tools/component.rs
```diff
@@ -107,6 +107,14 @@ pub struct NewOpts {
     /// Print the output in the WebAssembly text format instead of binary.
     #[clap(long, short = 't')]
     wat: bool,
+
+    /// Use memory.grow to realloc memory.
+    #[clap(long, short = 'r')]
+    realloc_via_memory_grow: bool,
+
+    /// Use memory.grow for stack allocation.
+    #[clap(long, short = 's')]
```
alexcrichton, line 116: I personally try to reserve `short` options for pretty common pieces of functionality, so I think it's ok to stick with just `long` for now and add `short` later if this ends up being more common.

src/bin/wasm-tools/component.rs
```diff
@@ -107,6 +107,14 @@ pub struct NewOpts {
...
+    alloc_stack_via_memory_grow: bool,
```
alexcrichton, line 117: Also, is it worth having a separate knob for this than the one above? I'm not sure when the state would want to be allocated with `cabi_realloc` and the stack with `memory.grow` specifically, so the above knob may be enough perhaps?

yowl: That works for me, removed

src/bin/wasm-tools/component.rs
yowl: removed, thanks

crates/wit-component/src/gc.rs
```diff
@@ -707,6 +707,7 @@ impl<'a> Module<'a> {
             num_func_imports += 1;
         }
 
+        let mut realloc_via_memory_grow_index = None;
```
alexcrichton, line 710: I think that the changes might no longer be required in this file? Or are they perhaps related to a bugfix otherwise?

yowl: Thanks, you are right, reverted.

alexcrichton (approved):
Thanks!

PR #1239, opened 2023-10-06
Make wit-component's `wat` dep optional

alexcrichton:
The `wast` crate is a little hefty and many uses of the `wit-component` crate don't need it, so make it optional and enable it from the CLI where it's used.

### DIFF
--- a/crates/wit-component/src/lib.rs
+++ b/crates/wit-component/src/lib.rs
@@ -3,11 +3,11 @@
 #![deny(missing_docs)]
 
 use std::str::FromStr;
-use std::{borrow::Cow, fmt::Display, path::Path};
+use std::{borrow::Cow, fmt::Display};
 
-use anyhow::{bail, Context, Result};
+use anyhow::{bail, Result};
 use wasm_encoder::{CanonicalOption, Encode, Section};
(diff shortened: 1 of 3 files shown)


crates/wit-component/src/lib.rs
```diff
@@ -85,7 +85,13 @@ pub(crate) fn base_producers() -> wasm_metadata::Producers {
 
 /// Parse a WIT file from a path that represents a top level 'wit' directory,
 /// normally containing a 'deps' folder.
-pub fn parse_wit_from_path(path: impl AsRef<Path>) -> Result<(Resolve, PackageId)> {
+#[cfg(feature = "wat")]
```
peterhuene, line 88: It feels like this function should continue to be defined without the `wat` feature and instead `is_wasm_binary_or_wat` should _only_ check for the binary preamble when the `wat` feature is disabled?

peterhuene: Oh and we'd obviously skip the `wat::parse_bytes` call.

Not worth the complexity?

alexcrichton: Nah makes sense to me!

peterhuene (approved):

PR #1245, opened 2023-10-12
`wasmparser`: Fix `From<wasmparser::StructType> for StructType`

fitzgen:
`Box<[T]>` doesn't implement `into_iter`, it is actually just calling `&[T]`'s version of `into_iter` which doesn't actually give ownership.

Also, test `wasm-encoder` with this feature enabled in CI.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -43,6 +43,7 @@ jobs:
       shell: bash
     - run: cargo test --locked --all
     - run: cargo test --locked -p wasmparser --benches
+    - run: cargo test --locked -p wasm-encoder --all-features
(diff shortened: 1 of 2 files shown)


alexcrichton (approved):

peterhuene:
Oops!

Once I realized that it was a `Box<[T]>` and not `Vec<T>`, I originally suggested using `into_vec().into_iter()` which should be a super cheap way of avoiding the clone of the elements, but then you pushed up my original suggestion and it passed CI so I just resolved my comment.

I failed to realize CI wasn't checking the new feature flag 😞. Sorry about that!

fitzgen:
Not your fault! My b for not testing the change and getting CI exercising this stuff in the original PR.

PR #1246, opened 2023-10-12
Enable threads, multi-memory, and relaxed-simd by default

alexcrichton:
This commit reflects the stage 4 advancement of these proposals and updates the default feature set to include these three proposals. Each proposal can still be individually be disabled if desired.


fitzgen (approved):
LGTM, can we land this after https://github.com/bytecodealliance/wasm-tools/pull/1247 tho?

alexcrichton:
👍

PR #1247, opened 2023-10-12
Bump `wasm-encoder` to 0.34.0 and `wasmparser` to 0.114.0

fitzgen:


alexcrichton (approved):
Thanks!

tamird:
I think this is the cause of https://github.com/bytecodealliance/wit-bindgen/issues/697 - two crates got minor version bumps, but all the other crates got patch bumps.

Then today `wit-component` got a patch release, but bumped its dependency on `wit-parser` by a minor version. That means that any dependent that uses both (even indirectly) is hosed until the rest of the ecosystem cuts new releases.

Another way to avoid this would be to not expose foreign types in the APIs, but that ship has likely sailed.

alexcrichton:
Yes thanks for the heads up, and this is my mistake as I didn't catch this. I posted https://github.com/bytecodealliance/wasm-tools/pull/1250 to fixup the versions and additionally future-proof this (hopefully). I'll yank the broken versions once those are published to avoid future breakage.

tamird:
Thanks.

> I'll yank the broken versions once those are published to avoid future breakage.

Has this happened?

alexcrichton:
I've just now done the yanks, yes.

PR #1251, opened 2023-10-13
Extend the set of exported type ids with defined resources

elliottt:
Currently, validation of the following component fails, as the resource `r` is not viewed as exported when validating the export of `$types'`. (More specifically, it's the use of `r` in the method `r.foo` that causes the error, as `r` is not viewed as exported at that point.)

```lisp
(component
  (type $types (component
    (type $types' (instance
      (export $r "r" (type (sub resource)))
      (export "[method]r.foo" (func (param "self" (borrow $r))))
    ))
    (export (interface "local:demo/types") (type (eq $types')))
  ))
  (export "types" (type $types))
)
```

This PR addresses that by augmenting the set of named types when processing an `ComponentInstance` export, extending the set with the explicitly exported resources in that instance.

An alternate approach would be to extend the special handling of instances in `validate_and_register_named_types` to the `ComponentEntityType::Type` case as well, which would be a more targeted change than the one this PR currently proposes.


crates/wasmparser/src/validator/types.rs
```diff
@@ -725,6 +725,20 @@ pub struct ComponentInstanceType {
...
+impl ComponentInstanceType {
+    /// [`TypeId`]s for all resources defined by this instance.
+    pub fn explicit_resources(&self) -> impl Iterator<Item = TypeId> + '_ {
+        self.explicit_resources.iter().map(|(res, path)| {
+            println!("explicit: {res:?}, {path:?}");
+            assert_eq!(path.len(), 1);
+            TypeId {
+                index: path[0] as u32,
+                unique_id: res.contextually_unique_id,
+            }
```
elliottt, line 737: This is almost certainly incorrect, @alexcrichton what's the right way to go from a `ResourceId` to a `TypeId`?

crates/wasmparser/src/validator/types.rs
```diff
@@ -725,6 +725,19 @@ pub struct ComponentInstanceType {
...
 
+impl ComponentInstanceType {
+    /// [`TypeId`]s for all resources defined by this instance.
+    pub fn explicit_resources(&self) -> impl Iterator<Item = TypeId> + '_ {
+        self.explicit_resources.iter().map(|(res, path)| {
+            assert_eq!(path.len(), 1);
+            TypeId {
+                index: path[0] as u32,
+                unique_id: res.contextually_unique_id,
+            }
```
elliottt, line 736: This is almost certainly incorrect, @alexcrichton what's the right way to go from a `ResourceId` to a `TypeId`?

crates/wasmparser/src/validator/component.rs
```diff
@@ -711,19 +711,31 @@ impl ComponentState {
...
-                    | ComponentEntityType::Type { created: id, .. }
-                    | ComponentEntityType::Instance(id)
-                    | ComponentEntityType::Component(id) => *id,
-                    ComponentEntityType::Value(ComponentValType::Primitive(_)) => return true,
+            // Instances must recursively have all referenced types named. Resources defined by
+            // this interface are considered defined
+            Type::ComponentInstance(ty) => {
+                let mut with_resources = HashSet::new();
+
+                let set = if ty.explicit_resources.is_empty() {
```
alexcrichton, line 719: While a bit subtle, I think that this check may want to look at the `defined_resources` list rather than look at the `explicit_resources` map. The reason for that is that concrete instances will have an empty list there but the instance types, which this PR is fixing, may not have an empty `defined_resources` list. On the other hand though it's frequent for any kind of instance to have an `explicit_resources` array set, so this I don't think is the right branch in the implementation.

The reason for this is that the set of exported resources should be only judiciously added to and for normal instances everywhere the previous additions to the set should already be taken care of. In effect by adding all `explicit_resources` it's forcing this check to basically pass.

So, in other words, I think this check should be refactored to look more like:

```rust
                let set = if ty.defined_resources.is_empty() {
                    set
                } else {
                    with_resources = set.clone();
                    for resource in ty.defined_resources.iter() {
                        with_resources.insert(ty.type_id_of(resource));
                    }
                    &with_resources
                };
```

or something along those lines where `type_id_of` is the body of the loop in the `explicit_resources` method you added. Either that or rename the added `explicit_resources` to `defined_resources` and have the same loop you have today returning an iterator (actually after typing this out I prefer the latter style here which you already have)

alexcrichton (commented):
Thanks for this! I believe that this is the correct approach instead of `validate_and_register_named_types`. The distinguishing feature there is that when an instance type is exported it shouldn't mutate `self` and add more exported resources since the resources are only considered exported for that one singular type itself and nothing else.

crates/wasmparser/src/validator/component.rs
elliottt: Wouldn't only looking at `defined_resources` mean that the test you mentioned above wouldn't work?
```
(component
  (type (export "x") (instance
    (export $i "i" (instance
      (export "x" (type (sub resource)))
    ))

    (alias export $i "x" (type $x))
    (export "f" (func (param "x" (borrow $x))))
  ))
)
```

tests/cli/validate-exported-instance.wat
```diff
@@ -0,0 +1,30 @@
+;; RUN: validate % --features component-model
+
+(component
```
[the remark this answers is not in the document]
elliottt: I've moved the tests into `tests/local/component-model`, and added this one.

elliottt:
After some discussion with @alexcrichton I'm going to close this. We can work around this bug with the upcoming wit encoding changes, and there are some dragons lurking in how we convert resource ids back to type ids.

PR #1252, opened 2023-10-13
Change the encoding of wit definitions

elliottt:
Address https://github.com/WebAssembly/component-model/pull/248 by implementing support for the new encoding proposal. Additionally, this PR maintains backwards compatibility with decoding the previous encoding, by detecting when the newer encoding was used.

The rollout of the v2 encoding scheme will follow roughly the same plan as for semicolons in wit: there's a new environment variable introduced that controls the encoding scheme, `WIT_COMPONENT_ENCODING_V2` that can be used to control when we encode in the new format, with the default behavior being what's done on main. After tooling has caught up to expecting the new format we can switch the default behavior to be the new format, and then completely remove support for v1 after we're confident nothing depends on it anymore.


crates/wit-component/tests/interfaces/type-alias.wat
```diff
@@ -35,10 +40,10 @@
           (export (;1;) (interface "foo:foo/foo") (instance (type 1)))
         )
       )
-      (export (;0;) (interface "foo:foo/my-world") (component (type 1)))
+      (export (;1;) (interface "foo:foo/my-world") (type (eq 0)))
```
alexcrichton, line 43: Could this stay as `(component (type ...))` instead of a `(type (eq ..))`? Basically mirror the instance change for interfaces onto worlds too

crates/wit-component/tests/interfaces/upstream-deps-same-name/foo.wit.print
```diff
@@ -1,7 +1,8 @@
 package foo:foo
 
 interface a {
-  use foo:a/the-name.{ty}
-  use foo:b/the-name.{ty as ty2}
+  type ty = u8
+
+  type ty2 = u8
```
alexcrichton, line 6: I think this may indicate a bug? (losing that these are `use` and instead generating new types)

alexcrichton (commented):
All looks reasonable to me! There's some changes in tests though where `*.wit.print` shouldn't change but some type definitions are being duplicated through `use`

elliottt:
@alexcrichton I've switched the fuzzer over to using the V2 format, and also modified it to never produce an empty package.

alexcrichton (approved):

PR #1257, opened 2023-10-19
wasmparser: Use type-specific identifiers pervasively

fitzgen:
This commit changes `wasmparser` from the uni-typed `TypeId` identifier to a myriad of type-specific identifiers, such as `ComponentInstanceTypeId` and `ComponentDefinedTypeId`. There are also identifiers for any kind of type within some group, such as `ComponentAnyTypeId` for any kind of component type and `AnyTypeId` for any kind of type, component or core.

The `wasmparser::Types` arena can be indexed by any of these leaf type identifiers, which all implement the `TypeIdentifier` trait. The resulting indexed data is the particular type, e.g. a `ComponentFuncType`, rather than the uni-typed `wasmparser::Type`.

There was a lot of ocean boiling involved here. Apologies in advance for the pain of this upgrade! That said, it should provide type safety benefits moving forward and sets the stage for further identifier refactors for core types that are necessary for Wasm GC.

If you are upgrading and were using `TypeId` before, and don't want to spend time gaining type safety on your type lookups, then just use `AnyTypeId` instead of the old `TypeId`. They are morally identical. Alternatively, if you know a particular `TypeId` was always a component function, for example, and you always did `types[id].unwrap_component_func()`, then you can switch to using `ComponentFuncTypeId`.

### DIFF
--- a/crates/wasmparser/src/readers/core/types.rs
+++ b/crates/wasmparser/src/readers/core/types.rs
@@ -760,6 +760,16 @@ pub enum StructuralType {
     Struct(StructType),
 }
 
+impl StructuralType {
+    /// Unwrap a `FuncType` or panic.
+    pub fn unwrap_func(&self) -> &FuncType {
+        match self {
+            Self::Func(f) => f,
+            _ => panic!("not a func"),
+        }
+    }
+}
+
 /// Represents a subtype of possible other types in a WebAssembly module.
 #[derive(Debug, Clone)]
 pub struct SubType {
@@ -769,6 +779,13 @@ pub struct SubType {
     pub supertype_idx: Option<u32>,
     /// The structural type of the subtype.
     pub structural_type: StructuralType,
+}
+
+impl SubType {
+    /// Unwrap a `FuncType` or panic.
+    pub fn unwrap_func(&self) -> &FuncType {
+        self.structural_type.unwrap_func()
+    }
 }
 
 /// Represents a recursive type group in a WebAssembly module.
--- a/crates/wasmparser/src/validator/types.rs
+++ b/crates/wasmparser/src/validator/types.rs
@@ -6,9 +6,8 @@ use super::{
 };
 use crate::validator::names::KebabString;
 use crate::{
-    ArrayType, BinaryReaderError, Export, ExternalKind, FuncType, GlobalType, Import, MemoryType,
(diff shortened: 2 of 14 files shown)


fitzgen:
FWIW: 3.5 million fuzz iterations and no problems discovered thus far.

crates/wasmparser/src/readers/core/types.rs
```diff
@@ -771,6 +781,13 @@ pub struct SubType {
     pub structural_type: StructuralType,
 }
 
+impl SubType {
+    /// Unwrap a `FuncType` or panic.
+    pub fn unwrap_func(&self) -> &FuncType {
+        self.structural_type.unwrap_func()
+    }
```
alexcrichton, line 788: Should this perhaps also check that `is_final` is true or false? (I forget which it should check for)

crates/wasmparser/src/validator/types.rs
```diff
@@ -140,58 +139,651 @@ fn push_primitive_wasm_types(ty: &PrimitiveValType, lowered_types: &mut LoweredT
...
+///
+/// This is the data that can be retreived by indexing with the associated
+/// [`TypeIdentifier`].
+pub trait TypeData: TypeDataImpl + std::fmt::Debug {
+    /// The identifier for this type data.
+    type Id: TypeIdentifier<Data = Self>;
+}
+
+/// A type that can be aliased in the component model.
+pub trait Aliasable: AliasableImpl {}
```
alexcrichton, line 164: Wasmparser has a breaking major version bump all the time, so I think it's fine to skip the `sealed` bits perhaps and go straight to just making everything `pub`. If we change it then it's not much different than any other breaking change in `wasmparser` sort of

alexcrichton (approved):

crates/wasmparser/src/readers/core/types.rs
fitzgen: The default is `final == true` but there is nothing stopping one from defining a non-final function type, and I'd imagine this would still want to return the underlying function type.

crates/wasmparser/src/validator/types.rs
fitzgen: I guess I can make it `pub` but `doc(hidden)` at least so that it doesn't clutter the API docs and confuse people.

PR #1261, opened 2023-10-24
implement resource support for `wasm-compose`

dicej:
This aims to support the specific case of passing handles to imported resource types across component boundaries.  ~~It does not include any support for guest-exported resources (as would be required for most virtualization use cases).~~ Update: guest-exported resources can now be connected to imports, and I've updated `wasi-virt` to use that, with all tests passing.

I've updated the middleware example to use resources to demonstrate the concept. To make it more realistic, we could consider using real `wasi-http` request and response resources.

TODO items:

~~- Add tests~~

~~- Add missing docs~~


crates/wasm-compose/example/middleware/Cargo.toml
```diff
@@ -6,7 +6,7 @@ publish = false
 
 [dependencies]
 flate2 = "1.0.24"
-cargo-component-bindings = { git = "https://github.com/bytecodealliance/cargo-component" }
+cargo-component-bindings = { git = "https://github.com/dicej/cargo-component", branch = "update-deps" }
```
peterhuene, line 9: We can probably set this to `0.3.1` now (which I'm about to publish).

crates/wasm-compose/example/server/config.yml
```diff
@@ -4,4 +4,5 @@ search-paths:
 instantiations:
   $input:
     arguments:
-      example:service/handler@0.1.0: svc
+      # TODO: add package version (but note that `wit-bindgen` 0.13.0 doesn't seem to handle it correctly)
```
peterhuene, line 7: Was the error relating to this coming from building the components? I think https://github.com/bytecodealliance/cargo-component/pull/164 might be part of the problem, which should be addressed with `0.3.1` of `cargo-component-bindings`.

crates/wasm-compose/example/service.wit
```diff
@@ -1,19 +1,26 @@
-package example:service@0.1.0
+// TODO: add package version (but note that `wit-bindgen` 0.13.0 doesn't seem to handle it correctly)
```
peterhuene, line 1: Same comment as above.

peterhuene (commented):
This is looking good so far! Just a few minor comments.

Regarding guest-exported resources, what's the plan to get that implemented for the wasi-virt use case?

crates/wasm-compose/example/server/config.yml
dicej: I didn't dig into it; I'll try again with the new release.

dicej:
FYI, we ran into some bugs with a more complex example involving `wasi-http`.  I'll push updates once I've got them straightened out.

morenol:
Thanks for this PR!

dicej:
FYI, I won't have much time to work on this PR in the next few weeks since I've got other things to focus on.  Hoping to come back to it starting mid November.  Which is to say: feel free to jump in and help if this is blocking you :)

dicej:
> Regarding guest-exported resources, what's the plan to get that implemented for the wasi-virt use case?

Sorry, just noticed your question.  I don't have an immediate plan for that.  I agree it's a super important case, but not as urgent for me as guest-imported resources at the moment.  I'd love to help out with it when I'm a bit less busy.

crates/wasm-compose/example/server/config.yml
dicej: I've updated to `cargo-component-bindings` 0.4.0 and all is well.

dicej:
FYI, I've started to work on resource exports and virtualization-style composition today.  Will push updates as I make progress.

guybedford (approved):
This has now been tested to work in WASI-Virt's upgrade path, and is passing those integration tests correctly.

dicej:
I've added docs and tests, squashed, and rebased.  Here's squashed commit message, for reference:

> The basic approach is as follows:
> 
> - Start with an empty resource remapping (i.e. no given pair of distinct
>   `ResourceId`s are considered equivalent)
> 
> - For instance import and export pair we're considering connecting to each
>   other, match up any resource types by name and optimistically produce a
>   remapping that sets identically-named pairs equal to each other, provided that
>   remapping does not contradict any previous remappings.
> 
> - If the above fails (either because a consistent remapping could not be created
>   or because the instances were incompatible for other reasons), throw away the
>   changes and reject the connection (which may mean giving up on the whole
>   composition if we can't do what the user requested).  Otherwise, store the
>   remapping changes and accept the connection.
> 
> - Once all such connections have been made, update the remapping to set all
>   remaining, unconnected, identically-named resource imports in
>   identically-named instances equal to each other.
> 
> - Finally, encode the output component using the remapping produced above.
> 
> Note that I'm using `RefCell` for interior mutability since the places I needed
> to update the remapping only had a shared reference to the `CompsitionGraph`,
> and refactoring them to provide unique references felt like overkill (and I've
> heard a major refactor of `wasm-compose` is planned anyway, so perhaps this can
> be revisited then).
> 
> I've brought in `im-rc` as a dependency since it makes rolling back tentative
> changes trivial.  There may be more efficient approaches if that's a concern.
> 
> Per Alex's advice, I've changed a few types and traits in `wasmparser::types`
> from `pub(crate)` to `pub` in order to use them in the algorithm described
> above.

crates/wasmparser/src/validator/types.rs
```diff
@@ -504,6 +504,11 @@ impl Aliasable for AliasableResourceId {
 }
 
 impl AliasableResourceId {
+    /// Create a new instance with the specified resource and alias IDs.
+    pub fn new(id: ResourceId, alias_id: u32) -> Self {
+        Self { id, alias_id }
+    }
```
alexcrichton, line 510: If possible I'd prefer to avoid exporting this since the `alias_id` bits are intended to be internal implementation details which aren't exposed. 

Where this is used it looks like it's just swapping out the `ResourceId` bits, so instead of a `new` method could a method such as `fn with_resource_id(&self, id: ResourceId) -> AliasableResourceId` be added instead?

crates/wasmparser/src/validator/types.rs
```diff
@@ -3462,6 +3491,16 @@ where
...
 impl Remapping {
+    /// Add a mapping from the specified old resource ID to the new resource ID
+    pub fn add(&mut self, old: ResourceId, new: ResourceId) {
+        self.resources.insert(old, new);
+    }
+
+    /// Clear any type remappings
+    pub fn reset(&mut self) {
+        self.types.clear()
+    }
```
alexcrichton, line 3502: This feels a bit odd in that it only clears out an internal cache as opposed to additionally clearing out the resources mapping. Could this grow a new name to emphasize it's not resetting the entire state but only part of it?

crates/wasm-compose/src/encoding.rs
```diff
@@ -830,7 +882,9 @@ impl<'a> TypeEncoder<'a> {
                 .cur
                 .type_defs
                 .insert(key, state.cur.encodable.type_count());
-            assert!(prev.is_none());
+            // Resource aliases will already have been added to the table, but
+            // other types should not have been:
+            assert!(prev.is_none() || matches!(id, ComponentAnyTypeId::Resource(_)));
```
alexcrichton, line 887: Could this assert that `prev == Some(what_was_inserted)`? 

I also vaguely remember talking about this and somewhat concluding that this is the assertion to have. That being said reading this again I can't immediately determine why it's ok to replace a previous value in the map for resources, so if it's intended could you expand the comment here with why that's ok?

crates/wasm-compose/src/graph.rs
```diff
@@ -737,17 +911,38 @@ impl<'a> CompositionGraph<'a> {
...
+
+                let mut a_ty = export_ty;
+                context.a.remap_component_entity(&mut a_ty, remapping);
+                remapping.reset();
+
+                let mut b_ty = import_ty;
+                context.b.remap_component_entity(&mut b_ty, remapping);
+                remapping.reset();
+
+                if context.component_entity_type(&a_ty, &b_ty, 0).is_err() {
```
alexcrichton, line 936: This same sequence of code shows up in 2 other places I've seen so far now, could it instead be refactored into a shared helper used by this crate?

crates/wasm-compose/src/graph.rs
```diff
@@ -457,6 +572,65 @@ impl<'a> CompositionGraph<'a> {
         Self::default()
     }
 
+    pub(crate) fn unify_imported_resources(&self) {
```
alexcrichton, line 575: Could this grow a doc block explaining what the method is doing?

crates/wasm-compose/src/graph.rs
```diff
@@ -438,6 +472,86 @@ pub struct EncodeOptions {
...
+    map: im_rc::HashMap<ResourceId, (ComponentId, ResourceId)>,
+}
+
+impl ResourceMapping {
+    pub(crate) fn add_pairs(
+        mut self,
+        export_component: ComponentId,
+        export_type: ComponentEntityType,
+        export_types: TypesRef,
+        _import_component: ComponentId,
```
peterhuene, line 486: Nit: unused parameter?

```suggestion
```

peterhuene (approved):
This looks great! Thanks for implementing this as it is sorely needed.

crates/wasm-compose/src/graph.rs
```diff
@@ -438,6 +472,86 @@ pub struct EncodeOptions {
...
+        import_types: TypesRef,
+    ) -> Option<Self> {
+        if let (
+            ComponentEntityType::Instance(export_type),
+            ComponentEntityType::Instance(import_type),
+        ) = (export_type, import_type)
+        {
+            let mut exports = HashMap::new();
+            for (export_name, ty) in &export_types[export_type].exports {
+                // TODO: support nested instances
```
alexcrichton, line 497: How difficult do you think it would be to solve this TODO and the ones below which derive from a similar code structure?

I was thinking there could be something along the lines of "invoke this callback argument on all resources found" or similar?

alexcrichton (approved):
Agreed looks great, thanks for working on this! I've mostly got some cosmetic things, it's ok to defer anything to later if you'd prefer

dicej:
Thanks for the feedback, @peterhuene and @alexcrichton.  I'll work on addressing it soon.

First, though: my colleague found a way to make it panic during encoding, so I'm going debug that.

crates/wasm-compose/src/encoding.rs
dicej: This had to do with the `created` vs. `referenced` fields in `wasmparser::types::ComponentEntityType::Type`, which I'll admit I still don't understand.  If you look at `TypeEncoder::component_entity_type` we used `created` for resource types and `referenced` for all other types.  That's the only way I could get it working, but it had the side effect of triggering this assertion.  I'm happy to expand the comment, but I'm afraid I don't understand the situation well enough to explain it.

I'll add the assertion clause you suggested, at least.

crates/wasm-compose/src/graph.rs
dicej: Not difficult, but I'm lazy so I've opened an issue instead: https://github.com/bytecodealliance/wasm-tools/issues/1288

dicej:
> First, though: my colleague found a way to make it panic during encoding, so I'm going debug that.

Turns out this was related to the incomplete state of `wasi-http` virtualization in `wasi-virt` as discussed [here](https://bytecodealliance.zulipchat.com/#narrow/stream/219900-wasi/topic/.60wasi-http.60.20middleware.20is.20infeasible.3F/near/402081439), which resulted in an inconsistent composition graph.  That's something `wasm-compose` should catch while creating the graph rather than panic during encoding.  I'll see if I can reduce that to a manageable test case, but meanwhile I've created #1289.

alexcrichton:
Ok I'm going to go ahead and merge this since I think it's in a good state, and it's ok to follow-up with fixes if necessary of course

PR #1262, opened 2023-10-24
Remove "interface imports" from binary syntax, expand accepted strings

alexcrichton:
This PR is the spiritual successor of https://github.com/bytecodealliance/wasm-tools/pull/1146 and is an implementation of these PRs together for this repository:

* https://github.com/WebAssembly/component-model/pull/222
* https://github.com/WebAssembly/component-model/pull/259
* https://github.com/WebAssembly/component-model/pull/263

This is technically a breaking change to the binary format, but the breaking change is so minor that a gradual rollout is implemented here. Previously import/export names had a discriminator byte of 0x00 or 0x01 depending on if they were a "plain name" or an "interface name". They no longer have a discriminator byte and instead always use 0x00. This means that old binaries are easily read by discarding the 0x00 or 0x01 byte. New binaries are, for now, temporarily created with the 0x01 byte still to ensure that binaries produced with this tooling is still compatible with older tooling. Once this tooling has permeated enough I plan to change the default emission to be 0x00 instead of 0x01.

The text syntax was changed in this PR as well. Previously where `(interface "a:b/c")` was printed only `"a:b/c"` is printed now. Like the binary format the text format still parses the old form with the intention of being deleted eventually.

Finally this PR additionally expands the set of accepted syntax within import and export strings. This now includes imports of implementations (locked/unlocked deps), imports of URLs (relative and not), and integrity-based imports.


alexcrichton:
TODO items:

* [ ] Validate that exports can't use urls/implementations/etc, basically only imports get the new syntax
* [ ] Parsing `integrity-metadata` currently does nothing but is defined by w3c. Should figure out what to do in the component model

alexcrichton:
Ok I've updated with parsing of what I believe is the format of `integrity-metadata` so this is good to go I think

crates/wit-component/src/metadata.rs
```diff
@@ -207,6 +207,7 @@ impl Bindgen {
             DecodedWasm::WitPackage(resolve, pkg) => (resolve, pkg),
             DecodedWasm::Component(..) => bail!("expected an encoded wit package"),
         };
+        println!("{world_name}");
```
peterhuene, line 210: Looks like a debugging change?

```suggestion
```

tests/cli/print-core-wasm-wit.wit.stdout
```diff
@@ -1,3 +1,4 @@
+my-world
```
peterhuene, line 1: Looks like a result of that `println!` from before?

```suggestion
```

peterhuene (commented):
This looks great! I'm really happy with where we landed with the strings and relaxing the restriction on conflicting names for all externs.

Just a few feedback comments.

peterhuene (approved):

PR #1264, opened 2023-10-25
`wasmparser`: Define core Wasm types in a macro

fitzgen:
This sets the stage for a second set of these definitions that don't use `u32` indices, but instead use something similar to `CoreTypeId`. This new set of types will be used in validation and will support canonicalization.

At the same time, I've taken the liberty to rename `StructuralType` to `CompositeType` to match the latest spec and to change `HeapType::Indexed` to `HeapType::Concrete` for the same reason.

The macro prevents users from having to worry about type parameters while still letting us avoid duplication.

### DIFF
--- a/crates/wasmparser/src/define_types.rs
+++ b/crates/wasmparser/src/define_types.rs
@@ -20,20 +20,20 @@ macro_rules! define_core_wasm_types {
         /// Represents a recursive type group in a WebAssembly module.
         #[derive(Debug, Clone)]
         pub struct RecGroup {
-            pub(crate) types: smallvec::SmallVec<[SubType; 1]>,
+            inner: RecGroupInner,
+        }
 
-            /// Whether or not this rec group was explicitly encoded in the
-            /// binary or was implicitly created for a type that was not
-            /// contained in a `(rec ...)` in the types section.
-            pub(crate) explicit_rec_group: bool,
+        #[derive(Debug, Clone)]
+        enum RecGroupInner {
+            Implicit(SubType),
+            Explicit(Vec<SubType>),
         }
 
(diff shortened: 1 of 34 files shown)


crates/wasmparser/src/define_types.rs
```diff
@@ -0,0 +1,746 @@
...
+    ($index_type:ty) => {
+        /// Represents a recursive type group in a WebAssembly module.
+        #[derive(Debug, Clone)]
+        pub struct RecGroup {
+            pub(crate) types: smallvec::SmallVec<[SubType; 1]>,
+
+            /// Whether or not this rec group was explicitly encoded in the
+            /// binary or was implicitly created for a type that was not
+            /// contained in a `(rec ...)` in the types section.
+            pub(crate) explicit_rec_group: bool,
```
alexcrichton, line 28: Given that this is the only use of `smallvec::` and we still need the explicit `bool` here anyway, could `SmallVec` be dropped as a dep?

alexcrichton (approved):

crates/wasmparser/src/define_types.rs
fitzgen: I find that working with smallvec is much easier and we already have it as a dep in this workspace and in wasmtime. How strongly do you feel here?

alexcrichton: If there was more to use than just this one location I think I'd agree, but if this is the only location to use `smallvec` I don't think it's worth it personally. Are you planning on adding more usages or is this thie only one?

fitzgen: I'll just switch it back to the old `enum`.

PR #1275, opened 2023-11-01
`wasmparser`: Perform type canonicalization for Wasm GC

### LINKED ISSUE #923: wasmparser: Validation of function references can take ~infinite time
This module:

```wasm
(module
  (type $t1 (func))
  (type $t2 (func
    (param (ref $t1) (ref $t1) (ref $t1) (ref $t1) (ref $t1) (ref $t1) (ref $t1) (ref $t1) (ref $t1) (ref $t1))
  ))
  (type $t3 (func
    (param (ref $t2) (ref $t2) (ref $t2) (ref $t2) (ref $t2) (ref $t2) (ref $t2) (ref $t2) (ref $t2) (ref $t2))
  ))
  (type $t4 (func
    (param (ref $t3) (ref $t3) (ref $t3) (ref $t3) (ref $t3) (ref $t3) (ref $t3) (ref $t3) (ref $t3) (ref $t3))
  ))
  (type $t5 (func
    (param (ref $t4) (ref $t4) (ref $t4) (ref $t4) (ref $t4) (ref $t4) (ref $t4) (ref $t4) (ref $t4) (ref $t4))
  ))
  (type $t6 (func
    (param (ref $t5) (ref $t5) (ref $t5) (ref $t5) (ref $t5) (ref $t5) (ref $t5) (ref $t5) (ref $t5) (ref $t5))
  ))
  (type $t7 (func
    (param (ref $t6) (ref $t6) (ref $t6) (ref $t6) (ref $t6) (ref $t6) (ref $t6) (ref $t6) (ref $t6) (ref $t6))
  ))
  (type $t8 (func
    (param (ref $t7) (ref $t7) (ref $t7) (ref $t7) (ref $t7) (ref $t7) (ref $t7) (ref $t7) (ref $t7) (ref $t7))
  ))
  (type $t9 (func
    (param (ref $t8) (ref $t8) (ref $t8) (ref $t8) (ref $t8) (ref $t8) (ref $t8) (ref $t8) (ref $t8) (ref $t8))
  ))
  (type $t10 (func
    (param (ref $t9) (ref $t9) (ref $t9) (ref $t9) (ref $t9) (ref $t9) (ref $t9) (ref $t9) (ref $t9) (ref $t9))
  ))
  (type $t11 (func
    (param (ref $t9) (ref $t9) (ref $t9) (ref $t9) (ref $t9) (ref $t9) (ref $t9) (ref $t9) (ref $t9) (ref $t9))
  ))

  (table 1 1 (ref null $t10))

  (func (param (ref null $t11))
      i32.const 0
      local.get 0
      table.set)


)
```

Currently takes ~infinite time to validate with the `function-references` proposal, but it shouldn't

fitzgen:
The unit of canonicalization is a recursion group. Having "unnecessary" types in a recursion group can "break" canonicalization of other types within that same recursion group, as can reordering types within a recursion group.

It is an invariant that all types defined before the recursion group we are currently canonicalizing have already been canonicalized themselves.

Canonicalizing a recursion group then proceeds as follows:

* First we walk each of its `SubType` elements and put their type references (i.e. their `PackedIndex`es) into canonical form. Canonicalizing a `PackedIndex` means switching it from indexing into the Wasm module's types space into either

  1. Referencing an already-canonicalized type, for types outside of this recursion group. Because inter-group type references can only go towards types defined before this recursion group, we know the type is already canonicalized and we have a `CoreTypeId` for each of those types. This updates the `PackedIndex` into a `CoreTypeId`.

  2. Indexing into the current recursion group, for intra-group type references.

  Note that (2) has the effect of making the "same" structure of mutual type
  recursion look identical across recursion groups:

      ;; Before
      (rec (struct (field (module-type 1))) (struct (field (module-type 0))))
      (rec (struct (field (module-type 3))) (struct (field (module-type 2))))

      ;; After
      (rec (struct (field (rec-group-type 1))) (struct (field (rec-group-type 0))))
      (rec (struct (field (rec-group-type 1))) (struct (field (rec-group-type 0))))

* Now that the recursion group's elements are in canonical form, we can "simply" hash cons whole rec groups at a time. The `TypesList` morally maintains a hash map from `Vec<SubType>` to `RecGroupId` and we can do get-or-create operations on it. I say "morally" because we don't actually duplicate the `Vec<SubType>` key in that hash map since those elements are already stored in the `TypeList`'s internal `SnapshotList<CoreType>`. This means we need to do some low-level hash table fiddling with the `hashbrown` crate.

And that's it! That is the whole canonicalization algorithm.

Some more random things to note:

* Because we essentially already have to do the check to canonicalize, and to avoid additional passes over the types, the canonicalization pass also checks that type references are in bounds. These are the only errors that can be returned from canonicalization.

* Canonicalizing requires the `Module` to translate type indices to actual `CoreTypeId`s.

* It is important that *after* we have canonicalized all types, we don't need the `Module` anymore. This makes sure that we can, for example, intern all types from the same store into the same `TypeList`. Which in turn lets us type check function imports of a same-store instance's exported functions and we don't need to translate from one module's canonical representation to another module's canonical representation or perform additional expensive checks to see if the types match or not (since the whole point of canonicalization is to avoid that!).

-------------------------------------

I initially tried to have two different Rust types for each Wasm core type (`SubType`, `FuncType`, etc...): one for the version that contains raw type space indices that are produced directly from the reader and another that contains `CoreTypeId`s after canonicalization. This approach is essentially what we do for component model types. However, this was getting really painful, because even `ValType` would have to have two different versions. The amount of places I was touching, including in downstream crates, was getting out of hand.

So instead I opted to make a new index type that is morally the following enum:

```rust
enum Index {
    ModuleTypesSpaceIndex(u32),
    RecGroupLocalIndex(u32),
    CoreTypeId(CoreTypeId),
}
```

Of course, we have to be very frugal with bits to keep `RefType` fitting in 24 bits and `ValType` in 32 bits, so it is actually a bit-packed version of that. We can still represent the maximum number of Wasm types in a module. However a `TypeList` can only have `2 * MAX_WASM_TYPES` stored in it now (or at least that is how many are addressable; you could add more and then never stuff their `CoreTypeId`s into these bit-packed indices). We could free up some more bits here if we started bit-packing `ValType`, but the loss in ergonomics of matching on `ValType` would be pretty bad.

Anyways, I also added an unpacked version of these bit-packed indices for ergonomics. The bit-packed version can infallibly be converted to the unpacked version, and the unpacked version can fallibly be converted to the bit-packed version (it checks that the indices are representable in the number of bits we actually have available).

Finally, because we are back to only having a single Rust type for each core Wasm type, I removed the `define_core_wasm_types!` macro and inlined the definitions. Sorry for the churn! But it is definitely nicer not having them inside a macro at the end of the day.

-----------------------------------

This also fixes #923, since canonicalization avoids the exponential behavior observed there.


fitzgen:
Beyond the private types failures because I've been testing in nightly, I accidentally left a test file partly commented out and uncommenting it has revealed a couple more test failures. Will work on them, probably fine to give feedback on the PR at the same time because I don't expect things to change wildly.

fitzgen:
Okay I resolved the private-types error and got the failing tests passing.

crates/wasmparser/src/readers/core/types/matches.rs
```diff
@@ -0,0 +1,380 @@
...
+
+impl Matches for WithRecGroup<ValType> {
+    fn matches(types: &TypeList, a: Self, b: Self, offset: usize) -> Result<bool> {
+        match (*a, *b) {
+            (ValType::Ref(ra), ValType::Ref(rb)) => types.matches(
+                WithRecGroup::map(a, |_| ra),
+                WithRecGroup::map(b, |_| rb),
+                offset,
+            ),
+            (a, b) => Ok(a == b),
```
alexcrichton, line 300: Could this expand `a` to `a @ (ValType::I32 | ValType::I64 ...)` to avoid the catch-all?

crates/wasmparser/src/readers/core/types/matches.rs
```diff
@@ -0,0 +1,380 @@
...
+                subtype(b, ib)?.composite_type,
+                CompositeType::Array(_) | CompositeType::Struct(_)
+            )),
+
+            (HT::NoFunc, HT::Concrete(ib)) => Ok(matches!(
+                subtype(b, ib)?.composite_type,
+                CompositeType::Func(_)
+            )),
+
+            _ => Ok(false),
```
alexcrichton, line 377: Could this also be expanded to avoid the catch-all?

crates/wasmparser/src/validator/core.rs
```diff
@@ -1164,8 +1171,18 @@ impl WasmModuleResources for OperatorValidatorResources<'_> {
         self.module.element_types.get(at as usize).cloned()
     }
 
-    fn matches(&self, t1: ValType, t2: ValType) -> bool {
-        self.module.matches(t1, t2, self.types)
+    fn matches(&self, mut a: ValType, mut b: ValType) -> bool {
+        let canonicalizer = TypeCanonicalizer::new(&self.module, usize::MAX);
+        canonicalizer.canonicalize_val_type(&mut a).unwrap();
+        canonicalizer.canonicalize_val_type(&mut b).unwrap();
```
alexcrichton, line 1177: Could this have a brief comment on why the `unwrap()` is ok?

alexcrichton: (same below too)

crates/wasmparser/src/validator/types.rs
```diff
@@ -2408,12 +2442,149 @@ impl TypeList {
     where
         T: TypeData,
     {
+        // eprintln!("FITZGEN: TypeList::push({ty:?})");
```
alexcrichton, line 2445: debug prints

crates/wasmparser/src/validator/types.rs
```diff
@@ -2438,6 +2612,9 @@ impl TypeList {
             component_funcs: component_funcs.len(),
             core_modules: core_modules.len(),
             core_instances: core_instances.len(),
+            core_type_to_rec_group: core_type_to_rec_group.len(),
+            rec_group_elements: rec_group_elements.len(),
+            canonical_rec_groups: canonical_rec_groups.clone(),
```
alexcrichton, line 2617: This is cloning an entire hash table on all checkpoints, right? That seems like it could get expensive?

alexcrichton: Oh I remember we talked a bit about asserting that the hash table just doesn't change during the checkpoint and I suppose that didn't make its way here yet

alexcrichton (commented):
Thanks for this! Generally all looks good to me, some additional questions though:

* One thing I'm a bit worried about is that we have 20 bits for an index and allow up to a million types. This'll definitely causes problems with components right now where each module can have up to a million types so we can blow this somewhat easily. I didn't see too many unwraps though and I think that errors are plumbed, so wanted to confirm? (ideally we'd have a million-type-limit-per-component rather than per-module, but that's best for another PR)
* I'm a bit surprised by the recursion group hashing, because that means that equivalent types across recursion groups which are slightly different (e.g. one has one extra type) won't hash the same? That could open back up the exponential validation time problem, right?

crates/wasmparser/Cargo.toml
```diff
@@ -13,6 +13,7 @@ edition.workspace = true
 exclude = ["benches/*.wasm"]
 
 [dependencies]
+hashbrown = { version = "0.14.2", default-features = false, features = ["ahash"] }
```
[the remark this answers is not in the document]
fitzgen: The standard hashmap doesn't allow for not actually storing the key in the table the way that we are doing here with `hashbrown::HashTable`.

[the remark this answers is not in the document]
fitzgen: And querying the hash table with types that aren't `Borrow` of the actual "key" type that isn't actually present inside the hash table.

crates/wasmparser/src/readers/core/types/matches.rs
```diff
@@ -0,0 +1,380 @@
...
+    }
+}
+
+impl<'a> Matches for WithRecGroup<&'a SubType> {
+    fn matches(types: &TypeList, a: Self, b: Self, offset: usize) -> Result<bool> {
+        types.matches(
+            WithRecGroup::map(a, |a| &a.composite_type),
+            WithRecGroup::map(b, |b| &b.composite_type),
+            offset,
+        )
```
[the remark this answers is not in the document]
fitzgen: Matching isn't quite the same as subtyping AFAICT. The finality checks only happen when we check the actual type definition, not inside matching, in the spec.

crates/wasmparser/src/readers/core/types/matches.rs
```diff
@@ -0,0 +1,380 @@
...
+        if a.params().len() != b.params().len() || a.results().len() != b.results().len() {
+            return Ok(false);
+        }
+
+        let params_match =
+            a.params()
+                .iter()
+                .zip(b.params())
+                .try_fold(true, |matches, (pa, pb)| {
+                    // Parameters are contravariant.
```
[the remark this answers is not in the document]
fitzgen: Done!

xortoast:
Is this really the best way to solve this problem? Changing representation of types in this way (just to make canonicalization simpler) makes working with them much uglier in all other cases, since you need to handle cases that can never happen. This is already evident with this PR, with all the changes from `From` to `TryFrom` inside `wasm-encoder` and added `.as_module_index().unwrap()` calls in a few other crates.

I have another idea: Instead of rewriting `SubType` elements ahead of time, we do it on-the-fly inside `eq` and `hash` functions.

We would need to somehow pass additional context to them, in particular: (1) index of the recursion group itself (to distinguish between intra-group and inter-group type references) and (2) a mapping from module indices to canonical indices. Since we would need to do some low-level hash table fiddling in any case, this shouldn't be too hard.

xortoast:
> I initially tried to have two different Rust types for each Wasm core type (`SubType`, `FuncType`, etc...): one for the version that contains raw type space indices that are produced directly from the reader and another that contains `CoreTypeId`s after canonicalization.

Why would you need two different Rust types? Wouldn't it be enough to make each of the types generic over index type `I`, with default being `u32`?

fitzgen:
@xortoast having two different types was *very* unwieldy, and also requires essentially copying all the types during canonicalization. I don't think it has a future.

The idea of rewriting on the fly is interesting, but I think will be a bit painful for consumers that rely on the results of validation across multiple modules since they will have to maintain and thread module state through all their various calls and they won't be able to compare types without providing the context and all that. In contrast, with this approach, they have the invariant that all types returned from the validator are canonicalized and they only need to provide the `wasmparser::Types` arena, which they needed anyways since that is where the types themselves are stored. Similarly, API users just using the readers have the invariant that they only get the raw module indices. It is true that it requires some unwraps, but it feels like the best of the options available.

xortoast:
@fitzgen You misunderstood my idea. The context I mentioned is used only for canonicalization (i.e. building the arena). Once a type is added to the arena, all the module-space indices are replaced with canonical indices.

Anyway, I think we both over complicated our solutions.

The first thing is that we don't need to explicitly distinguish between `ModuleTypesSpaceIndex` and `CoreTypeId`. When read from module, all indices are `ModuleTypesSpaceIndex`, and when stored in the arena, all indices (that are not `RecGroupLocalIndex`) are `CoreTypeId`. Therefore, your index type could be conceptually simplified as follows:

```rust
enum Index {
    AbsoluteIndex(u32),
    RecGroupLocalIndex(u32),
}
```

The second thing is that we can also get rid of `RecGroupLocalIndex`. Group-local indices are only relevant for canonicalization, and for that you can calculate them on the fly (using additional context as mentioned before).

Therefore, I think that even your existing solution could be implemented without modifying `RefType`. The only real change is the additional context that would be passed to `eq` and `hash` functions during canonicalization (just the index of the recursion group itself; the mapping between module indices and canonical indices would not be needed).

crates/wasmparser/src/validator/operators.rs
```diff
@@ -959,15 +979,24 @@ impl<'resources, R: WasmModuleResources> OperatorValidatorTemp<'_, 'resources, R
     fn params(&self, ty: BlockType) -> Result<impl PreciseIterator<Item = ValType> + 'resources> {
         Ok(match ty {
             BlockType::Empty | BlockType::Type(_) => Either::B(None.into_iter()),
-            BlockType::FuncType(t) => Either::A(self.func_type_at(t)?.inputs()),
+            BlockType::FuncType(t) => Either::A(self.func_type_at(t)?.inputs().map(|mut ty| {
+                self.resources.canonicalize_valtype(&mut ty);
```
alexcrichton, line 983: Is this necessary? I would have expected that `func_type_at` is looking something up which has already been canonicalized

crates/wasmparser/src/validator/operators.rs
```diff
@@ -959,15 +979,24 @@ impl<'resources, R: WasmModuleResources> OperatorValidatorTemp<'_, 'resources, R
...
         Ok(match ty {
             BlockType::Empty => Either::B(None.into_iter()),
-            BlockType::Type(t) => Either::B(Some(t).into_iter()),
-            BlockType::FuncType(t) => Either::A(self.func_type_at(t)?.outputs()),
+            BlockType::Type(mut t) => {
+                self.resources.canonicalize_valtype(&mut t);
+                Either::B(Some(t).into_iter())
+            }
+            BlockType::FuncType(t) => Either::A(self.func_type_at(t)?.outputs().map(|mut ty| {
+                self.resources.canonicalize_valtype(&mut ty);
```
alexcrichton, line 997: ditto to above

crates/wasmparser/src/validator/operators.rs
```diff
@@ -1357,20 +1393,22 @@ where
...
         self.push_operand(ty)?;
         Ok(())
     }
     fn visit_local_get(&mut self, local_index: u32) -> Self::Output {
-        let ty = self.local(local_index)?;
+        let mut ty = self.local(local_index)?;
         if !self.local_inits[local_index as usize] {
             bail!(self.offset, "uninitialized local: {}", local_index);
         }
+        self.resources.canonicalize_valtype(&mut ty);
```
alexcrichton, line 1411: Could this be avoided by canonicalizing locals when they're declared rather than on each instruction?

crates/wasmparser/src/validator/operators.rs
```diff
@@ -1396,7 +1434,9 @@ where
     }
     fn visit_global_get(&mut self, global_index: u32) -> Self::Output {
         if let Some(ty) = self.resources.global_at(global_index) {
-            self.push_operand(ty.content_type)?;
+            let mut ty = ty.content_type;
+            self.resources.canonicalize_valtype(&mut ty);
```
alexcrichton, line 1438: Ditto to above, could the type as returned by `global_at` be pre-canonicalized?

crates/wasmparser/src/validator/operators.rs
```diff
@@ -2291,10 +2333,12 @@ where
         // FIXME(#924) this should not be conditional based on enabled
         // proposals.
         if self.features.function_references {
```
alexcrichton, line 2335: FWIW I think you may have fixed this and this branch may be able to be deleted

(awhile back, not in this PR specifically)

crates/wasmparser/src/validator/operators.rs
```diff
@@ -3298,8 +3342,10 @@ where
             Some(ty) => ty.element_type,
             None => bail!(self.offset, "table index out of bounds"),
         };
+        let mut ty = ValType::Ref(ty);
+        self.resources.canonicalize_valtype(&mut ty);
```
alexcrichton, line 3346: Same concern as prior instances, ideally this would avoid canonicalizing.

alexcrichton (approved):
Ok I talked a bit more with @fitzgen about this and I'm comfortable with this landing. Some resolutions of my prior concerns:

* Exponential matching time is avoided because the structural matching "bottoms out" in "is A a supertype of B" which is a precalculated, in effect, predicate. The `Matches` trait "looks" recursive but isn't because it always bottoms out in this predicate.
* As for two modules each will a million types in a component, we can handle that later.
* As for recursion group hashing, that's just how the spec is. While I found it a bit counterintuitive at first an explanation of what the spec is cleared this up for me.

One bit I've commened on here is that there looks like there's a lot of canonicalization happening during validation but I think ideally that would all be front-loaded as much as possible. Only instructions with immediates that are types should need canonicalization I think (e.g. typed `select` or `ref.null`).

As for @xortoast's concerns and a different representation, personally I feel a bit uneasy about having a `u32` which could simultaneously be three different things depending on where you are. While I don't disagree that the approach in this PR is wordy and probably not the most concise I feel that it's at least a solid starting point. @xortoast how would you feel about landing this PR's approach and if you're interested you could have a follow-up with further refactorings?

xortoast:
@alexcrichton Sure. You can land this PR.

crates/wasmparser/src/validator/operators.rs
fitzgen: I honestly don't know what this comment is about because that issue is fixed but also had nothing to do with this comment afaict?

alexcrichton: Ah no worries, I'll remove it later. I got confused and thought this comment was here for typechecks possibly becoming exponential. I was wrong though and this was here for another reason, unrelated to this PR entirely.

PR #1279, opened 2023-11-09
wast: Update parsing of try_table for latest version

eqrion:


eqrion:
The latest version can be found [here](https://github.com/WebAssembly/exception-handling/blob/main/proposals/exception-handling/Exceptions.md#control-flow-instructions). The change is basically to have variants that capture or don't capture the exnref, along with allowing multiple catch_all's.

crates/wast/src/core/expr.rs
```diff
@@ -1228,42 +1228,31 @@ impl<'a> Parse<'a> for BlockType<'a> {
...
-                    p.parse::<kw::catch_all_ref>()?;
-                    TryTableCatchKind::Ref
-                } else {
+                    TryTableCatchKind::Catch(p.parse()?)
+                } else if parser.peek::<kw::catch_all>()? {
                     p.parse::<kw::catch_all>()?;
-                    TryTableCatchKind::Drop
+                    TryTableCatchKind::CatchAll
+                } else {
+                    p.parse::<kw::catch_all_ref>()?;
```
alexcrichton, line 1251: If you'd like, the [`parser.lookahead1()` method](https://docs.rs/wast/latest/wast/parser/struct.Parser.html#method.lookahead1) can be used here as well to provide a nicer error message if none of the cases match. 

(optional, happy to merge either way)

crates/wast/src/core/expr.rs
```diff
@@ -1273,32 +1262,38 @@ impl<'a> Parse<'a> for TryTable<'a> {
...
+    CatchAllRef,
+}
+
+impl<'a> TryTableCatchKind<'a> {
+    #[allow(missing_docs)]
+    pub fn tag_index_mut(&mut self) -> Option<&mut Index<'a>> {
+        match self {
+            TryTableCatchKind::Catch(tag) |
+            TryTableCatchKind::CatchRef(tag) => Some(tag),
+            _ => None
```
alexcrichton, line 1288: Mind making this exhaustive as well?

alexcrichton (approved):
Thanks! Some minor comments, but also I think this may need a `rustfmt`

crates/wast/src/core/expr.rs
eqrion: Hmm, so I don't think that would improve the error messages here as we only do these peeks here if the earlier peek2() checks succeeded and we entered the parentheses, so these peeks are infallible and no error will be generated here.

alexcrichton: Oh good point!

alexcrichton (approved):

alexcrichton:
> I always forget that plain cargo fmt isn't enough, or just do it wrong

I think this is due to an unfortunate limitation of rustfmt where the `wast` crate defines modules internally through a macro which thwarts rustfmt's discovery of files to format. We should probably update the crate structure to not be defined with macros and write things out a bit more verbosely.

alexcrichton:
Oh and would you like a release of this as well?

eqrion:
Yes, that would be great!

PR #1281, opened 2023-11-10
Enable more GC spec tests 

fitzgen:


crates/wast/src/core/binary.rs
```diff
@@ -212,6 +212,11 @@ impl Encode for Type<'_> {
 
 impl Encode for Rec<'_> {
     fn encode(&self, e: &mut Vec<u8>) {
+        if self.types.len() == 1 {
+            self.types[0].encode(e);
+            return;
+        }
```
alexcrichton, line 218: Was this required to get tests working, and if so which one? Otherwise I'd prefer to not do this to avoid possibly-surprising changes to the AST of the binary emitted

tests/local/missing-features/rec-group.wast
```diff
@@ -2,9 +2,12 @@
...
 
-(assert_invalid
-  (module (rec (type (func))))
-  "requires `gc` proposal to be enabled")
+;; As a size optimization, we encode explicit rec groups of one type into the
+;; equivalent implicit rec group of just that type's direct encoding.
+;;
+;; (assert_invalid
+;;   (module (rec (type (func))))
+;;   "requires `gc` proposal to be enabled")
```
alexcrichton, line 10: This seems a result of the `wast` change above, but is a test I'd ideally like to keep. If the `wast` change can't be reverted then can this switch to using `module binary` instead of the text format?

crates/wast/src/core/binary.rs
fitzgen: This was needed to get round-tripping of `types-subtyping.wast` working, I'll add the `module binary` thing you suggested below.

alexcrichton: Which test needed this? Is it something that should be changed upstream?

alexcrichton (approved):

PR #1285, opened 2023-11-13
wasm-encoder: Field and tag names

xortoast:
As defined in WebAssembly/gc#193 and WebAssembly/exception-handling#213.


alexcrichton (approved):
Thanks!

PR #1293, opened 2023-11-16
Fix failing assertion in compose encoding

fibonacci1729:
This assertion was expanded while addressing the final feedback in the PR to [Implement Resources in Compose](https://github.com/bytecodealliance/wasm-tools/pull/1261).

Currently, with the expanded assertion, when composing a component that exports `wasi:http/incoming-handler` with another that imports it, the assertion fails with:

`thread 'main' panicked at 'assertion failed: prev.is_none() ||\n  (matches!(id, ComponentAnyTypeId :: Resource(_)) && prev == Some(value))`

And specifically the condition that fails `prev == Some(value)` where, in this particular instance, `prev=Some(5)` != `value=6`.

I'm not entirely sure of the "why" here but have confirmed that reverting to the old assertion seems to "correct" this.

### DIFF
--- a/crates/wasm-compose/src/encoding.rs
+++ b/crates/wasm-compose/src/encoding.rs
@@ -880,10 +880,7 @@ impl<'a> TypeEncoder<'a> {
             let prev = state.cur.type_defs.insert(key, value);
             // Resource aliases will already have been added to the table, but
             // other types should not have been:
-            assert!(
-                prev.is_none()
-                    || (matches!(id, ComponentAnyTypeId::Resource(_)) && prev == Some(value))
-            );
+            assert!(prev.is_none() || matches!(id, ComponentAnyTypeId::Resource(_)));
             state.cur.add_type_export(key, name);
         }
         export


fibonacci1729:
cc @alexcrichton @dicej

alexcrichton:
Hm interesting! Before merging I think I'd like to dig in a bit more here to see if I can't bottom this out a bit more. Do you have a reproduction I could poke around at to trigger this?

fibonacci1729:
@alexcrichton Testing with: https://github.com/fermyon/http-auth-middleware/

Using: `cargo-component-component 0.4.0 (821c329 2023-10-27 wasi:2da78ca)`

To reproduce:
```
cargo component build --manifest-path github-oauth/Cargo.toml --release
cd example && ./build.sh
```

alexcrichton:
Ok thanks! I came up with https://github.com/bytecodealliance/wasm-tools/pull/1297 which ends up tightening the assertion here though some slight restructuring of the code. I've tested that the example you gave gets composed with that change as well. I suspect there may still be an issue lurking but I don't know of any off the top of my head at this time

fibonacci1729:
Closing in favor of #1297

PR #1294, opened 2023-11-16
stop kebab-casing and validate component names

macovedj:
Currently wasm-compose is kebab-casing package names, as this used to make sense before recent changes to import syntax.

With the updated changes, if a package is imported with the name:
`locked-dep=<foo:add@1.0.0>,integrity=<sha256-d780a28ec35f001cf1e61401ec1ee22dad5821ccf8280affcf135074c7c53cf2>`,

 wasm-compose will change it to
`locked-dep-foo-add-1-0-0-integrity-sha256-d780a28ec35f001cf1e61401ec1ee22dad5821ccf8280affcf135074c7c53cf2`.  

This PR stops package names from being kebab-cased, and validates their names using `ComponentName` from wasmparser.


alexcrichton (commented):
Thanks! Could you add a test for this as well?

macovedj:
> Thanks! Could you add a test for this as well?

Sure thing!  Looks like there are some existing test failures I'll have to investigate as well.

macovedj:
Ok I just went ahead and enabled validation for `$input` specifically for the use case that is `wasm-compose` because it felt like the best option to me... happy to to go with another option though if that makes sense.

crates/wasmparser/src/validator/names.rs
```diff
@@ -547,6 +548,10 @@ impl<'a> ComponentNameParser<'a> {
             return Ok(ParsedComponentNameKind::Hash);
         }
 
+        if self.eat_str("$input") {
+            return Ok(ParsedComponentNameKind::Hash);
+        }
+
```
alexcrichton, line 554: Oh wasm-compose-specific inputs shouldn't make their way into wasmparser. I don't know how `$input` is used much myself, but the fix for this should be in wasm-compose rather than in validation

alexcrichton (commented):
cc @peterhuene, you might have a better idea of how to handle `$input` perhaps (I also don't fully understand the implications of changing it)

macovedj:
After conferring with Peter, went ahead and reverted the new validation and changed `$input` to `composition-root`

crates/wasm-compose/src/graph.rs
```diff
@@ -51,9 +52,8 @@ impl<'a> Component<'a> {
     pub fn from_file(name: impl Into<String>, path: impl AsRef<Path>) -> Result<Self> {
         let path = path.as_ref();
         log::info!("parsing WebAssembly component file `{}`", path.display());
-
         let component = Self::parse(
-            name.into(),
+            ComponentName::new(&name.into(), 0)?.to_string(),
```
alexcrichton, line 56: This change, and the one below, may not be necessary any more?

crates/wasm-compose/src/graph.rs
```diff
@@ -1554,4 +1558,344 @@ mod test {
 
         Ok(())
     }
+
+    #[test]
+    fn it_encodes_a_simple_composition_with_locked_dep() -> Result<()> {
```
alexcrichton, line 1563: Would it be possible to move these tests into `tests/compositions`? Otherwise it'll be pretty difficult to update these test expectations if they start failing in the future

alexcrichton (commented):
Sounds reasonable to me!

crates/wasm-compose/src/graph.rs
macovedj: Went ahead and changed the original tests in `tests/compositions` to include `invalid-semver` in their name, and created new ones there that capture the spirit of these, called `locked-dep-simple` and `unlocked-dep-simple`.

crates/wasm-compose/src/encoding.rs
```diff
@@ -1083,7 +1083,7 @@ impl<'a> ImportMap<'a> {
             assert!(self
                 .0
                 .insert(
-                    entry.component.name.to_kebab_case().into(),
+                    entry.component.name.clone().into(),
```
peterhuene, line 1086: I think we might be able to lose the `Cow<'a, str>` key for the import map and just have `&'a str` now that we're no longer needing to create a new string by `to_kebab_case()`.

peterhuene (commented):
Generally looks great! Just a few feedback comments.

peterhuene (approved):

PR #1297, opened 2023-11-16
Tighten and fix an assertion defining resource types

alexcrichton:
This commit goes back to some comments I had on #1261 where an assertion was loosened there. My requested modifications ended up spawning #1293 so I've taken another look at this assertion to try to determine what's going on. I've managed to re-tighten the assertion while fixing the situation coming up in #1293 as well.

The fix in this PR is to restructure the `_ty` method to iterate over the alias chain once and only define the type at the very end which should fix this for resources. I'm still having a difficult time fully explaining everything involved here but this feels like a more appropriate structure for the code at this time anyway.


dicej (approved):
I still won't claim to understand this code, but I've tested it with a few non-trivial examples and it's working.

fibonacci1729 (approved):
Tested on my auth-component and is working as expected,  thanks!

PR #1299, opened 2023-11-17
Canonicalize types earlier in validation

alexcrichton:
This commit refactors the internals of validation to centralize canonicalization in one location. This means that stages such as operator validation no longer need to perform canonicalization. Additionally types are always stored in their canonical forms.

The refactoring here is done when a `RecGroup` is inserted into a `TypeList`. The rec-group-local offsets are used to intern the group but when actually adding to a `TypeList` all rec-group-local offsets are updated to an ID-based form. This is then plumbed in many other locations as well by changing many `t: ValType` validators to `t: &mut ValType` to internally canonicalize.

This is originally motivated by trying to update Wasmtime to the latest `wasmparser` but many APIs which previously returned `&FuncType` for example returned `FuncType` instead due to the
canonicalization-on-the-fly. This change means that the old `&FuncType`-style APIs can return because types are always canonicalized at-rest.

This change additionally changes the behavior of one test. It was previously considered valid and is now considered valid. I've confirmed that `wasm-opt` considers the test invalid, so I've updated the test to `assert_invalid` instead.

### DIFF
--- a/crates/wasmparser/src/validator/core.rs
+++ b/crates/wasmparser/src/validator/core.rs
@@ -970,15 +970,10 @@ impl Module {
                 *type_index = UnpackedIndex::Id(id);
                 Ok(())
             }
-            UnpackedIndex::RecGroup(_) => {
-                todo!()
-            }
-            UnpackedIndex::Id(_) => {
-                // If the type index has already been canonicalized,
-                // then we already checked that it was in bounds and
-                // valid at that time.
-                Ok(())
-            }
+            // Types at this stage should not be canonicalized. All
+            // canonicalized types should already be validated meaning they
+            // shouldn't be double-checked here again.
+            UnpackedIndex::RecGroup(_) | UnpackedIndex::Id(_) => unreachable!(),
         }
     }
 
--- a/crates/wasmparser/src/validator/core.rs
+++ b/crates/wasmparser/src/validator/core.rs
@@ -3,15 +3,13 @@
 
 mod canonical;
 
-use self::{
-    arc::MaybeOwned,
-    canonical::{canonicalize_and_intern_rec_group, TypeCanonicalizer},
-};
+use self::{arc::MaybeOwned, canonical::canonicalize_and_intern_rec_group};
 use super::{
     check_max, combine_type_sizes,
     operators::{ty_to_str, OperatorValidator, OperatorValidatorAllocations},
     types::{CoreTypeId, EntityType, RecGroupId, TypeAlloc, TypeList},
 };
+use crate::validator::types::TypeIdentifier;
 use crate::{
     limits::*, BinaryReaderError, CompositeType, ConstExpr, Data, DataKind, Element, ElementKind,
     ExternalKind, FuncType, Global, GlobalType, HeapType, MemoryType, PackedIndex, RecGroup,
@@ -128,13 +126,13 @@ impl ModuleState {
 
     pub fn add_global(
         &mut self,
-        global: Global,
+        mut global: Global,
         features: &WasmFeatures,
         types: &TypeList,
         offset: usize,
     ) -> Result<()> {
         self.module
-            .check_global_type(&global.ty, features, offset)?;
+            .check_global_type(&mut global.ty, features, offset)?;
         self.check_const_expr(&global.init_expr, global.ty.content_type, features, types)?;
         self.module.assert_mut().globals.push(global.ty);
         Ok(())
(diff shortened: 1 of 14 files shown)


crates/wasmparser/src/validator/core/canonical.rs
```diff
@@ -140,18 +140,18 @@ impl<'a> TypeCanonicalizer<'a> {
...
-    pub fn within_rec_group(&mut self, types: &TypeList, rec_group_id: RecGroupId) -> &mut Self {
-        assert_eq!(self.mode, CanonicalizationMode::OnlyIds);
-        self.within_rec_group = Some(types[rec_group_id].clone());
-        self
-    }
+    // pub fn within_rec_group(&mut self, types: &TypeList, rec_group_id: RecGroupId) -> &mut Self {
+    //     assert_eq!(self.mode, CanonicalizationMode::OnlyIds);
+    //     self.within_rec_group = Some(types[rec_group_id].clone());
+    //     self
+    // }
```
fitzgen, line 154: A little surprised to see this stuff commented out. Haven't dug into the `types.rs` changes yet, but I'd expect that we would still want to canonicalize all type references to `CoreTypeId`s before exposing them to users. Like, I don't think they really know what to do with a rec-group local index.

crates/wasmparser/src/validator/core.rs
```diff
@@ -956,20 +961,23 @@ impl Module {
...
+                Ok(())
+            }
+            UnpackedIndex::RecGroup(_) => {
+                todo!()
+            }
+            UnpackedIndex::Id(_) => {
+                // If the type index has already been canonicalized,
+                // then we already checked that it was in bounds and
+                // valid at that time.
+                Ok(())
```
fitzgen, line 980: These should both be unreachable after this PR, right? Since the idea is that we check once and that does the canonicalization, and then we never check again?

crates/wasmparser/src/validator/types.rs
```diff
@@ -2403,12 +2403,12 @@ pub struct TypeList {
...
-    rec_group_elements: SnapshotList<std::ops::Range<CoreTypeId>>,
+    rec_group_elements: SnapshotList<Range<CoreTypeId>>,
     // A hash map from rec group elements to their canonical `RecGroupId`.
     //
-    // This hash map is queried by the full `RecGroup` structure but actually
-    // only stores the range of the rec group's elements as a key.
-    canonical_rec_groups: hashbrown::HashTable<(std::ops::Range<CoreTypeId>, RecGroupId)>,
+    // This is `None` when a list is "committed" meaning that no more insertions
+    // can happen.
+    canonical_rec_groups: Option<HashMap<RecGroup, RecGroupId>>,
```
fitzgen, line 2411: This is going to keep two copies of each `SubType` around, until we commit the the types list.

That was why I was using `hashbrown` before, to avoid that memory overhead.

crates/wasmparser/src/validator/types.rs
```diff
@@ -2466,93 +2466,104 @@ impl TypeList {
...
-                    self.core_type_to_rec_group.push(rec_group_id);
-                }
+            self.core_type_to_supertype
+                .push(ty.supertype_idx.map(|idx| match idx.unpack() {
+                    UnpackedIndex::RecGroup(offset) => CoreTypeId::from_index(start.index + offset),
+                    UnpackedIndex::Id(id) => id,
+                    UnpackedIndex::Module(_) => unreachable!("in canonical form"),
+                }));
+            let mut ty = ty.clone();
+            canonicalize_sub_type(&mut ty, start);
```
fitzgen, line 2497: This doesn't make sense to me -- it should already be in canonical form when we call this function? Otherwise the hash map lookup makes no sense in the first place.

fitzgen (commented):
Two things

1. the memory overhead of duplicated types (I guess this is okay if we decide it is okay, but we should be cognizant of the choice we are making)

2. I don't understand how hashconsing still works and/or I am misunderstanding something eles.

crates/wasmparser/src/validator/core.rs
alexcrichton: Ah yes good point!

crates/wasmparser/src/validator/types.rs
alexcrichton: True! That was a tradeoff I'd like to lean into with this PR though. Effectively one copy of the type using `RecGroup`-local indices is kept around and one copy with `Id`-based indices is kept around. The `TypeList` in theory can discard the `RecGroup`-local version once the type section is fully processed and that way only the `Id`-based versions are retained. I haven't refactored to do that here in this PR though.

crates/wasmparser/src/validator/types.rs
alexcrichton: We talked about this but I wanted to write this down here for others too, but this is changing `RecGroup` canonicalization into `Id` canonicalization. I'll work on getting this to share code with the preexisting mapping of indices.

alexcrichton:
Nick's out on vacation for a few weeks so I'm going to ask for a rubber stamp from @pchickey on this. I believe I've addressed Nick's feedback here but Nick if I missed anything or if you'd like me to update anything please let me know of course and I'd be happy to do so!

pchickey (approved):

PR #1301, opened 2023-11-20
Make wasmparser an optional dependency of wasm-smith

eqrion:
I ran into an issue importing the latest `wast` (68.0.0) into SpiderMonkey. We vendor all rust deps into mozilla-central and there's a rule that we need to avoid vendoring duplicate dependencies with different versions if we can at all avoid it.

Today we have two explicit deps of wasm-tools, `wast` and `wasm-smith`. Those then pull in other deps like `wasm-encoder`. It looks like `wasm-smith` has a dep on `wasmparser` which then pulled in a new major version of `hashbrown`, which resulted in a duplicate with other deps of `hashbrown` in mozilla-central.

I think the easiest way for us to resolve it would be if we could avoid importing `wasmparser` for `wasm-smith`. It looks like it's very lightly used in the library (just for specifying fixed imports) and so this PR makes it optional.

I'm open to other solutions, this was just a quick thing to hack up.

Note, one issue was importing KebabStr/KebabString from wasmparser. I feel bad, but it was small enough I just made a copy inside of wasm-smith. I think having it in a shared crate would make sense, but there is not one today and so I'd need to create a new one.


alexcrichton (approved):
Thanks! I've pushed up a few commits here of my own (hope that's ok!) to refactor things along the lines of reducing the amount of `#[cfg]` and additionally ensure that this is tested on CI.

eqrion:
Not a problem, less work for me. Thanks!

PR #1307, opened 2023-11-27
Display magic number mismatch in error

rylev:
I'm not sure if allocations in the unhappy path are something we'd rather avoid, but I ran into an issue where this change would have been somewhat helpful.


alexcrichton (approved):
Thanks! This should be fine yeah, but mind using `#x` as the formatter to prefix the versions with `0x` to avoid hex-vs-decimal confusion?

rylev:
@alexcrichton done!

PR #1308, opened 2023-11-27
Enable new format for embedding component types by default

alexcrichton:
This commit enables the support for the new format of type information embedded in custom sections for `wit-component` first implemented in #1260. The new format minimizes type information to only that which was bound for the `world` selected as opposed to all packages and interfaces regardless of whether they're referenced or not.

A bug with this format was fixed with #1270 and shipped in Wasmtime 15 so there's at least one stable release that supports the old and the new without bugs. This commit turns on the new format by default while leaving in an environment variable to turn it off. This isn't expected to have any practical breakage at this time.


alexcrichton:
I believe that this is the final of the in-flight change to `wasm-tools` so some time in January of the new year I plan on deleting all the supporting code for the old now-phased-out pieces. (e.g. removing all the env-var-based knobs)

pchickey (approved):

PR #1316, opened 2023-12-06
implement from_reader for ComponentInfo for incremental wit decoding

macovedj:
For use cases with memory limitations, this PR enables the wit decoding from a wasm binary without needing to buffer the entirety of the binary.  For the original conversation see the [zulip thread](https://bytecodealliance.zulipchat.com/#narrow/stream/327223-wit-bindgen/topic/wit-decode.20on.20streams/near/397759913)

### DIFF
--- a/crates/wit-component/src/decoding.rs
+++ b/crates/wit-component/src/decoding.rs
@@ -1,39 +1,39 @@
 use anyhow::{anyhow, bail, Context, Result};
 use indexmap::{IndexMap, IndexSet};
-use std::collections::HashMap;
 use std::mem;
+use std::{collections::HashMap, io::Read};
 use wasmparser::{
     names::{ComponentName, ComponentNameKind},
     types, ComponentExport, ComponentExternalKind, ComponentImport, Parser, Payload,
     PrimitiveValType, ValidPayload, Validator, WasmFeatures,
 };
+use wasmparser::{Chunk, ComponentImportName};
 use wit_parser::*;
 
 use crate::encoding::docs::{PackageDocs, PACKAGE_DOCS_SECTION_NAME};
 
 /// Represents information about a decoded WebAssembly component.
-struct ComponentInfo<'a> {
+struct ComponentInfo {
     /// Wasmparser-defined type information learned after a component is fully
     /// validated.
     types: types::Types,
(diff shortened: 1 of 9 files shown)


crates/wit-component/src/decoding.rs
```diff
@@ -1,30 +1,37 @@
...
     package_docs: Option<PackageDocs>,
 }
 
-enum Extern<'a> {
-    Import(ComponentImport<'a>),
-    Export(ComponentExport<'a>),
+struct DecodingExport {
+    name: String,
+    kind: ComponentExternalKind,
+    index: u32,
```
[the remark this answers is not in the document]
macovedj: Yeah I was hoping to get rid of them, but they're used places like [here](https://github.com/bytecodealliance/wasm-tools/blob/main/crates/wit-component/src/decoding.rs#L134) and [here](https://github.com/bytecodealliance/wasm-tools/blob/main/crates/wit-component/src/decoding.rs#L101)

crates/wit-component/src/decoding.rs
```diff
@@ -562,7 +621,7 @@ impl WitPackageDecoder<'_> {
                 created,
             } => {
                 let id = self
-                    .register_type_export(name, owner, referenced, created)
+                    .register_type_export(&name, owner, referenced, created)
```
alexcrichton, line 624: I think the leading ampersand here and above can be removed?

crates/wit-component/src/decoding.rs
```diff
@@ -575,30 +634,30 @@ impl WitPackageDecoder<'_> {
...
             }
             types::ComponentEntityType::Instance(i) => {
                 let ty = &types[i];
                 let (name, id) = if name.contains('/') {
-                    let id = self.register_import(name, ty)?;
+                    let id = self.register_import(&name, ty)?;
                     (WorldKey::Interface(id), id)
                 } else {
-                    self.register_interface(name, ty, package)
+                    self.register_interface(&name, ty, package)
```
alexcrichton, line 660: I think the leading ampersands in this function can be removed as well

alexcrichton (approved):
Thanks!

alexcrichton (approved):

PR #1322, opened 2023-12-07
Allow renaming imports in `ComponentEncoder`. 

peterhuene:
This commit adds an `import_name_map` option to `ComponentEncoder` to allow
renaming imports in the output component.

Language toolchains can use this option to substitute `unlocked-dep` and
`locked-dep` names in the components they build.

PR #1325, opened 2023-12-09
Make import_name CLI arg optional

yowl:
I could be wrong, but I think the intention was for this to be optional in https://github.com/bytecodealliance/wasm-tools/pull/1322

This PR makes it an `Option<>`

### DIFF
--- a/src/bin/wasm-tools/component.rs
+++ b/src/bin/wasm-tools/component.rs
@@ -116,7 +116,7 @@ pub struct NewOpts {
     ///
     /// If the old import name is not found, it is ignored.
     #[clap(long = "import-name", value_name = "[OLD]=NEW", value_parser = parse_import_name)]
-    import_names: Option<HashMap<String, String>>,
+    import_names: Vec<(String, String)>,
 
     #[clap(flatten)]
     io: wasm_tools::InputOutput,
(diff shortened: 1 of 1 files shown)


src/bin/wasm-tools/component.rs
```diff
@@ -116,7 +116,7 @@ pub struct NewOpts {
     ///
     /// If the old import name is not found, it is ignored.
     #[clap(long = "import-name", value_name = "[OLD]=NEW", value_parser = parse_import_name)]
-    import_names: HashMap<String, String>,
+    import_names: Option<HashMap<String, String>>,
```
peterhuene, line 119: I don't think this will work as intended; originally I expected clap to treat `HashMap` as `Vec` with respect to multivalue options.

I think we'll need to have `Vec<(String, String)>` and do a `self.import_names.into_iter().collect()` on it.

peterhuene:
```suggestion
    import_names: Vec<(String, String)>,
```

src/bin/wasm-tools/component.rs
```diff
@@ -153,7 +153,7 @@ impl NewOpts {
         encoder = encoder.realloc_via_memory_grow(self.realloc_via_memory_grow);
 
         let bytes = encoder
-            .import_name_map(self.import_names)
+            .import_name_map(self.import_names.unwrap_or_else(|| HashMap::new()))
```
peterhuene, line 156:
```suggestion
            .import_name_map(self.import_names.into_iter().collect())
```

peterhuene (commented):
@yowl thanks very much for fixing this and sorry I didn't catch it sooner!

I think with the suggested changes we should be good.

peterhuene (approved):

yowl:
Thanks for correcting my newbie rust 😊

PR #410, opened 2021-12-03
`wasm-smith`: `SwarmConfig` should enable bulk memory if enabling reference types

fitzgen:
The reference types proposal depends on the bulk memory proposal.

This fixes #409 because we enable/disable wasm feature validation based on what the `SwarmConfig` has enabled or disabled.

@cfallin mind giving this a quick review?

cc @Jacarte 

Fix #409

PR #1326, opened 2023-12-11
Sync wasmparser/wasm-smith on bulk-memory

### LINKED ISSUE #1324: Smith generates wasm files with elem.drop even with bulk memory disabled
`elem.drop` is part of [bulk memory](https://github.com/WebAssembly/bulk-memory-operations/blob/master/proposals/bulk-memory-operations/Overview.md#instruction-encoding), but it is emitted by this command:

```
$ wasm-tools smith input.dat -t -o /dev/stdout --bulk-memory false | grep elem.drop
      elem.drop 1
      elem.drop 1
        elem.drop 2
      elem.drop 2
            elem.drop 2

```

Tested on latest release, `wasm-tools-1.0.54-x86_64-linux/`.

[input.dat.gz](https://github.com/bytecodealliance/wasm-tools/files/13622209/input.dat.gz)

alexcrichton:
Historically in #410 fuzzing for `wasm-tools smith` was updated to unconditionally enable the bulk-memory proposal if the reference-types proposal was enabled. This doesn't prevent configuration on the CLI, however, as exposed in #1324. This change undoes #410 effectively and then syncs the `wasm-smith` crate's understanding of bulk-memory and reference-types with `wasmparser`'s understanding.

I'm not actually sure if `wasmparser`'s understanding of bulk-memory-vs-reference-types is correct. It was always a bit murky to me precisely which instructions where in which proposal, so I've opted here to take whatever `wasmparser` does today as the source of truth. Any edits necessary to move items between proposals can be done in `wasmparser` and then reflected in `wasm-smith`.

Closes #1324


crates/wasm-smith/src/core.rs
```diff
@@ -1149,11 +1149,13 @@ impl Module {
 
         // Reference types allows us to create passive and declared element
         // segments.
-        if self.config.reference_types_enabled() {
+        if self.config.bulk_memory_enabled() {
```
fitzgen, line 1152: Update comment to account for the new subtlety here?

fitzgen (approved):

PR #1298, opened 2023-11-16
[wast] Rename extern.internalize and extern.externalize

jpages:
The following instructions were renamed in wasm-gc (https://github.com/WebAssembly/gc/pull/447):

extern.internalize -> any.convert_extern
extern.externalize -> extern.convert_any

PR #1327, opened 2023-12-11
Update testsuite submodule to dc27dad3e34e466bdbfea32fe3c73f5e31f88560

fitzgen:


alexcrichton:
I was a bit curious where all the new tests came from and now I'm a bit confused. The [comparison of the submodules](https://github.com/WebAssembly/testsuite/compare/26d62d0ad297b14218807cd20d9b568e11ec7bd9...dc27dad3e34e466bdbfea32fe3c73f5e31f88560) doesn't actually contain any meaningful changes, which makes me curious what's going on here.

I don't think that the snapshot logic accounts for deleted tests well, so are these extra files perhaps leftover on your machine from some previous runs? (I suppose if you delete them and re-run with `BLESS=1` do they reappear?) 

Although you're adding new code here like `AssertMalformedCustom` which I presume you ran into an issue to do that. So now that I type this out, is there perhaps a forgotten submodule update here?

fitzgen:
Not sure what's going on. I blew away the snaphots, ran `git clean -f -x -d` in the submodule, and then re-ran tests with `BLESS=1` and there aren't any changes.

fitzgen:
Also, that comparison between submodules you linked seems wrong? Because I know that there were changes to GC tests, which is the whole reason I'm doing this PR. Not sure what is going on here.

fitzgen:
Or did someone else update the testsuite repo in the last couple weeks?

alexcrichton:
https://github.com/bytecodealliance/wasm-tools/pull/1298 was the latest PR to update the submodule which I think may have included the changes you're looking for? The diff is the difference of the submodule commit on this repo's `main` branch vs the `testsuite` repo's `main` branch.

Updating the submodule locally and blowing away the `tests/snapshots` folder doesn't yield any changes relative to `main` locally for me, so I'm also preplexed as to what's happening...

alexcrichton:
This is adding a `tests/snapshots/testsuite/repos` folder which doesn't currently exist which I think is generated from the update script, so it may be that the `clean` in the submodule itself didn't clean up this folder? I think that may be because `git clean` doesn't remove submodules

alexcrichton:
As in, `tests/testsuite/repos/*` probably has a bunch of stuff for you in the submodule of the `wasm-tools` checkout you have, and if you remove all that manually I think most of these files should disappear.

That being said I think that many changes here may not be necessary after #1298?

PR #1329, opened 2023-12-12
Update skipping of tests from spec test suite

alexcrichton:
* Enable the `obsolete-keywords.wast` test by removing parsing of `anyfunc`.
* Move all tests out of `skip_test` now that everything in the spec testsuite submodule can be parsed with `wast`.
* Update the logic in `skip_validation` slightly.


fitzgen (approved):
Nice

PR #1330, opened 2023-12-12
Use wasm-encoder-defined-conversions in more crates

alexcrichton:
Enable the `wasmparser` feature for a number of crates which convert from `wasmparser` types to `wasm-encoder` types.


fitzgen (approved):
Nice!

PR #1332, opened 2023-12-12
Add `wasmparser` support for `ref.{test,cast}`

fitzgen:
This also splits `wasm-encoder::Intruction::Ref{Test,Cast}` into nullable and
non-nullable forms, since that more directly aligns with the binary format where
there are actually different opcodes for nullable vs non-nullable variations (as
opposed to a bit on some immediate or something).

### DIFF
--- a/crates/wasmparser/src/binary_reader.rs
+++ b/crates/wasmparser/src/binary_reader.rs
@@ -1001,6 +1001,23 @@ impl<'a> BinaryReader<'a> {
     {
         let code = self.read_var_u32()?;
         Ok(match code {
+            0x14 => {
+                let ht = HeapType::from_reader(self)?;
+                visitor.visit_ref_test_non_null(ht)
+            }
+            0x15 => {
+                let ht = HeapType::from_reader(self)?;
+                visitor.visit_ref_test_nullable(ht)
+            }
+            0x16 => {
+                let ht = HeapType::from_reader(self)?;
+                visitor.visit_ref_cast_non_null(ht)
+            }
+            0x17 => {
+                let ht = HeapType::from_reader(self)?;
+                visitor.visit_ref_cast_nullable(ht)
+            }
+
             0x1c => visitor.visit_ref_i31(),
             0x1d => visitor.visit_i31_get_s(),
             0x1e => visitor.visit_i31_get_u(),
--- a/crates/wasmparser/src/binary_reader.rs
+++ b/crates/wasmparser/src/binary_reader.rs
@@ -1001,22 +1001,10 @@ impl<'a> BinaryReader<'a> {
     {
         let code = self.read_var_u32()?;
         Ok(match code {
-            0x14 => {
-                let ht = HeapType::from_reader(self)?;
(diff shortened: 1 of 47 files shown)


crates/wasmparser/src/binary_reader.rs
```diff
@@ -1001,6 +1001,23 @@ impl<'a> BinaryReader<'a> {
...
+                visitor.visit_ref_test_nullable(ht)
+            }
+            0x16 => {
+                let ht = HeapType::from_reader(self)?;
+                visitor.visit_ref_cast_non_null(ht)
+            }
+            0x17 => {
+                let ht = HeapType::from_reader(self)?;
+                visitor.visit_ref_cast_nullable(ht)
+            }
```
alexcrichton, line 1019: I think these should be possible to do `visitor.visit_ref_cast_nullable(self.read()?)` to be a bit more compact

crates/wasmparser/src/validator/operators.rs
```diff
@@ -970,6 +970,48 @@ where
...
 
+    /// Common helper for `ref.test` and `ref.cast` downcasting/checking
+    /// instructions. Returns the given `heap_type` as a `ValType`.
+    fn check_downcast(&mut self, nullable: bool, mut heap_type: HeapType) -> Result<ValType> {
+        self.resources
+            .check_heap_type(&mut heap_type, self.offset)?;
+
+        let sub_ty = RefType::new(nullable, heap_type)
+            .map(ValType::from)
+            .ok_or_else(|| BinaryReaderError::new("TODO FITZGEN", self.offset))?;
```
alexcrichton, line 981: "TODO FITZGEN"

crates/wasmprinter/src/operator.rs
```diff
@@ -316,6 +316,30 @@ macro_rules! define_visit {
...
+        let rty = RefType::new(false, $hty)
+            .ok_or_else(|| anyhow!("implementation limit: type index too large"))?;
+        $self.printer.print_reftype(rty)?;
+    );
+    (payload $self:ident RefCastNullable $hty:ident) => (
+        $self.push_str(" ");
+        let rty = RefType::new(true, $hty)
+            .ok_or_else(|| anyhow!("implementation limit: type index too large"))?;
+        $self.printer.print_reftype(rty)?;
+    );
```
alexcrichton, line 342: Huh. Interesting! I think this is the first instance of two different opcodes having the same name. That seems not great from an AST/tooling perspective... you wouldn't happen to know if this is being bikeshedded and/or discussed elsewhere would you?

alexcrichton (approved):

crates/wasmprinter/src/operator.rs
fitzgen: I wasn't aware of any discussion and didn't find anything when searching. It is definitely a bit funky.

fitzgen: Filed https://github.com/WebAssembly/gc/issues/491 to ask for some background

PR #1333, opened 2023-12-12
Update the exception-handling validator implementation

alexcrichton:
This commit updates the implementation of the exception-handling proposal to the latest version of the proposal. This was already implemented in the text format but all other crates needed updating. The changes were:

* A new `exn` heap type is added for `exnref`. Currently this isn't a subtype with anything else, and this probably isn't correct but it's at least conservative for now.

* The `try`, `delegate`, `catch`, `catch_all`, and `rethrow` instructions were all removed as they're no longer present.

* New `try_table` and `throw_ref` instructions were added.

* Support for the name section subsection for tags has been added along with printing tag names in the text format.

* All exception-handling spec tests are now re-enabled.


crates/wast/src/core/resolve/names.rs
```diff
@@ -514,25 +514,25 @@ impl<'a, 'b> ExprResolver<'a, 'b> {
...
                 for catch in &mut try_table.catches {
                     if let Some(tag) = catch.kind.tag_index_mut() {
                         self.resolver.resolve(tag, Ns::Tag)?;
                     }
                     self.resolve_label(&mut catch.label)?;
                 }
+                self.blocks.push(ExprBlock {
+                    label: try_table.block.label,
+                    pushed_scope: false,
+                });
```
alexcrichton, line 535: cc @eqrion on this change because this will change the text-to-wasm translation for `try_table`. I believe this is fixing a mistake because previously labels reference by name were off-by-one because the `self.blocks` depth was modified before the catch labels were resolved. By modifying them afterwards I think it fixes that.

crates/wasmparser/src/validator.rs
```diff
@@ -352,6 +352,15 @@ impl WasmFeatures {
                     Err("heap types not supported without the gc feature")
                 }
             }
+
+            // These types were added in the exception-handling proposal.
+            (HeapType::Exn, _) => {
+                if self.exceptions {
+                    Ok(())
+                } else {
+                    Err("heap types not supported without the gc feature")
```
fitzgen, line 361:
```suggestion
                    Err("exception refs not supported without the exception handling feature")
```

crates/wasmparser/src/validator/types.rs
```diff
@@ -2708,6 +2708,10 @@ impl TypeList {
             | (HT::Struct, _)
             | (HT::Array, _)
             | (HT::I31, _) => false,
+
+            // TODO: this probably isn't right, this is probably related to some
+            // gc type.
+            (HT::Exn, _) => false,
```
fitzgen, line 2714: I expect either there will be a `NoExn` bottom type and this will be a fourth type hierarchy or this will slot under `Any`, but until those wrinkles get worked out, this is fine.

fitzgen (approved):
I expect that Firefox folks might want `wasmprinter` and `wast` to continue to support the old exception instructions for the indefinite future, which would mean that `wasmparser` would also have to support them for the indefinite future. Would be good to get confirmation/clarification of this from them before merging.

osa1:
> I expect that Firefox folks might want wasmprinter and wast to continue to support the old exception instructions for the indefinite future

FWIW, V8 main branch currently supports both old and new instructions.

I don't know how long they plan to support both versions.

alexcrichton:
AFAIK no one's using the historical version of the validation in wasmparser. Wasmtime for example has never had an implementation of the exception handling proposal. In that sense while it would make sense to keep the text parsing if Firefox is using it I'm not sure it's necessary to keep anything else if there aren't users.

crates/wast/src/core/resolve/names.rs
eqrion: With your change, is it possible for a `catch` to refer to the `try_table` itself? i.e `try_table $me (catch_all $me)`.

My understanding of the instruction would be that this is allowed. But the spec appears to match your implementation here and validate the catch clauses without the try_table added to the labels vector. [1]

Gonna file an upstream issue.

[1] https://webassembly.github.io/exception-handling/core/valid/instructions.html#xref-syntax-instructions-syntax-instr-control-mathsf-try-table-xref-syntax-instructions-syntax-blocktype-mathit-blocktype-xref-syntax-instructions-syntax-catch-mathit-catch-ast-xref-syntax-instructions-syntax-instr-mathit-instr-ast-xref-syntax-instructions-syntax-instr-control-mathsf-end

crates/wasmparser/src/validator/types.rs
eqrion: My understanding of the longer term plan is for it be a fourth type hierarchy. Having it under `any` adds some difficulties for AOT translators like wasm2c which want it to have a different representation than any. There would also be some difficulties with JS thrown non-object values if under `any`.

eqrion:
> AFAIK no one's using the historical version of the validation in wasmparser. Wasmtime for example has never had an implementation of the exception handling proposal. In that sense while it would make sense to keep the text parsing if Firefox is using it I'm not sure it's necessary to keep anything else if there aren't users.

Firefox will need the text parsing of legacy instructions around for the forseeable future. We don't use wasmparser for anything.

I do personally use wasm-tools to disassemble binaries from the wild, so would like if at-least the binary to text pipeline continues to work.

eqrion:
> > I expect that Firefox folks might want wasmprinter and wast to continue to support the old exception instructions for the indefinite future
> 
> FWIW, V8 main branch currently supports both old and new instructions.
> 
> I don't know how long they plan to support both versions.

The browser vendors will keep supporting both versions until usage of the old drops to the point of being able to be deprecated. That's a fuzzy line though, and could take many years or never be achievable.

crates/wast/src/core/resolve/names.rs
eqrion: https://github.com/WebAssembly/exception-handling/issues/286

alexcrichton: Ah yeah I was also suprised as I expected to be able to catch-to-yourself but the spec tests required otherwise which is why I changed this. I'd be happy to update one way or another depending on the resolution of your issue, thanks for filing that!

alexcrichton:
Ok I'll look to preserve the text parsing, although I'll perhaps remove some of the more sugared forms such as `(try (do ..` since removing that simplified things greatly. (unless y'all need that in Firefox?). I'll remove all validation from `wasmparser` since no one needs it (I think) and it shouldn't be too bad to keep the instructions around for awhile and print them as well.

I'll get back to this in the new year.

crates/wast/src/core/resolve/names.rs
eqrion: @alexcrichton Reading the room in the issue, not sure it's going to change. So I'd say merging this so it matches the spec is the best way forward.

eqrion:
> Ok I'll look to preserve the text parsing, although I'll perhaps remove some of the more sugared forms such as `(try (do ..` since removing that simplified things greatly. (unless y'all need that in Firefox?). I'll remove all validation from `wasmparser` since no one needs it (I think) and it shouldn't be too bad to keep the instructions around for awhile and print them as well.
> 
> I'll get back to this in the new year.

I don't think we need the sugared forms. If we do use them in tests, I can rewrite them to avoid it. Thanks!

eqrion:
```rust
impl<'a> Parse<'a> for TryTable<'a> {
    fn parse(parser: Parser<'a>) -> Result<Self> {
        let block = parser.parse()?;
        let mut catches = Vec::new();
        while parser.peek2::<kw::catch>()?
            || parser.peek2::<kw::catch_ref>()?
            || parser.peek2::<kw::catch_all>()?
            || parser.peek2::<kw::catch_all_ref>()?
        {
            catches.push(parser.parens(|p| {
```

This one's my fault, unrelated to your PR here, the above code is incorrect for the following test case:
```
try_table
  try
  catch_all
  end
end
```

peek2 in try_table will see catch_all and think that it should be parsing a parentheses.

Would you be able to fix that as well, while you're here?

alexcrichton:
> I'll get back to this in the new year.

Ok I've added a commit now which means that the PR now should only remove the s-expression-based form of the old opcodes (e.g. `(try (do ...`) as well as validation. Otherwise parsing the text format, parsing the binary format, and printing to the text format should all continue to work.

> Would you be able to fix that as well, while you're here?

Sure!

I suspect it'd still fall over on `try_table (catch_all)` which technically is valid if both the old-and-the-new are supported, but unless you need support for that I think I'll probably just leave that in.

alexcrichton:
Oh oops here I was waiting for an approval I already got, I'll queue this up for merge now

eqrion:
@alexcrichton Would you be able to spin up a release sometime for this?

alexcrichton:
Sure! I'll do that in https://github.com/bytecodealliance/wasm-tools/pull/1366

PR #1334, opened 2023-12-12
Print the name of label targets

### LINKED ISSUE #841: Block names for `br` are not preserved

alexcrichton:
This commit updates `wasmprinter` to print the target label name for instructions such as `br` and friends. This is done by maintaining a stack matching the stack of blocks that are entered, and the relative depth is looked up in this stack to then consult the name label name map contained in the label name subsection.

Closes #841

### DIFF
--- a/crates/wasmprinter/src/lib.rs
+++ b/crates/wasmprinter/src/lib.rs
@@ -62,7 +62,6 @@ struct CoreState {
     tags: u32,
     globals: u32,
     tables: u32,
-    labels: u32,
     modules: u32,
     instances: u32,
     func_names: HashMap<u32, Naming>,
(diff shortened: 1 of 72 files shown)


pchickey (approved):
These comments should win some sort of award

d3lm:
Oh wow, thanks for implementing this @alexcrichton 🙏

PR #1337, opened 2023-12-13
Enable the `ref_{test,cast}.wast` GC spec tests

fitzgen:
This involved implementing a few new instructions:

* `struct.new_default`
* `array.new_default`
* `any.convert_extern`
* `extern.convert_any`

These new instructions will be more thoroughly tested when I get to enabling their relevant spec tests (which require even more new instructions).


crates/wasmparser/src/resources.rs
```diff
@@ -213,10 +251,10 @@ pub trait WasmModuleResources {
     /// The global's value type must be canonicalized.
     fn global_at(&self, at: u32) -> Option<GlobalType>;
 
-    /// Returns the `FuncType` associated with the given type index.
+    /// Returns the `SubType` associated with the given type index.
     ///
-    /// The function type must be canonicalized.
-    fn func_type_at(&self, type_idx: u32) -> Option<&Self::FuncType>;
+    /// The sub type must be canonicalized.
+    fn sub_type_at(&self, type_index: u32) -> Option<&Self::SubType>;
```
alexcrichton, line 257: I feel like we might have talked about this in the past but I think it's probably time to remove these `*Resources` traits if we can since I don't think anything is actually using them. Especially as the type system gets more complicated with GC. Their intention was always to allow recording type information externally and then using that external type information to validate (which may not necessarily be wasmparser's type information), but I don't think there's any current users of that.

All-in-all I'd personally be more in favor of "just" returning `&wasmparser::SubType` here and nixing all the traits/impls.

crates/wasmparser/src/validator/operators.rs
```diff
@@ -3442,6 +3473,79 @@ where
...
+            if !val_ty.is_defaultable() {
+                bail!(
+                    self.offset,
+                    "invalid `struct.new_default`: {val_ty} field is not defaultable"
+                );
+            }
+        }
+
+        let mut heap_ty = HeapType::Concrete(UnpackedIndex::Module(type_index));
+        self.resources.check_heap_type(&mut heap_ty, self.offset)?;
```
alexcrichton, line 3490: I think this can be skipped since `struct_type_at` above would have otherwise failed?

crates/wasmparser/src/validator/operators.rs
```diff
@@ -3442,6 +3473,79 @@ where
...
+        let val_ty = field.element_type.unpack();
+        if !val_ty.is_defaultable() {
+            bail!(
+                self.offset,
+                "invalid `array.new_default`: {val_ty} field is not defaultable"
+            );
+        }
+
+        let mut heap_ty = HeapType::Concrete(UnpackedIndex::Module(type_index));
+        self.resources.check_heap_type(&mut heap_ty, self.offset)?;
```
alexcrichton, line 3510: Like above I think this can be skipped

crates/wasmparser/src/validator/operators.rs
```diff
@@ -3442,6 +3473,79 @@ where
...
+            Some(r) if self.resources.top_type(&r.heap_type()) == HeapType::Extern => r,
+            Some(r) => bail!(
+                self.offset,
+                "type mismatch: expected (ref null? extern), found {r}"
+            ),
+            None => bail!(
+                self.offset,
+                "type mismatch: expected (ref null? extern), found bottom"
+            ),
+        };
```
alexcrichton, line 3530: I'm a little surprised that subtyping checks didn't kick in here, could this not be `self.pop_operand` with `RefType::EXTERNREF` or similar?

alexcrichton: (same below for popping `anyref`)

alexcrichton (approved):

crates/wasmparser/src/validator/operators.rs
fitzgen: This does take subtyping into account: when checking if `t <: Any` we can just do `top(t) == Any`, which is going to be faster.

We can't use `self.pop_operand(Some(RefType::EXTERNREF))` because these instructions are polymorphic over nullability. We don't know ahead of time whether to expect a nullable or non-nullable reference.

crates/wasmparser/src/validator/operators.rs
fitzgen: This is to canonicalize the module index into an id. I'll add a comment to this effect.

crates/wasmparser/src/validator/operators.rs
alexcrichton: Oh sorry I mean moreso for I would have expected that `self.pop_operand` to work since it has built-in subtyping checks. I'm not sure what you mean though precisely about the instructions being polymorphic? If it's what I think though `pop_operand` returns the precise type which I think can be tested for being nullable or non-nullable?

crates/wasmparser/src/resources.rs
fitzgen: I thought maybe `wasmi` used it, but it seems not.

I'll do this in a follow up PR.

crates/wasmparser/src/validator/operators.rs
fitzgen: Its that we can't pass in an expected type to pop because we can pop either a `(ref extern)` or a `(ref null extern)`

alexcrichton: Isn't `(ref extern)` a subtype of `(ref null extern)` though so you could pop a `(ref null extern)` and see what came out?

fitzgen: Ah now I see what you are getting at. Pushed a new commit that does this.

PR #1339, opened 2023-12-17
wit-component: Re-export `_start` while shared-everything link

kateinoigakukun:
This is a bit of a hack to componentize a WASI preview1 command module with shared-everything linking. The `_start` function was not being exported, so `wasm-tools component link` was failing to link modules with `wasi_snapshot_preview1.command.wasm` adapter.


dicej (approved):
Makes sense to me.  Thanks for the PR!

PR #1341, opened 2023-12-18
wit-component: Reserve table offset 0 for the null function pointer

kateinoigakukun:
Placing a function at zero address violates most application's assumptions about null pointers. This patch reserves table offset 0 for the null function pointer. This convention follows wasm-ld's table layout.

CC: @dicej


dicej (commented):
LGTM.  Would you please update the tests so CI is green?  It should just be a matter of running `BLESS=1 cargo test --all` and checking in the results.

kateinoigakukun:
Sure!

dicej (approved):

PR #1342, opened 2023-12-18
`wasmparser`: Get the `array.wast` GC spec test passing

fitzgen:
This involved adding support for the following instructions:

* `array.new`
* `array.new_data`
* `array.new_elem`
* `array.new_fixed`
* `array.get`
* `array.get_s`
* `array.get_u`
* `array.set`
* `array.len`


crates/wasmprinter/src/operator.rs
```diff
@@ -424,10 +440,37 @@ macro_rules! define_visit {
...
+    (payload $self:ident ArrayNewElem $type_index:ident $elem_index:ident) => (
+        $self.type_index_no_type($type_index)?;
+        $self.elem_index_no_elem($elem_index)?;
+    );
+    (payload $self:ident ArrayGet $type_index:ident) => (
+        $self.type_index_no_type($type_index)?;
+    );
+    (payload $self:ident ArrayGetS $type_index:ident) => (
+        $self.type_index_no_type($type_index)?;
+    );
```
alexcrichton, line 469: To avoid needing to list all the instructions manually could these all have a slightly different name for their field? Maybe something like `array_type_index` or something like that? Either way works but the intention of these macros was to avoid case-per-instruction variants where possible.

alexcrichton (approved):

PR #1346, opened 2023-12-21
`wasmparser`: Get more GC spec tests passing

fitzgen:
Newly passing tests:

* `br_on_cast.wast`
* `br_on_cast_fail.wast`
* `struct.wast`

Newly implemented instructions:

* `struct.new`
* `struct.get`
* `struct.get_s`
* `struct.get_u`
* `struct.set`
* `br_on_cast`
* `br_on_cast_fail`

### DIFF
--- a/crates/wasm-encoder/src/core/code.rs
+++ b/crates/wasm-encoder/src/core/code.rs
@@ -1,4 +1,4 @@
-use crate::{encode_section, Encode, HeapType, Section, SectionId, ValType};
+use crate::{encode_section, Encode, HeapType, RefType, Section, SectionId, ValType};
 use std::borrow::Cow;
 
 /// An encoder for the code section.
@@ -593,17 +593,13 @@ pub enum Instruction<'a> {
     RefCastNullable(HeapType),
     BrOnCast {
         relative_depth: u32,
-        from_type_nullable: bool,
-        from_heap_type: HeapType,
-        to_type_nullable: bool,
-        to_heap_type: HeapType,
+        from_ref_type: RefType,
+        to_ref_type: RefType,
     },
     BrOnCastFail {
         relative_depth: u32,
-        from_type_nullable: bool,
-        from_heap_type: HeapType,
-        to_type_nullable: bool,
-        to_heap_type: HeapType,
+        from_ref_type: RefType,
+        to_ref_type: RefType,
     },
     AnyConvertExtern,
     ExternConvertAny,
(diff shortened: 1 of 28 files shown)


fitzgen:
Added another commit and now all GC spec tests are passing!

crates/wasmprinter/src/operator.rs
```diff
@@ -238,6 +238,11 @@ impl<'a, 'b> PrintOperator<'a, 'b> {
         self.printer.print_idx(&self.state.core.type_names, idx)
     }
 
+    fn field_index(&mut self, idx: u32) -> Result<()> {
+        write!(&mut self.printer.result, "{idx}")?;
+        Ok(())
+    }
```
alexcrichton, line 244: Would you be up for threading field names through here? (or is there not a subsection of the `name` custom section to specify field names yet?)

crates/wasmprinter/src/operator.rs
```diff
@@ -471,6 +476,48 @@ macro_rules! define_visit {
...
+        $self.relative_depth($relative_depth)?;
+        $self.push_str(" ");
+        let from_ty = RefType::new($from_type_nullable, $from_heap_type)
+            .ok_or_else(|| anyhow!("implementation limit: type index too large"))?;
+        $self.printer.print_reftype(from_ty)?;
+        $self.push_str(" ");
+        let to_ty = RefType::new($to_type_nullable, $to_heap_type)
+            .ok_or_else(|| anyhow!("implementation limit: type index too large"))?;
+        $self.printer.print_reftype(to_ty)?;
+    );
```
alexcrichton, line 499: Using `RefType` as the intermediate storage for the type-related fields might help remove this special case I think

alexcrichton (approved):
🎊

crates/wasm-encoder/src/core/code.rs
```diff
@@ -575,11 +587,24 @@ pub enum Instruction<'a> {
...
     RefTestNonNull(HeapType),
     RefTestNullable(HeapType),
     RefCastNonNull(HeapType),
     RefCastNullable(HeapType),
+    BrOnCast {
+        relative_depth: u32,
+        from_type_nullable: bool,
+        from_heap_type: HeapType,
+        to_type_nullable: bool,
+        to_heap_type: HeapType,
```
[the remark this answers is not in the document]
fitzgen: This would mean enforcing the index implementation limit during decoding rather than validation, but I think that's fine and I think we do it elsewhere already, so not a big deal.

crates/wasmprinter/src/operator.rs
fitzgen: There is a fields subsection of the name custom section in the spec appendix, but we don't parse it at all in `wasmprinter` yet. I'd prefer to leave this for a follow up PR.

PR #1348, opened 2024-01-01
Add parsing for `linking` custom section

### LINKED ISSUE #785: Full support for parsing the linking section
There is a `LinkingSectionReader` type, but it seems to be for a severely outdated version of the tool conventions and barely parses anything.

Likely necessary for https://github.com/gimli-rs/object/issues/471.

andylizi:
Closes #785. Only metadata version 2 is supported.

I'm not sure how to add tests for this. #1135 chose to invent a new WAST format for the custom section, which I'd rather avoid doing myself. I tried the parser locally on `WASM_SYMBOL_TABLE` and it works fine, but I don't know where to find wasm files with `WASM_SEGMENT_INFO`, etc.


crates/wasmparser/src/readers/core/linking.rs
```diff
@@ -0,0 +1,405 @@
...
+    pub fn range(&self) -> Range<usize> {
+        self.range.clone()
+    }
+
+    /// Returns the iterator for advancing through the subsections.
+    ///
+    /// You can also use [`IntoIterator::into_iter`] directly on this type.
+    pub fn subsections(&mut self) -> &mut Subsections<'a, Linking<'a>> {
+        &mut self.subsections
+    }
```
alexcrichton, line 395: Could this return a fresh copy of the iterator rather than a mutable view into the internals? Idiomatically most of `wasmparser` tries to avoid internal mutation and return sub-view except for `Iterator`-like things on `next` methods.

alexcrichton (commented):
Thanks for this! One minor suggestion API-wise but otherwise looks good to me 👍 

For testing I agree it's probably best to avoid inventing text syntax (in retrospect not my best idea). Although that being said if an annotation such as `@wasm-tools-linking` or something like that was added to indicate it's specific to this tool that may work out well enough too. Either way though it's ok to skip the test side for now.

andylizi:
Done!

By the way, now that we have more significant uses for symbol flags `WASM_SYM_*`, I'd like to change them to `bitflags`, as suggested in https://github.com/bytecodealliance/wasm-tools/pull/1135#discussion_r1275448580. But that'd involve breaking changes like `pub flags: u32` to `pub flags: <newtype>`. Would that be okay?

crates/wasmparser/src/readers/core/linking.rs
```diff
@@ -0,0 +1,405 @@
...
+    /// Creates a new reader for the linking section contents starting at
+    /// `offset` within the original wasm file.
+    pub fn new(data: &'a [u8], offset: usize) -> Result<Self> {
+        let mut reader = BinaryReader::new_with_offset(data, offset);
+        let range = reader.range();
+
+        let version = reader.read_var_u32()?;
+        if version != 2 {
+            return Err(BinaryReaderError::new(
+                format!("unsupported linking section version: {}", version),
```
andylizi, line 361: Do you think it'd be useful to provide an easier way for users to differentiate between
* the `linking` section is there, but its contents are malformed — we report an error and abort
* the `linking` section is there, but we don't support parsing this version — we print a warning and continue

alexcrichton (approved):

alexcrichton:
> I'd like to change them to bitflags

That'd be great, thanks! That's ok to happen at any time, this crate regularly has breaking API changes (e.g. we have an `enum` of all wasm opcodes and those change regularly as proposals are implemented).

crates/wasmparser/src/readers/core/linking.rs
alexcrichton: Up to you! Currently `wasmparser` doesn't do a great job of making errors introspectable though, right now they're all simply treated as "fatal please print somewhere and move on" as opposed to being specifically actionable one way or another.

alexcrichton:
Also I'm happy to merge this as-is if you'd prefer to leave those other items as follow-up PRs, or I don't mind holding this off if you'd like to implement them. Let me know which you'd prefer

andylizi:
Thanks for the explanation! I don't mind merging as-is if you don't mind `flags: u32` being temporary. I just realized my use case actually needs `reloc.*` as well, so I'll probably do that next :joy:.

alexcrichton:
Sounds good!

jbms:
@andylizi Are you still working on a Reloc section reader?  Otherwise I may submit one.  I assume it would be a welcome addition to this repo?

andylizi:
> Are you still working on a Reloc section reader? Otherwise I may submit one.

@jbms I haven't made any progress on it for a long time since I've been busy with other projects. It'd be great if anyone wants to work on it instead!

Below is my unfinished take on this, feel free to build on it or start fresh. No citations necessary.

<details>
  <summary>Click to expand: <code>reloc.rs</code></summary>

  ```rust
use std::cmp;

use crate::{BinaryReader, FromReader, Result, SectionLimited};

// Credit: https://stackoverflow.com/a/57578431
macro_rules! back_to_enum {
    ($(#[$meta:meta])* $vis:vis enum $name:ident {
        $($(#[$vmeta:meta])* $vname:ident $(= $val:expr)?,)*
    }) => {
        $(#[$meta])*
        $vis enum $name {
            $($(#[$vmeta])* $vname $(= $val)?,)*
        }

        impl ::std::convert::TryFrom<u32> for $name {
            type Error = ();

            #[inline]
            fn try_from(v: u32) -> ::std::result::Result<Self, Self::Error> {
                match v {
                    $(x if x == Self::$vname as u32 => Ok(Self::$vname),)*
                    _ => Err(()),
                }
            }
        }
    }
}

// https://github.com/llvm/llvm-project/blob/llvmorg-17.0.6/llvm/include/llvm/BinaryFormat/WasmRelocs.def
back_to_enum! {
    /// Represent the type of relocation.
    #[allow(missing_docs)]
    #[repr(u32)]
    #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
    pub enum RelocationType {
        /// A function index used for the immediate argument of a `call` instruction.
        FunctionIndexLeb = 0,
        /// A function table index used to refer to the immediate argument of a `i32.const`` instruction, e.g. taking the address of a function.
        TableIndexSleb = 1,
        /// A function table index, e.g. taking the address of a function in a static data initializer.
        TableIndexI32 = 2,
        /// A linear memory index used for the immediate argument of a `load` or `store` instruction, e.g. directly loading from or storing to a C++ global.
        MemoryAddrLeb = 3,
        /// A linear memory index used for the immediate argument of a `i32.const` instruction, e.g. taking the address of a C++ global.
        MemoryAddrSleb = 4,
        /// A linear memory index, e.g. taking the address of a C++ global in a static data initializer.
        MemoryAddrI32 = 5,
        /// A type index, e.g. the type immediate in a `call_indirect` instruction.
        TypeIndexLeb = 6,
        /// A global index, e.g. the index immediate in a `get_global` instruction.
        GlobalIndexLeb = 7,
        /// A byte offset within code section for the specific function. The offsets start at the actual function code excluding its size field.
        FunctionOffsetI32 = 8,
        /// A byte offset from start of the specified section.
        SectionOffsetI32 = 9,
        /// An event index used for the immediate argument of a `throw` or `if_except` instruction.
        TagIndexLeb = 10,
        MemoryAddrRelSleb = 11,
        TableIndexRelSleb = 12,
        /// A global index.
        GlobalIndexI32 = 13,
        /// A 64-bit linear memory index used for the immediate argument of a `load` or `store` instruction on a 64-bit linear memory array. The counterpart of [`Self::MemoryAddrLeb`].
        MemoryAddrLeb64 = 14,
        /// A 64-bit linear memory index used for the immediate argument of a `i64.const` instruction. The counterpart of [`Self::MemoryAddrSleb`].
        MemoryAddrSleb64 = 15,
        /// A 64-bit linear memory index, e.g. taking the address of a C++ global in a static data initializer. The counterpart of [`Self::MemoryAddrI32`].
        MemoryAddrI64 = 16,
        MemoryAddrRelSleb64 = 17,
        /// A function table index used to refer to the immediate argument of a `i64.const` instruction, e.g. taking the address of a function in Wasm64. The counterpart of [`Self::TableIndexSleb`].
        TableIndexSleb64 = 18,
        /// A function table index, e.g. taking the address of a function in a static data initializer. The counterpart of [`Self::TableIndexI32`].
        TableIndexI64 = 19,
        /// A table number used for the table immediate argument in `table.*` instructions.
        TableNumberLeb = 20,
        MemoryAddrTlsSleb = 21,
        FunctionOffsetI64 = 22,
        MemoryAddrLocrelI32 = 23,
        TableIndexRelSleb64 = 24,
        MemoryAddrTlsSleb64 = 25,
        FunctionIndexI32 = 26,
    }
}

/// A reader for the `reloc.*` custom sections of a WebAssembly module.
///
/// This format is currently defined upstream at
/// <https://github.com/WebAssembly/tool-conventions/blob/main/Linking.md>.
#[derive(Debug, Clone)]
pub struct RelocSectionReader<'a> {
    section: u32,
    entries: RelocationEntryReader<'a>,
}

/// Represents a relocation entry from `reloc.*` sections.
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub struct RelocationEntry {
    /// The type of this entry.
    pub ty: RelocationType,
    /// The offset of the value to rewrite.
    ///
    /// Note: This is relative to the relevant section's contents.
    /// Offset zero is immediately after the id and size of the section.
    pub offset: u32,
    /// The index of the symbol used (or, for [`RelocationType::TypeIndexLeb`] relocations, the index of the type).
    pub index: u32,
    /// The addend to add to the address, zero if not applicable.
    pub addend: i64,
}

/// Represents a reader for relocation entries from `reloc.*` sections.
pub type RelocationEntryReader<'a> = SectionLimited<'a, RelocationEntry>;

impl FromReader<'_> for RelocationEntry {
    fn from_reader(reader: &mut BinaryReader<'_>) -> crate::Result<Self> {
        let ty_pos = reader.original_position();
        let ty = reader.read_var_u32()?;
        let ty = RelocationType::try_from(ty)
            .map_err(|_| format_err!(ty_pos, "invalid relocation type {}", ty))?;

        let offset = reader.read_var_u32()?;
        let index = reader.read_var_u32()?;

        let addend_pos = reader.original_position();
        let addend = if ty.has_addend() {
            let addend = reader.read_var_i64()?;
            if !ty.has_addend_64() && i32::try_from(addend).is_err() {
                bail!(addend_pos, "addend too large for var_i32");
            }
            addend
        } else {
            0
        };

        Ok(Self {
            ty,
            offset,
            index,
            addend,
        })
    }
}

impl PartialOrd for RelocationEntry {
    fn partial_cmp(&self, other: &Self) -> Option<cmp::Ordering> {
        Some(self.cmp(other))
    }
}

impl Ord for RelocationEntry {
    fn cmp(&self, other: &Self) -> cmp::Ordering {
        self.offset.cmp(&other.offset)
    }
}

impl RelocationType {
    /// Returns the number of bytes being relocated for this relocation type.
    pub const fn size(self) -> u32 {
        // Values taken from:
        // https://github.com/llvm/llvm-project/blob/llvmorg-17.0.6/llvm/lib/Object/WasmObjectFile.cpp#L1043
        use RelocationType as R;
        match self {
            R::MemoryAddrLeb64 | R::MemoryAddrSleb64 | R::MemoryAddrRelSleb64 => 10,
            R::TableIndexI32
            | R::MemoryAddrI32
            | R::MemoryAddrLocrelI32
            | R::SectionOffsetI32
            | R::FunctionOffsetI32
            | R::FunctionIndexI32
            | R::GlobalIndexI32 => 4,
            R::TableIndexI64 | R::MemoryAddrI64 | R::FunctionOffsetI64 => 8,
            _ => 5,
        }
    }

    const fn has_addend(self) -> bool {
        // Values taken from:
        // https://github.com/llvm/llvm-project/blob/llvmorg-17.0.6/llvm/lib/BinaryFormat/Wasm.cpp#L66
        use RelocationType as R;
        matches!(
            self,
            R::MemoryAddrLeb
                | R::MemoryAddrLeb64
                | R::MemoryAddrSleb
                | R::MemoryAddrSleb64
                | R::MemoryAddrRelSleb
                | R::MemoryAddrRelSleb64
                | R::MemoryAddrI32
                | R::MemoryAddrI64
                | R::MemoryAddrTlsSleb
                | R::MemoryAddrTlsSleb64
                | R::FunctionOffsetI32
                | R::FunctionOffsetI64
                | R::SectionOffsetI32
                | R::MemoryAddrLocrelI32
        )
    }

    const fn has_addend_64(self) -> bool {
        use RelocationType as R;
        matches!(
            self,
            R::MemoryAddrLeb64
                | R::MemoryAddrSleb64
                | R::MemoryAddrRelSleb64
                | R::MemoryAddrI64
                | R::MemoryAddrTlsSleb64
                | R::FunctionOffsetI64
        )
    }
}

impl<'a> RelocSectionReader<'a> {
    /// Creates a new reader for the `reloc.*` section contents starting at
    /// `offset` within the original wasm file.
    pub fn new(data: &'a [u8], offset: usize) -> Result<Self> {
        let mut reader = BinaryReader::new_with_offset(data, offset);
        let section = reader.read_var_u32()?;
        let entries =
            RelocationEntryReader::new(reader.remaining_buffer(), reader.original_position())?;

        Ok(Self { section, entries })
    }

    /// Returns the index of the target section.
    pub fn section_index(&self) -> u32 {
        self.section
    }

    /// Returns the iterator for advancing through the entries.
    pub fn entries(&self) -> RelocationEntryReader<'a> {
        self.entries.clone()
    }
}
  ```
</details>

jbms:
Thanks, created PR here: https://github.com/bytecodealliance/wasm-tools/pull/1511

PR #1350, opened 2024-01-03
Use `bitflags` for symbol flags

andylizi:
Follow-up on https://github.com/bytecodealliance/wasm-tools/pull/1348#issuecomment-1874687578.

### DIFF
--- a/crates/wast/src/core/custom.rs
+++ b/crates/wast/src/core/custom.rs
@@ -360,6 +360,7 @@ fn parse_sym_flags(parser: Parser<'_>) -> Result<u32> {
                 }
             })*};
         }
+        // N.B.: Keep in sync with `print_dylink0_flags` in `crates/wasmprinter/src/lib.rs`.
         parse_flags! {
             "binding-weak" = 1 << 0,
             "binding-local" = 1 << 1,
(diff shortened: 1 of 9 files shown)


crates/wit-component/src/linking/metadata.rs
```diff
@@ -131,7 +131,7 @@ pub struct Import<'a> {
     pub module: &'a str,
     pub name: &'a str,
     pub ty: Type,
-    pub flags: u32,
+    pub flags: SymbolFlags,
```
andylizi, line 134: I'm not familiar with the component model, is there a reason we need to use `BTreeSet` to keep imports & exports sorted? It's indirectly requiring `SymbolFlags` to also derive `Ord`, which I'd prefer to avoid as it doesn't make sense semantically.

crates/wast/src/core/custom.rs
```diff
@@ -368,6 +369,8 @@ fn parse_sym_flags(parser: Parser<'_>) -> Result<u32> {
             "exported" = 1 << 5,
             "explicit-name" = 1 << 6,
             "no-strip" = 1 << 7,
+            "tls" = 1 << 8,
+            "absolute" = 1 << 9,
```
andylizi, line 373: Since the `wast` crate doesn't depend on `wasmparser`, I'm assuming there's no way to avoid the code duplication here?

alexcrichton: Unfortunately, yeah, but that's ok for now

crates/wit-component/src/linking/metadata.rs
alexcrichton: Yes and no, it depends on the location, but it's probably easiest to just leave `Ord` in there

alexcrichton (approved):
Thanks for this!

PR #1351, opened 2024-01-03
`wasm-smith`: Deduplicate and devirtualize configuration

fitzgen:
This makes it so that we don't have three or four places where each config option needs to be enumerated anymore and also removes the `Config` trait and replaces it with a `Config` struct (what used to be the `SwarmConfig` struct). This struct implements `Arbitrary` and therefore you can still do swarm testing by generating an arbitrary config and then using that to generate a module. There is also no difference between `Module` and `ConfiguredModule` anymore.

Overall, I feel like this is a big clean up, and also should in theory make module generation a little faster since there won't be so many indirect calls.


alexcrichton (approved):
Nice!

PR #1355, opened 2024-01-04
Document using wit-components for C and C++ programs.

sunfishcode:
Add instructions and an example to wit-component's README.md showing how to use it for C/C++ programs.


alexcrichton (commented):
Thanks! Would this perhaps be best on the `wit-bindgen` repository though rather than here in `wit-component`? This README is somewhat buried and doesn't cover any other languages for example, whereas `wit-bindgen` seems like a more prominent location with room to grow for other languages too

PR #1356, opened 2024-01-04
Accept `f32`/`f64` as aliases for `float32`/`float64`.

sunfishcode:
In WebAssembly/component-model#277 there seems to be consensus emerging to rename `float32`/`float64` to `f32`/`f64`. This PR just adds support for parsing `f32`/`f64`, and changes nothing else for now, to start preparing for this change.


esoterra (approved):

alexcrichton:
Before committing to this could a rough plan of how to make this transition be written down first? I suspect it'd look pretty similar to other historical transition plans, but it'd be good to have it in words somewhere. I'm retroactively wishing we had a list of active transitions happening at any one point in time because I've long forgotten what's needed any more at this point, but that can be handled later.

Can you additionally add some tests which assert that the syntax parses and works correctly?

esoterra:
Maybe we could get the `;`, `///`, and `f32`/`f64` stuff put onto some sort of combined WIT transitions roadmap.

I think they're the main changes and we can probably try to synchronize them to some extent.

sunfishcode:
> Before committing to this could a rough plan of how to make this transition be written down first? I suspect it'd look pretty similar to other historical transition plans, but it'd be good to have it in words somewhere. I'm retroactively wishing we had a list of active transitions happening at any one point in time because I've long forgotten what's needed any more at this point, but that can be handled later.

A plan for `f32`/`f64` is:
 - Make wit-parser accept `f32` and `f64` as aliases for `float32` and `float64`. https://github.com/bytecodealliance/wasm-tools/pull/1356
 - Wait until that's in a release and the major tools have updated.
 - Update Wit files, documentation, pretty-printers, etc. to `f32`/`f64`.
 - After some time, add a deprecation warning for `float32`/`float64` pointing users to `f32`/`f64`
 - After some time, remove `float32`/`float64`

The other transition in flight is documentation comments. That one's simpler because wit-parser already accepts `///` and most Wit files and documentation already use it.
 - Update the Wit printer to use it. https://github.com/bytecodealliance/wasm-tools/pull/1357
 - After some time, add a deprecation warning for `//` comments pointing uses to `///`.
 - After some time, make the warning an error.

The semicolon transition is already done. Wit-parser is already requiring semicolons.

Do these look reasonable? If so, I can file new issues in this repo to track them.

> Can you additionally add some tests which assert that the syntax parses and works correctly?

I've now added a test.

alexcrichton (approved):
Sounds good to me, and thanks! And yeah if you wouldn't mind opening issues that's probably the best way to track things for now.

Like with `//` as well this is probably a good time to go ahead and implement wearnings for `float{32,64}` and just have them disabled by default until we decide to turn them on.

PR #1357, opened 2024-01-04
Switch to `///` for printing documentation comments.

sunfishcode:
Switch to `///` for printing documentation comments.
    
Per the discussion in WebAssembly/component-model#286, switch to `///`
for documentation comments in the Wit printer.
    
`//` comments are still accepted and interpreted as documentation
comments for now.

And update the copy of wasi-http in the tests, to a newer version that uses `///` doc comments.


alexcrichton (approved):
Thanks!

Would you be up for implementing a warning as well for usage of `//`? It's not something we would enable just yet by default but it'd be good to go ahead and get the code in so it can be opted-in to if desired.

Issue #1362, opened 2024-01-09
wit: migrate to only accepting `///` for doc comments


sunfishcode:
Per the discussion in https://github.com/WebAssembly/component-model/issues/286, there seems to be consensus for using `///` as the documentation comment syntax.

This syntax is already recognized, so the plan for using `///` is:

 - [x] Update the Wit printer to use it. https://github.com/bytecodealliance/wasm-tools/pull/1357
 - [ ] After some time, add a deprecation warning for // comments pointing uses to ///.
 - [ ] After some time, make the warning an error.

PR #1363, opened 2024-01-09
Silently ignore non-doc-comment comments.

sunfishcode:
As part of bytecodealliance/wasm-tools#1362, silently ignore comments that aren't `///` comments. This won't break any actual code; it just means that documentation comments won't show up in generated bindings unless the `///` syntax is used.

All wit files I've seen already use `///` syntax for comments, so I expect the impact of this change will be low.


alexcrichton:
Given the recent switch to `//` away from `///` and that this is going back, could this be turned off by default for now? Effectively allow either `//` or `///` during this transitionary period with an env var to force `//` to get ignored

Issue #135, opened 2020-11-10
wast: How do you look up function indices when `Id::new(...)` is not public?


matthew-a-thomas:
If the following code is the idiomatic way to look up a function index then there's a problem. I can't create an `Id`:

```rust
let wat = "...";
let module = parse::<Wat>(&wat)?.module;
let names = module.resolve()?;
names.resolve_func(&mut Index::Id(
    Id::new( // associated function `new` is private
        "$my-name",
        Span::from_offset(0)
    )
));
```

Is there a better way to look up the `$my-name` function index?

alexcrichton:
Currently this was intended to help resolve other nodes that already had an `Id`, but exposing other methods seems fine by me!

martinitus:
Just stumbled over the same thing - I want to create AST nodes programatically, i.e. without parsing.

I see a couple of options: 
- `pub new(name: &'a str, span: Span)->Id<'a>` without any checks. This could cause downstream bugs, because i could e.g. provide an invalid name `$foo` or `foo bar`.
- `pub new(name: &'a str, span: Span)->Id<'a>` and internally parse the name, panic if the name is not a valid identifier. However I think that might not be possible due to requirements with the parser buffer / parser.
- `pub new(name: &'a str, span: Span)->Result<Id<'a>>`  as above, but instead of panic return the parse error.

martinitus:
I think this can be closed :) Thanks again @alexcrichton

PR #1369, opened 2024-01-13
Provide public new method for Id AST node

martinitus:
Proposed fix for #135.

### DIFF
--- a/crates/wast/src/token.rs
+++ b/crates/wast/src/token.rs
@@ -2,9 +2,9 @@
 //! associated specifically with the wasm text format per se (useful in other
 //! contexts too perhaps).
 
-use crate::annotation;
-use crate::lexer::Float;
+use crate::lexer::{Float, Lexer, TokenKind};
 use crate::parser::{Cursor, Parse, Parser, Peek, Result};
+use crate::{annotation, Error};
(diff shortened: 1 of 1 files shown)


martinitus:
Alternatively one could use `pub(crate)` fields in `Id` and drop the `new_unchecked` completely?

alexcrichton (commented):
Thanks for this! Given how low-level this function is though I'd be tempted to say that the `new` function as-is is probably ok to export. If you think the behavior you added here though is useful, perhaps a `new` and `new_checked`?

martinitus:
You are welcome! Thanks for all the work you are putting into this!<3

I thought about this once more and changed it to the pub(crate) variant as it aligns best with the rest of the code (imo) and avoids exposing a potential pitfall in the public API (e.g. passing in a `"invalid argument"` as name). The wast crate can simply instantiate the Id directly!

alexcrichton:
Ok I think that's fine, but I think the `pub(crate)` can be removed from the fields since it's not necessary to construct an `Id` within the same module. Replacing the one instance of `Id::new` that was there prior I think is ok to use struct-construction syntax, however.

alexcrichton (approved):

PR #1373, opened 2024-01-16
feat: add CLI completion script generation subcommand

### LINKED ISSUE #1371: Add shell completion script generation to the CLI
`clap` provides a package for this called available at https://crates.io/crates/clap_complete.

The package would be added as a runtime dependency. The CLI could then provide a new command called `completion`, accepting the shell name and printing the completion script to stdout.

If this feature is desired, I can open a PR.

mainrs:
Closes #1371

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -256,6 +256,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "clap_complete"
+version = "4.4.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


pchickey (approved):
Thanks! This is a nice helpful improvement.

mainrs:
Let me know if changes are necessary.

mainrs:
I have chosen fish since its the only shell I know that has a specific path for installing completion scripts. Other shells I am aware of like bash and zsh use different paths depending on OS, system configuration and so on.

PR #1382, opened 2024-01-23
`wasm-smith`: Implement support for generating GC instructions

fitzgen:
Found and fixed a bunch of bugs in the validator, encoder, and the generator itself. Since then, this has been fuzzing for about 18 hours without finding any new bugs.

### DIFF
--- a/crates/wasm-encoder/src/core/code.rs
+++ b/crates/wasm-encoder/src/core/code.rs
@@ -1614,8 +1614,8 @@ impl Encode for Instruction<'_> {
                 sink.push(0x18);
                 let cast_flags =
                     (from_ref_type.nullable as u8) | ((to_ref_type.nullable as u8) << 1);
-                relative_depth.encode(sink);
                 sink.push(cast_flags);
+                relative_depth.encode(sink);
                 from_ref_type.heap_type.encode(sink);
                 to_ref_type.heap_type.encode(sink);
             }
@@ -1628,8 +1628,8 @@ impl Encode for Instruction<'_> {
                 sink.push(0x19);
                 let cast_flags =
                     (from_ref_type.nullable as u8) | ((to_ref_type.nullable as u8) << 1);
-                relative_depth.encode(sink);
                 sink.push(cast_flags);
+                relative_depth.encode(sink);
                 from_ref_type.heap_type.encode(sink);
                 to_ref_type.heap_type.encode(sink);
             }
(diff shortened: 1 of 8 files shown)


fitzgen:
Okay, CI is green now.

crates/wasm-encoder/src/core/code.rs
```diff
@@ -1614,8 +1614,8 @@ impl Encode for Instruction<'_> {
                 sink.push(0x18);
                 let cast_flags =
                     (from_ref_type.nullable as u8) | ((to_ref_type.nullable as u8) << 1);
-                relative_depth.encode(sink);
                 sink.push(cast_flags);
+                relative_depth.encode(sink);
```
alexcrichton, line 1618: I'm sort of baffled how this didn't get caught by the spec test suite...

crates/wasmparser/src/validator/operators.rs
```diff
@@ -1015,18 +1072,10 @@ where
...
-        };
-        let sup_ty = RefType::new(
-            sup_ty.is_nullable(),
-            self.resources.top_type(&sup_ty.heap_type()),
-        )
-        .unwrap();
+        let sup_ty = self
+            .pop_ref()?
+            .unwrap_or_else(|| sub_ty.as_reference_type().unwrap());
+        let sup_ty = RefType::new(true, self.resources.top_type(&sup_ty.heap_type())).unwrap();
```
alexcrichton, line 1078: Mind leaving some comments here as to why these `.unwrap()` calls won't panic?

crates/wasmparser/src/validator/operators.rs
```diff
@@ -3968,26 +3990,20 @@ where
...
     fn visit_i31_get_s(&mut self) -> Self::Output {
-        match self.pop_ref()? {
-            Some(ref_type) => match ref_type.heap_type() {
-                HeapType::I31 => self.push_operand(ValType::I32),
-                _ => bail!(self.offset, "ref heap type mismatch: expected i31"),
-            },
-            _ => bail!(self.offset, "type mismatch: expected (ref null? i31)"),
-        }
+        self.pop_operand(Some(ValType::Ref(
+            RefType::new(true, HeapType::I31).unwrap(),
```
alexcrichton, line 3994: Could this perhaps be an associated const on `RefType` e.g. `RefType::I31` or similar?

alexcrichton (approved):
This is an impressive amount of work, thanks for going through all this as I'm sure it was a bit of a slog! Looking forward though to seeing how this gets used :)

crates/wasm-encoder/src/core/code.rs
fitzgen: Yeah...

fitzgen:
Thanks for taking a look at this!

PR #1384, opened 2024-01-24
Break wasm-smith's default dependency on wasmparser

### LINKED ISSUE #1383: wasm-smith depends on wasmparser by default again
#1301 made wasmparser an optional dependency of wasm-smith, #1330 landed a bit later and seems to have inadvertently brought this back.

[Cargo.toml](https://github.com/bytecodealliance/wasm-tools/blob/main/crates/wasm-smith/Cargo.toml#L27). The issue seems to be that while wasmparser is still optional for wasm-smith, wasm-smith by default enables the wasmparser feature for wasm-encoder.

@alexcrichton Do you know of a way to fix this? I'm pretty sure it's not our fault, we're not enabling anything odd when depending on wasm-smith.

alexcrichton:
This fixes a mistake introduced in #1330 which accidentally undid #1301.

Closes #1383

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -119,6 +119,11 @@ jobs:
       - run: cargo check --no-default-features --features metadata
       - run: cargo check --no-default-features --features wit-smith
       - run: cargo check --no-default-features --features addr2line
+      - run: |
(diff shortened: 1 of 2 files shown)


fitzgen (approved):
👍

PR #1386, opened 2024-01-26
`wast`: Fix pointer provenance bug reported by MIRI

fitzgen:
We need to re-derive the str pointer after moving the original `Box<str>` it was derived from.

Thanks to @Robbepop for finding this error!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2232,6 +2232,7 @@ name = "wast"
 version = "70.0.1"
 dependencies = [
  "anyhow",
+ "bumpalo",
  "leb128",
  "memchr",
  "rayon",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,9 @@ repository = "https://github.com/bytecodealliance/wasm-tools"
 readme = "README.md"
 exclude = ['tests/wabt', 'tests/testsuite', 'tests/snapshots', 'ci']
 
+[lints]
+workspace = true
+
 [workspace]
 members = [
   'crates/c-api',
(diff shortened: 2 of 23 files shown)


fitzgen:
This PR does not add any testing under MIRI to CI.

We (Alex and I) experimented with running some subset of our tests under MIRI, but it is just too slow to go through the `roundtrip.rs` test runner. Even just a single test is too slow, and also feels too brittle.

Our thinking at this point is to

* generate some number of wasm test cases with wasm-smith offline / ahead of time
  * this could be manual and then checked in to the repo
  * or in a `build.rs` (but we would want to make a utility testing crate to avoid downstream dependers from having to run a build script that isn't useful for them)
* run these test cases through `wat` / `wast` / `wasmprinter`

fitzgen:
Notably, generating the test cases offline means that we don't need to run `wasm-smith` under MIRI, as that wouldn't be too interesting and also would be very slow.

fitzgen:
The alternative to this plan would be to remove the single `unsafe` block from `wast`. It exists to essentially have an arena and extend certain lifetimes. We can use an existing and well-vetted and MIRI-tested crate for that, such as `bumpalo`.

fitzgen:
I've pushed two new commits to go with the remove-unsafe-blocks-via-bumpalo approach and to configure lints to deny unsafe code for all crates in the workspace by default.

This allows us to sidestep the MIRI testing challenges because we simply don't have unsafe code anymore, and the crates we pull in to replace the old unsafe code are themselves already tested with MIRI.

alexcrichton (approved):

Robbepop:
I am glad this bug got tracked down so quickly! Thanks a lot @fitzgen for keeping me in the loop. :)

PR #1390, opened 2024-01-28
Use different limits for items in components

### LINKED ISSUE #1389: Relax external MAX_WASM_MEMORIES check for imported memory count
Hi! I've been using `wasmencoder` by way of `componentize-py`. In particular, I've been using `componentize-py` to compile the native dependencies of a Python package to Wasm shared libraries – following along the pattern @dicej established with [`wasi-wheels`](https://github.com/dicej/wasi-wheels).

It's been pretty successful so far, but I ran into a snag: it looks like imported memories – such as the ones defined by Python `wasm32-wasi` "native" packages – count against [the validation](https://github.com/bytecodealliance/wasm-tools/blob/main/crates/wasmparser/src/validator/component.rs#L2309-L2318) the encoder performs.

With the packages I've translated thus far, there are something like 81 memory imports; I believe `componentize-py` adds some more, pushing the resulting component across the `MAX_WASM_MEMORIES` check threshold. However, the component that `componentize-py` produces only contains a single core `(memory)`, shared many times. I found it surprising that these references would count against the maximum memory count. (However, it's possible I'm misunderstanding the purpose of the limit!)

alexcrichton:
Right now the validation limit for the number of memories or tables in a component is quite small, only 100. This is the core wasm limit which is intended to limit the number of definitions in core wasm, but in a component items taking up the index space are just references, not definitions. That means they don't need to have similarly small limits and can instead be much larger.

Closes #1389


dicej (approved):
Thanks, Alex!

PR #1398, opened 2024-02-01
`wasm-encoder`: Support extended constant expressions proposal

### LINKED ISSUE #1397: wasm-encoder: support extended constant expressions
I've noticed that #687 added the groundwork for implementing the extended constant expressions proposal, but `wasm-encoder` still does not support them. Are there any plans for this, and if so, when can I expect them to be available? This isn't a pressing issue for me, hence why I don't open a PR at this point, but I wanted to check in on this.

ThePuzzlemaker:
Closes #1397.

See [this comment](https://github.com/bytecodealliance/wasm-tools/issues/1397#issuecomment-1921533659).

### DIFF
--- a/crates/wasm-encoder/src/core/code.rs
+++ b/crates/wasm-encoder/src/core/code.rs
@@ -3188,6 +3188,11 @@ impl ConstExpr {
         Self { bytes }
     }
 
+    fn with_insn(mut self, insn: Instruction) -> Self {
+        insn.encode(&mut self.bytes);
+        self
+    }
+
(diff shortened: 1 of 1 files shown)


fitzgen (approved):
Thanks!

I think long-term we probably want to stop adding a hundred different constructors and methods for the specific instructions that are allowed in const expressions, and instead expose essentially the `with_insn` method and maybe assert that `inst.is_const()` in there. Maybe even split out `ConstExpr` and `ConstExprBuilder`.

But this is absolutely fine as-is for now.

Thanks!

PR #1402, opened 2024-02-06
Add mutations to remove local usage

alexcrichton:
Won't remove locals just yet but now removal of `local.{set,tee}` can happen.


fitzgen (approved):
Could you add a test over here: https://github.com/bytecodealliance/wasm-tools/blob/403f943985a7a608756909d2e2d4517f10d45089/crates/wasm-mutate/src/mutators/peephole.rs#L556

PR #1404, opened 2024-02-08
Add `--name-unnamed` to `wasm-tools print`

alexcrichton:
Sometimes I've wanted to twiddle with wasm files but I don't enjoy having to update a bunch of indices. This option can be used to automatically assign a name to everything in a file to more easily make the file human editable.


fitzgen (approved):
nice

PR #1407, opened 2024-02-09
Add a `semver-check` subcommand

### LINKED ISSUE #1399: Add the ability to test whether one version of an interface is a semver-compatible evolution of another
This was discussed at the BA plumbers summit this past week. WASI proposals do not want to break during the 0.2.x release series and we want to provide a CI-backed guarantee that this doesn't happen. There needs to be functionality added to `wit-parser` (and/or `wit-component`) to load two versions of WASI and assert that one is a semver-compatible superset of the other.

One way of implementing this would be to load two `Resolve` instances and manually iterate over them. Another vague idea would be to use the strategy of `wasm-tools component targets` and effectively rely on `wasmparser` validation. Something along the lines of if a component is created from the old version of a WASI proposal it should still "target" the new world because everything in the old should still be in the new. Wrapping that up in a single subcommand I think would still be useful however, even if that implementation route is chosen.

alexcrichton:
This issue is intended to address #1399. The assumption is that this will be used by WASI repositories (eventually) to ensure that the current `main` branch of the repository is semver-compatible with the last release of the repository.

This commit adds a new subcommand which looks like:

    wasm-tools component semver-check ./wit --prev old-world --new new-world

This will check that `new-world` is a semver-compatible evolution of `new-world`. Internally this works the same way as the `targets` subcommand where a component is synthesized on-the-fly which is then fed to `wasmparser` for validation to determine whether it's valid or not. Specifically a component of shape `old-world` is fed into an import of shape `new-world`, effectively answering the question of whether a component taking the "old" world can be instantiated with the "new" world.

Some basic CLI tests are added here as well showcasing various bits and bobs. Some further tooling will be needed to actually integrate this into WASI repositories but this should be at least one chunk of the necessary work.

Closes #1399

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -112,7 +112,7 @@ rustc-demangle = { version = "0.1.21", optional = true }
 cpp_demangle = { version = "0.4.0", optional = true }
 
 # Dependencies of `component`
-wit-component = { workspace = true, optional = true, features = ['dummy-module', 'wat'] }
+wit-component = { workspace = true, optional = true, features = ['dummy-module', 'wat', 'semver-check'] }
 wit-parser = { workspace = true, optional = true }
 wast = { workspace = true, optional = true }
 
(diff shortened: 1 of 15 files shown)


pchickey (approved):
Looks like a great MVP. If/when it becomes time to give this more humane error messages, it will need some changes to the tactics, but since this is a pretty complex operation I like that it leans on functionality we already have confidence in.

PR #1408, opened 2024-02-09
Support decoding wasm-encoded WIT packages in wit-parser

alexcrichton:
This commit does a bunch of shuffling internally and refactoring to move the decoding functionality in `wit-component` into the `wit-parser` crate. The goal of this is to handle the use cases identified at the recent BA plumber's summit to where WIT packages can be consumed as binary artifacts rather than exclusively text artifacts. Text artifacts are unfortunately easy to edit and binary artifacts are more amenable to distribution and external management.

To that extent the end goal of thse changes are:

* Entries in the `deps` folder are now allowed to be `*.wit` files (package-in-one-file), `*.wasm` (package-encoded-to-wasm) or `*.wat` (text format of prior).
* The `wit-parser` crate has the above supported in the preexisting `Resolve::push_dir` method which all bindings generators already use.
* There is a new `Resolve::push_path` to ingest either the directory form we're used to today (with a `deps` folder) or a single file which is WIT-encoded-to-wasm. 

The next step after this change is to update all consumers of `wit-parser` to use `Resolve::push_path` instead of `Resolve::push_dir`. This'll enable putting binary artifacts anywhere in the spots for ingestion of WIT packages. At that point it should be possible to switch Wasmtime, for example, to exclusively using binary artifacts produced by CI on various WASI proposals.


alexcrichton:
cc @peterhuene as you're likely interested in this and/or may have opinions on organization

crates/wat/src/lib.rs
```diff
@@ -221,6 +222,73 @@ fn _parse_str(wat: &str) -> Result<Vec<u8>> {
...
+
+        let lexer = Lexer::new(text);
+        let mut iter = lexer.iter(0);
+
+        while let Some(next) = iter.next() {
+            match next.map(|t| t.kind) {
+                Ok(TokenKind::Whitespace)
+                | Ok(TokenKind::BlockComment)
+                | Ok(TokenKind::LineComment) => {}
+                Ok(TokenKind::LParen) => return Detect::WasmText,
```
peterhuene, line 275: Should this check go further to see if the next token is `module` or `component`, or would that be too specific?

crates/wit-parser/src/decoding.rs
```diff
@@ -412,10 +414,59 @@ pub fn decode(bytes: &[u8]) -> Result<DecodedWasm> {
...
+        _ => bail!("expected an exported component type"),
+    };
+
+    // let (resolve, world) = crate::decoding::decode_world(types, ty)?;
+    // Ok((resolve, world, string_encoding))
+    // }
+    // pub(crate) fn decode_world(
+    // types: &types::Types,
+    // world: types::ComponentTypeId,
+    // ) -> Result<(Resolve, WorldId)> {
```
peterhuene, line 469:
```suggestion
```

src/bin/wasm-tools/component.rs
```diff
@@ -505,31 +507,34 @@ impl WitOpts {
...
-            if wasmparser::Parser::is_component(&input) {
-                wit_component::decode(&input)
-            } else {
-                let (_wasm, bindgen) = wit_component::metadata::decode(&input)?;
-                Ok(DecodedWasm::Component(bindgen.resolve, bindgen.world))
+        match Detect::from_bytes(&input) {
+            Detect::WasmBinary | Detect::WasmText => {
+                // Use `wat` to possible translate the text format, and then
+                // afterwards use either `decode` or `metadata::decode` depending on
+                // if the input is a component or a core wasm mdoule.
```
peterhuene, line 514:
```suggestion
                // if the input is a component or a core wasm module.
```

peterhuene (commented):
Looks good! Just a few comments.

peterhuene (approved):

crates/wat/src/lib.rs
alexcrichton: Ah meant to write more here but got distracted, so wanted to fill this in before I merge. I read this and thought as well, "that's a good idea!", but [according to this](https://webassembly.github.io/spec/core/text/modules.html):

> In a source file, the toplevel surrounding the module body may be omitted.

Meaning that we'd have to support more than just `module` or `component` here but any top-level item, so it's probably best to keep this as-is and see if it causes issue and adapt as necessary from there.

PR #1417, opened 2024-02-15
Bump env_logger dependency to 0.11.0

alexcrichton:


pchickey (approved):
Dep upgrade is fine, but im curious why those test failures resulted from this change

alexcrichton:
Ah the CI failure here is https://github.com/bytecodealliance/wasm-tools/pull/1419, and I think we just got lucky not running into that before

Issue #1405, opened 2024-02-08
wasm-smith: introduce `modify_exports` option


adambratschikaye:
Would you be up for taking a PR to add an optional field `modify_export` to [`wasm_smith::Config`](https://docs.rs/wasm-smith/latest/wasm_smith/struct.Config.html#) which takes a callback to be applied to each export of the generated modules?

The motivation for this is that we'd like to generate modules which are valid for running on the IC, where there are restrictions on the names and types of exports (see [here](https://internetcomputer.org/docs/current/references/ic-interface-spec#system-api-module)). So we we'd like the option to remove or modify the names of some exports.

alexcrichton:
I think this is an excellent idea, and I think it'd be reasonable to add!

Now excuse me while I also take a moment here to digress a bit and pontificate slightly about other possible designs here. First though I think it's reasonable to have a PR for this and we can always iterate on it in the future, no need for you to take on my ideas below here.

One nice part about `wasm-smith` right now is that it's possible to use the CLI to configure all the various options, notably there's a "serializable form" of `Config` which can make it kind of nice to share with others. This option naturally would not be configurable on the CLI since it's a custom callback. A possible alternative, though, would be to take a leaf from the notebook of `available_imports`. My thinking would be that we could generalize the `available_imports` option to "please generate a module of this shape". For example we could rename `available_imports` to `desired_shape` and then `wasm-smith` would guarantee that it would generate a module with at most the imports provided and exactly the exports provided.

That way you could generate a desired shape and then pass it to `wasm-smith` to manufacture a module of that shape. One consequence of this is that you'd have to have your own version of, for example, deciding whether the module should have an export called `canister_heartbeat`. The main reason I say this is that I'm not sure how to encode "maybe this can be exported but it also doesn't have to be" as a request to `wasm-smith`, but having "the module must export this function of this signature" makes sense to me from a wasm-smith point of view. Do you think it'd be easy enough for y'all to generate a random "shape" which adheres to the IC spec?

(and again if it's easiest for you to implement `modify_exports` that's totally ok to send a PR for, this is just me thinking and something I might make an update towards later)

fitzgen:
> My thinking would be that we could generalize the `available_imports` option to "please generate a module of this shape". For example we could rename `available_imports` to `desired_shape` and then `wasm-smith` would guarantee that it would generate a module with at most the imports provided and exactly the exports provided.

I like this a lot :+1:

adambratschikaye:
[...]

PR #1418, opened 2024-02-15
Add `exports` field to `Config`

adambratschikaye:
Add an optional `exports` field to `Config` which describes exactly which exports the generated module should have.

This is similar to the solution described [here](https://github.com/bytecodealliance/wasm-tools/issues/1405#issuecomment-1934328598).


adambratschikaye:
This is failing on a seemingly unrelated test for `wit-parser`, but I also get that same failure on `main` commit https://github.com/bytecodealliance/wasm-tools/commit/2573bb28cc897d64ce33739e3b46ac63c9588267.

alexcrichton:
Ah yeah no worries on the wit-parser failure, that's due to https://github.com/bytecodealliance/wasm-tools/pull/1419

crates/wasm-smith/tests/exports.rs
```diff
@@ -0,0 +1,246 @@
...
+                }
+            } else if let wasmparser::Payload::GlobalSection(rdr) = payload {
+                for global in rdr {
+                    let global = global.expect("Failed to read global");
+                    globals.push(global.ty);
+                }
+            } else if let wasmparser::Payload::ExportSection(rdr) = payload {
+                exports = rdr.into_iter().collect::<Result<_, _>>().unwrap();
+            }
+        }
```
alexcrichton, line 206: One thing I might recommend is to use the [result of validation](https://docs.rs/wasmparser/latest/wasmparser/types/struct.Types.html) which should avoid the need to manually walk over these sections and record the results.

alexcrichton: You might also be able to skip the `expected_exports` field and instead run the validator over the `export_module` field and then assert everything there is also present in the wasm-smith output

alexcrichton (commented):
Thanks again for this! Overall everything looks reasonable to me, but if you're up for it could you look into using `wasmparser::types::Types`, the result of validation, instead of parsing the input module directly? I left some comments below on the tests but I think this is also applicable to where the example module is parsed and read. I think that might help simplify some of the logic here where you would only need to buffer up the results of the `ExportSection` and then that could be passed to `Types` directly to get all the type information (via [`entity_type_from_export`](https://docs.rs/wasmparser/latest/wasmparser/types/struct.TypesRef.html#method.entity_type_from_export) and [`EntityType`](https://docs.rs/wasmparser/latest/wasmparser/types/enum.EntityType.html)).

If you feel though that it doesn't end up simplifying much and it's about as complex I'm also happy to land this as-is.

alexcrichton:
Also, as a minor note, after https://github.com/bytecodealliance/wasm-tools/pull/1416 we're changing a bit how CI works on this repository, so CI won't fully finish on this PR just yet until it's rebased. After rebasing on `main` though it should be good to go.

crates/wasm-smith/tests/exports.rs
adambratschikaye: Thanks, I switched to using `Validator` to get `Types` in both the test and the implementation.

alexcrichton (approved):
Thank you again for this, it's very much appreciated!

PR #1421, opened 2024-02-15
Remove working tree files for publication

alexcrichton:
Avoids an error encountered during publishing where everything published except for `wasm-tools` itself because Cargo was warning about a dirty source tree.


alexcrichton:
Found while watching [this log](https://github.com/bytecodealliance/wasm-tools/actions/runs/7923558805/job/21633510153). I published the `wasm-tools` crate itself manually to get everything working.

fitzgen (approved):

PR #1423, opened 2024-02-20
Add pointer and length types to `WasmType`.

sunfishcode:
In anticipation of memory64, provenance in Rust, and guest bindings with instrumented pointers, add `Pointer` and `Length` types to `WasmType`, and use it in the ABI for list, strings, and argument/return value pointers. Consumers that don't have anything special to do with these can handle them both the same as `i32`.

And, because the variant type Canonical ABI can unify pointers with integer types like `i64`, add a `Pointer64` type as well, which represents a conceptual union of a pointer and an `i64`. Consumers that don't have anything special to do with this type can handle it the same as an `i64`.

### DIFF
--- a/crates/wit-component/src/dummy.rs
+++ b/crates/wit-component/src/dummy.rs
@@ -128,6 +128,9 @@ pub fn dummy_module(resolve: &Resolve, world: WorldId) -> Vec<u8> {
                 WasmType::I64 => dst.push_str("i64"),
                 WasmType::F32 => dst.push_str("f32"),
                 WasmType::F64 => dst.push_str("f64"),
+                WasmType::Pointer => dst.push_str("i32"),
+                WasmType::Pointer64 => dst.push_str("i64"),
+                WasmType::Length => dst.push_str("i32"),
             }
         }
         dst.push(')');
--- a/crates/wit-component/src/dummy.rs
+++ b/crates/wit-component/src/dummy.rs
@@ -129,7 +129,7 @@ pub fn dummy_module(resolve: &Resolve, world: WorldId) -> Vec<u8> {
                 WasmType::F32 => dst.push_str("f32"),
                 WasmType::F64 => dst.push_str("f64"),
                 WasmType::Pointer => dst.push_str("i32"),
-                WasmType::Pointer64 => dst.push_str("i64"),
+                WasmType::PointerOrI64 => dst.push_str("i64"),
                 WasmType::Length => dst.push_str("i32"),
             }
         }
--- a/crates/wit-component/src/encoding.rs
+++ b/crates/wit-component/src/encoding.rs
@@ -107,7 +107,7 @@ fn to_val_type(ty: &WasmType) -> ValType {
         WasmType::F32 => ValType::F32,
         WasmType::F64 => ValType::F64,
         WasmType::Pointer => ValType::I32,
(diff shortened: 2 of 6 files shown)


sunfishcode:
One awkward thing about this is that we no longer get signature deduping between two types that differ only in `i32` vs. pointer, as seen in the test change here.

pchickey:
Can you say more about `provenance in Rust, and guest bindings with instrumented pointers`?

I'm apprehensive that this is the right way to solve this problem - right now the WasmType corresponds to the spec [`Number Types`](https://webassembly.github.io/spec/core/syntax/types.html#number-types) and these pointer and length types don't appear anywhere in the spec. I would need to understand the intended way these new AST nodes are used to give a more concrete suggestion, though.

alexcrichton:
Not deduping some types I think is fine, everything is dedup'd at the runtime layer anyway typically. 

One thing that I'd find helpful to review this, which I think might also help address @pchickey's concern, could this be integrated with `wit-bindgen` to see what the changes there might be? I think that's the main consumer of `WasmType` apart from Wasmtime and Wasmtime would probaby look quite similar to `wit-bindgen`

sunfishcode:
[Rust is adopting](https://github.com/rust-lang/rfcs/blob/master/text/3559-rust-has-provenance.md) a semantic model where pointers conceptually carry additional "provenance" information which describes what objects may be accessed. We should ideally make the bindings use pointer types like `*mut c_void` instead of `i32` when carrying pointer values at the Rust level. Pointers codegen down to `i32` at the Wasm level, but give the Rust compiler better information at the Rust level.

For instrumented pointers, I'm thinking about systems that use fat-pointer techniques to add spatial or temporal memory safety bookkeeping to pointers. Not everyone will want to do this, but for people who do, the bindings generator will want to know which values are pointers.

Here's a preview of what this looks like it wit-bindgen:

https://github.com/sunfishcode/wit-bindgen/commit/a939252a90195df6faa6a7e219af3c5b563ec697

It's not done yet; in particular, the `RET_AREA` variable needs bigger changes, but it does illustrate the high-level idea.

sunfishcode:
`Pointer64` is for when there's eg. a `variant<string, i64>` and the Canonical ABI unifies the `string`'s pointer with the `i64`. Neither `*mut T` nor `i64` is sufficient, because `*mut T` might not be 64-bit, and `i64` doesn't carry the provenance. So it seems we need a `Pointer64` type which is like a union between a pointer and an `i64` and able to carry either.

alexcrichton:
Ah I see, I misunderstood `Pointer64`. Perhaps that could be named `PointerOrI64` to help disambiguate a bit?

I'm also not sure how this will be represented in Rust. In the bindings you're representing it currently as `(*mut T, *mut T)` but I'm not sure that'll do the right thing of expanding to two arguments? (or if it does is it guaranteed to do so? I feel like I still haven't lived down "Alex made `wasm32-unknown-unknown` entirely incompatible with C" so I'm hesitant to rely on ABI lowerings as-is in rustc again.

sunfishcode:
`PointerOrI64` makes sense.

Yeah, `(*mut T, *mut T)` isn't great. I still need to figure out something to do there, because it turns out `#[repr(transparent)]` on a `union` is unstable. But that's entirely on the wit-bindgen side.

alexcrichton:
Ok looks good to me, and thanks again for this!

PR #1425, opened 2024-02-23
wasmprinter: print the end offset

sunfishcode:
With `--print-offsets`, print the offset at the end of the module or component. For example, with an empty module, print this:

```
(module)
(;@8     ;)
```

This change also ensures that top-level modules and components are printed with a newline at the end, which makes using them on the command line tidier.


alexcrichton (approved):
Thanks!

PR #1426, opened 2024-02-24
Refactor generation of tables/elements in wasm-smith

alexcrichton:
This commit refactors `wasm-smith` and its generation of both table types and element segments. The goal is to help generate modules of shapes that Wasmtime does not currently support but should. Notably constant expressions are allowed to use `global.get`, even in element segments, and Wasmtime does not currently support this.

Furthermore this commit additionally enables generating tables with initialization expressions which was not previously supported by `wasm-smith`.

Internally this refactors a number of pieces of code that work with constant expressions to instead use a shared helper for generating constant expressions. This takes into account subtyping and such to try to generate interesting shapes of expressions when GC is enabled in particular.

### DIFF
--- a/crates/wasm-smith/src/core.rs
+++ b/crates/wasm-smith/src/core.rs
@@ -2353,16 +2353,9 @@ pub(crate) fn arbitrary_table_type(
     if config.disallow_traps {
         assert!(minimum > 0);
     }
-    let element_type = if !config.reference_types_enabled {
-        // Without reference types the only available type is `funcref` as
-        // that's all that was available with MVP wasm.
-        RefType::FUNCREF
-    } else if !config.gc_enabled || module.is_none() {
-        // With reference types, but without GC, there's only two options.
-        *u.choose(&[RefType::FUNCREF, RefType::EXTERNREF])?
-    } else {
-        // With GC we get the full suite of reference types available to us.
-        module.unwrap().arbitrary_ref_type(u)?
+    let element_type = match module {
+        Some(module) => module.arbitrary_ref_type(u)?,
+        None => RefType::FUNCREF,
     };
     Ok(TableType {
         element_type,
--- a/crates/wasm-smith/src/core.rs
+++ b/crates/wasm-smith/src/core.rs
@@ -143,6 +143,9 @@ pub struct Module {
     /// Reusable buffer in `self.arbitrary_const_expr` to amortize the cost of
     /// allocation.
     const_expr_choices: Vec<Box<dyn Fn(&mut Unstructured, ValType) -> Result<ConstExpr>>>,
+
+    /// What the maximum type index that can be referenced is.
+    max_type_limit: MaxTypeLimit,
 }
 
 impl<'a> Arbitrary<'a> for Module {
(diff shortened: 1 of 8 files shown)


crates/wasm-smith/src/core.rs
```diff
@@ -968,7 +979,7 @@ impl Module {
 
         if self.config.gc_enabled && !self.types.is_empty() && u.arbitrary()? {
             let type_ref_limit = u32::try_from(self.types.len()).unwrap();
-            let idx = u.int_in_range(0..=type_ref_limit)?;
+            let idx = u.int_in_range(0..=type_ref_limit - 1)?;
```
fitzgen, line 982: I think this was not `-1` to allow type references to `self` when we are constructing a type, which isn't yet in `self.types`. But that only works when we are constructing a new type to add to `self.types`, not for arbitrary callers. Maybe it makes sense to take a parameter to control this behavior?

fitzgen (approved):
Thanks! Didn't realize we were missing this :grimacing:

crates/wasm-smith/src/core.rs
```diff
@@ -2266,12 +2353,19 @@ pub(crate) fn arbitrary_table_type(u: &mut Unstructured, config: &Config) -> Res
...
+        // Without reference types the only available type is `funcref` as
+        // that's all that was available with MVP wasm.
+        RefType::FUNCREF
+    } else if !config.gc_enabled || module.is_none() {
+        // With reference types, but without GC, there's only two options.
+        *u.choose(&[RefType::FUNCREF, RefType::EXTERNREF])?
+    } else {
+        // With GC we get the full suite of reference types available to us.
+        module.unwrap().arbitrary_ref_type(u)?
+    };
```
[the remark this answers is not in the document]
alexcrichton: Good point!

crates/wasm-smith/src/core.rs
alexcrichton: I opted to refactor the various infrastructure here a bit more to ensure that this is handled throughout all type generation, mind taking a look again?

crates/wasm-smith/src/core.rs
```diff
@@ -549,16 +569,20 @@ impl Module {
...
             }
         } else {
             let type_ref_limit = u32::try_from(self.types.len()).unwrap();
-            let ty = self.arbitrary_sub_type(u, type_ref_limit)?;
+            self.max_type_limit = MaxTypeLimit::Num(type_ref_limit);
+            let ty = self.arbitrary_sub_type(u)?;
             self.add_type(ty);
         }
 
+        self.max_type_limit = MaxTypeLimit::ModuleTypes;
```
fitzgen, line 584: Might be worth it to have 

```rust
fn with_max_type_limit<T>(&mut self, limit: MaxTypeLimit, f: impl FnMut(&mut Self) -> T) -> T {
    ...
}
```

fitzgen (approved):
I really like this clean up, much better than threading the old `type_ref_limit` parameter everywhere!

crates/wasm-smith/src/core.rs
alexcrichton: With only one method that modifies the max type limit I think I'll stick to this for now because I never like to rename `self` to `me` or `this`, but if this grows in the future I agree a helper should be added

PR #1431, opened 2024-02-29
wit-component: Support Asyncified modules with shared-everything link

kateinoigakukun:
This commit adds support for modules Asyncified with `--pass-arg=asyncify-relocatable` to the shared-everything linker. It adds the `__asyncify_state` and `__asyncify_data` mutable globals definitions as it does for `__stack_pointer` so that the Asyncify global state can be shared across all modules.

This unlocks CRuby to link shared-everything modules.


kateinoigakukun:
CC: @dicej

dicej (approved):
LGTM.  Thanks for doing this!

PR #1432, opened 2024-03-01
Make wast parse results serializable

### LINKED ISSUE #1428: Making wast parse results serializable
I'd love to use the parser and encoder from `wast` in JavaScript. The easiest way for that to work would be for the parsed module to be serializable and deserializable as JSON, so that the string representations can be passed to JS via wasm-bindgen. Could the various AST types be made serializable?

I know adding serde adds some heft, but I think it could be done conditionally with an off-by-default feature.

I tried doing this myself by sprinkling `#[derive(Serialize, Deserialize)]` everywhere, but got stuck making the lifetimes work inside the [`instructions` macro](https://github.com/bytecodealliance/wasm-tools/blob/90161a9b5fbfeaa40e9b4ba2339d7cd1bd52deff/crates/wast/src/core/expr.rs#L344). I don't have enough experience with rust to figure out how to get that working.

bakkot:
Fixes #1428 by liberally sprinkling `derive(Serialize, Deserialize)` and friends on things until it built.

Concurrently with making this PR, I found a different approach for the high-level thing I wanted this for, and therefore don't have any immediate use for it. If you therefore don't think it's worth the hit to code readability, feel free to close without reading further. I figured I'd submit it anyway since I was halfway done at that point.

Some notes:
- `#[serde(borrow)]` needs to go on at least one field of a structure or variant of enum which uses a lifetime parameter, but it suffices to put it on any one of them in the case that there are multiple. I put it on only the first but could do something else if you want.
- For enums, I've gone with the ["adjacently tagged"](https://serde.rs/enum-representations.html#adjacently-tagged) representation, since the default "externally tagged" representation is annoying for consumers and the "internally tagged" representation doesn't work when there's a sequence type in one of the variants. I went with a consistent `tag = "type", content = "val"` on all the enums.
- No tests. I don't think CI will even confirm that building with the new feature actually works, at the moment. Happy to add some if you tell me how you want that done - this is my first PR to a Rust project so I have no idea what the conventions are.
- A couple of the types which needed to be made serializable - e.g. `CoreItemRef` - are generic and have a field whose type is given by the generic. That perforce means that the generic type needs to be constrained to be serializable as well, but only when the feature is enabled. I've done this, but it's a bit ugly, and if you have a better approach to suggest I'd happily take it:
    ```rust
    #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
    pub struct CoreItemRef<'a,
        #[cfg(feature = "serde")] K: SerializeT,
        #[cfg(not(feature = "serde"))] K,
    > {
        /// The item kind being parsed.
        pub kind: K,
        // ...
    }
    ```
    and it's even worse for the places which use it:
    ```rust
    fn parse_trailing_item_ref<
        #[cfg(feature = "serde")] T: SerializeT,
        #[cfg(not(feature = "serde"))] T,
    >(kind: T, parser: Parser) -> Result<CoreItemRef<T>> {
      // ...
    }
    ```


alexcrichton:
Thanks for this!

Does your use case require `Deserialize`? If possible I'd prefer to slim this down to just being `Serialize` if it works for you and I think that would at least solve the `#[serde(borrow)]` part I think.

Using an internal tag I think is fine, I think it's ok to just pick something that works. As for CI though we'll defintely want something there, otherwise this'll probably break in a few weeks. Could you add a [check here](https://github.com/bytecodealliance/wasm-tools/blob/e2e853a494a1e6e46bdac020223c70ff154415e9/.github/workflows/main.yml#L139-L169)?

I'll take a closer look at this in a bit and see if I have any ideas about the trait bounds, that looks a bit worrisome from a maintainability standpoint but I'll need to take a closer look.

bakkot:
> Does your use case require `Deserialize`?

Unfortunately yes - the goal was to manipulate the parse result and then synthesize it back to `wasm`, which means I need to be able to take a JSON representation of the parse result and deserialize it so I can then call `.encode`. (Though as I say I did end up finding a different way to do the high-level thing I wanted.)

> Could you add a [check here](https://github.com/bytecodealliance/wasm-tools/blob/e2e853a494a1e6e46bdac020223c70ff154415e9/.github/workflows/main.yml#L139-L169)?

Done. I added only `--features serde,wasm-module` because the relevant code is only exercised at all when both features are present.

alexcrichton:
Ah ok, thanks for adding the check! If you're ok though given that you've already got an alternative solution I'd at least personally prefer to not land this for now. This is a fair bit of code without a clear in-repo use-case so while we can maintain it for external users if the need isn't there just yet by you or others I'd be tempted to hold off on merging.

bakkot:
Yeah, that's fine, not to worry. I'll close so as not to clutter up the repo, but I'll leave the branch around.

PR #1434, opened 2024-03-01
Remove text parsing support for let/func.bind

alexcrichton:
These are historical vestiges of revisions of the function-references proposal I believe, and I'm not aware of any current proposals using these, so remove them.

### DIFF
--- a/crates/wast/src/core/expr.rs
+++ b/crates/wast/src/core/expr.rs
@@ -184,7 +184,6 @@ impl<'a> ExpressionParser<'a> {
                         // seen
                         i @ Instruction::Block(_)
                         | i @ Instruction::Loop(_)
-                        | i @ Instruction::Let(_)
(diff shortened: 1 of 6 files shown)


crates/wast/src/core/expr.rs
```diff
@@ -468,8 +467,6 @@ instructions! {
         // function-references proposal
         CallRef(Index<'a>) : [0x14] : "call_ref",
         ReturnCallRef(Index<'a>) : [0x15] : "return_call_ref",
-        FuncBind(FuncBindType<'a>) : [0x16] : "func.bind",
-        Let(LetType<'a>) : [0x17] : "let",
```
pchickey, line 472: Can we delete `FuncBindType` and `LetType` as well?

alexcrichton: Can do, updated!

fitzgen (approved):

PR #1437, opened 2024-03-07
Add support for using symlink for dependency dir in the `deps` dir

Thaumy:
Using a symlink as a dependency directory in the `deps` directory will result in "package not found", which is confusing and there is no reason not to support it.
This patch fixes this behavior to provide a better experience for tools using the `wit-parser` crate.

### DIFF
--- a/crates/wit-parser/src/resolve.rs
+++ b/crates/wit-parser/src/resolve.rs
@@ -236,10 +236,11 @@ impl Resolve {
         for dep in entries {
             let path = dep.path();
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
This works too yeah, thanks! Instead of duplicating the `if` clause though mind throwing it as `|| path.metadata()?.is_dir()` on the preexisting clause? Happy to r+ and merge afterwards.

alexcrichton (approved):

PR #1438, opened 2024-03-07
Make self argument for guest exported resources a Pointer (rep)

cpetig:
This solution feels inelegant, ideally you would distinguish between imported and exported resource Handles and then use either I32 or Pointer in abi::push_flat().

But it clearly fixes the problem of the self argument in respect to that rep is 64 bit on wasm64 (and potential native uses).

Apart from removing the related casting in https://github.com/bytecodealliance/wit-bindgen/blob/main/crates/rust/src/bindgen.rs#L444 and some linker symbol changes in wit-bindgen this is all which is needed to enable native (e.g. x86_64) WIT interfaces.

See https://github.com/bytecodealliance/wit-bindgen/blob/73e5c68d88714e5073bfa93836f8dd6d456b4b0b/crates/rust/src/interface.rs#L581-L589 for above mentioned linker symbol code.

### DIFF
--- a/crates/wit-parser/src/abi.rs
+++ b/crates/wit-parser/src/abi.rs
@@ -1,5 +1,3 @@
-use std::assert_matches::assert_matches;
-
 use crate::{Function, Handle, Int, Resolve, Type, TypeDefKind};
 
 /// A core WebAssembly signature with params and results.
@@ -162,7 +160,7 @@ impl Resolve {
                 // But this contextual information isn't available, yet.
                 // See https://github.com/bytecodealliance/wasm-tools/pull/1438 for more details.
                 params.get_mut(0).map(|p| {
-                    assert_matches!(p, &WasmType::I32);
+                    assert!(matches!(*p, WasmType::I32));
                     *p = WasmType::Pointer;
                 });
             }
--- a/crates/wit-parser/src/abi.rs
+++ b/crates/wit-parser/src/abi.rs
@@ -1,3 +1,5 @@
+use std::assert_matches::assert_matches;
+
 use crate::{Function, Handle, Int, Resolve, Type, TypeDefKind};
 
 /// A core WebAssembly signature with params and results.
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for this! I agree that this isn't the best solution but should at least get the job done for exported resources. To be clear this isn't a complete solution since borrowed-handles-to-exported-resources can show up anywhere, not just as the first argument to an exported method. The first argument of exported methods, however, are guaranteed to be exported resources since otherwise the component wouldn't validate.

Fixing this "for real" I think is not in the cards at this time. There needs to be the ability to answer the question of "is this resource exported" and that's not a question we can answer at this time because the same type is both used for imports and exports. That means that more contextual information is needed to answer the question of whether it's exported or not, and I don't think that this method can reasonably take it at this time.

All that being said, some requests if you're ok with it:

* Can you assert that the previous value was `I32`?
* Can you add a comment, with a `// TODO` or similar, about why this transformation is guaranteed to be correct as well as how it's best to fix this in the future since it's not a complete solution?

cpetig:
I hope the TODO is understandable.

And of course I have to correct myself:
I do have a working example of compiling a Rust project to a shared library and accessing it via a WIT interface, but for lists and string lengths the offset is still wrong, as well as the size and alignment of ret areas. So additional work is needed for these.

cpetig:
Oh, I see, I used an unstable Rust feature ... going to fix it.

crates/wit-parser/src/abi.rs
```diff
@@ -148,6 +148,22 @@ impl Resolve {
...
+                // Guest exported methods always receive resource rep as first argument
+                //
+                // TODO: Ideally you would distinguish between imported and exported
+                // resource Handles and then use either I32 or Pointer in abi::push_flat().
+                // But this contextual information isn't available, yet.
+                // See https://github.com/bytecodealliance/wasm-tools/pull/1438 for more details.
+                params.get_mut(0).map(|p| {
+                    assert!(matches!(*p, WasmType::I32));
+                    *p = WasmType::Pointer;
+                });
```
alexcrichton, line 165: Could this instead be:

```rust
assert!(matches!(params[0], WasmType::I32));
params[0] = WasmType::Pointer;
```

to further assert that the first element in the list is defined? (in that it's a bug if the closure here isn't executed)

alexcrichton (approved):
Thanks!

crates/wit-parser/src/abi.rs
cpetig: Yes, I think your proposal is more readable. My panic avoidance habits kicked in, but for code generators panics even improve the DX.

alexcrichton:
Ah the current state is that imported functions can refer exclusively to imported resources, but exported functions may refer to either exported or imported resources. That means that the problem is for `push_flat` in exports and when it sees a resource it's not clear whether it's imported or exported. Only borrows to exported resources become pointers, everything else is a 32-bit handle.

This is also a problem for all `borrow<T>` types anywhere in the function signature, not just the first location, so this is a problem for ABI bindings generation in addition to `push_flat`. This is where I was getting at larger refactorings are required to "truly fix" this issue I think within the `wit-parser` crate. I'd like to get to a state where a `TypeId` is either imported or exported, we're just not quite there yet.

(also it appears there's some build errors now cropping in)

cpetig:
> (also it appears there's some build errors now cropping in)

No, it was just me being too eager to get this committed before riding to work.

PR #1439, opened 2024-03-08
Fix `--name-unnamed` and branch labels

alexcrichton:
This commit fixes an issue where `wasm-tools print --name-unnamed` would produce an invalid `*.wat` file. This PR fixes a few off-by-one issues locally and commits a test as well.

Finally this additionally updates the printing of `@N` "helper labels" to avoid printing that if there's no actual target named `@N`, for example in the case of name conflicts or when a branch depth goes directly to the function itself.


fitzgen (approved):
LGTM

PR #1450, opened 2024-03-13
Replace custom "test harnesses" with `libtest-mimic`

alexcrichton:
There's a number of tests suites in this repository which are based on "drop a file here and it's a test now". Up to now these have been powered by a custom "test harness" copied around which implements at filtering mechanism but little else. The `libtest-mimic` crate provides a `test`-crate-lookalike API which enables dynamically building the list of tests at program start time (aka read the filesystem) and then afterwards looks and feels similar to the `test` crate's CLI and UX.

This enables a nicer dev experience when running these tests and additionally enables using runners like `cargo nextest` with this repository.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -617,6 +617,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "escape8259"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


alexcrichton:
@dicej would you be ok taking a look at this? Happy to circle around if you're busy though!

dicej (approved):

PR #1454, opened 2024-03-13
`wasm-smith`: Generate "interesting" constants

fitzgen:


crates/wasm-smith/src/core.rs
```diff
@@ -2224,6 +2233,168 @@ impl Module {
...
+                    }
+                    if let Some(x) = max.checked_sub(1 << i) {
+                        interesting(x);
+                    }
+                }
+            }
+        }
+
+        self.interesting_values32.extend(interesting_values32);
+        self.interesting_values64.extend(interesting_values64);
```
alexcrichton, line 2346: Mind sorting these after this to ensure they don't inherit the nondeterminism of the sets?

crates/wasm-smith/src/core.rs
```diff
@@ -2224,6 +2233,168 @@ impl Module {
...
+    }
+
+    fn arbitrary_const_instruction(
+        &self,
+        ty: ValType,
+        u: &mut Unstructured<'_>,
+    ) -> Result<Instruction> {
+        match ty {
+            ValType::I32 => Ok(Instruction::I32Const(
+                if self.interesting_values32.len() > 0 && u.arbitrary()? {
```
alexcrichton, line 2356: Should this, and below, assert that the interesting sets are nonzero in size?

alexcrichton (approved):
Nice!

While this is probably covered in the bit patterns you added can you also add in the signed `::MIN` constants too?

crates/wasm-smith/src/core.rs
fitzgen: D'oh, of course

PR #1460, opened 2024-03-19
Implement a `json-from-wast` subcommand

### LINKED ISSUE #1395: wasm-tools parse support for .wast file.
From current README, I got the impression that both wat and wast crates are accessible from subcommand parse:

https://github.com/bytecodealliance/wasm-tools/blob/d15fda33f309a27cd507d7739d7b5208575e9b78/README.md#L128

But as I try out the command from the shell and also looking through the code, this doesn't seem currently possible:

https://github.com/bytecodealliance/wasm-tools/blob/49753602683a539b66d0a65ffa11acb402f148bb/src/lib.rs#L67-L72

Is it intended? If not, may I contribute a feature addition?

alexcrichton:
This commit is an implementation of #1395 which brings wabt's `wast2json` functionality to `wasm-tools`. This is implemented entirely in the CLI since if using the crate it's already got all the functionality necessary for this. Additionally this PR attempts to be pretty close to `wast2json` from wabt in terms of structure and syntax. The main difference is that wabt fills out the expected return types of an operation when the operation isn't supposed to return (e.g. `assert_trap`), but `wasm-tools` won't do that as we don't have that information so easily available.

This is tested by adding `*.json` files to the `snapshots` directory with example outputs from all known `*.wast` tests. Everything should be included except for the `component-model` tests at this time which don't have support in printing the args/results.

Closes #1395


alexcrichton:
I'll note that I worry that calling this `wast2json` will cause confusion with the actual `wast2json` tool, so I'm curious if others share the same concern and/or have a good idea about naming. Otherwise I might rename to a bland `wast-to-json` but that still feels confusing as it's pronounced the same

fitzgen:
> I'll note that I worry that calling this `wast2json` will cause confusion with the actual `wast2json` tool, so I'm curious if others share the same concern and/or have a good idea about naming. Otherwise I might rename to a bland `wast-to-json` but that still feels confusing as it's pronounced the same

Flip it on its head with `json-from-wast`?

crates/wast/src/wast.rs
```diff
@@ -219,11 +218,23 @@ pub enum WastExecute<'a> {
...
     Wat(Wat<'a>),
     Get {
+        span: Span,
         module: Option<Id<'a>>,
         global: &'a str,
     },
 }
 
+impl<'a> WastExecute<'a> {
+    /// Returns the first span for htis execute statement.
```
fitzgen, line 228:
```suggestion
    /// Returns the first span for this execute statement.
```

fitzgen (approved):

alexcrichton:
I like it!

PR #1466, opened 2024-03-21
Rename `float32`/`float64` to `f32`/`f64`.

sunfishcode:
Rename `float32`/`float64` to `f32`/`f64`.
    
As discussed [here], and following up on #1364 which added `f32`
and `f64` parsing support, this PR renames the types `float32`
and `float64` to `f32` and `f64`.
    
The old names are still accepted by the parser for compatibility.
    
[here]: https://github.com/WebAssembly/component-model/issues/277


alexcrichton (approved):
Thanks for this!

PR #1472, opened 2024-03-27
Don't generate exports named `memory` in wit-smith

alexcrichton:
This fixes a fuzz bug discovered last night and this applies a similar fix for interfaces which avoids using `memory` that clashes with the canonical name for linear memory itself.


fitzgen (approved):
Maybe worth doing a `with_capacity` call instead of `new` in initial definition of `export_names` just above. Or a `from_iter(Some("memory".into()))` or something.

PR #1473, opened 2024-03-27
Bail when doing `wit_component::decode` on a component

rylev:
Ran into an issue where I was accidentally calling `wasm-tools component new` on a wasm binary that was already a component. Unfortunately, there's nothing that prevents this, and things only blow up deep inside the parser where an EOF error is returned. I believe there's no good reason for calling `wit_component::decode` with a component so this feels like the right place to return this error.


alexcrichton:
Nice! Mind adding a CLI test as well for this? Dropping something in `tests/cli/*.wat` should suffice where it contains `(component)` and tries to do a `component new` with the result.

alexcrichton (approved):

PR #1474, opened 2024-03-27
Add additional error context to ComponentEncoder::encode

rylev:
This would have helped me when investigating an issue. Changes to `wasm-parser`'s `BinaryReaderError` would be needed to make the errors here even clearer.  

Before:

```
error: failed to encode a component from module

Caused by:
    unexpected end-of-file (at offset 0xc)
```

After: 
```
error: failed to encode a component from module

Caused by:
    1: module was not valid
    2: unexpected end-of-file (at offset 0xc)
```

### DIFF
--- a/crates/wit-component/src/encoding.rs
+++ b/crates/wit-component/src/encoding.rs
@@ -2078,7 +2078,7 @@ impl ComponentEncoder {
             bail!("a module is required when encoding a component");
         }
 
-        let world = ComponentWorld::new(self)?;
+        let world = ComponentWorld::new(self).context("failed to decode world from module")?;
         let mut state = EncodingState {
             component: ComponentBuilder::default(),
             module_index: None,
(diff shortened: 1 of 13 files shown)


fitzgen (approved):
Thanks!

fitzgen:
Test failure in CI: https://github.com/bytecodealliance/wasm-tools/actions/runs/8456061199/job/23164960594?pr=1474#step:4:924

```
 ---- component_encoding_via_flags stdout ----
testing error-empty-module-import
thread 'component_encoding_via_flags' panicked at crates/wit-component/tests/components.rs:241:17:
assertion failed: `(left == right)`: failed baseline comparison (tests/components/error-empty-module-import/error.txt)

Diff < left / right > :
<no top-level imported function `foo` specified
>failed to decode world from module
>
>Caused by:
>    0: module was not valid
>    1: no top-level imported function `foo` specified
```

alexcrichton (approved):

PR #1478, opened 2024-03-28
add `--use-built-in-libdl` option to `component link` subcommand

dicej:
This tells `wit-component` to include a prebuilt libdl.so when linking.  This library provides `dlopen`, `dlsym`, etc. using the lookup tables generated by the linker when one or more `--dl-openable` options are provided.


dicej:
FYI @kateinoigakukun

crates/wit-component/dl/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[lib]
+crate-type = [ "staticlib" ]
+
+[dependencies]
+
+[workspace]
```
alexcrichton, line 12: Mind throwing this in the same toplevel workspace to avoid an extra `Cargo.lock` file?

crates/wit-component/dl/build.sh
```diff
@@ -0,0 +1,7 @@
+# This script will rebuild libdl.so from source.
+#
+# To run this, you'll need Rust nightly 2023-07-27 or later, including the
+# `wasm32-wasi` target, plus `wasi-sdk` 21 or later, installed at $WASI_SDK_PATH
+
+RUSTFLAGS="-C relocation-model=pic" cargo +nightly build -Z build-std=panic_abort,std --release --target=wasm32-wasi
```
alexcrichton, line 6: Could this drop the `+nightly` flag to enable running it with custom nightlies/toolchains as needed?

Also, could you update CI to have a check that runs this script and verifies that the wasm file hasn't changed?

crates/wit-component/dl/src/lib.rs
```diff
@@ -0,0 +1,180 @@
...
+    count: u32,
+    libraries: *const Library,
+}
+
+struct Pointer<T>(*const T);
+
+unsafe impl<T> Sync for Pointer<T> {}
+
+static mut ERROR: Pointer<c_char> = Pointer(ptr::null());
+static mut LIBRARIES: Pointer<Libraries> = Pointer(ptr::null());
```
alexcrichton, line 52: I think you can probably get away with `*const c_char` and `*const Libraries` here directly, I believe it's just `static` that has restrictions on `Sync`, not `static mut`

crates/wit-component/dl/src/lib.rs
```diff
@@ -0,0 +1,180 @@
...
+    let library = library as *const Library;
+    if (0..(*LIBRARIES.0).count).any(|index| {
+        (*LIBRARIES.0)
+            .libraries
+            .add(usize::try_from(index).unwrap())
+            == library
+    }) {
+        false
+    } else {
+        ERROR.0 = b"invalid library handle\0" as *const _ as _;
```
alexcrichton, line 71: While you're here you may be interested to try out `c"foo"` which I think will be stable in 1.78 but should already be stable on nightly

crates/wit-component/dl/build.sh
alexcrichton: Also, would it make sense to change this to perhaps build with `-Copt-level=s` and/or LTO to slim things down?

crates/wit-component/tests/components/link-dl-openable-builtin-libdl/component.wat
alexcrichton, line 1: This is a pretty hefty file to check in, what do you think of perhaps adding something to `tests/cli/*` and piping the result to `wasm-tools validate` to basically assert that it's valid but don't go further than that? I'm not opposed to checking this in but I figure a test otherwise wouldn't be too bad either.

crates/wit-component/dl/build.sh
dicej: I tried a few things (`-Copt-level=s`, `-Clinker-plugin-lto`, and `#![no_std]` with a `panic_handler` that just calls `core::arch::wasm32::unreachable!()`) but the size of the output never changed at all.

dicej: Oops -- my mistake.  I was still pointing `clang` at target/wasm32-wasi/release/libdl.a after adding `dl` to the workspace; now it's under the target dir under the root directory.  I'll try again.

dicej: Okay, `-Copt-level=s` actually made it larger for some reason.  And I couldn't figure out the right `clang` incantation to make LTO work (kept getting `wasm-ld: error: Invalid record (Producer: 'LLVM18.1.2-rust-1.79.0-nightly' Reader: 'LLVM 17.0.6')` errors).  But `#![no_std]` with a trivial panic handler got it down to 2703 bytes.

crates/wit-component/tests/components/link-dl-openable-builtin-libdl/component.wat
dicej: Now that libdl.so is only 2703 bytes, component.wat is much more manageable, so I think I'll leave it if that's okay with you.

crates/wit-component/src/linking.rs
```diff
@@ -1268,6 +1279,14 @@ impl Linker {
             })
             .collect::<Result<Vec<_>>>()?;
 
+        eprintln!(
```
alexcrichton, line 1282: Stray eprintln?

alexcrichton (approved):
Nice 👍

crates/wit-component/src/linking.rs
dicej: Ugh, thanks for catching that.

PR #1479, opened 2024-03-29
Add a mechansim for optimizing stream `read` functions.

sunfishcode:
The wasi-io input-stream function `read()` returns a `list<u8>` by value. In practice, that means that it has to call `cabi_realloc` on every call to allocate memory for the list, and then copy the data from that memory into the user's buffer.

This patch adds a mechanism for defining components with a `cabi_stream_buffer` function, which can be used as an alternative to `cabi_realloc` in canonical options.

`cabi_stream_buffer` returns the value of a global named `CABI_STREAM_BUFFER_PTR`, which bindings can use to store user-supplied buffers in. This allows them to avoid any `cabi_realloc` call.

This is similar to an optimization done in the preview1-to-preview2 adapter.

### DIFF
--- a/crates/wit-component/src/encoding.rs
+++ b/crates/wit-component/src/encoding.rs
@@ -340,10 +340,11 @@ pub struct EncodingState<'a> {
     ///
     /// If `None`, then the realloc function has not yet been aliased.
     default_realloc_index: Option<u32>,
-    /// The index in the core function index space for the stream realloc function.
+    /// The index in the core function index space for the a realloc function
+    /// designed for use with caller-provided buffers.
     ///
     /// If `None`, then the realloc function has not yet been aliased.
-    stream_buffer_index: Option<u32>,
+    use_caller_buffer_index: Option<u32>,
     /// The index of the shim instance used for lowering imports into the core instance.
     ///
     /// If `None`, then the shim instance how not yet been encoded.
@@ -379,8 +380,8 @@ pub struct EncodingState<'a> {
     export_func_type_map: HashMap<types::FunctionKey<'a>, u32>,
 
     /// Interface-name and function-name pairs identifying functions which
-    /// should use the optimized stream ABI with `cabi_stream_buffer`.
-    optimize_stream_read: IndexSet<(String, String)>,
+    /// should use the caller-buffer optimization with `cabi_use_caller_buffer`.
+    optimize_using_caller_buffer: IndexSet<(String, String)>,
 
     /// Metadata about the world inferred from the input to `ComponentEncoder`.
     info: &'a ComponentWorld<'a>,
@@ -1460,7 +1461,7 @@ impl<'a> EncodingState<'a> {
     }
 
     /// Decide whether to use the regular `cabi_realloc` or some alternative
-    /// (eg. `cabi_stream_buffer`) for a function named `name` in an imported
+    /// (eg. `cabi_use_caller_buffer`) for a function named `name` in an imported
     /// interface `interface_name`.
     fn select_realloc_function(
         &self,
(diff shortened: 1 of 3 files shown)


sunfishcode:
Opening as a draft for now, as I haven't yet added a testcase.

I also have an [experimental branch of wit-bindgen](https://github.com/sunfishcode/wit-bindgen/tree/sunfishcode/stream-realloc) that uses this to implement the full optimization.

alexcrichton:
Nice! I'm wondering though if this might be best to perhaps generalize a bit more? Looking at this and the wit-bindgen pieces it seems like this is enabling an option to wit-bindgen which auto-optimizes APIs like `read` which is nice from a "it's just a flag perspective", but it also feels pretty brittle in terms of it's not clear when the flag can be applied or not.

Instead what would you think of perhaps generalizing how `wit-component` searches for a realloc function? For example it could detect a symbol of the name `cabi_realloc[import][wasi:io/streams#read]` (or something like that) and that way any realloc function could be specified for any lifted/lowered function. That'd help apply this to other locations as well as necessary, and the `wasi` crate or `wasi-libc` could skip auto-generated bindings and add specialized ones with a custom realloc

sunfishcode:
I'm unsure how to make that work. If a Rust program has multiple crates that use generated bindings for eg. `wasi:io/streams#[method]input-stream.read`, then it would include multiple `cabi_realloc_wasi:io/streams#[method]input-stream.read` definitions. wit-bindgen-rt uses C code and some clever tricks to arrange a weak symbol to fix this problem for `cabi_realloc` itself, but that's not something I'm eager to expand to more functions.

alexcrichton:
That's a good point, and it would need further changes yeah. Imports with a customized cabi_realloc would need to be a customized import as well as a customized import, and then wit-component would need to recognize that. There's sort of basic support for this right now in that wit-component knows that a single component import might be lowered multiple times with different options (e.g. cabi-realloc for the main module and for an adapter).

I'm basically pretty wary to make this specific to a single use case for a custom cabi_realloc. If you feel it's too much to take on a more general design though I can try to help out here as well

sunfishcode:
Thinking about it more, what should happen if there are multiple copies of the bindings, and each copy uses a different custom realloc? If the resolution is based on symbol names based on the import name, they'd collide. If we make them weak, the linker can pick one, but that would mean that someone's silently not getting the realloc function they asked for. 

Is there a use case is for per-import customizable realloc functions? I'm hesitant to design something significantly more complex here, when we just have this one particularly acute use case that can use a relatively simply approach.

The most brittle part of the patches I have here is on the wit-bindgen side, and that part will be unavoidable, because it needs to generate different public-facing APIs in order to accept a buffer from the user.

alexcrichton:
Oh sure I'm not thinking we should do anything related to weak symbols or anything like that, I agree we need to avoid that. My thinking is something like:

```wasm
(module 
  (import "wasi:proposal" "[realloc=my_realloc]the-function-name" (func ...))

  ;; ..

  (func (export "my_realloc") ...)
)
```

or something along those lines.

> Is there a use case is for per-import customizable realloc functions? I'm hesitant to design something significantly more complex here, when we just have this one particularly acute use case that can use a relatively simply approach.

Sort of and sort of not. I know where you're coming from, but I think there's clear benefit in being able to support custom realloc per-import/export since that's what the component model supports. Just as there's an argument for deferring complexity to later there's the same argument for solving the complexity early on to enable future use cases. In that sense I think it's worth seeing if there is a not-super-complex design (or basically roughly the same complexity as this PR) to handle these cases.

What I might propose is:

* Imports and exports can all start with an optional `[realloc=...]`
* If present, that specifies the custom realloc function to use for this symbol if needed.
* Otherwise all symbol names are processed as normal in wit-component

I suspect something like that may require some refactoring in wit-component, but I would predict it's not an order-of-magnitude more than what you've already implemented?

> The most brittle part of the patches I have here is on the wit-bindgen side, and that part will be unavoidable, because it needs to generate different public-facing APIs in order to accept a buffer from the user.

Yeah I agree the bindgen side of this is going to get very tricky. What I might recommend is to skip this part entirely, literally. For example using `skip` to avoid generating bindings for read and then writing a handwritten function with a handwritten import/export.

ydnar:
Curious: why were `read` methods implemented as returning `list<u8>` instead of accepting a [BYOB-style](https://github.com/whatwg/streams/blob/main/byte-streams-explainer.md) `list<u8>` argument?

Will this change in WASI 0.3?

PR #1480, opened 2024-03-29
threads: add `shared` globals

abrown:
This change allows wasm-tools to encode and decode `shared` global types as a part of the [shared-everything-threads] proposal. It includes some initial fuzzing support, though it all should be turned off by default.

[shared-everything-threads]: https://github.com/WebAssembly/shared-everything-threads

### DIFF
--- a/crates/fuzz-stats/src/bin/failed-instantiations.rs
+++ b/crates/fuzz-stats/src/bin/failed-instantiations.rs
@@ -102,8 +102,9 @@ impl State {
         let mut config = wasm_smith::Config::arbitrary(&mut u)?;
         config.allow_start_export = false;
 
-        // Wasmtime doesn't support this proposal yet.
+        // Wasmtime doesn't support these proposals yet.
         config.gc_enabled = false;
+        config.shared_everything_threads_enabled = false;
 
         let mut wasm = wasm_smith::Module::new(config, &mut u)?;
         wasm.ensure_termination(10_000).unwrap();
--- a/crates/wasm-mutate/src/mutators/peephole/eggsy/encoder/expr2wasm.rs
+++ b/crates/wasm-mutate/src/mutators/peephole/eggsy/encoder/expr2wasm.rs
@@ -23,6 +23,8 @@ pub enum ResourceRequest {
         tpe: PrimitiveTypeInfo,
         /// If its mutable
         mutable: bool,
+        /// If its shared
+        shared: bool,
     },
     // TODO add other needed resources here, for example, needed locals, needed
     // memory etc. Notice that how this resources are translated to Wasm code,
@@ -363,6 +365,7 @@ pub fn expr2wasm(
                             index: global_idx as usize,
                             tpe: PrimitiveTypeInfo::I32,
                             mutable: true,
+                            shared: false,
                         };
                         resources.push(request);
 
@@ -375,6 +378,7 @@ pub fn expr2wasm(
                             index: global_idx as usize,
                             tpe: PrimitiveTypeInfo::I64,
                             mutable: true,
+                            shared: false,
                         };
                         resources.push(request);
 
@@ -387,6 +391,7 @@ pub fn expr2wasm(
                             index: global_idx as usize,
                             tpe: PrimitiveTypeInfo::F32,
                             mutable: true,
+                            shared: false,
                         };
                         resources.push(request);
 
@@ -399,6 +404,7 @@ pub fn expr2wasm(
                             index: global_idx as usize,
                             tpe: PrimitiveTypeInfo::F64,
                             mutable: true,
+                            shared: false,
                         };
                         resources.push(request);
 
--- a/crates/wasm-smith/src/core.rs
+++ b/crates/wasm-smith/src/core.rs
@@ -1424,6 +1424,7 @@ impl Module {
         Ok(GlobalType {
             val_type: self.arbitrary_valtype(u)?,
             mutable: u.arbitrary()?,
+            shared: self.config.shared_everything_threads_enabled && u.arbitrary()?,
         })
     }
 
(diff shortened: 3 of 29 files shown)


crates/wit-component/src/linking/metadata.rs
```diff
@@ -89,6 +89,7 @@ impl TryFrom<&FuncType> for FunctionType {
 pub struct GlobalType {
```
alexcrichton, line 89: @dicej I forget if I asked this already, but do you remember why this module duplicates some of the type structures of wasmparser/wasm-encoder?  Would it be possible to migrate this module to using wasmparser directly or wasm-encoder directly for example?

(no need to change anything in this PR yourself @abrown)

tests/local/shared-everything-threads/global.wast
alexcrichton, line 1: In addition to the positive test here, can you add a test to `tests/local/missing-features` which uses a `shared` global and make sure it's gated with an error?

crates/wast/src/core/types.rs
```diff
@@ -368,22 +368,35 @@ pub struct GlobalType<'a> {
...
 
 impl<'a> Parse<'a> for GlobalType<'a> {
     fn parse(parser: Parser<'a>) -> Result<Self> {
-        if parser.peek2::<kw::r#mut>()? {
+        if parser.peek2::<kw::shared>()? || parser.peek2::<kw::r#mut>()? {
             parser.parens(|p| {
-                p.parse::<kw::r#mut>()?;
+                let mut shared = false;
+                let mut mutable = false;
+                if parser.peek::<kw::shared>()? {
```
alexcrichton, line 381: s/parser/p/ here and below

crates/wasmparser/src/readers/core/globals.rs
```diff
@@ -37,13 +37,15 @@ impl<'a> FromReader<'a> for Global<'a> {
 
 impl<'a> FromReader<'a> for GlobalType {
     fn from_reader(reader: &mut BinaryReader<'a>) -> Result<Self> {
+        let content_type = reader.read()?;
+        let flags = reader.read_u8()?;
+        if flags > 0b11 {
+            bail!(reader.original_position() - 1, "malformed mutability")
```
alexcrichton, line 43: Mind updating this to a different error message along the lines of "malformed global flags byte" or similar?

(might as well go ahead and implement it as-if shared everything was stable)

crates/wasm-smith/src/core.rs
```diff
@@ -1424,6 +1424,7 @@ impl Module {
         Ok(GlobalType {
             val_type: self.arbitrary_valtype(u)?,
             mutable: u.arbitrary()?,
+            shared: self.config.shared_everything_threads_enabled && u.arbitrary()?,
```
alexcrichton, line 1427: FWIW I'd recommend leaving wasm-smith out of the equation entirely rather than piecemeal updating it. I find it's best to do wasm-smith as a dedicated project once validation is done to fuzz the two in tandem. 

For example wasm-smith is going to need a lot of updates to avoid accessing non-shared items from a shared context, so while this precise change may remain there's going to need to be a fair bit of infrastructure elsewhere too. 

In the meantime it's ok to ignore wasm-smith entirely while you're first implementing the shared-everything proposal.

crates/wit-component/src/linking/metadata.rs
dicej: I don't totally remember, but I think I wanted a `ValueType` that only represented the types we actually support in this module (i.e. no V128 and no reference types).  I.e. I wanted unsupported variants to be unrepresentable.  I'm not necessarily opposed to reusing types from another crate, though.

crates/fuzz-stats/src/bin/failed-instantiations.rs
```diff
@@ -102,8 +102,9 @@ impl State {
         let mut config = wasm_smith::Config::arbitrary(&mut u)?;
         config.allow_start_export = false;
 
-        // Wasmtime doesn't support this proposal yet.
+        // Wasmtime doesn't support these proposals yet.
         config.gc_enabled = false;
+        config.shared_everything_threads_enabled = false;
```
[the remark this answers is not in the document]
abrown: Ok, I'll just leave it in for now.

crates/wasmprinter/tests/all.rs
```diff
@@ -277,3 +277,17 @@ fn no_panic_non_func_type() {
...
+    const MODULE: &str = r#"
+    (module
+        (global (;0;) (shared f32))
+    )"#;
+    let bytes = wat::parse_str(MODULE).unwrap();
+    let result = wasmprinter::print_bytes(&bytes).unwrap();
+    assert_eq!(
+        result.replace(" ", "").trim(),
+        MODULE.replace(" ", "").trim()
+    );
```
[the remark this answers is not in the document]
abrown: I hear you, but this test wasn't added for completeness: rather, in the midst of troubleshooting `shared` flags and wandering aimlessly through `roundtrip.rs`, I realized this approach was a more direct way to understand what was going wrong. For foundational bits like `shared` flags, can we leave this in for the time being and clean them up once we have more confidence (e.g., after fuzzing)?

crates/wasm-mutate/src/mutators/peephole/eggsy/encoder/expr2wasm.rs
```diff
@@ -23,6 +23,8 @@ pub enum ResourceRequest {
         tpe: PrimitiveTypeInfo,
         /// If its mutable
         mutable: bool,
+        /// If its shared
+        shared: bool,
     },
```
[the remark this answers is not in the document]
abrown: Sounds good!

PR #1481, opened 2024-03-29
Accept `RTLD_GLOBAL` flag in `dlopen` function

kateinoigakukun:
We always resolve symbols in the global scope in AOT linking, so we can safely accept the `RTLD_GLOBAL` flag in the `dlopen` function.

This change makes `--use-built-in-libdl` work with CRuby :)


kateinoigakukun:
CC @dicej Thank you for your work! It works very well for our use case!

alexcrichton (approved):

dicej (approved):

PR #1483, opened 2024-04-02
threads: add atomic versions of `global.get|set`

abrown:
The shared-everything-threads [proposal] allows marking globals as `shared` and thus adds `global.atomic.get|set` for choosing the memory ordering of these accesses. These changes also include the memory `ordering` flags for marking accesses as sequentially consistent (`seq_cst`) or acquire-release (`acq_rel`).

### DIFF
--- a/crates/wasmparser/src/binary_reader.rs
+++ b/crates/wasmparser/src/binary_reader.rs
@@ -277,6 +277,18 @@ impl<'a> BinaryReader<'a> {
         })
     }
 
+    fn read_ordering(&mut self) -> Result<Ordering> {
+        let byte = self.read_u8()?;
+        match byte {
+            0 => Ok(Ordering::SeqCst),
+            1 => Ok(Ordering::AcqRel),
+            x => Err(BinaryReaderError::new(
+                &format!("invalid atomic consistency ordering {}", x),
+                self.original_position() - 1,
+            )),
+        }
+    }
+
     fn read_br_table(&mut self) -> Result<BrTable<'a>> {
         let cnt = self.read_size(MAX_WASM_BR_TABLE_SIZE, "br_table")?;
         let start = self.position;
@@ -1625,6 +1637,10 @@ impl<'a> BinaryReader<'a> {
             0x4c => visitor.visit_i64_atomic_rmw8_cmpxchg_u(self.read_memarg(0)?),
             0x4d => visitor.visit_i64_atomic_rmw16_cmpxchg_u(self.read_memarg(1)?),
             0x4e => visitor.visit_i64_atomic_rmw32_cmpxchg_u(self.read_memarg(2)?),
+
+            // Decode shared-everything-threads proposal.
+            0x4f => visitor.visit_global_atomic_get(self.read_ordering()?, self.read_var_u32()?),
+            0x50 => visitor.visit_global_atomic_set(self.read_ordering()?, self.read_var_u32()?),
 
             _ => bail!(pos, "unknown 0xfe subopcode: 0x{code:x}"),
         })
--- a/crates/wasmparser/src/binary_reader.rs
+++ b/crates/wasmparser/src/binary_reader.rs
@@ -278,7 +278,7 @@ impl<'a> BinaryReader<'a> {
     }
 
     fn read_ordering(&mut self) -> Result<Ordering> {
-        let byte = self.read_u8()?;
+        let byte = self.read_var_u32()?;
         match byte {
             0 => Ok(Ordering::SeqCst),
             1 => Ok(Ordering::AcqRel),
--- a/crates/wasmparser/src/validator/operators.rs
+++ b/crates/wasmparser/src/validator/operators.rs
@@ -1066,18 +1066,14 @@ where
 
     /// Common helper for checking the types of global types accessed
     /// atomically.
-    fn check_atomic_global_ty(&self, global_index: u32, supertype: Option<ValType>) -> Result<()> {
+    fn check_atomic_global_ty(&self, global_index: u32) -> Result<()> {
         let ty = self
             .resources
             .global_at(global_index)
             .expect("existence should be checked prior to this point");
         let ty = ty.content_type;
-        let is_valid_subtype = if let Some(supertype) = supertype {
-            self.resources.is_subtype(ty, supertype)
-        } else {
-            false
-        };
-        if !(ty == ValType::I32 || ty == ValType::I64 || is_valid_subtype) {
+        let supertype = RefType::ANYREF.into();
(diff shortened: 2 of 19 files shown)


crates/wasmparser/src/binary_reader.rs
```diff
@@ -277,6 +277,18 @@ impl<'a> BinaryReader<'a> {
         })
     }
 
+    fn read_ordering(&mut self) -> Result<Ordering> {
+        let byte = self.read_u8()?;
```
alexcrichton, line 281: Unsure how much in flux this is, but [cross referencing this](https://github.com/WebAssembly/shared-everything-threads/blob/main/proposals/shared-everything-threads/Overview.md#instructions) this should technically be `read_var_u32` as specified currently.

That being said I also think that the u32 is probably overkill, so the spec may also want to switch to `u8` perhaps.

crates/wasmparser/src/validator/operators.rs
```diff
@@ -1612,6 +1613,32 @@ where
...
+        global_index: u32,
+    ) -> Self::Output {
+        // The following is almost identical to `visit_global_get` except for
+        // some additional type restrictions. No validation of `ordering` is
+        // needed because `global.atomic.get` can be used on both shared and
+        // unshared globals.
+        if let Some(ty) = self.resources.global_at(global_index) {
+            let ty = ty.content_type;
+            // TODO: need to express that any ValType::Ref(..) must be a subtype
+            // (<:) of `anyref`.
```
alexcrichton, line 1628: IIRC there's various methods we have to get the "top type" and for `ValType::Ref` types you're interested in the ones where the "top type" is `Anyref`

alexcrichton (approved):

alexcrichton:
Lemme know if you'd like to handle the TODO for anyref here or defer that to later, I'm fine either way

crates/wasmparser/src/binary_reader.rs
abrown: Good catch!

abrown: I can still push a `u8` during encoding, right?

crates/wasmparser/src/validator/operators.rs
abrown: Stumbled across `self.resources.is_subtype(...)` which explains things well &mdash; what do you think of that?

abrown:
> Lemme know if you'd like to handle the TODO for anyref here or defer that to later, I'm fine either way

I couldn't find what I was looking for yesterday. I think d683529 should resolve this?

abrown:
> it should be validated on the definition of the global

Yup, that's more "right". I'll tweak this all next week.

abrown:
Scratch that: I think we may have to validate at the instruction level because different instructions have different type restrictions.

alexcrichton:
That'd be two layers of validation, right? One layer for when the global is defined and then another layer for specific opcodes having more restrictions?

crates/wasmparser/src/validator/operators.rs
```diff
@@ -1625,6 +1649,17 @@ where
...
     }
+    fn visit_global_atomic_get(
+        &mut self,
+        _ordering: crate::Ordering,
+        global_index: u32,
+    ) -> Self::Output {
+        self.visit_global_get(global_index)?;
+        // No validation of `ordering` is needed because `global.atomic.get` can
+        // be used on both shared and unshared globals.
+        self.check_atomic_global_ty(global_index, Some(RefType::ANYREF.into()))?;
```
alexcrichton, line 1660: Ok so apparently I was confused about this and sorry if that caused issues.

I only just now realize that the intention is to be able to use `global.atomic.get` with non-shared globals in addition to shared globals. If that's the goal then that means that the type-checking code is duplicated both at shared-global-definition-time but additionally here.

Spec-level questions for you:

* Could the spec be stricter and say that `global.atomic.*` instructions only work on `shared` globals? IIRC the rationale for `memory.atomic.*` working on non-atomic memories isn't quite as applicable to globals since globals, ironically, aren't as "global" as memory since there's only one linear memory but often many globals.
* Failing that, could the spec allow any type with non-shared globals and `global.atomic.*`?
* Failing that, mind refactoring the validation here to have a shared helper that verifies that the type used here is valid for atomic operations? That would then be shared with the validation of a global declaration.

Also, an orthogonal spec-level question: is it valid to `global.get` a shared global? Mostly want to confirm if preexisting instructions need to be updated in their validation.

crates/wasmparser/src/validator/operators.rs
```diff
@@ -1064,6 +1064,28 @@ where
         self.push_operand(sub_ty)
     }
 
+    /// Common helper for checking the types of global types accessed
+    /// atomically.
+    fn check_atomic_global_ty(&self, global_index: u32, supertype: Option<ValType>) -> Result<()> {
```
alexcrichton, line 1069: Instead of passing in a `supertype` here it might be best to leave this without an argument for now. Then it could grow an argument when equality-related globals are required (e.g. compare-and-swap instructions). In the future this might be best represented as a boolean flag rather than a `supertype` so indicate that it's basically just indicating the top type is one of two choices: anyref or eqref.

alexcrichton (approved):
Looks good 👍, just want to clarify some spec things

crates/wasmparser/src/validator/operators.rs
abrown: > is it valid to global.get a shared global?

Yes, it's the way to do an unordered access.

> Could the spec be...?

I think we confused each other here. I was a little too in the weeds figuring out how to do all of this in wasm-tools and didn't see the big picture: actually, defining a global &mdash; shared or otherwise &mdash; can be done with _any_ possible type. E.g., someone might want to store a `v128` in a shared global or a `funcref` or a float... We shouldn't really be putting any extra validation on the global definition.

The values in those shared globals can be accessed as usual by `global.get|set` in an unordered way. It's only if we want to enforce atomic access that we limit the types to `i32 | i64 | <: anyref`. So we really only need the extra validation there.

For context From [here](https://github.com/WebAssembly/shared-everything-threads/blob/main/proposals/shared-everything-threads/Overview.md#globals)

> The new instructions are valid for both shared and unshared globals.

abrown: See c15c281.

alexcrichton (commented):
Ah ok those clarifications make sense, thanks!

One final request, and sorry I only just now remembered this. There is still a filter on the definition of a shared global, right? Which is to say that only `shared` types are allowed? All primitives are shared but `funcref`/`anyref` as-is aren't shared. If this is right could you add a `validate_type_is_shared(...)` validation step (or something like that) and call it from the definition of a global?


(that'd make the eqref/anyref tests invalid, but that's ok I think)

alexcrichton (approved):

PR #1485, opened 2024-04-03
Add `nullexnref` and `noexn` types to exception-handling

eqrion:
The latest overview has a `noexn` bottom type like the other hierarchies now.

https://github.com/WebAssembly/exception-handling/blob/main/proposals/exception-handling/Exceptions.md#exnref


alexcrichton (approved):
Thank you for this!

PR #1489, opened 2024-04-09
Add support for the custom-page-sizes proposal

fitzgen:
This pull request adds support for [the custom-page-sizes proposal](https://github.com/WebAssembly/custom-page-sizes) to:

* [x] `wasm-encoder`
* [x] `wasmparser`
* [x] `wast`
* [x] `wasmprinter`
* [x] `wasm-smith`

It additionally adds some `.wast` tests that I intend to eventually upstream into the proposal's spec test suite.

### DIFF
--- a/crates/wasmparser/src/validator/core.rs
+++ b/crates/wasmparser/src/validator/core.rs
@@ -876,7 +876,7 @@ impl Module {
         offset: usize,
     ) -> Result<()> {
         self.check_limits(ty.initial, ty.maximum, offset)?;
-        let page_size = if let Some(page_size_log2) = ty.page_size_log2 {
+        let (page_size, page_size_log2) = if let Some(page_size_log2) = ty.page_size_log2 {
             if !features.custom_page_sizes {
                 return Err(BinaryReaderError::new(
                     "the custom page sizes proposal must be enabled to \
@@ -884,17 +884,17 @@ impl Module {
                     offset,
                 ));
             }
+            if page_size_log2 > 16 {
+                return Err(BinaryReaderError::new("invalid custom page size", offset));
+            }
             let page_size = 1_u64 << page_size_log2;
             debug_assert!(page_size.is_power_of_two());
-            if page_size > DEFAULT_WASM_PAGE_SIZE {
-                return Err(BinaryReaderError::new(
-                    "invalid custom page size: {page_size}",
-                    offset,
-                ));
-            }
-            page_size
+            debug_assert!(page_size <= DEFAULT_WASM_PAGE_SIZE);
+            (page_size, page_size_log2)
         } else {
-            DEFAULT_WASM_PAGE_SIZE
+            let page_size_log2 = 16;
+            debug_assert_eq!(DEFAULT_WASM_PAGE_SIZE, 1 << page_size_log2);
+            (DEFAULT_WASM_PAGE_SIZE, page_size_log2)
         };
         let (true_maximum, err) = if ty.memory64 {
             if !features.memory64 {
@@ -907,7 +907,7 @@ impl Module {
                 max_wasm_memory64_pages(page_size),
                 format!(
                     "memory size must be at most 2**{} pages",
-                    64 - page_size.ilog2()
+                    64 - page_size_log2
                 ),
             )
         } else {
--- a/crates/wasmparser/src/validator/core.rs
+++ b/crates/wasmparser/src/validator/core.rs
@@ -876,7 +876,7 @@ impl Module {
         offset: usize,
     ) -> Result<()> {
         self.check_limits(ty.initial, ty.maximum, offset)?;
-        let page_size = if let Some(page_size) = ty.page_size {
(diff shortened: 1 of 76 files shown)


fitzgen:
Feedback addressed!

crates/wasmparser/src/validator/core.rs
```diff
@@ -876,6 +876,26 @@ impl Module {
...
+            }
+            let page_size = 1_u64 << page_size_log2;
+            debug_assert!(page_size.is_power_of_two());
+            if page_size > DEFAULT_WASM_PAGE_SIZE {
+                return Err(BinaryReaderError::new(
+                    "invalid custom page size: {page_size}",
+                    offset,
+                ));
+            }
+            page_size
```
alexcrichton, line 895: I think this will want to validate that `page_size_log2` is <64 first or otherwise I think the over-wide shift will panic in debug mode

alexcrichton: (and might as well bundle in the check that it's less than 16 to begin with)

crates/wasmparser/src/validator/core.rs
```diff
@@ -884,13 +904,17 @@ impl Module {
                 ));
             }
             (
-                MAX_WASM_MEMORY64_PAGES,
-                "memory size must be at most 2**48 pages",
+                max_wasm_memory64_pages(page_size),
+                format!(
+                    "memory size must be at most 2**{} pages",
+                    64 - page_size.ilog2()
```
alexcrichton, line 910: Could the `ilog2` be avoided by plumbing the log around?

crates/wasmprinter/src/lib.rs
```diff
@@ -1076,6 +1076,10 @@ impl Printer {
         if ty.shared {
             self.result.push_str(" shared");
         }
+        if let Some(p) = ty.page_size_log2 {
+            let p = 1 << p;
+            write!(self.result, "(pagesize {p:#x})")?;
+        }
```
alexcrichton, line 1082: This'll need validation to avoid panicking in debug mode like the validator.

tests/local/custom-page-sizes/custom-page-sizes-invalid.wast
```diff
@@ -0,0 +1,12 @@
...
+   "\00asm" "\01\00\00\00"
+   "\05\04\01"                ;; Memory section
+
+   ;; memory 0
+   "\08"                      ;; flags w/ custom page size
+   "\00"                      ;; minimum = 0
+   "\11"                      ;; pagesize = 2**17
+ )
+ "invalid custom page size"
+)
```
alexcrichton, line 12: Can this be rewritten as `*.wast` now that it's no longer an error there?

Also mind adding a test where the page size is >=17, e.g. `1<<64`?

alexcrichton: Also can you add a test for `(pagesize 3)` for a non-power-of-two test?

alexcrichton (approved):

tests/local/custom-page-sizes/custom-page-sizes-invalid.wast
fitzgen: > Also can you add a test for `(pagesize 3)` for a non-power-of-two test?

We can't do that because we still need to check that the page size is a power of two in the wast parser because we cannot delay the error until encoding time because the wast encoding methods are infallible but because we can't losslessly turn a non-power-of-two into its ilog2 we would need to be able to fail at encoding time. So instead we still eagerly check for powers of two at parse time, which means that the whole `.wast` test file gets rejected as invalid text format rather than the particular module being invalid and passing the `assert_invalid` directive.

alexcrichton: Oh sorry, I mean to use `assert_malformed` for this, not `assert_invalid`, which specifically tests text-level errors

fitzgen: Even with `assert_malformed` instead of `assert_invalid`:

```
failed test: tests/local/custom-page-sizes/custom-page-sizes-invalid.wast

Caused by:
    0: failed to run `json-from-wast` cli subcommand
    1: failed to run "/home/nick/wasm-tools/target/debug/wasm-tools" "json-from-wast" "tests/local/custom-page-sizes/custom-page-sizes-invalid.wast" "--pretty" "--wasm-dir" "/tmp/.tmphI5u29"
       stdout: 
       stderr: error: invalid custom page size: 3
            --> tests/local/custom-page-sizes/custom-page-sizes-invalid.wast:2:31
             |
           2 |   (module (memory 0 (pagesize 3)))
             |                               ^
```

fitzgen: I don't think we can resolve this without making `wast`'s encoding fallible, which doesn't seem worth it.

fitzgen: oh nice!

PR #1490, opened 2024-04-10
Do not strip the `dylink.0` section from `libdl.so`

kateinoigakukun:
The `dylink.0` section is required to correctly layout the main memory space and provide `__memory_base` and `__table_base` symbols. The section was being stripped by the `strip` tool, and it led to the unpredictable corruption of `LIBRARIES` global variable.


kateinoigakukun:
I've tried to write test cases to kill this case, but I couldn't make a maintainable one 😢

kateinoigakukun:
CC: @dicej

crates/wit-component/dl/build.sh
```diff
@@ -7,4 +7,4 @@
 
 CARGO_PROFILE_RELEASE_LTO=true RUSTFLAGS="-C relocation-model=pic" cargo build --release --target=wasm32-wasi
 $WASI_SDK_PATH/bin/clang -shared -o $1 -Wl,--whole-archive ../../../target/wasm32-wasi/release/libdl.a -Wl,--no-whole-archive
-cargo run --manifest-path ../../../Cargo.toml -- strip -a $1 -o $1
+cargo run --manifest-path ../../../Cargo.toml -- strip $1 -o $1
```
dicej, line 10: Optional idea: we could run `strip` a second time using `--delete name`, which would remove the `name` section but keep the others.

dicej (approved):
LGTM -- thanks!

crates/wit-component/dl/build.sh
kateinoigakukun: Sounds reasonable to me 👍

kateinoigakukun:
Thanks!

PR #1492, opened 2024-04-10
Ensure that we have the full length of component and module sections

fitzgen:
Fixes the root cause of https://github.com/bytecodealliance/wasmtime/issues/8322

### DIFF
--- a/crates/wasm-metadata/src/lib.rs
+++ b/crates/wasm-metadata/src/lib.rs
@@ -418,8 +418,14 @@ impl Metadata {
                         }
                     }
                 }
-                ModuleSection { unchecked_range: range, .. } => metadata.push(Metadata::empty_module(range)),
-                ComponentSection { unchecked_range: range, .. } => metadata.push(Metadata::empty_component(range)),
+                ModuleSection {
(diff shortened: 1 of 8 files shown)


fitzgen:
Hm I'm not sure how to preserve the `nested_section_too_big` test but I think it is basically subsumed by the new fuzzing?

The `single_module` test is a bit more annoying since it is valid, but we give each little bit of data to parse a little at a time and the way the test is written is kind of incompatible with this PR's change.

Should we delete these tests or should we go back to the drawing board on this change and the property it is guaranteeing?

alexcrichton:
Reading over this I think I'm reaching the conclusion that this change is incompatible with the previous design, namely the streaming compilation aspect. The intention is to avoid needing to buffer up entire modules/components into memory, which I think we want to preserve, so perhaps instead this should be a doc-level change on the module/component section that the range reported isn't guaranteed to be valid until validation has finished?

fitzgen:
Updated to revert the range checking for these sections and renamed their `range` fields to `unchecked_range` and added some docs.

alexcrichton (approved):

PR #1493, opened 2024-04-11
Support `no_std` in the `wasmparser` crate

alexcrichton:
This commit adds a `std` feature to the `wasmparser` crate, enables it
by default, and then enables building the crate when the feature is
disabled. This works by switching imports of `wasmparser` from `std` to
either the `core` or `alloc` crates which are available in the `no_std`
context. The main change made in this PR is that maps used by
`wasmparser`, e.g. `HashMap` and `IndexMap`, are now abstracted with
type aliases in `wasmparser` which unconditionally customize a hash
algorithm parameter. When the `std` feature is enabled this parameter is
implemented using the standard library's default hashing algorithm. When
`std` is disabled then this uses the `ahash` crate internally.

This commit additionally updates CI to both check that wasmparser builds
when the `std` feature is disabled and that it builds on a target
without `std`.


Robbepop:
I really like this direction!
One thing that I did in the `wasmparser-nostd` fork was to replace all `HashMap` usage with `BTreeMap` where possible. The reason for this is that `HashMap` is potentially attackable in a real `no_std` environment where there is no possibility to randomly initialize the `HashMap`'s state if the users can control the inputs which is very much possible in a Wasm parser unfortunately.
On the flip side, I did not see any changes in any of the `wasmparser` benchmarks by doing so. Generally I think Rust's `BTreeMap` is highly underrated.

The biggest downside is that the `indexmap` dependency uses `hashbrown::HashMap` for it's `no_std` support which required me to also fork it with `indexmap_nostd`. However, that was very burdensome and I'd prefer not to.

alexcrichton:
I filled out a bit more information about hashers at https://github.com/bytecodealliance/wasmtime/issues/8341#issuecomment-2052240265 inspired by your comment, thanks for bringing it up!

If possible I'd prefer to keep hash maps as foundational data structures and avoid switching data structures just based on no-std. I think that the utilities provided by `ahash` and various other possible tricks we can do should be able to largely retain the DoS guarantees.

alexcrichton:
I've updated this to be a landable version for https://github.com/bytecodealliance/wasmtime/issues/8341. This now includes CI checks and a bit more documentation.

crates/wasmparser/src/lib.rs
```diff
@@ -27,6 +27,28 @@
 //! the examples documented for [`Parser::parse`] or [`Parser::parse_all`].
 
 #![deny(missing_docs)]
+#![no_std]
+
+extern crate alloc;
+#[cfg(feature = "std")]
+extern crate std;
```
Robbepop, line 34: I'd suggest going with what has been described in this comment: https://github.com/rust-lang/rust/issues/121708#issuecomment-1974964139

I did the refactoring in Wasmi and the code reads a lot better as we can use `std` for nearly all imports. Also the warnings mentioned in this issue are no longer a deal.

alexcrichton: Sorry I'm a bit confused, but isn't that what's happening here?

Robbepop: Yeah sorry, the approach I actually wanted to suggest was to use:
```rust
#![no_std]

#[cfg(not(feature = "std"))]
extern crate alloc as std;

#[cfg(feature = "std")]
extern crate std;
```
So that you can keep most of the `std` imports which reduces the PR diff. Also works with the warnings issue that the linked thread is about.

Robbepop: fair enough!

crates/wasmparser/src/map.rs
```diff
@@ -0,0 +1,126 @@
...
+
+#[cfg(not(feature = "std"))]
+impl BuildHasher for RandomStateImpl {
+    type Hasher = ahash::AHasher;
+
+    #[inline]
+    fn build_hasher(&self) -> ahash::AHasher {
+        ahash::RandomState::new().build_hasher()
+    }
+}
```
fitzgen, line 126: Does it make sense expose a method for manually setting the random state for `nostd` users that happen to have some source of randomness we don't (and can't) know about?

fitzgen (approved):

crates/wasmparser/src/map.rs
alexcrichton: This is sort of already handled by [the ahash crate itself](https://docs.rs/ahash/latest/ahash/random_state/fn.set_random_source.html) although that's not quite the same as providing it through wasmparser itself. My thinking is given that this is going to be shared across Wasmtime and wasmparser it might be best to leave this to ahash for now and add it here if necessary in the future

Issue #1491, opened 2024-04-10
Transparent `FuncToValidate` type


Robbepop:
In Wasmi I am using the `wasmparser` crate in order to parse and validate Wasm input.
Since more recent versions users can choose to do so lazily which means that the `wasmi::Engine` needs to store `FuncToValidate` instances until the associated function is used for the first time and thus lazily validated and compiled.

However, I noticed that doing so is a bit wasteful memory-wise, e.g. `FuncToValidate` stores `WasmFeatures` and also the `index` field overlaps with what Wasmi already has to store anyways so I am looking into either of two things:

1. Make `FuncToValidate` fields `pub` which allows me to easily destructure them and only store what I am going to need later. This might be preferrable since it is the smaller change.
2. Add getter API to `FuncToValidate` for the same reason. This might be preferrable to `pub` fields since this would to discourage doing what I am doing on an API-level.

The `FuncToValidate` type already has a `pub` constructor, so those APIs or `pub` fields are the only thing I'd need to support my use case.

alexcrichton:
I think it's reasonable to go ahead and make all the fields `pub` and probably remove the constructor as well, this is definitely a "bag of names" struct which doesn't benefit much from privacy. 

Also I think it'd be reasonable to shrink `WasmFeatures` to a bitflags representation instead of array-of-bools, I've been meaning to get around to that for awhile as the list of features keeps growing. Not that such a change would solve your problem here of course, only insofar as it's a good change to make in addition to making the fields here `pub`!

alexcrichton:
I think using `bitflags` as-is is fine yeah, I don't think we need to get too too fancy with it.

Robbepop:
Okay, I will file a PR later.

PR #1495, opened 2024-04-15
Make `FuncToValidate` fields public

Robbepop:
Partially closes https://github.com/bytecodealliance/wasm-tools/issues/1491.

I will implement the bit-field based `WasmFeatures` in another PR.


alexcrichton (approved):
Thanks!

PR #1496, opened 2024-04-15
Turn `WasmFeatures` into a bitflags type

Robbepop:
This PR reduces the size of `WasmFeatures` from 23 bytes down to just 4 bytes.

As proposed by @alexcrichton here: https://github.com/bytecodealliance/wasm-tools/issues/1491#issuecomment-2049892243

### DIFF
--- a/crates/wasm-encoder/src/core/types.rs
+++ b/crates/wasm-encoder/src/core/types.rs
@@ -650,6 +650,8 @@ impl Section for TypeSection {
 
 #[cfg(test)]
 mod tests {
+    use wasmparser::WasmFeatures;
+
     use super::*;
     use crate::Module;
 
@@ -666,10 +668,8 @@ mod tests {
         module.section(&types);
         let wasm_bytes = module.finish();
 
-        let mut validator = wasmparser::Validator::new_with_features(wasmparser::WasmFeatures {
-            gc: false,
-            ..Default::default()
-        });
+        let mut validator =
+            wasmparser::Validator::new_with_features(WasmFeatures::default() | WasmFeatures::GC);
 
         validator.validate_all(&wasm_bytes).expect(
             "Encoding pre Wasm GC type should not accidentally use Wasm GC specific encoding",
--- a/crates/wasm-encoder/src/core/types.rs
+++ b/crates/wasm-encoder/src/core/types.rs
@@ -669,7 +669,7 @@ mod tests {
         let wasm_bytes = module.finish();
 
         let mut validator =
-            wasmparser::Validator::new_with_features(WasmFeatures::default() | WasmFeatures::GC);
+            wasmparser::Validator::new_with_features(WasmFeatures::default() & !WasmFeatures::GC);
 
         validator.validate_all(&wasm_bytes).expect(
             "Encoding pre Wasm GC type should not accidentally use Wasm GC specific encoding",
--- a/crates/wasm-shrink/src/lib.rs
+++ b/crates/wasm-shrink/src/lib.rs
@@ -8,6 +8,7 @@ use std::collections::HashSet;
 use anyhow::{Context, Result};
 use rand::{rngs::SmallRng, Rng, SeedableRng};
 use wasm_mutate::WasmMutate;
+use wasmparser::WasmFeatures;
 
 #[rustfmt::skip]
 static EMPTY_WASM: &'static [u8] = &[
@@ -216,33 +217,24 @@ impl ShrinkRun {
     }
 
     fn validate_wasm(&self, wasm: &[u8]) -> Result<()> {
-        let mut validator = wasmparser::Validator::new_with_features(wasmparser::WasmFeatures {
-            // TODO: we should have CLI flags for each Wasm proposal.
-            reference_types: true,
-            multi_value: true,
-            bulk_memory: true,
-            simd: true,
-            threads: true,
-            shared_everything_threads: false,
-            tail_call: true,
-            multi_memory: true,
-            exceptions: true,
-            memory64: true,
-            relaxed_simd: true,
-            extended_const: true,
-            mutable_global: true,
-            saturating_float_to_int: true,
-            sign_extension: true,
-            component_model: false,
-            function_references: false,
-            gc: false,
-            custom_page_sizes: false,
-            component_model_values: false,
-            component_model_nested_names: false,
-
-            floats: true,
-            memory_control: true,
-        });
+        let mut validator = wasmparser::Validator::new_with_features(
+            WasmFeatures::REFERENCE_TYPES
+                | WasmFeatures::MULTI_VALUE
+                | WasmFeatures::BULK_MEMORY
+                | WasmFeatures::SIMD
+                | WasmFeatures::THREADS
+                | WasmFeatures::TAIL_CALL
+                | WasmFeatures::MULTI_MEMORY
+                | WasmFeatures::EXCEPTIONS
+                | WasmFeatures::MEMORY64
+                | WasmFeatures::RELAXED_SIMD
+                | WasmFeatures::EXTENDED_CONST
+                | WasmFeatures::MUTABLE_GLOBAL
+                | WasmFeatures::SATURATING_FLOAT_TO_INT
+                | WasmFeatures::SIGN_EXTENSION
+                | WasmFeatures::FLOATS
+                | WasmFeatures::MEMORY_CONTROL,
+        );
 
         validator.validate_all(wasm)?;
         Ok(())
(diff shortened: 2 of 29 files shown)


crates/wasm-encoder/src/core/types.rs
```diff
@@ -666,10 +668,8 @@ mod tests {
         module.section(&types);
         let wasm_bytes = module.finish();
 
-        let mut validator = wasmparser::Validator::new_with_features(wasmparser::WasmFeatures {
-            gc: false,
-            ..Default::default()
-        });
+        let mut validator =
+            wasmparser::Validator::new_with_features(WasmFeatures::default() | WasmFeatures::GC);
```
alexcrichton, line 672: I think this should turn off GC rather than turning it on?

crates/wasm-shrink/src/lib.rs
```diff
@@ -216,33 +217,24 @@ impl ShrinkRun {
...
+                | WasmFeatures::EXCEPTIONS
+                | WasmFeatures::MEMORY64
+                | WasmFeatures::RELAXED_SIMD
+                | WasmFeatures::EXTENDED_CONST
+                | WasmFeatures::MUTABLE_GLOBAL
+                | WasmFeatures::SATURATING_FLOAT_TO_INT
+                | WasmFeatures::SIGN_EXTENSION
+                | WasmFeatures::FLOATS
+                | WasmFeatures::MEMORY_CONTROL,
+        );
```
alexcrichton, line 237: Mind replacing this with `::all()` since that's now available?

crates/wasm-smith/tests/common/mod.rs
```diff
@@ -2,32 +2,34 @@ use wasm_smith::Config;
...
-        threads: false,
-        shared_everything_threads: false,
-        floats: true,
-        extended_const: false,
-        component_model: false,
-        memory_control: false,
-        component_model_values: false,
-        component_model_nested_names: false,
-    }
+    let mut features = WasmFeatures::MUTABLE_GLOBAL;
```
alexcrichton, line 5: I think this may also want to start with the `floats` feature

crates/wasm-smith/tests/core.rs
```diff
@@ -143,14 +143,12 @@ fn smoke_test_wasm_gc() {
...
-        ..WasmFeatures::default()
-    }
+    WasmFeatures::default()
+        | WasmFeatures::MULTI_MEMORY
+        | WasmFeatures::RELAXED_SIMD
+        | WasmFeatures::MEMORY64
+        | WasmFeatures::EXCEPTIONS
+        | WasmFeatures::TAIL_CALL
+        | WasmFeatures::FUNCTION_REFERENCES
+        | WasmFeatures::GC
```
alexcrichton, line 153: This might actually be another good spot to use `::all()` if you're up for it

crates/wasmparser/src/validator.rs
```diff
@@ -197,98 +198,71 @@ impl Default for State {
...
-    pub component_model_values: bool,
-    /// Support for the nested namespaces and projects in component model names.
-    pub component_model_nested_names: bool,
+bitflags! {
+    /// Flags for features that are enabled for validation.
+    #[derive(Hash, Debug, Copy, Clone)]
+    pub struct WasmFeatures: u32 {
+        /// The WebAssembly `mutable-global` proposal (enabled by default)
+        const MUTABLE_GLOBAL = 1;
+        /// The WebAssembly `mutable-global` proposal (enabled by default)
```
alexcrichton, line 207: copy/paste typoe

crates/wasmparser/src/validator.rs
```diff
@@ -375,34 +349,18 @@ impl WasmFeatures {
...
+            | Self::SIGN_EXTENSION
+            | Self::REFERENCE_TYPES
+            | Self::MULTI_VALUE
+            | Self::BULK_MEMORY
+            | Self::SIMD
+            | Self::RELAXED_SIMD
+            | Self::THREADS
+            | Self::TAIL_CALL
+            | Self::FLOATS
+            | Self::MULTI_MEMORY
```
alexcrichton, line 363: I think this may be missing the component model feature?

fuzz/src/validate.rs
```diff
@@ -27,31 +27,7 @@ pub fn validate_maybe_invalid_module(u: &mut Unstructured<'_>) -> Result<()> {
...
-        saturating_float_to_int: u.arbitrary()?,
-        sign_extension: u.arbitrary()?,
-        memory_control: u.arbitrary()?,
-        function_references: u.arbitrary()?,
-        gc: u.arbitrary()?,
-        custom_page_sizes: u.arbitrary()?,
-        component_model_values: u.arbitrary()?,
-        component_model_nested_names: u.arbitrary()?,
-    });
+    let validator = Validator::new_with_features(WasmFeatures::from_bits_truncate(u.arbitrary()?));
```
alexcrichton, line 30: Nice I like this 👍

src/bin/wasm-tools/validate.rs
```diff
@@ -97,31 +97,31 @@ impl Opts {
...
     let mut ret = WasmFeatures::default();
 
-    const FEATURES: &[(&str, fn(&mut WasmFeatures) -> &mut bool)] = &[
-        ("reference-types", |f| &mut f.reference_types),
-        ("function-references", |f| &mut f.function_references),
-        ("simd", |f| &mut f.simd),
-        ("threads", |f| &mut f.threads),
-        ("shared-everything-threads", |f| {
-            &mut f.shared_everything_threads
+    const FEATURES: &[(&str, fn(&mut WasmFeatures, new_value: bool))] = &[
```
alexcrichton, line 100: If you're up for it this might actuall be best modeled as `&[(&str, WasmFeatures)]` instead of a bunch of closures and that might help clean this up a bit

tests/roundtrip.rs
```diff
@@ -576,87 +576,84 @@ impl TestState {
...
-            sign_extension: true,
-            mutable_global: true,
-            function_references: true,
-            memory_control: true,
-            gc: true,
-            custom_page_sizes: true,
-            component_model_values: true,
-            component_model_nested_names: false,
-        };
+        let mut features = WasmFeatures::empty()
```
alexcrichton, line 579: Reading over this again this'd actually be a good use for `::all()` I think

alexcrichton (approved):
I like how this turned out, thanks!

crates/wasm-shrink/src/lib.rs
Robbepop: we can but then we'd need to disable 7 of the flags again. Not sure if that's more readable.

fuzz/src/validate.rs
Robbepop: I wasn't sure if `from_bits_truncate` is to be preferred because this will silently ignore some bits generated by the fuzzer. Instead we could use `from_bits` and bail out if any unknown bits are set. But I thought this method was more resilient and probably more efficient for fuzzing.

tests/roundtrip.rs
```diff
@@ -576,87 +576,84 @@ impl TestState {
...
+                    features.remove(WasmFeatures::SIGN_EXTENSION);
+                    features.remove(WasmFeatures::SATURATING_FLOAT_TO_INT);
+                    features.remove(WasmFeatures::MUTABLE_GLOBAL);
+                    features.remove(WasmFeatures::BULK_MEMORY);
+                    features.remove(WasmFeatures::FUNCTION_REFERENCES);
+                    features.remove(WasmFeatures::GC);
+                    features.remove(WasmFeatures::CUSTOM_PAGE_SIZE);
+                    features.remove(WasmFeatures::COMPONENT_MODEL);
+                    features.remove(WasmFeatures::COMPONENT_MODEL_VALUES);
+                    features.remove(WasmFeatures::SHARED_EVERYTHING_THREADS);
```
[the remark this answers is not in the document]
Robbepop: I scanned it and I think floats is the only feature needed.

Robbepop:
@alexcrichton Implemented nearly all of your code review comments.

src/bin/wasm-tools/validate.rs
Robbepop: Technically we could maybe even make use of this method if all names match:
https://docs.rs/bitflags/2.5.0/bitflags/trait.Flags.html#method.from_name

alexcrichton (approved):
All looks good to me, thanks again!

fuzz/src/validate.rs
alexcrichton: Nah yeah the `*_truncate` one is the right one to use here to avoid throwing out too many test cases I think

fitzgen:
This change itself is good, but I think we will additionally want to have an `expand` method that creates basically the old struct as well, because there are a few places in Wasmtime that exhaustively match on the struct fields to make sure that we aren't accidentally missing any. For example, when serializing compiled Wasm we also serialize the Wasm features so that we can check whether the compiled code is compatible with the current config when loading that code again. In this scenario, we really don't want to forget a feature.

I can look into this.

fitzgen:
Opened https://github.com/bytecodealliance/wasm-tools/pull/1499 to resolve this issue.

PR #1498, opened 2024-04-16
`wasmparser`: Add `TypesRef` methods to expose some rec group-related data

fitzgen:
We already compute and store all of this internally for validation, just need to expose some of it so that Wasmtime can re-canonicalize rec groups from a Wasm module context to a whole engine context.

### DIFF
--- a/crates/wasmparser/src/validator/types.rs
+++ b/crates/wasmparser/src/validator/types.rs
@@ -1518,6 +1518,22 @@ impl<'a> TypesRef<'a> {
         self.list.get(id)
     }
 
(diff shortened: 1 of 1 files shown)


fitzgen:
After merging this, I'd like to do another `wasm-tools` release. Any blockers on that?

fitzgen:
> After merging this, I'd like to do another `wasm-tools` release. Any blockers on that?

https://github.com/bytecodealliance/wasm-tools/pull/1496#issuecomment-2059497166

alexcrichton (approved):

PR #1506, opened 2024-04-18
`wasmparser`: Add validator identifiers and a `reset` method

fitzgen:
Lets users reuse validators (and their typing contexts) across different Wasm modules. The IDs are for helping them assert they are using the correct context with their `CoreTypeId`s and all that.

### DIFF
--- a/crates/wasmparser/src/validator.rs
+++ b/crates/wasmparser/src/validator.rs
@@ -20,6 +20,7 @@ use crate::{
 use bitflags::bitflags;
 use std::mem;
 use std::ops::Range;
+use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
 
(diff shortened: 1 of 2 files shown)


crates/wasmparser/src/validator.rs
```diff
@@ -86,6 +87,30 @@ fn combine_type_sizes(a: u32, b: u32, offset: usize) -> Result<u32> {
...
+        ValidatorId(ID_COUNTER.fetch_add(1, Ordering::AcqRel))
+    }
+}
+
+impl From<ValidatorId> for usize {
+    #[inline]
+    fn from(id: ValidatorId) -> Self {
+        id.0
+    }
+}
```
alexcrichton, line 112: Is this necessary? We might want to change to u64 one day perhaps

alexcrichton (approved):

PR #1511, opened 2024-04-23
`wasmparser`: add `RelocSectionReader`

jbms:


crates/wasmparser/src/readers/core/reloc.rs
```diff
@@ -0,0 +1,307 @@
...
+    fn from_reader(reader: &mut BinaryReader<'a>) -> Result<Self> {
+        let range = reader.range().clone();
+        let section = reader.read_var_u32()?;
+        Ok(Self {
+            section,
+            range,
+            entries: SectionLimited::new(reader.remaining_buffer(), reader.original_position())?,
+        })
+    }
+}
```
alexcrichton, line 50: Mind dropping this impl? The `FromReader` trait is mostly intended for "parse the next few bytes as this type" but this implementation is taking the whole `BinaryReader` contents, so I think it'd be best to package this up in the `new` constructor above

crates/wasmparser/src/readers/core/reloc.rs
```diff
@@ -0,0 +1,307 @@
...
+back_to_enum! {
+
+    /// Relocation entry type. Each entry type corresponds to one of the
+    /// `R_WASM_*` constants defined at
+    /// https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/BinaryFormat/WasmRelocs.def
+    /// and
+    /// https://github.com/WebAssembly/tool-conventions/blob/main/Linking.md#relocation-sections.
+    #[derive(Debug, PartialEq, Eq, Hash, Clone, Copy)]
+    #[repr(u8)]
+    pub enum RelocationType {
```
alexcrichton, line 83: The number of relocations is probably going to change over time so I'm curious if you have thoughts on having a catch-all `Unknown(u32)` case? I think it's good idea to have named variants no matter what, but my thinking is that it might be reasonable to have a catch-all case for forwards-compatibility perhaps

alexcrichton (approved):
Thanks for this! This all looks good to me 👍, just some minor questions below

crates/wasmparser/src/readers/core/reloc.rs
jbms: It isn't possible to parse an unknown relocation type because we need to know whether it has an addend.  Otherwise we might attempt to parse the first byte of the addend as the next relocation type.

crates/wasmparser/src/readers/core/reloc.rs
jbms: Done.

crates/wasmparser/src/readers/core/reloc.rs
alexcrichton: Good point yeah, sounds good!

PR #1520, opened 2024-04-27
wast: Fix offsets computation for branch-hinting proposal.

jpages:
The branch hinting proposal uses offsets from the "function body". In a previous patch, I implemented this offset as starting from the beginning of the function body (before locals).

Offsets were actually supposed to be relative to the start of the locals vector, see: 
https://github.com/WebAssembly/branch-hinting/pull/26

### DIFF
--- a/crates/wasmprinter/src/lib.rs
+++ b/crates/wasmprinter/src/lib.rs
@@ -1254,7 +1254,6 @@ impl Printer {
         let mut first = true;
         let mut local_idx = 0;
         let mut locals = NamedLocalPrinter::new("local");
-        let func_start = body.original_position();
         for _ in 0..body.read_var_u32()? {
             let offset = body.original_position();
             let cnt = body.read_var_u32()?;
@@ -1279,6 +1278,7 @@ impl Printer {
         }
         locals.finish(&mut self.result);
 
+        let func_start = body.original_position();
         let nesting_start = self.nesting;
         body.allow_memarg64(true);
 
--- a/crates/wast/src/core/binary.rs
+++ b/crates/wast/src/core/binary.rs
@@ -712,7 +712,8 @@ impl Func<'_> {
         // encodes its length first then the body.
         let mut tmp = Vec::new();
         locals.encode(&mut tmp);
-        let branch_hints = expr.encode(&mut tmp, 0);
+        let offset = tmp.len();
+        let branch_hints = expr.encode(&mut tmp, offset);
         tmp.encode(e);
 
         branch_hints


alexcrichton:
Thanks for the PR! I want to confirm though because I'm still a little confused myself. In https://github.com/WebAssembly/branch-hinting/pull/26 the wording was changed to:

> the |U32| byte offset of the hinted instruction, relative to the beginning of the function locals declaration,

I believe this is what wasmparser/wast/wasmprinter currently do today. They're currently calculating an offset relative to the start of the locals declaration, namely after the varint32 which indicates the size of the function body (locals + expressoin). This PR appears to be calculating the offset relative to the end of the section of locals which my my read is not what the clarification PR to the proposal was indicating.

Is this perhaps causing issues in some tools though? Is there another implementation using after-local relative offsets rather than pre-local relative offsets?

jpages:
Yes, it's still a bit confusing...

This related Github issue explains a bit more clearly the problem: https://github.com/WebAssembly/branch-hinting/issues/25

My understanding is that we need to include the locals in the offsets for Branch Hinting. An offset should be: size of locals + offset in the body after that.

We have the following code in the current implementation (in binary.rs):
```
let mut tmp = Vec::new();
locals.encode(&mut tmp);
let branch_hints = expr.encode(&mut tmp, 0);
```

I think we are not including the locals in the offsets right now.

It's not causing any big problem because we're doing the same (miss?)-interpretation of the offsets in SpiderMonkey. It works in SpiderMonkey, but I don't think the offsets are supposed to be as they are now.

alexcrichton:
Hm I still think that wasm-tools is matching the intention of the proposal as-is today without this PR. The code you linked is passing in a 0 indicating that the relative offsets are calculated from position 0 in the `tmp` buffer, where `tmp` contains the entire function, locals and all. Additionally the code modified here in `wasmprinter` is changing the relatative position of the branch hint to starting after the locals declaration since the position is taken after the locals are printed, not before.

I opened https://github.com/bytecodealliance/wasm-tools/pull/1540 which might help showcase this a bit more. That expands `wasm-tools dump` with a bit more logic to handle the branch hints custom section. The dump from that PR looks like:

```
  0x0 | 00 61 73 6d | version 1 (Module)
      | 01 00 00 00
  0x8 | 01 04       | type section
  0xa | 01          | 1 count
--- rec group 0 (implicit) ---
  0xb | 60 00 00    | [type 0] SubType { is_final: true, supertype_idx: None, composite_type: Func(FuncType { params: [], results: [] }) }
  0xe | 03 02       | func section
 0x10 | 01          | 1 count
 0x11 | 00          | [func 0] type 0
 0x12 | 00 23       | custom section
 0x14 | 19 6d 65 74 | name: "metadata.code.branch_hint"
      | 61 64 61 74
      | 61 2e 63 6f
      | 64 65 2e 62
      | 72 61 6e 63
      | 68 5f 68 69
      | 6e 74      
 0x2e | 01          | 1 count
 0x2f | 00          | func: 0
 0x30 | 02          | 2 count
 0x31 | 03 01 00    | BranchHint { func_offset: 3, taken: false }
 0x34 | 06 01 01    | BranchHint { func_offset: 6, taken: true }
 0x37 | 0a 0c       | code section
 0x39 | 01          | 1 count
============== func 0 ====================
 0x3a | 0a          | size of function
 0x3b | 00          | 0 local blocks
 0x3c | 41 00       | i32_const value:0
 0x3e | 04 40       | if blockty:Empty
 0x40 | 0b          | end
 0x41 | 04 40       | if blockty:Empty
 0x43 | 0b          | end
 0x44 | 0b          | end
 ```

where I believe both `BranchHint` entries are correctly pointing to the `if` instruction as a relative offset from the byte 0x3b, the local block start (which is just a single byte here saying zero locals).

If I apply this PR then the output changes like so:

```diff
diff --git a/tests/cli/dump-branch-hints.wat.stdout b/tests/cli/dump-branch-hints.wat.stdout
index 7fe75765..3c01340a 100644
--- a/tests/cli/dump-branch-hints.wat.stdout
+++ b/tests/cli/dump-branch-hints.wat.stdout
@@ -18,8 +18,8 @@
  0x2e | 01          | 1 count
  0x2f | 00          | func: 0
  0x30 | 02          | 2 count
- 0x31 | 03 01 00    | BranchHint { func_offset: 3, taken: false }
- 0x34 | 06 01 01    | BranchHint { func_offset: 6, taken: true }
+ 0x31 | 02 01 00    | BranchHint { func_offset: 2, taken: false }
+ 0x34 | 05 01 01    | BranchHint { func_offset: 5, taken: true }
  0x37 | 0a 0c       | code section
  0x39 | 01          | 1 count
 ============== func 0 ====================
```

but I don't think that's correct because offset 2 for the first `if` instruction yields 0x3c going backwards, which is the start of the function's expression and not its local block.

jpages:
You're right!

Indeed it's fine as it is, thanks for clearing up the confusion and for https://github.com/bytecodealliance/wasm-tools/pull/1540

I'm gonna close this pull-request, thanks again

Issue #1517, opened 2024-04-25
`wasmparser`: Add `no-hash-maps` crate feature & support


Robbepop:
This issue follows https://github.com/bytecodealliance/wasm-tools/pull/1493 and is about the implementation proposals introduced in https://github.com/bytecodealliance/wasmtime/issues/8341#issuecomment-2061519772 and detailed in https://github.com/bytecodealliance/wasmtime/issues/8341#issuecomment-2061975331.

My personal questions about this before I start with the implementation:

- What should the crate feature be called?
- When will `btreemap`-based data structures be enabled?
    - If `#[cfg(feature = "x")]`: Simple design but may be sub-optimal in cases where `std` is enabled.
    - If `#[cfg(and(feature = "x", not(feature = "std"))]`: Less transparent design but uses more efficient hashmaps whenever possible. For Wasmi this would be fine I suppose. I think this is the better design since you only really needs to avoid hashmaps for `no_std` mode since under `std` mode the crate has a proper random source.

Currently the type aliases can be found in this file:
https://github.com/bytecodealliance/wasm-tools/blob/main/crates/wasmparser/src/map.rs

So if I understood correctly the PR for `btreemap`-based data structures shall introduce a new type for each `IndexMap` and `HashMap` that either wraps `hashbrown::HashMap` or `BTreeMap` or `indexmap::IndexMap` or our custom solution, right?

Is this required so that we can have unified trait bounds that require both `T: Ord + Hash` to make uses always compile whatever mode is enabled?

Also we may want to rename the `HashMap` and `HashSet` type aliases to just `Map` and `Set` since then they are not necessarily hash based.

alexcrichton:
Thanks for opening an issue for this!

How about `no-hash-maps` for a name? It'd be an off-by-default feature and when enabled wasmparser would switch all its maps to a btree map?

To ensure that this works well with Cargo features we'll need to change the type aliases for `Index{Map,Set}` to being a `struct Index{Map,Set}` along with all the iterator types and such in the module. Internally it'd then `#[cfg]` between the two implementations. 

Renaming `Hash{Map,Set}` sounds good to me too!

Robbepop:
> How about `no-hash-maps` for a name? It'd be an off-by-default feature and when enabled wasmparser would switch all its maps to a btree map?

[...]

PR #1521, opened 2024-04-27
`wasmparser`: Implement `no-hash-maps` crate feature & support

Robbepop:
Closes https://github.com/bytecodealliance/wasm-tools/issues/1517 .

This PR tries to implement `Set`, `Map` `IndexSet` and `IndexMap` with as similar APIs as possible
to each other but also to Rust's `HashMap`, `BTreeMap` and `indexmap::IndexMap`. 

## TODO

- [x] Restructure `wasmparser::map` module to `wasmparser::collections`.
- [x] Opaque `Set` type:
    - [x] Extended shared API between `hash-set` and `btree-set`.
    - [x] Iterator API
- [x] Opaque `Map` type:
    - [x] Extended shared API between `hash-map` and `btree-map`.
    - [x] Entry API
    - [x] Iterator API
- [x] Opaque `IndexSet` type:
    - [x] Minimally required API
    - [x] Iterator API
    - [x] `IndexMap<T, ()>`-based custom Implementation
- [x] Opaque `IndexMap` type:
    - [x] Minimally required API
    - [x] Entry API
    - [x] Iterator API
    - [x] `btree`-based Implementation
        - [x] Minimally required API
        - [x] Entry API
        - [x] Iterator API
        - [x] Unit tests


alexcrichton:
Thanks again for this! Wanted to drop by and say I'm traveling this week but I'll review when I get back.

Robbepop:
> Thanks again for this! Wanted to drop by and say I'm traveling this week but I'll review when I get back.

That's fine.

I just finished implementation of all opaque types for `Set`, `Map`, `IndexSet` and `IndexMap` with all the methods and trait implementations required to make them work for the `wasmparser` crate (and workspace).

What is missing are the custom implementations for `IndexSet` and `IndexMap` based on `BTreeMap` which will probably take me some time with proper testing.

One downside with the `indexmap_nostd::IndexMap` fork I just remembered is that it adds another trait bound `K: Clone`. I did this to keep the implementation trivially simple because performance was not critical for me for these data structures. However, it is obviously suboptimal to have an additional `K: Clone` bound. @alexcrichton What would you recommend? Going forward with `K: Clone` or coming up with something better, e.g. wrapping `K` in `Arc` to avoid this which comes with yet another set of problems; OR actually using `unsafe` Rust code to avoid cloning.

Link to source code: https://github.com/Robbepop/indexmap-nostd/blob/main/src/map.rs#L78

Robbepop:
All opaque types for `Map`, `Set`, `IndexMap` and `IndexSet` are now done and the workspace seems to compile and tests are passing.
I think this PR is now ready for the initial review and I will write unit tests once the current design is approved by you @alexcrichton .

Robbepop:
@alexcrichton Inspired by this PR I went forward and implemented something similar on the Wasmi side and it already is used there in the newest beta release: https://docs.rs/wasmi_collections/0.1.0/wasmi_collections/

One downside I probably want to fix is the conditional compilation part so that the custom implementations are always compiled and not just when they are in use. While this increases compilation time of the crate it makes it simpler to maintain with `rust-analyzer`.

crates/wasmparser/src/collections/index_set.rs
```diff
@@ -0,0 +1,241 @@
...
+mod detail {
+    use crate::collections::hash;
+
+    pub type IndexSetImpl<T> = indexmap::IndexSet<T, hash::RandomState>;
+    pub type IterImpl<'a, T> = indexmap::set::Iter<'a, T>;
+    pub type IntoIterImpl<T> = indexmap::set::IntoIter<T>;
+}
+
+#[cfg(feature = "no-hash-maps")]
+mod detail;
```
alexcrichton, line 15: Would it make sense to build this module on `crate::collections::IndexMap<T, ()>` instead of having this `detail` split as well?

crates/wasmparser/src/collections/map.rs
```diff
@@ -0,0 +1,468 @@
...
+    inner: detail::IterImpl<'a, K, V>,
+}
+
+impl<'a, K, V> Iterator for Iter<'a, K, V> {
+    type Item = (&'a K, &'a V);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.inner.next()
+    }
+}
```
alexcrichton, line 330: Mind adding `size_hint` overrides which forward to the underlying iterator as well to this and other impls?

crates/wasmparser/src/collections/map.rs
```diff
@@ -0,0 +1,468 @@
...
+
+impl<K, V> ExactSizeIterator for IntoIter<K, V> {
+    fn len(&self) -> usize {
+        self.inner.len()
+    }
+}
+
+impl<K, V> FusedIterator for IntoIter<K, V> {}
```
alexcrichton, line 402: For these `FusedIterator` impls could this have `where detail::IntoIterImpl<K, V>: FusedIterator` to static-assert that the internal iterator is also fused?

crates/wasmparser/src/collections/map.rs
```diff
@@ -0,0 +1,468 @@
...
+    pub type EntryImpl<'a, K, V> = btree_map::Entry<'a, K, V>;
+    pub type OccupiedEntryImpl<'a, K, V> = btree_map::OccupiedEntry<'a, K, V>;
+    pub type VacantEntryImpl<'a, K, V> = btree_map::VacantEntry<'a, K, V>;
+    pub type IterImpl<'a, K, V> = btree_map::Iter<'a, K, V>;
+    pub type IterMutImpl<'a, K, V> = btree_map::IterMut<'a, K, V>;
+    pub type IntoIterImpl<K, V> = btree_map::IntoIter<K, V>;
+    pub type KeysImpl<'a, K, V> = btree_map::Keys<'a, K, V>;
+    pub type ValuesImpl<'a, K, V> = btree_map::Values<'a, K, V>;
+    pub type ValuesMutImpl<'a, K, V> = btree_map::ValuesMut<'a, K, V>;
+}
```
alexcrichton, line 39: Nice I like how this impl turned out 👍

alexcrichton (commented):
This all looks fantastic to me, thank you again so much for taking on this change!

> In order to implement an efficient BTreeMap based IndexMap one would need something like a RawBrTable similar to [hashbrown's RawTable](https://docs.rs/hashbrown/0.14.5/hashbrown/raw/struct.RawTable.html). @alexcrichton What would you recommend?

I like the current strategy of just requiring `K: Clone`, that's pretty reasonable to me. As you say that can always be relaxed in the future, and that seems fine for a future perf-related PR where it can focus on just improving the perf.

>  I think this PR is now ready for the initial review and I will write unit tests once the current design is approved by you.

For testing, in addition to some unit tests (which I think are ok to be minimal, no need to be super exhaustive), I think it'd be nice to also run the integration tests of other crates with `no-hash-maps` enabled. For example `cargo test --workspace --features no-hash-maps` added to CI. That'll be a nice check that everything in the repository works fine with this feature's implementation of data structures. I'd recommend adding this as a separate job on CI to get some better parallelism on CI (no need for the windows/mac/linux matrix for this sort of test either)

> One downside I probably want to fix is the conditional compilation part so that the custom implementations are always compiled and not just when they are in use

I'm fine to go either way on this. I also like having more code compiled-in by default as it's checked-by-default more often. This crate is already large-ish and has a nontrivial compile time so I suspect that the inclusion of mostly-dead modules won't really break the bank either way. In any case I'll leave this up to you, I'm fine either way.

crates/wasmparser/src/collections/map.rs
Robbepop: You probably meant `where detail::IterImpl<K, V>: FusedIterator`, right?

crates/wasmparser/src/collections/map.rs
Robbepop: Yes, I also did this in my Wasmi implementation already.

crates/wasmparser/src/collections/index_set.rs
Robbepop: I will try it out and do so if it is simple.

crates/wasmparser/src/collections/map.rs
Robbepop: done

Robbepop:
So I just tried to implement `IndexSet<T>` in terms of `IndexMap<T, ()>` but unfortunately it causes `IndexMap` API surface to increase significantly which kinda defeats the purpose imo.

Robbepop:
@alexcrichton I tried to implement `IndexSet<T>` in terms of `IndexMap<T, ()>` as proposed by you. Generally I like the idea, however, the `IndexSet::replace` API that has 1 use in the workspace is not easy to implement this way. The original `IndexSet` uses the low-level `RawMap` for this which we cannot do. Naturally `Map`s do not allow to change their underlying keys.

crates/wasmparser/src/collections/index_set.rs
Robbepop: I tried very hard to implement this, however, especially the `IndexSet::replace` method is very hard to properly implement in terms of `IndexMap<T, ()>` since it requires to change the underlying keys which is not possible directly with `IndexMap`. For this the original `IndexSet` data structure uses the low-level underlying `RawMap` which we cannot simply do. Also tests indicate that even the current `replace` API is bugged since for similar reasons it clones the key instead of actually replacing it. I thought it was enough but tests mandate a change. I am not yet sure how to fix this with the current architecture tbh.

alexcrichton: If `replace` is used rarely then it might be an option to implement that as a simple remove-then-insert perhaps? That should at least semantically be correct and it could always be sped up in the future if needed

alexcrichton:
I think it's reasonable to go ahead and add in more API surface area yeah, especially if it's easy enough to implement

Robbepop:
@alexcrichton I have now done the following

- implemented `IndexSet` in terms of `IndexMap<T, ()>` as suggested by you.
- completed the shared API surface for both `Set` and `Map` types (just what I did in `wasmi_collections`)
- `IndexSet::replace` is implemented in terms of `remove` and `insert` which is inefficient but works as intended

Testing of the custom `IndexMap` implementation still needs to be done.

One question:
Shall we put `#[inline]` on all forwarding methods of `Map`, `Set`, `IndexMap` and `IndexSet`? I did the same for `wasmi_collections` but wanted to ask you first. I think for `wasmparser` in particular it would be beneficial since it seems that the intention is to re-export the `collections` module for use in other crates.

alexcrichton:
Technically the methods don't need `#[inline]` to be candidates for cross-crate inlining due to the use of generics, but for such small methods like these it also wont' hurt too much so if you'd like I think it's reasonable to add.

Robbepop:
@alexcrichton I just added tests for the custom `IndexMap` implementation so this PR is ready for merge from my side.

alexcrichton (approved):
Looks great to me, thanks again for your work on this!

Robbepop:
> Looks great to me, thanks again for your work on this!

Thanks you for making wasmparser compativle for Wasmi. I am very happy to contribute this. :)

PR #1524, opened 2024-04-30
`wasmparser`: Enable the extended constant expressions proposal by default

fitzgen:
It is phase 4, so let's enable it by default.


alexcrichton (approved):
As a note we'll need to remember to disable this in Wasmtime when updating because Wasmtime doesn't fully implement it yet.

PR #1531, opened 2024-05-03
Implement the table64 extension to the memory64 proposal

alexcrichton:
Brings wasm-tools and all the various bits and pieces here up-to-date with the latest version of the spec. This includes support for a toggle between using i32 or i64 as the index type for tables along with all the consequences therein.


alexcrichton:
I'll note that this was mostly motivated by keeping up to date with the testsuite submodule which now has table64 tests, coupled with the fact that implementing it wasn't too too bad. 

In updating the spec testsuite submodule that uncovered a minor issue with the gc proposal's validation of `local.tee` which is fixed in the first commit. The second commit actually updates the submodule and re-blesses all output, and the final commit gets all the tests passing for the spec testsuite submodule.

I'm running the `validate_valid_module` fuzzer locally for awhile to try to suss out any lingering issues there.

crates/wast/src/core/binary.rs
fitzgen, line 1: We really gotta bit the bullet and get this crate using `wasm-encoder` one of these days... Wish I did it years ago...

fitzgen (approved):

crates/fuzz-stats/src/bin/failed-instantiations.rs
```diff
@@ -99,6 +99,10 @@ impl State {
         let mut config = wasm_smith::Config::arbitrary(&mut u)?;
         config.allow_start_export = false;
 
+        // NB: just added "table64" support to this and wasmtime doens't
+        // implement that yet
```
FrankReh, line 103: doesn't

PR #1536, opened 2024-05-07
`wasmparser`: fix `ahash::RandomState` usage in `no_std` mode

Robbepop:
The `ahash::RandomState` usage so far was incorrect in that it always used a new random initialization for each hash causing the same `HashMap` to create different hashes for the same hash inputs.

Discovered in https://github.com/wasmi-labs/wasmi/issues/1018.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,7 +30,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
 dependencies = [
  "cfg-if",
- "const-random",
  "once_cell",
  "version_check",
  "zerocopy",
@@ -285,26 +284,6 @@ name = "colorchoice"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "acbf1af155f9b9ef647e42cdc158db4b64a1b61f743629225fde6f3e0be2a7c7"
-
-[[package]]
-name = "const-random"
-version = "0.1.18"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "87e00182fe74b066627d63b85fd550ac2998d4b0bd86bfed477a0ae4c7c71359"
-dependencies = [
- "const-random-macro",
-]
-
-[[package]]
-name = "const-random-macro"
-version = "0.1.16"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f9d839f2a20b0aee515dc581a6172f2321f96cab76c1a38a4c584a194955390e"
-dependencies = [
- "getrandom",
- "once_cell",
- "tiny-keccak",
-]
 
 [[package]]
 name = "constant_time_eq"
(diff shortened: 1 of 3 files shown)


alexcrichton:
Hm can you detail a bit more what's going on here? I don't think we want a compile-time rng because that means that the output of compiling the same source twice will be different, right? (e.g. builds are no longer determinstic)

I had a similar issue show up in Wasmtime recently with no_std work which was fixed by updating the `indexmap` dependency from 2.0.0 to the latest 2.2.6. I didn't bottom out why it fixed the issue, but I presumed it was a bugfix somewhere along the lines. Could that perhaps be applicable to the issue you're seeing too?

Robbepop:
Maybe I misunderstood how `ahash::RandomState` works but I found this issue to be causing the following behavior: For `StringInterner` (which uses `hashbrown::HashMap` under the hood) it caused calls to `make_hash` return different hashes when using this `ahash::RandomState` implementation thus leading to incorrect behavior whereas with my fixed version it correctly returned the same hash.

1. https://github.com/Robbepop/string-interner/blob/c2d9ad2099fb4f4dae0ebbc0575ca866aafcaec0/src/interner.rs#L175
2. https://github.com/Robbepop/string-interner/blob/c2d9ad2099fb4f4dae0ebbc0575ca866aafcaec0/src/interner.rs#L206

Maybe `StringInterner` uses the underlying `hashbrown::HashMap` type improperly but I fail to see how.

alexcrichton:
Do you rely on `make_hash` returning the same value across different instances of a hash map? If that's the case then that's where the issue lies. The `RandomState` from ahash means that all hash maps are seeded differnetly, whereas in the standard library only hash maps on different threads are seeded differently

Robbepop:
> Do you rely on `make_hash` returning the same value across different instances of a hash map? If that's the case then that's where the issue lies. The `RandomState` from ahash means that all hash maps are seeded differnetly, whereas in the standard library only hash maps on different threads are seeded differently

No, I do not rely or assume that different instances hash the same values equally. But I do rely on the assumption that the same `HashMap` instance hashes values equally.

I thought the change introduced by this PR was trivial since now `RandomStateImpl` under both configs is stateful whereas before it was stateless in `no_std` mode and stateful in `std` mode, so kinda missing the symmetry. Also the change fixed all issues for me.

Though after this conversation I am now super confused and need to take yet another look at the whole thing.

alexcrichton:
Ok makes sense, and rereading this this is definitely needed as a fix I agree. Mind removing the `compile-time-rng` feature though? Other than that I think this should still be sufficient to fix the issue

Robbepop:
> Ok makes sense, and rereading this this is definitely needed as a fix I agree. Mind removing the `compile-time-rng` feature though? Other than that I think this should still be sufficient to fix the issue

Okay glad to hear. I am currently not at home but will try out if it works without the feature. I used it for ahash::RandomState to have its Default implementation. Otherwise we can only instantiate it via fixed seed or we try to use `runtime-rng` crate feature if possible.

alexcrichton:
Ah I remember that now and I believe we'll need a manual `Default` calling `ahash::RandomState::new()`. I believe they left out `Default` intentionally to alert users to the various guarantees around that which is ok for our use case.

(sorry again for the runaround I was confused at the start of this looking back)

alexcrichton (approved):

alexcrichton:
This fixes an issue in a project I was working with locally as well so I hope you don't mind but I pushed up a second commit which removes the compile-time-rng feature and adds a manual `Default` impl.

Robbepop:
> This fixes an issue in a project I was working with locally as well so I hope you don't mind but I pushed up a second commit which removes the compile-time-rng feature and adds a manual `Default` impl.

Yes totally fine, glad this is fixed now. Sorry, I simply was not at home to do it myself.

alexcrichton:
Oh no worries! I mostly try to avoid pushing to others' PRs directly, but wanted to get the wheels turning in this case getting the fixes out.

PR #1541, opened 2024-05-09
Initial support for value section parsing

primoly:
This draft adds support for component value sections to wasmparser.
https://github.com/WebAssembly/component-model/pull/336

Questions, issues and remarks:

The validator currently does not support validation of the actual values:

The parsing of values depends on the component type space, so a value section can’t be parsed in isolation https://github.com/WebAssembly/component-model/issues/352 How should this be handled? Currently, `ComponentValue` only parses the type and keeps the bytes of the `val`. To read the actual value the `val` method takes a slice of `ComponentType`s which must be in the order they were read from the current components `ComponentTypeSectionReader`.

wasmparser duplicates the types in validator and reader. Currently `values.rs` uses the ones from reader. Should it use `types::Types` from validator instead?

Validation only needs to call `val`, as the parser then checks if the types of values are correct.

Crates depending on wasmparser (such as wasmprinter) need to keep track of the component type space. wasmprinter already keeps some global module/component information in the `state` variable. wasmparser itself does that in its validator.

Parsing of `flag`s that set labels past 64 to `true` currently fails. Needs support for arbitrarily large unsigned LEB128.

How should `flags` be stored in wasmparser? Explainer.md: list of `label`s of fields set to `true`. For wasmparser I think this will likely confuse people into thinking the `u32`s are just `bool`s encoded as integers. So I instead chose `Vec<bool>`, which would treat it as a special `record` with only bool fields (it already is). More symmetrical and ergonomic.

Should `Record`, `Flags`, `Variant` and `Enum` store the label names? They are already provided in the `ComponentType`. `Flags` could then also be a `Vec<&str>`.

`BinaryReaderError`s use the offset `reader.original_position()`. Might not always be correct (off by one or couple bytes).

### DIFF
--- a/crates/wasmparser/src/readers/component/values.rs
+++ b/crates/wasmparser/src/readers/component/values.rs
@@ -255,7 +255,7 @@ fn read_val<V: Val>(
             record.end();
         }
         ComponentDefinedType::Variant(variant_ty) => {
-            let label = read_label(variant_ty.cases.len() as u32, reader)?;
+            let label = reader.read_var_u32()?;
             if let Some((name, case_ty)) = variant_ty.cases.get_index(label as usize) {
                 if let Some(case_ty) = case_ty.ty {
                     read_val(reader, case_ty, types, visitor.variant_case(label, name))?;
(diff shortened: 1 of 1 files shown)


alexcrichton (commented):
It looks to me like you got most of the bits and pieces all sorted out and this looks great to me, thanks!

alexcrichton:
The serde crates/traits might be a good inspiration to draw from for this use case? They're solving a pretty similar problem and might be good to model after.

PR #1542, opened 2024-05-09
Use libtest-mimic for wit-component tests

alexcrichton:
Makes it easier to run individual tests and provides better output of the test binaries as well.


pchickey (approved):
Cool, I didn't know about that library but its a nice improvement!

PR #1548, opened 2024-05-09
 Store a full WasmFeatures inside of BinaryReader 

alexcrichton:
Historically wasmparser has had a pretty clean split between validation
and parsing with the idea that the parser produces a bunch of stuff and
validation weeds it all out depending on activated features. Parsing has
historically not had access to the set of active features to have it
work the same no matter what. Over time, though, this has become quite
the burden.

The integration with the spec test repo in this repository tries to
make sure that any invalid wasm binary matches the same error message,
with a degree of flexibility. Not being able to take wasm features into
account when parsing makes this quite difficult. For example WebAssembly
binaries invalid before a feature is implemented might be valid after a
feature is implemented. They might also have an entirely different class
of error after a feature is implemented. Effectively WebAssembly does
not guarantee stability of the error message for invalid wasms as it
evolves, only that valid wasms all remain interpreted the same way.

Wasmparser has historically had a number of hacks around this. The
`roundtrip.rs` test suite has a very large function trying to match
error messages between wasmparser and the spec interpreter. Lots of
special cases happen here for wasmparser's feature-agnostic parser when
run against the spec test suite where each proposal has its own copy of
the spec interpreter, possibly with subtly different binary decoders.
Wasmparser additionally has hacks present in the public API such as
`BinaryReader::allow_memarg64` and the `table_byte` field of the
`CallIndirect` instruction. These have historically exclusively been
around to get the spec tests passing but otherwise serve no purpose.

Overall, there's been a lot of pain historically from not being able to
understand active wasm features when parsing wasm. It's quite easy to
write a code path along the lines of "if this feature is active parse
this way otherwise parse this way", but the features have never been
available to test this. Additionally `BinaryReader` instances were
created in quite a few locations throughout `wasmparser` which would
mean threading these around would be quite difficult. This commit
changes all of this.

This commit replaces the `allow_memarg64: bool` field of `BinaryReader`
with `features: WasmFeatures`. This means that all instances of
`BinaryReader` have access to all activated features and will be able to
parse differently when a feature is disabled or enabled. This should
make it much easier to pass the various stages of spec tests depending
on what happens to be merged into the spec interpreter.

To assist with threading this value around and to ensure it's accurate
the `new_with_offset` constructor of `BinaryReader` was removed and the
only remaining `new` constructor now requires that features must be
specified. Much of `wasmparser` has now additionally been refactored to
take a `BinaryReader` as a constructor rather than a slice-and-offset.
This centralizes the construction of `BinaryReader` and cuts down on
constant conversions in-and-out of a `BinaryReader`. For example much of
wasmparser now forwards along `BinaryReader` instances as-is. This is a
large-ish API change, too, for anyone constructing these manually
(shouldn't affect those exclusively using the validator).

The intention is that it's generally ok to pass `WasmFeatures::all()` to
the construction of a `BinaryReader`. For example the
`wasmparser::Parser` type defaults to having all features active. This
means that errors will be different (or not present) for newer wasm
features but they'll all still get filtered out by the validator. The
features in the `BinaryReader` are not intended to be a strict gating
procedure such that it's an exact wasm MVP parser, for example, instead
only being available as necessary for spec tests and other relevant edge
cases.

### DIFF
--- a/crates/wasm-encoder/src/core/code.rs
+++ b/crates/wasm-encoder/src/core/code.rs
@@ -83,11 +83,13 @@ impl CodeSection {
     /// into a new code section encoder:
     ///
     /// ```
+    /// # use wasmparser::{BinaryReader, WasmFeatures, CodeSectionReader};
     /// //                  id, size, # entries, entry
     /// let code_section = [10, 6,    1,         4, 0, 65, 0, 11];
     ///
     /// // Parse the code section.
-    /// let reader = wasmparser::CodeSectionReader::new(&code_section, 0).unwrap();
+    /// let reader = BinaryReader::new(&code_section, 0, WasmFeatures::all());
+    /// let reader = CodeSectionReader::new(reader).unwrap();
     /// let body = reader.into_iter().next().unwrap().unwrap();
     /// let body_range = body.range();
     ///
--- a/crates/wasm-metadata/src/lib.rs
+++ b/crates/wasm-metadata/src/lib.rs
@@ -9,8 +9,8 @@ use std::mem;
 use std::ops::Range;
 use wasm_encoder::{ComponentSection as _, ComponentSectionId, Encode, Section};
 use wasmparser::{
-    ComponentNameSectionReader, KnownCustom, NameSectionReader, Parser, Payload::*,
-    ProducersSectionReader,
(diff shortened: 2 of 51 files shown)


pchickey (approved):
Looks good. Thanks for all the documentation and explanation, as always.

PR #1561, opened 2024-05-15
Fix outdated common `Resolve::push`

### LINKED ISSUE #1559: obsolete comment for `push`
The part of the comment describing the `deps` argument is wrong now that the function signature has changed and there is no longer a `deps` argument.

I didn't want to just remove the line in a PR, hoping a maintainer would have a newer rewrite for the whole function doc string, as the act of pushing an unresolved package is such a pivotal  part of building a complete set of resolved WIT packages. There may be nuances that the function signature cannot capture.

https://github.com/bytecodealliance/wasm-tools/blob/c6732bb781f1d5bcb51bf1f175a5fdd623c3652e/crates/wit-parser/src/resolve.rs#L339-L340

alexcrichton:
Closes #1559

### DIFF
--- a/crates/wit-parser/src/resolve.rs
+++ b/crates/wit-parser/src/resolve.rs
@@ -336,11 +336,13 @@ impl Resolve {
     /// Appends a new [`UnresolvedPackage`] to this [`Resolve`], creating a
     /// fully resolved package with no dangling references.
     ///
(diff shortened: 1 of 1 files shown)


FrankReh:
Thanks. Nice description!

pchickey (approved):

PR #1562, opened 2024-05-16
Add manifest metadata for `cargo binstall`

alexcrichton:
I've seen this recommended a number of times for installing precompiled binaries and this enables `cargo binstall` to pull from the precompiled artifacts produced on CI rather than needing a third-party source.


alexcrichton:
I've tested this out a bit [over on wasm-component-ld](https://github.com/bytecodealliance/wasm-component-ld/pull/18) and it seems to work well there so copying over here too.

pchickey (approved):

PR #1563, opened 2024-05-16
threads: add `global.atomic.rmw.*` instructions

abrown:
This change adds support for the various `global.atomic.rmw.*` instructions that are now available as a part of the shared-everything-threads [proposal].


abrown:
@alexcrichton, I'm playing around with generating all of the test cases programmatically for completeness and efficiency. But there's a lot there. Feel free to recommend a different way to organize those tests!

crates/wasmparser/src/validator/operators.rs
```diff
@@ -1676,6 +1705,63 @@ where
...
+    ) -> Self::Output {
+        let ty = self.check_atomic_global_rmw_ty(global_index)?;
+        self.check_atomic_binary_op(ty, ty)
+    }
+    fn visit_global_atomic_rmw_xchg(
+        &mut self,
+        _ordering: crate::Ordering,
+        global_index: u32,
+    ) -> Self::Output {
+        let ty = self.check_atomic_global_rmw_ty(global_index)?;
```
alexcrichton, line 1753: I think this isn't quite right since this should additionally allow any `anyref` (or subtype) global according to the spec

crates/wasmparser/src/validator/operators.rs
```diff
@@ -1676,6 +1705,63 @@ where
...
+        self.check_atomic_unary_op(ty)
+    }
+    fn visit_global_atomic_rmw_cmpxchg(
+        &mut self,
+        _ordering: crate::Ordering,
+        global_index: u32,
+    ) -> Self::Output {
+        let ty = self.check_atomic_global_rmw_ty(global_index)?;
```
alexcrichton, line 1761: Like above I think that this isn't quite right where it should allow any `eqref` (or subtype) global

crates/wasmparser/src/validator/operators.rs
```diff
@@ -1163,6 +1186,12 @@ where
             .ok_or_else(|| format_err!(self.offset, "unknown tag {}: tag index out of bounds", at))
     }
 
+    fn global_type_at(&self, at: u32) -> GlobalType {
+        self.resources
+            .global_at(at)
+            .expect("existence should be checked prior to this point")
+    }
```
alexcrichton, line 1193: This'll want to return a `Result<_>` to handle out-of-bounds global indices. Also if you're up for it there's probably some other instructions which could use this helper as well, but it's fine to defer that too

crates/wasmparser/src/validator/operators.rs
```diff
@@ -938,17 +938,31 @@ where
...
+    /// Checks the validity of atomic binary operator on memory.
+    fn check_atomic_binary_memory_op(&mut self, memarg: MemArg, op_ty: ValType) -> Result<()> {
         let ty = self.check_shared_memarg(memarg)?;
+        self.check_atomic_binary_op(op_ty, ty)
+    }
+
+    /// Checks the validity of a common atomic binary operator.
+    fn check_atomic_binary_op(&mut self, op_ty: ValType, ty: ValType) -> Result<()> {
+        // TODO: why did the pre-existing code treat `op_ty` and `ty` as
+        // different types?
```
alexcrichton, line 950: Not the best names but for the previous code one type was the index type of memory and another was the type of the operation.

tests/local/shared-everything-threads/global.wast
```diff
@@ -12,26 +48,363 @@
...
+    local.get $y
+    global.atomic.rmw.sub seq_cst $b)
+  (func (export "rmw-sub-i32-acq_rel-$a") (param $x i32) (param $y i32) (result i32)
+    local.get $x
+    local.get $y
+    global.atomic.rmw.sub acq_rel $a)
+  (func (export "rmw-sub-i32-acq_rel-$b") (param $x i32) (param $y i32) (result i32)
+    local.get $x
+    local.get $y
+    global.atomic.rmw.sub acq_rel $b)
```
alexcrichton, line 219: Shouldn't this, and others, only take one operand on the stack? It's the global -= one stack value, not two stack values? (I think cmpxchg is the only one taking two operands)

crates/wasmparser/src/validator/operators.rs
abrown: The other possible users of this were `global.get` and `global.set`, so I refactored them to use the updated version of this function.

alexcrichton (approved):
Thanks! Could you also add some tests for gc references too? For example externref with cmpxchg/rmw.add doesn't work, but xchg works with anyref and cmpxchg works with any eqref

abrown:
> Thanks! Could you also add some tests for gc references too? For example externref with cmpxchg/rmw.add doesn't work, but xchg works with anyref and cmpxchg works with any eqref

I don't think we can express that yet because of:

https://github.com/bytecodealliance/wasm-tools/blob/c034aafe906766f379ffad9b7447004a74b7364f/crates/wasmparser/src/readers/core/types.rs#L822-L828

abrown:
And it's kind of a weird parsing issue because I'm not sure we want to write: `(shared mut (shared anyref))`, right? It seems like references should "inherit" sharedness from the context they're in?

Issue #1462, opened 2024-03-19
Don't require the `wit/` root directory to contain a package


lann:
As an incremental step along the lines of #1461, it would be nice to allow a `wit/` directory to _not_ contain a root/default package.

The scope of this issue would keep the `deps/*/*.wit` structure in place to simplify the change.

Issue #1461, opened 2024-03-19
Relax WIT directory structure


lann:
Zulip context: https://bytecodealliance.zulipchat.com/#narrow/stream/327223-wit-bindgen/topic/WIT.20directory.20structure

Currently, most WIT tooling expects a specific directory layout, typically rooted in a directory named `wit/`:

`wit/*.wit`: Mandatory "default" package
`wit/deps/*/*.wit`: Optional "dependency" packages, which can be referenced by other packages

The only way to store multiple packages in the same place is to (often arbitrarily) partition them into one default package and all others as "deps". This is awkward for repositories that don't have exactly one natural default package, such as the [WASI 0.2 repo](https://github.com/WebAssembly/WASI/tree/main/preview2) (ref #1433).

I'm proposing two changes:

* Relax this directory structure into basically one rule: any directory under a "WIT root" (including the root itself) that contains at least one `.wit` file will be parsed as a package. 
* Update tooling to accept multiple WIT root directories to resolve.

These changes should be backward compatible with the existing layout and permit some new conventions, such as:

- The WASI 0.2 repo doesn't have to change; if someone wants to consume it directly they can point at the entire `preview2/` directory or individual `preview2/*` packages as WIT roots.
- Projects with many dependencies can organize them by namespace, e.g. `wit/wasi/{http,io}/*.wit`, `wit/my-namespace/my-package@1.0.0/*.wit`
- Registry tooling can use unambiguous paths with e.g. content-addressed files

PR #1577, opened 2024-05-26
Support multiple package in a single WIT source

azaslavsky:
This change implements the narrow version of the work proposed at https://github.com/WebAssembly/component-model/issues/313, and added to the language specification in https://github.com/WebAssembly/component-model/pull/340. Multiple packages may now be described in a single WIT file, also known as "explicit" package declaration (ie, a block scope for every package), in addition to the already support "implicit" mode (a single package, spread over one or more files, with no block scope at the top level).

Relevant tools, like `component embed` and `component semver-check`, have been updated and given tests where required.

A great number of APIs that previously assumed that they would receive exactly one `Package`/`PackageId`/other package-describing object, or a `Result` thereof, now receive vectors of these objects. This spidered out to various call-sites and tools in a fairly broad manner.  I've made an effort to have these tools gracefully handle multiple packages where necessary, or, in cases where the tool cannot accept multiple packages, gracefully fail on these sorts of inputs.

The AST has been modified a good deal to  accommodate these new changes. Specifically, the concept of a `DeclList` has been introduced, to contain all of the non-`package` declarations in a package scope. At the top-level, the AST is now a union of either a list of `ExplicitPackage`s or a single `PartialImplicitPackage`; we try to deduce which of these two modes the given WIT file is in as early as possible, so that we can properly establish scope and start building the relevant package-level/file-level `DeclList`, respectively. Care is taken throughout the parser to catch scenarios where the explicit and implicit styles have been mixed and return a descriptive error to the user explaining that this is not allowed.

As of this PR, single packages written in the explicit style (`package my-only-package:-in-this-file { ... }`) are allowed. My feeling is that the parser should accept this, but a future linter should discourage it and "auto-unwrap" such spellings into the implicit style.


azaslavsky:
I believe this is ready for review.

Afaict, the failing tests are all failing on head, with the exception of `Rustfmt`. However, when I run `cargo fmt` locally it produces no changes, so I'm not sure where the difference in formatting settings may be occurring.

azaslavsky:
Also worth noting for future reviewers: I realize this is a rather large PR, which I think is due to this change modifying a lot of internal APIs to be be vectors of `Package`, rather than singletons. I've tried to make the individual commits meaningful in their own right, so they each have their own commit messages and so on. The first 11  commits in this PR can be reviewed in order if more bite-sized review chunks are desired.

alexcrichton:
> Afaict, the failing tests are all failing on head

Ah looks like we're getting bitten by a warning in a new Rust nightly, and I think this should be fixed by https://github.com/bytecodealliance/wasm-tools/pull/1581

> Also worth noting for future reviewers: I realize this is a rather large PR, which I think is due to this change modifying a lot of internal APIs to be be vectors of Package, rather than singletons

No worries! That comes with the territory of changes like this. I'll dig in today for review.

crates/wit-parser/src/resolve.rs
```diff
@@ -158,69 +191,48 @@ impl Resolve {
...
+            }
+
+            ret.push((last.unwrap(), files));
         }
+
+        Ok(ret)
     }
 
+    // TODO: this is a pretty naive implementation, since for example it doesn't re-use dependencies
+    // between "multi-"packages. Could maybe do better going forward?
```
alexcrichton, line 235: I'm a bit worried about this in that `.insert` below will silently overwrite preexisting packages. Instead I think it'd be best to preserve everything to get errors if necessary.

In refactoring this, what about:

* This method returns `Vec<UnresolvedPackage>` instead of a `BTreeMap`
* The list of packages from the original parse that callers have is appended to this array
* This one big array of `UnresolvedPackage` is top-o sorted with `visit` plus perhaps a `sort_by_key` or similar to generate a sorted array
* The big array is iterated over with `.into_iter()` and pushed iteratively into `Resolve`

crates/wit-component/src/printing.rs
```diff
@@ -50,35 +50,41 @@ impl WitPrinter {
...
-        for (name, id) in pkg.interfaces.iter() {
-            self.print_docs(&resolve.interfaces[*id].docs);
-            self.output.push_str("interface ");
-            self.print_name(name);
-            self.output.push_str(" {\n");
-            self.print_interface(resolve, *id)?;
-            writeln!(&mut self.output, "}}\n")?;
-        }
+    /// Print a set of one or more WIT packages into a string.
+    pub fn print(&mut self, resolve: &Resolve, pkg_ids: Vec<PackageId>) -> Result<String> {
```
alexcrichton, line 54: Conventionally a Rust array is typically passed as `&[T]` instead of `Vec<T>` to avoid taking unnecessary ownership.

Additionally should this method dynamically switch between the `package foo:bar;` syntax and `package foo:bar { ... }` depending on how many entries there are in `pkg_ids`?

crates/wit-component/src/lib.rs
```diff
@@ -79,6 +79,43 @@ impl From<StringEncoding> for wasm_encoder::CanonicalOption {
...
+/// Handles world name resolution for cases when multiple packages may have been resolved. If this
+/// is the case, and we're dealing with input that contains a user-supplied world name (like via a
+/// CLI command, for instance), we want to ensure that the world name follows the following rules:
+///
+///   * If there is a single resolved package with a single world, the world name name MAY be
+///     omitted.
+///   * If there is a single resolved package with multiple worlds, the world name MUST be supplied,
+///     but MAY or MAY NOT be fully-qualified.
+///   * If there are multiple resolved packages, the world name MUST be fully-qualified.
+pub fn resolve_world_from_name(
```
alexcrichton, line 91: This is more-or-less a replacement of `Resolve::select_world`. In that case what would you think of basically replacing `select_world` with this? That way callers can largely stay the same where they'll pass in a list of ids here rather than a single id and the `"foo"` selector will only work if the list of packages is of length 1.

crates/wit-parser/src/resolve.rs
```diff
@@ -341,13 +352,31 @@ impl Resolve {
...
+    /// will be returned here, if successful a package identifier is returned
+    /// which corresponds to the package that was just inserted.
+    pub fn append(&mut self, unresolved: Vec<UnresolvedPackage>) -> Result<Vec<PackageId>> {
+        let mut ids = Vec::new();
+        for mut pkg in unresolved {
+            let source_map = mem::take(&mut pkg.source_map);
+            ids.push(source_map.rewrite_error(|| Remap::default().append(self, pkg))?);
+        }
+        Ok(ids)
+    }
```
alexcrichton, line 378: Since this is "just" a simple loop calling `push`, could this be left out instead?

Either that or this might perhaps need some extra refactoring to make it work. Namely if this method take a list of unresolved packages then I'd expect it to automatically topologically sort the packages for insertion into `Resolve` so it's not on the user to manually sort the packages.

One idea perhaps is to have `push_dir` bottom out here where the topo-sort performed there is now performed here instead?

crates/wit-parser/src/resolve.rs
```diff
@@ -158,69 +191,48 @@ impl Resolve {
...
+            let mut files = Vec::new();
+            let mut pkg = Some(pkg);
+            for name in order {
+                let pkg = deps.remove(&name).unwrap_or_else(|| pkg.take().unwrap());
+                files.extend(pkg.source_files().map(|p| p.to_path_buf()));
+                let pkgid = self.push(pkg)?;
+                last = Some(pkgid);
+            }
+
+            ret.push((last.unwrap(), files));
```
alexcrichton, line 228: This is me musing a bit here, but I wanted to highlight this a bit. No action needed in this PR, but figured I'd jot down my ideas.

Here the previous API returned the `PackageId` of the "top" package. That's being much more ambiguous after this PR because the concept of a "top" package is sort of gone where the top thing might have a whole bunch of packages. Once we buy into the concept of multiple "top" packages then some other related issues like https://github.com/bytecodealliance/wasm-tools/issues/1461 and https://github.com/bytecodealliance/wasm-tools/issues/1462 become more digestable (IMO at least).

AFAIK the only downside to having multiple "top" packages is the "world selection" process. Currently that allows the shorthand of specifying a single `foo` world name (or no name at all) and an appropriate world is selected. If we can reframe that functionality somehow to work with a set of "top" packages then it'd be pretty easy to implement those issues. That being said I don't know how best to rationalize how to select a world from a set of "top" packages.

I also bring this up because I can see here that maintaining a return value of the "top" packages is kinda awkward here and might be more awkward with one of my suggestions below of using `append` here. Anyway just wanted to muse a bit here in case you were inspired by something and had an idea.

crates/wit-parser/src/decoding.rs
```diff
@@ -359,16 +359,16 @@ impl DecodedWasm {
...
-            DecodedWasm::WitPackage(resolve, _) => resolve,
+            DecodedWasm::WitPackages(resolve, _) => resolve,
             DecodedWasm::Component(resolve, _) => resolve,
         }
     }
 
-    /// Returns the main package of what was decoded.
-    pub fn package(&self) -> PackageId {
+    /// Returns the main packages of what was decoded.
+    pub fn packages(&self) -> Vec<PackageId> {
```
alexcrichton, line 368: This might be best to return `&[PackageId]` instead of a `Vec`. You can turn `&T` into `&[T]` with [`slice::from_ref`](https://doc.rust-lang.org/stable/std/slice/fn.from_ref.html)

crates/wit-parser/src/ast.rs
```diff
@@ -14,29 +15,129 @@ pub mod toposort;
...
+///   interface i {}
+///   world w {}
+///   /* END SECOND DECL LIST */
+/// }
+/// ```
+pub struct DeclList<'a> {
     items: Vec<AstItem<'a>>,
 }
 
+fn parse_explicit_package_items<'a>(tokens: &mut Tokenizer<'a>) -> Result<DeclList<'a>> {
```
alexcrichton, line 72: Could this, and `parse_implicit_package_items` below, be scoped under a `impl DeclList<'_> { ... }` block?

crates/wit-parser/src/ast.rs
```diff
@@ -1226,15 +1330,58 @@ impl SourceMap {
...
+                        }
+
+                        implicit_resolver.push_partial(partial).with_context(|| {
+                            format!("failed to start resolving path: {}", src.path.display())
+                        })?;
+                    }
+                }
+            }
+
+            if let Some(pkg) = implicit_resolver.resolve()? {
```
alexcrichton, line 1367: If you'd be ok with something slightly wordier here I'm a bit worried about this mixing of partial-and-explicit resolvers and using the wrong resolver by accident. Could this function perhaps have:

```rust
enum PackageResolver {
    Unknown,
    Explicit,
    Implicit(Resolver),
}
```

where it'd start as `Unknown` and transition to other states once a package was seen. That way there's no way for the `ExplicitPackages` side to accidentally use the implicit resolver since it won't ever match on it. Additionally here this method wouldn't be called for explicit packages, only implicit ones, so the `resolve()` method wouldn't need to return an `Option`

alexcrichton (commented):
This looks fantastic to me, thank you again so much for working on this! I'm definitely looking forward to being able to refactor tools to use this and take advantage of it. I think this is going to unlock some quite nice refactorings and downstream abilities of other tooling.

crates/wit-parser/src/lib.rs
```diff
@@ -17,6 +17,7 @@ pub mod abi;
 mod ast;
 use ast::lex::Span;
 pub use ast::SourceMap;
+pub use ast::{parse_use_path, ParsedUsePath};
```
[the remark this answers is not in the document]
azaslavsky: My reasoning was that a `use` path and a "fully-qualified" name, used for identifying something in a WIT file that we are interested for whatever reason, are very similar. For example, there are reasons to want to be able to parse all of the following examples in contexts outside of WIT source files:
```
foo:name/my-cool-world
foo:name/my-cool-world@1.2.3
foo:name/my-cool-world@1.2.3-rc-2023-11-10
foo:name/my-cool-world@1.2.3-rc-2023-11-10.{my-cool-type}
```

My particular use case was for reading `--world [MAYBE_FULLY_QUALIFIED_WORLD_NAME]` from the command line.

I could imagine that this is the a premature abstraction though, since it would also be able to read something like the following, which is probably a bit excessive:
```
foo:name/my-cool-world@1.2.3-rc-2023-11-10.{my-cool-type as my-cool-alias}
```

It's also possible there's another library function for parsing identifiers in the general sense that I missed?

crates/wit-component/src/printing.rs
azaslavsky: > Conventionally a Rust array is typically passed as `&[T]` instead of `Vec<T>` to avoid taking unnecessary ownership.

Whoops, thanks for catching that! Fixed.

> Additionally should this method dynamically switch between the `package foo:bar;` syntax and `package foo:bar { ... }` depending on how many entries there are in `pkg_ids`?

I've made this change, though in doing so I noticed that we have no "pretty printing" tests in and of themselves? The printer seems to only be used in the service of other tools. As currently constituted, there are no tools (in this repo anyway) that would be able to emit multi-package output. For example, `component embed` requires a `--world` argument for multi-package input, narrowing the result to a single pacakge.

We could of course make pretty printing tests to cover this sort of thing more directly. Speaking of - does WIT currently have an official formatter? I couldn't find one, but that doesn't mean one does not exist.

crates/wit-parser/src/resolve.rs
```diff
@@ -122,11 +151,11 @@ impl Resolve {
     ///
     /// Returns the top-level [`PackageId`] as well as a list of all files read
     /// during this parse.
-    pub fn push_path(&mut self, path: impl AsRef<Path>) -> Result<(PackageId, Vec<PathBuf>)> {
+    pub fn push_path(&mut self, path: impl AsRef<Path>) -> Result<Vec<(PackageId, Vec<PathBuf>)>> {
```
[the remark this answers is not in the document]
azaslavsky: Done.

crates/wit-component/src/lib.rs
azaslavsky: I initially had the same instinct, but felt like the current approach may be a little more typesafe?

In either case, the caller has to make sure of something at the call site: in the current setup, they have to make sure to select the right `PackageId` for the `world_name` they're interested in (I can add a comment to the doc-string to help point them to the right helper for doing this), but at least the type signature gives a hint that they need to do this. In the alternative world where `select_world` always takes a `Vec<PackageId>`, I could easily imagine authors forgetting to check that their input `world_name` formatting conforms to the rules dictated by the `Vec<PackageId>` count, and hitting this error condition in (probably relatively rarely tested) multi package scenarios.

azaslavsky:
Thanks for the thorough feedback @alexcrichton!

I've squashed existing commits, merged onto HEAD, and fixed some of the more striaghtforward comments. I'll try to address the remainder over the weekend!

src/bin/wasm-tools/component.rs
```diff
@@ -206,11 +207,12 @@ pub struct EmbedOpts {
...
-    /// argument, to the `world` that the core wasm module works with. This can
-    /// either be a bare string which a document name that has a `default
-    /// world`, or it can be a `foo/bar` name where `foo` names a document and
-    /// `bar` names a world within that document.
+    /// This is the path, within the `WIT` source provided as a positional argument, to the `world`
+    /// that the core wasm module works with. This can either be a bare string which is a document
+    /// name that has a `default world`, or it can be a `foo/bar` name where `foo` names a document
+    /// and `bar` names a world within that document. If the `WIT` source provided contains multiple
+    /// packages, this option must be set, and must be of the fully-qualified form (ex:
+    /// "wasi:http/proxy")
```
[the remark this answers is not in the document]
azaslavsky: Tbh, I'm not entirely clear on which parts of this comment are no longer relevant, so I'm not totally comfortable making deletions :/ My knowledge of the existing state of the tooling still leaves a lot to be desired!

crates/wit-parser/src/lib.rs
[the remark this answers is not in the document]
alexcrichton: Nah makes sense to me! If you've got a need for it I think that it's fine to expose.

That being said for common translations my hope is that it'd go through helpers like `select_world` so most users wouldn't need to reach for this. That doesn't work for everyone though so seems reasonable to have this.

crates/wit-component/src/printing.rs
alexcrichton: For the testing bits, actually if you're up for it I think it'd be great to add a test. I think it'd work well in `tests/cli/*.wit` and the `// RUN: ...` annotation at the top would basically say "please re-print this WIT file". That way you could feed in a multi-package input and the "golden output" would be a multi-package one as well. You're right we're not testing this super well though!

For formatting, not at this time. I know some folks were working on a vscode extension but I'm not sure if it did formatting.

crates/wit-component/src/lib.rs
alexcrichton: I think you're right, but I also think we'll still want to change `select_world`, lemme expand on this though. You're right in that if a single `PackageId` was what the caller already had then it's best to have the current signature of `select_world`. Note though that the `world` argument of `select_world` is not just `"foo"` but it can also be the fully qualified name of `foo:bar/baz`.

Otherwise though, why I think this still needs to change, is that the `select_world` is a method that has not aged well. The intention of the method was to be "here are the two inputs bindings generators typically have and then here's the canonical way to acquire a world". In that sense this is supposed to be **the** helper to acquire a `WorldId` to generate bindings for (and all bindings generators operate at the level of a world). The history of this method looked like:

* Originally there weren't worlds
* Then there were worlds, but there was a `default world` in each package
* Then we wanted the option to select a non-default world, hence the `world: Option<&str>` argument
* Then we removed `default world` but the `Option` remained for "if there's only one world otherwise disambiguate with the name here"
* Then we wanted to select worlds outside of this package so the string syntax was extended to `foo:bar/baz`.

Basically what I'm getting at is that the exact signature of `select_world` is not really all that important because it's pretty rare to have a single `PackageId` after this change. The input from all bindings generators is going to be "here's the set of packages that was defined in the 'main package' now please select a world from here". Given that I think it's best to switch this to `&[PackageId]`. I'd love one day to be able to remove the package argument entirely, but I think that's for a bit further down the road.

crates/wit-parser/src/ast.rs
azaslavsky: Done. I went with a slightly different structure for what I termed `ResolverKind`, which also tracks `Explicit` variants' parsed packages in the ADT as well.

crates/wit-parser/src/ast.rs
azaslavsky: Done.

crates/wit-parser/src/decoding.rs
azaslavsky: Done.

crates/wit-parser/src/resolve.rs
azaslavsky: Done, this is now topologically sorted.

crates/wit-component/src/lib.rs
azaslavsky: If possible, I'd like to save such a refactor for a follow on PR, to get a good baseline landed and free of merge conflicts here.

crates/wit-component/src/printing.rs
azaslavsky: Can I do this in a follow up PR? This one has already gotten a bit unwieldy, and I'd like to get it merged rather than repeatedly rebasing. It seems like adding test coverage is a separate concern anyway.

crates/wit-parser/src/resolve.rs
azaslavsky: I went with a similar approach, based on what you suggested in another thread:

- `Vec<UnresolvedPackage>` becomes `UnresolvedPackageGroup`. This stores a `Vec<UnresolvedPackage>`, plus the `SourceMap` for the entire WIT source. This means that `source_map` is no longer a property of `UnresolvedPackage`, which makes sense: multiple packages can potentially share the same source, so copying this around (or dropping it for packages 1..N, as I was erroneously doing in the previous draft) is wasteful.
- There are now two topological sorts that happen: one for top-level packages, and another for the deps. This made the shared `SourceMap` machinery easier: each `dep` gets a `SourceMap`, as does the top-level package, but there is no need for them to be shared beyond that, since each such `SourceMap` must be a discreet set of one or more complete packages. The sort-twice approach is safe, since the packages in the top-level set of files can never be referenced by a `deps` package, so doing separate sorts and then concatenating the results should be fine. This is essentially what we did before with the .chain call, but now we have to sort the top-level packages as well, since those may contain references into one another.

crates/wit-parser/src/resolve.rs
```diff
@@ -158,69 +191,48 @@ impl Resolve {
...
-                    }
-                    if let Some(dep) = deps.get(dep) {
-                        visit(dep, deps, order, visiting)?;
-                    }
-                    assert!(visiting.remove(dep));
-                }
-                assert!(order.insert(pkg.name.clone()));
-                Ok(())
-            })
+        for pkg in pkgs {
```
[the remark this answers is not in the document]
azaslavsky: Fixed, along with the `.insert` comment below, which describes the new approach.

crates/wit-parser/src/lib.rs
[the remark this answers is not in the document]
azaslavsky: Okay, I'll leave as is for now.

crates/wit-smith/src/lib.rs
```diff
@@ -22,19 +22,26 @@ pub fn smith(config: &Config, u: &mut Unstructured<'_>) -> Result<Vec<u8>> {
...
+                        panic!("bad wit parse: {e:?}")
+                    }
+                };
+                last = Some(id);
             }
-        };
-        last = Some(id);
+            // An empty vector means we had a bunch of empty wit files here.
+            None => return Err(arbitrary::Error::NotEnoughData),
+        }
```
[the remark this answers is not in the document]
azaslavsky: I didn't know enough about `smith` to know if creating all empty files was possible. But if it is not, this formulation works for me. I've made the change you suggest.

crates/wit-parser/src/resolve.rs
```diff
@@ -144,95 +180,118 @@ impl Resolve {
...
-            let id = self.push_file(path)?;
-            Ok((id, vec![path.to_path_buf()]))
+            let ids = self.push_file(path)?;
+            let mut pkg_ids = Vec::new();
+            let mut path_bufs = Vec::new();
+            for id in ids {
+                pkg_ids.push(id);
+                path_bufs.push(path.to_path_buf());
+            }
+            Ok((pkg_ids, path_bufs))
```
alexcrichton, line 190: I think this might be able to stay roughly as it was before, except renaming `id` to `ids`? There's no need to ensure that the two vectors have the same length, it's just that the second vector is the set of files to produce the first.

crates/wit-parser/src/resolve.rs
```diff
@@ -144,95 +180,118 @@ impl Resolve {
...
         }
     }
 
+    fn _sort_unresolved_packages(
```
alexcrichton, line 194: It's ok in general to skip the leading underscore for private methods in Rust since the visibility (or lack thereof here) covers that. Some of the other methods have this underscore prefix but that's just because the public and private versions both wanted to be called the same thing so I gave up on naming an gave the private one a leading `_`. For this case specifically though it's ok to just call this `sort_unresolved_packages`

crates/wit-component/src/printing.rs
alexcrichton: Sounds good to me 👍

crates/wit-component/src/lib.rs
alexcrichton: Sounds good to me 👍

alexcrichton (approved):
Thanks again for this! Just a few more minor things but other than that happy to land 👍

crates/wit-parser/src/resolve.rs
azaslavsky: Done.

crates/wit-parser/src/resolve.rs
azaslavsky: Done.

PR #1578, opened 2024-05-26
Check `...BACKTRACE` variables gracefully before running tests

azaslavsky:
This caught me a few times, especially since I had two terminals open, one with `export RUST_BACKTRACE=1` and one without. It took me longer than I'd like to admit to figure out why tests were passing in one and not the other!

If this is viewed as a desirable check to keep other contributors from tripping over this quirk, it may be useful to extract into its own `utils` crate, since the logic of dealing with the interaction of the `BLESS` and RUST_[LIB_]BACKTRACE` env variables occurs in a number of other internal crates' test suites as well.

### DIFF
--- a/crates/wit-parser/tests/all.rs
+++ b/crates/wit-parser/tests/all.rs
@@ -33,11 +33,11 @@ fn main() {
     // for debugging purposes.
     if bless && wasm_debug_backtrace {
         eprintln!("Cannot set both the `BLESS=1` and `WASM_DEBUG_BACKTRACE=1` env variables simultaneously; blessing a new set of golden values and debugging broken tests should be separate operations.");
-        return
+        return;
     }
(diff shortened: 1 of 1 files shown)


azaslavsky:
> Thanks for this! I've run into this before with anyhow and it's definitely an unfortunate interaction with the "I'd like to bless the output quickly" style of testing. One possible alternative to detecting various cases with env vars here is to change the formatting string [here](https://github.com/bytecodealliance/wasm-tools/blob/f1898f46bb9d96f0f09682415cb6ccfd6a4dca79/crates/wit-parser/tests/all.rs#L89) from `:?`, which includes the backtrace, to `:#` which does not include the backtrace but renders the error slightly differently. That might be the best solution for now since a new env var to debug the tests also wouldn't be necessary?

I wanted to avoid a large "we reformatted the codebase and made git blame sad" style change, but I can make the requested change if that's fine by you, just lmk!

PR #1579, opened 2024-05-26
Add .vscode to .gitignore

azaslavsky:
A nicety that allows VSCode-based developers to set custom settings and launch configurations for this project at their leisure.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ publish
 *.wat
 test.config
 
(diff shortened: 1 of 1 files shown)


pchickey:
Rather than wasmtime need to anticipate the files used by editors, could you use a global gitignore on your system to exclude `.vscode`? https://stackoverflow.com/questions/7335420/global-git-ignore

PR #1549, opened 2024-05-09
[wip] wit-encoder

yoshuawuyts:
I started work on a new `wit-encoder` crate a few weeks ago. I haven't found time to finish this work up, but @MendyBerger mentioned he might have time to progress some of this work - so I'm putting this up as a draft PR. Part as a reference, part so he can maybe push this forward.

Definitely not intended to be merged (yet?). Probably doesn't need to stay open for too long either. I just wanted to make sure to put this up so folks could take a look at it in a place that's easy to reference.

To share the stat of this work: I'd just started implementing the actual backing impls of the various types. To make sure I got this right I was in the progress of porting the parser tests and making sure we're able to generate those programs correctly using `wit-encoder`. To handle indentation I was thinking of implementing something similar to the [`html::Render`](https://docs.rs/html/latest/html/trait.Render.html) trait.

Anyway; despite not being done yet I hope it's okay to share this here. Perhaps someone can run with it, or if not I'll find time to pick it up once my current workload lifts. Thanks!

PR #1580, opened 2024-05-28
Add a wit-encoder tool

MendyBerger:
This PR is a collaboration between me and @yoshuawuyts.

Adds a new tool to allow building up wit with code.

It's not perfect, but I think that it's better than writing wit by hand.


yoshuawuyts:
To add some additional context here: this PR is a continuation of https://github.com/bytecodealliance/wasm-tools/pull/1549. Both Mendy and I found that independently of each other we had been encoding WIT documents by hand. And it seemed useful to have a WIT counterpart to the [`wasm-encoder` crate](https://github.com/bytecodealliance/wasm-tools/tree/main/crates/wasm-encoder). Because correctly encoding rich, structured formats by hand is hard - and so this crate aims to make that a little easier. Because of that, it also handles some of the trickier bits of WIT, like automatically escaping reserved keywords.

## Example

Say we wanted to encode the following document:

```rust
package mendyberger:encoder-example;

interface error-reporter {}

world %world {
    /// inline interface
    export example: interface {
        /// func docs
        do-nothing: func();
    }
    
    /// scan stuff
    export scan: func() -> list<u8>;
    import error-reporter;
    import print: func(s: string);
}
```

Using `wit-encoder` the code for that would look something like this (adapted from [this test](https://github.com/bytecodealliance/wasm-tools/blob/5a94ed1075853564cb33541583479c563598c4a3/crates/wit-encoder/tests/world.rs)):
```rust
use wit_encoder::{
    Interface,
    Package,
    PackageName,
    StandaloneFunction,
    Type,
    World,
};

let mut package = Package::new(PackageName::new("foo", "functions", None));
package.interface(Interface::new("error-reporter"));
package.world({
    let mut world = World::new("world");
    world.inline_interface_export({
        let mut interface = Interface::new("example");
        interface.docs(Some("inline interface"));
        interface.function({
            let mut func = StandaloneFunction::new("do-nothing");
            func.docs(Some("func docs"));
            func
        });
        interface
    });
    world.function_export({
        let mut func = StandaloneFunction::new("scan");
        func.results(Type::list(Type::U8));
        func.docs(Some("scan stuff"));
        func
    });
    world.named_interface_import("error-reporter");
    world.function_import({
        let mut func = StandaloneFunction::new("print");
        func.params(("s", Type::String));
        func
    });
    world
});
```

crates/wit-encoder/src/function.rs
```diff
@@ -0,0 +1,177 @@
...
+        match self {
+            Results::Named(params) => params.items().len(),
+            Results::Anon(_) => 1,
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize))]
+pub struct StandaloneFunction {
```
yoshuawuyts, line 147: Can we rename this to `Function`? That matches the term used in [wit-encoder](https://docs.rs/wit-parser/latest/wit_parser/struct.Function.html) too.

crates/wit-encoder/src/render.rs
```diff
@@ -0,0 +1,50 @@
+use std::fmt;
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct RenderOpts {
```
yoshuawuyts, line 4: Some docs explaining what this does would probably be helpful. It is not immediately clear why there are `RenderOpts`, why we have a `Render` trait rather than use `Display`, etc.

yoshuawuyts (approved):
Some minor nits around naming and docs. But overall this looks really useful - thank you for finishing this up! Being able to use this would very useful for my work on `wasm-http-tools`.

Cargo.toml
```diff
@@ -52,7 +53,10 @@ rayon = "1.3"
 serde = "1.0.166"
 serde_derive = "1.0.166"
 serde_json = { version = "1" }
-wasmtime = { version = "12.0.0", default-features = false, features = ['cranelift', 'component-model'] }
+wasmtime = { version = "12.0.0", default-features = false, features = [
+  'cranelift',
+  'component-model',
+] }
```
alexcrichton, line 59: Would it be ok to back out some of the formatting-related changes in this file?

alexcrichton (commented):
Thanks for continuing to push on this! All looks good to me 👍 

Does it make sense to leave in the serde bits here as well given that the serde bits are also present in wit-parser? (e.g. reducing the amount of json formats in play in theory)

crates/wit-encoder/src/render.rs
MendyBerger: Display is implemented for package.
Render/RenderOpts is mostly for internal use, but figured I'll make it public as we might wanna add in some options later.
I can make it private if you think it's confusing.

crates/wit-encoder/src/function.rs
MendyBerger: We have StandaloneFunction  and ResourceFunc, whereas wit-parser has a single Function type that takes a kind argument.
I made separate types so that the type system enforces where each function kind can be. I.e. that standalone functions are not part of a resource, and method/static/constructor are not without a resource.

(also, should probably rename StandaloneFunction to StandaloneFunc)

Cargo.toml
MendyBerger: Will do.
Was probably just cargo fmt.

MendyBerger:
@alexcrichton are you suggesting that I remove serde from wit-bindgen? Sure, I'm cool with that.

crates/wit-encoder/tests/world.rs
```diff
@@ -0,0 +1,56 @@
...
+
+#[test]
+fn worlds() {
+    let name = wit_encoder::PackageName::new("foo", "functions", None);
+    let mut package = wit_encoder::Package::new(name);
+
+    package.interface(wit_encoder::Interface::new("error-reporter"));
+
+    package.world({
+        let mut wold = wit_encoder::World::new("world");
```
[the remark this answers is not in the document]
MendyBerger: @yoshuawuyts sorry but I'm not sure I understand.
Are you suggesting that I rename the files?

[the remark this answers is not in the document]
yoshuawuyts: Oh haha, no sorry — it looks to me like you typo's the variable name on this file. You wrote "wold" where I assume you meant "world".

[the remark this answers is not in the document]
MendyBerger: Lol. I see it now.

alexcrichton:
Oh sorry no to clarify I mean removing the serde bits from this new crate `wit-encoder`

MendyBerger:
:facepalm:
I meant wit-encoder. No idea why I wrote wit-bindgen. :sweat_smile:

crates/wit-encoder/tests/world.rs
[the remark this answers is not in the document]
MendyBerger: Done

Cargo.toml
MendyBerger: Done

MendyBerger:
I removed serde

crates/wit-encoder/src/function.rs
MendyBerger: I  rename StandaloneFunction to StandaloneFunc

crates/wit-encoder/src/render.rs
MendyBerger: @yoshuawuyts think I should make the render stuff private?

crates/wit-encoder/src/result.rs
```diff
@@ -0,0 +1,60 @@
...
+        write!(f, "result")?;
+        if !self.is_empty() {
+            write!(f, "<")?;
+            if let Some(type_) = &self.ok {
+                type_.fmt(f)?;
+            }
+            if let Some(type_) = &self.err {
+                if self.ok.is_none() {
+                    write!(f, "_")?;
+                }
```
FrankReh, line 52: Maybe this write! could be moved to an else clause for the earlier 'if let' on self.ok? It might read a little better but maybe I'm missing something. 

I could guess this was written before the '!self.is_empty()` was added.

MendyBerger: Good catch @FrankReh!
I've updated it.

PR #1585, opened 2024-05-29
`wasmparser`: Fix validation of the `return_call` family of instructions

fitzgen:
We need to additionally check that the callee's results are an exact match of the caller's results. We were incorrectly allowing return calls that would push more values on the operand stack than would be returned. That is fine with a `call; return` sequence, where extra values on the stack are allowed to dangle, but not okay with a `return_call`. With a `return_call` it doesn't make sense because the callee might need a return pointer to put all its results into, but the caller can't supply one since its frame is going away, nor can the caller forward a return pointer that it received to the callee, since it might not return enough values to require a return pointer. This commit fixes the validation to match the spec and disallow `return_call`s that would leave dangling values on the operand stack.

cc https://github.com/bytecodealliance/wasmtime/issues/8704

### DIFF
--- a/crates/wasmparser/src/validator/operators.rs
+++ b/crates/wasmparser/src/validator/operators.rs
@@ -796,18 +796,22 @@ where
         }
     }
 
-    /// Validates a `call` instruction, ensuring that the function index is
-    /// in-bounds and the right types are on the stack to call the function.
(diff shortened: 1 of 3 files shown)


fitzgen:
Upstreaming these new tests to the spec testsuite here: https://github.com/WebAssembly/function-references/pull/116

PR #1591, opened 2024-06-04
wasm-encoder: fix encoding of f32 and f64 constant values in `TryFrom<wasmparser::ConstExpr>`

12101111:


alexcrichton (approved):
Good catch, thanks for this!

PR #1592, opened 2024-06-04
Refactor `wasmprinter` to rely less on a `String` output

alexcrichton:
This commit aims at starting to refactor `wasmprinter` to be able to output to a stream rather than unconditionally to a `String`. The fact that the current output is always an in-memory `String` is leveraged in a number of places in the internals to undo prints or capture temporary prints. These were all the easiest thing to do at the time but do not play well with the idea of streaming output. This commit instead refactors `wasmprinter` to remove all these bits and be more naturally oriented around "just stream out bytes at that's it". This is achieved with some internal refactorings, reorganization, and juggling some responsibilities.

A future PR will switch the internals to actually be `std::io::Write` or, my final goal, `termcolor::WriteColor`

### DIFF
--- a/crates/wasmparser/src/binary_reader.rs
+++ b/crates/wasmparser/src/binary_reader.rs
@@ -1712,6 +1712,12 @@ impl<'a> BinaryReader<'a> {
         self.visit_operator(&mut OperatorFactory::new())
     }
 
+    /// Returns whether there is an `end` opcode followed by eof remaining in
+    /// this reader.
+    pub fn is_end_then_eof(&mut self) -> bool {
+        self.remaining_buffer() == &[0x0b]
+    }
+
     fn read_lane_index(&mut self, max: u8) -> Result<u8> {
         let index = self.read_u8()?;
         if index >= max {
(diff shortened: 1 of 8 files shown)


crates/wasmparser/src/binary_reader.rs
```diff
@@ -1712,6 +1712,12 @@ impl<'a> BinaryReader<'a> {
         self.visit_operator(&mut OperatorFactory::new())
     }
 
+    /// Returns whether there is an `end` opcode followed by eof remaining in
+    /// this reader.
+    pub fn is_end_then_eof(&mut self) -> bool {
```
sunfishcode, line 1717: Can this be `&self` instead of `&mut self`?

crates/wasmparser/src/readers/core/operators.rs
```diff
@@ -209,6 +209,12 @@ impl<'a> OperatorsReader<'a> {
     pub fn get_binary_reader(&self) -> BinaryReader<'a> {
         self.reader.clone()
     }
+
+    /// Returns whether there is an `end` opcode followed by eof remaining in
+    /// this reader.
+    pub fn is_end_then_eof(&mut self) -> bool {
```
sunfishcode, line 215: Similarly, can this be `&self`?

crates/wasmparser/src/binary_reader.rs
alexcrichton: Oop, indeed!

sunfishcode (approved):

PR #1596, opened 2024-06-07
Add color by default to printing the text format

alexcrichton:
This commit builds on #1593 to provide colorized output of `wasm-tools print`, or any other subcommand that prints the text format, by default. Color settings are controlled on the CLI with `--color` and are automatically disabled in various situations such as outputting to a file or when an output pipe is used.

Internally `wasmprinter` has grown a new trait, `Print`, through which all output goes through. There are adapters for this trait for `std::fmt::Write`, `std::io::Write`, and `termcolor::WriteColor`. Custom trait implementations are also possible if necessary.

While here I've additionally moved custom custom section printing to the trait itself instead of being part of the printer configuration as it felt more appropriate as a per-sink configuration.

I'll note that I am by no means an artist and my sense of taste and color are probably objectively "not good". My hope is to add some basics here which lays the groundwork for someone more artistic than I to improve this in the future. Some sample screenshots of what this looks like are:

Screenshot 2024-06-07 at 5 05 42 PM

Screenshot 2024-06-07 at 5 06 44 PM

Screenshot 2024-06-07 at 5 06 54 PM

Screenshot 2024-06-07 at 5 10 25 PM

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1735,6 +1735,7 @@ dependencies = [
  "diff",
  "rayon",
  "tempfile",
+ "termcolor",
  "wasmparser 0.209.1",
  "wast",
  "wat",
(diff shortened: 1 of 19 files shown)


pchickey (approved):
Awesome!

Issue #1595, opened 2024-06-07
Online playground


bakkot:
I have occasionally wanted to use these tools on the web. I've used [wat2wasm](https://webassembly.github.io/wabt/demo/wat2wasm/) and the other playgrounds from wabt a lot, and the lack of a similar easy way of using wasm-tools keeps tripping me up.

Would you accept a PR which adds a GitHub action which uses the `wasm32-wasi` artifact from a release and wraps it up with a [browser-based wasi implementation](https://runno.dev/wasi) to live on GitHub Pages? I'm imagining something very simple, just a text box / file upload for input and a text box / download button for output plus a place to enter the command and see stdout/stderr.

alexcrichton:
Thanks for the report, and I'd love to see this! I've never been certain enough of what I'd want this to look like to implement it but I think it'd be great to have!

For some of the technical bits it should be ok to actually skip wasi support and use the `wasm32-unknown-unknown` target  since most of the libraries don't need the filesystem or env vars or things like that. If you're ok with it I think it would be best to architect this as something which lives in-tree and is deployed to github pages on all merges to `main`. I'm happy to help with the CI bits here too.

As for the actual integration with the web, I might recommend one of two approaches:

* Using `jco` would enable "componentizing" everything to be easily used from JS. A bit of a self-dogfood if you will.
* Using `wasm-bindgen` would be a bit more old-school but should work well for this purpose.

Whichever you find easiest I think would be a great place to start. If you need any help please just let me know!

bakkot:
Dogfooding sounds fun, but I'd probably need more help with it. I expect I could whip something up with `wasm-bindgen` largely on my own.

For the dogfooding option, what would the process actually look like, concretely? It looks like `wit-component` isn't meant to be used directly, in preference for `cargo component`, so would the idea be to do `cargo install cargo-component` in CI, and then `cargo component build  --target wasm32-unknown-unknown` followed by `jco transpile`?

alexcrichton:
Starting with `wasm-bindgen` sounds reasonable to me. My thinking is that a jco based experience would be as you mention, but that's mostly just to lean into components I don't think it is substantially different than wasm-bindgen at this time otherwise.

bakkot:
The `cargo-component` approach was actually fairly straightforward,

Unfortunately the `wasm32-unknown-unknown` target immediately panics (i.e., hits an `unreachable` wasm instruction) if I try to `wat::parse_str` anything nontrivial.

[...]

PR #1597, opened 2024-06-08
proof-of-concept playground

bakkot:
Fixes https://github.com/bytecodealliance/wasm-tools/issues/1595.

You can try [print playground](https://bakkot-org.github.io/wasm-tools/print) and the [parse playground](https://bakkot-org.github.io/wasm-tools/parse).

It's extremely basic and not very pretty, but it does work. I've only implemented `wasm-tools parse` ~~so far because I want to give maintainers a chance to weigh in on the approach before doing more work. I'm happy to implement another for `wasm-tools print` as well, and maybe a third for `dump`+`objdump`+`metadata show` if you think that's worth doing.~~ EDIT: and the `wasm-tools print` one.

I'm using `cargo-component` and `jco`, mostly because I thought that would be more fun than `wasm-bindgen`. Unfortunately I did need to target `wasm32-wasi` since I get an `unreachable` when targeting `wasm32-unknown-unknown`. And there does not seem to be a simple way to provide my own shims of the preview-1 functions when using `cargo component`; it unconditionally translates them to preview-2. But that's fine because `jco` has built in support for shimming preview-2. This adds about 50Kb of (unminimized) JavaScript, but whatever.

This includes a new workflow which a) ensures the playground builds in every PR and b) uploads the built artifact to gh-pages on every commit to `main`. I think you'll need to manually enable GitHub pages and set the source to "GitHub Actions" for this to actually do anything, but it's possible the included actions will take care of that.

I've commented out the main CI for the sake of not taking up a bunch of resources while iterating on this. I'll revert that before marking this as ready, of course.


playground/component/Cargo.lock
alexcrichton, line 1: Mind adding `playground/component` to the main workspace to avoid a duplicate `Cargo.lock`?

playground/component/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "component"
+edition = "2021"
+
```
alexcrichton, line 4: Might be worth adding a `publish = false` here, and as part of the main workspace thisi can also use `edition.workspace = true` along with `[lints] workspace = true` too

.github/workflows/playground.yml
```diff
@@ -0,0 +1,59 @@
...
+# at a time
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
```
alexcrichton, line 18: Mind adding this to the `main.yml` file as well? The final "gate everything here" job could gate on the `build` job here and the `deploy` job could continue to depend on `main`. Otherwise though this looks great!

I learned about `actions/{configure-pages,upload-pages-artifact,deploy-pages}` here, I didn't know what those were before!

alexcrichton (commented):
Thanks so much for this it looks great to me! I've got a few minor comments but I'm happy to land the overall structure and iterate in-tree as necessary.

I think I just frobbed the button to say github pages comes from actions, so I think that's ready to go?

playground/component/Cargo.lock
bakkot: Done.

playground/component/Cargo.toml
bakkot: Done.

.github/workflows/playground.yml
bakkot: I'm not totally sure what you're asking here - are you suggesting getting rid of `playground.yml` and moving both of these new jobs into `main.yml`? I was aiming to have this run after commits land on `main`, which that file doesn't (appear to) do. But I guess if everything goes through the merge queue it ends up being roughly equivalent?

bakkot:
I'll go ahead and make the `print` page as well, then, and add a root `index.html` which I guess just redirects to the github project (unless you want to have a separate page listing the tools, but that seems a bit redundant). Then I think think it should be fit to land.

playground/component/Cargo.toml
bakkot: Hm, actually, adding `[lints] workspace = true` [prevents building](https://github.com/bytecodealliance/wasm-tools/actions/runs/9433506842/job/25984523557) because the lints don't like the code that `cargo component` generates. I've removed it for the moment; happy to take other suggestions.

bakkot:
Alright, added the [printer](https://bakkot.github.io/wasm-tools/print) as well. This is good to go from my end once the CI thread above is resolved. Which I'm happy to do but if you'd find it easier to make the changes yourself rather than explaining them that's also fine.

playground/component/src/lib.rs
```diff
@@ -0,0 +1,33 @@
...
+struct Component;
+
+struct StringWriter(pub String);
+
+impl wasmprinter::Print for StringWriter {
+    fn write_str(&mut self, s: &str) -> std::io::Result<()> {
+        self.0.push_str(s);
+        Ok(())
+    }
+}
```
alexcrichton, line 15: If you're interested in getting particularly fancy here it might be neat to hook into the `start_*` methods and emit `<span>` entries to have highlighted colos on the web too. That's fine to do later though as a follow-up as well.

playground/pages/parse.html
```diff
@@ -0,0 +1,85 @@
+<!doctype html>
```
alexcrichton, line 1: For discoverability and sharing code do you think that it'd be reasonable to have only one landing page for the various tools and and sort of "tabs" between them to explore different functionality?

playground/component/Cargo.toml
alexcrichton: Ah ok no worries, I can try to work on that once this lands.

playground/pages/parse.html
bakkot: Can do. Are you thinking of one HTML page, or multiple pages that link to each other? Having multiple makes it easier to bookmark a specific tool, though I guess I could use the `#fragment` and some JS to make that possible with a single page.

alexcrichton: I was thinking one page, but I'm not wed to the idea either. It mostly just seemed like there was some duplication between the styles/js/etc. Having links to each page from the other I think is also fine.

.github/workflows/playground.yml
bakkot: > Do you know if it would be difficult to split it up like that?

Just looked and the answer appears to be that yes, it's (a little) difficult. [Artifacts](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts) are meant to be used within a single workflow. To make this work, you'd need to download and re-upload them in the publish job - there doesn't appear to be a (supported) way to publish an artifact from a different run.

You're already [downloading artifacts from a previous run in `publish.yml`](https://github.com/bytecodealliance/wasm-tools/blob/10d2e21b2171cf2a2227f9de04ac89985042a5f0/.github/workflows/publish.yml#L61-L72), but in trying it on my own fork I couldn't immediately get [that](https://github.com/bakkot-org/wasm-tools/blob/179bf220518ff2690948ce79cd10837d15e4cc14/.github/workflows/playground.yml#L51-L84) to work - it's failing to find the `run_id`, despite that working in this project, and I don't have enough experience or interest in GitHub actions to debug it further.

playground/pages/parse.html
bakkot: I added links from each page to the other. There's not really much duplicated that isn't already shared - they use the same core stylesheet and worker, it's just the actual relevant UI which differs.

playground/component/src/lib.rs
bakkot: [Done](https://bakkot-org.github.io/wasm-tools/print), following roughly the terminal colors used for the same parts.

alexcrichton (approved):
All looks reasonable to me, thanks again for providing the foundation for all this!

alexcrichton:
Looks like the [deploy worked](https://bytecodealliance.github.io/wasm-tools/parse), yay! I've got some [minor follow-ups](https://github.com/bytecodealliance/wasm-tools/pull/1605) but when you feel it's ready I think it'd be great to link this from the README in the root of the repo as well

PR #1599, opened 2024-06-09
wasm-tools, wit-parser: add --all-features to enable all features gated with @unstable

ydnar:
This PR adds the `--all-features` option to `wasm-tools component`, primarily so `wasm-tools component wit -j` can emit a JSON representation of the original WIT with all `@unstable` gated features intact.

It additionally adds support for `--features .` as an alias for `--all-features`.

Originally proposed here: https://bytecodealliance.zulipchat.com/#narrow/stream/327223-wit-bindgen/topic/wit-parser.20crate.3A.20JSON.20format.20changes/near/443516048


src/bin/wasm-tools/component.rs
```diff
@@ -202,6 +209,7 @@ impl WitResolve {
 
     fn load(&self) -> Result<(Resolve, Vec<PackageId>)> {
         let mut resolve = Self::resolve_with_features(&self.features);
+        resolve.all_features = self.all_features;
```
alexcrichton, line 212: Could setting this field be pushed into `resolve_with_features`? That way other callers don't have to worry about processing the argument.

alexcrichton (commented):
Thanks! Mind also adding a test along the lines of [this one](https://github.com/bytecodealliance/wasm-tools/blob/main/tests/cli/wit-with-features.wit)? (no need to be so exhaustive, but it'd be good to have at least one test for the flags)

src/bin/wasm-tools/component.rs
ydnar: Does WIT define the value constraints for `feature =`? Could `--all-features` alias `--features *`?

ydnar: Where can I find the formal definition of `id`?

alexcrichton: Sorry not sure I quite understand, can you elaborate on what you meant by value constraints? Also what `id` are you referring to?

As for `--features *` being the same as `--all-features`, it should be relatively easy to implement but it might be difficult for shells typically because `*` by default expands to all the files in the current directory.

ydnar: I just pushed a change that makes `--all-features` an alias for `--features *`, since `id` must be a lower kebab-case identifier.

ydnar: `*` isn’t great for shell expansion—would `--features .` be acceptable?

alexcrichton (commented):
Thanks! Can you elaborate on why to support `--features .` though? I've not personally seen precedent for something like that and it feels a bit odd to me that `.` means "everything". I'd personally be in favor of just sticking with `--all-features` myself.

ydnar:
> Thanks! Can you elaborate on why to support `--features .` though? I've not personally seen precedent for something like that and it feels a bit odd to me that `.` means "everything". I'd personally be in favor of just sticking with `--all-features` myself.

`.` for everything mimics regexp syntax, and Go uses `.` to mean "all" in a number of contexts in its CLI.

Happy to revert this if that’s what gets it done.

alexcrichton:
That's true about regexes, although I'd expect `.*` to match all features as opposed to just `.` for regexes. Would you be ok dropping `--features .` and just keeping `--all-features`?

alexcrichton (approved):

PR #1606, opened 2024-06-10
Add `lann/wasm-wave` to crates 

pchickey:
This is an import of https://github.com/lann/wasm-wave to a new home in this repository!

I imported the code at release 0.6.0 to `crates/wasm-wave`

I made minor fixes to: 
* fit in workspace
* delete dependency on wasmtime (those trait impls will be upstreamed into the wasmtime crate under a flag once this lands)
* and fix some benign warnings https://github.com/lann/wasm-wave/pull/36
* replace the use of snapbox/tryfn in the tests with a bespoke implementation, because snapbox and some of its deps dont support wasm32-wasi and the functionality the tests were really using from that crate was pretty minor.


pchickey:
@lann would appreciate your review and approval on this as well!

crates/wasm-wave/tests/ui.rs
```diff
@@ -0,0 +1,117 @@
...
+};
+use wit_parser::Resolve;
+
+fn test(path: &Path) -> Result<String> {
+    let filename = path.file_name().unwrap().to_string_lossy();
+    let inputs = std::fs::read_to_string(path)?;
+    let mut output = String::new();
+    let out = &mut output;
+    let inputs = inputs.trim_end().trim_end_matches(';');
+    for mut input in inputs.split(";\n") {
```
lann, line 18: Looks like this needs to account for Windows `\r\n` as well.

lann (approved):

crates/wasm-wave/tests/ui.rs
```diff
@@ -0,0 +1,117 @@
...
+                .flat_map(|(_, i)| &i.functions)
+                .map(|(name, func)| (name.clone(), resolve_wit_func_type(&resolve, func).unwrap()))
+                .collect::<HashMap<_, _>>()
+        })
+        .get(func_name)
+}
+
+#[test]
+fn ui() -> Result<()> {
+    for entry in fs::read_dir("tests/ui")? {
```
alexcrichton, line 90: A good TODO/issue to file perhaps would be to migrate this to `libtest-mimic` which other crates in this workspace are using as well

crates/wasm-wave/src/wasm/func.rs
```diff
@@ -0,0 +1,30 @@
...
+
+    /// Returns an iterator of the func's result types.
+    fn results(&self) -> Box<dyn Iterator<Item = Self::Type> + '_>;
+
+    /// Returns an iterator of the func's result names. Must be the same
+    /// length as the iterator returned by `results` or empty if there are no
+    /// named results or if this WasmFunc impl does not support result names.
+    fn result_names(&self) -> Box<dyn Iterator<Item = Cow<str>> + '_> {
+        Box::new(std::iter::empty())
+    }
```
alexcrichton, line 29: Given recent versions of rustc we may want to take a look at changing these to `-> impl Iterator`, but that's a concern for later

crates/wasm-wave/src/lib.rs
```diff
@@ -0,0 +1,69 @@
+//! Web Assembly Value Encoding
+//!
+//! WAVE is a human-oriented text encoding of WebAssembly Component Model values.
+//!
+//! For more information, see the [README](https://github.com/lann/wasm-wave#readme).
```
alexcrichton, line 5: another url to update

crates/wasm-wave/src/lex.rs
```diff
@@ -0,0 +1,137 @@
...
+/// A WAVE `logos::Lexer`
+pub type Lexer<'source> = logos::Lexer<'source, Token>;
+
+/// A WAVE token
+#[derive(Clone, Copy, Debug, PartialEq, Eq, logos::Logos)]
+#[logos(error = Option<Span>)]
+#[logos(skip r"[ \t\n\r]+")]
+#[logos(skip r"//[^\n]*")]
+#[logos(subpattern label_word = r"[a-z][a-z0-9]*|[A-Z][A-Z0-9]*")]
+#[logos(subpattern char_escape = r#"\\['"tnr\\]|\\u\{[0-9a-fA-F]{1,6}\}"#)]
```
alexcrichton, line 16: @lann I'm not familiar with logos but how difficult do you think it would be to hand-writet he lexer instead of using logos? I suspec that's a relatively big dep for Wamtime itself and while it's not the worst thing in the world to audit it might also be reasonable to hand-roll a lexer for this (there's also some precedent give the... two? ... other hand-rolled lexers in this repo)

If it's difficult to rewrite or logos gives some really nice features then I think it's fine to keep logos, though.

alexcrichton (approved):

crates/wasm-wave/src/lex.rs
lann: Wave started with a hand-rolled lexer. I only switched because of a plan to embed in WAC's grammer which uses Logos.

cc @fibonacci1729

alexcrichton: Ah ok makes sense! I that case I think it's reasonable to keep this. I should investigate and perahps we should change all the other lexers in this repository to also use logos...

fibonacci1729: Logos is quite nice; especially when wanting to compose lexers across different crates.

crates/wasm-wave/tests/ui.rs
pchickey: Done, but in a nasty way [b19582e](https://github.com/bytecodealliance/wasm-tools/pull/1606/commits/b19582e9b27fda697315a40d2a7ddc1ba1134228)

crates/wasm-wave/src/wasm/func.rs
pchickey: Just for my edification, how do recent rustcs change this?

crates/wasm-wave/tests/ui.rs
pchickey: Windows newlines ended up causing me too many headaches with these files, so I used `.gitattributes` so that they are not changed on windows: https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings

PR #1609, opened 2024-06-11
Enable binary encoding and printing of multi-package wit files

macovedj:
Wit files that leverage [multiple packages](https://github.com/bytecodealliance/wasm-tools/pull/1577) should be able to be encoded into binary and printed back out with the changes in this PR


alexcrichton (commented):
Thanks for this! On a skim this looks reasonable enough, but there's one part that will require careful review I haven't gotten to yet. Specifically this is affecting the binary format for wasm-encoded WIT packages and I need to review to see if this has an impact on how this interacts with older versions of `wasm-tools`. For example older versions need to be able to decode binaries produced by this PR and vice-versa, binaries produced here additionally need to be able to be decoded by older versions (assuming the new feature of more-than-one-package isn't used).

To perhaps help with this can you add some tests? There's no tests currently for this functionality and I think it would be good to have some in `crates/wit-component/tests/interfaces/*.wit`. Additionally would it be possible to avoid changing the `v1` encoding? That in theory isn't being emitted by deafult any more and may even be able to be removed, it's been a bit since I reviewed the state of that.

alexcrichton:
Ok reading over this some more I don't think that this is going to work as-is. It looks like this is basically adding a few loops in a few places, but I think that the changes required are going to be more significant than that. For example each interface/world in a package, when encoded today, is exported as the kebab-cased-name of the interface/world. Each kebab-cased export is a standalone component type which has full context for just that interface/world.

With multiple packages something will need to change about this format. Otherwise it's not possible to encode two packages which contains `interface foo` for example since that would generate two kebab-cased exports named `"foo"`. I haven't thought much about what the format should be, so I don't have an answer off the top of my head.

alexcrichton:
cc @lukewagner for possible brainstorming on the format here. The question being that today we have:

```
$ wasm-tools component wit -t <<< "package a:b; interface foo {}"
(component
  (type (;0;)
    (component
      (type (;0;)
        (instance)
      )
      (export (;0;) "a:b/foo" (instance (type 0)))
    )
  )
  (export (;1;) "foo" (type 0))
  (@custom "package-docs" "\00{}")
  (@producers
    (processed-by "wit-component" "0.210.0")
  )
)
```

What's the best way to represent this WIT package:

```wit
package a:b1 { interface foo {} }
package a:b2 { interface foo {} }
```

I'm tempted to say:

```wasm
(component
  (component $a:b1 (; output of `wasm-tools component wit -t for just `package a:b1` ;) )
  (component $a:b2 (; output of `wasm-tools component wit -t for just `package a:b2` ;) )

  (export "a:b1" (component $a:b1))
  (export "a:b2" (component $a:b2))
)
```

This isn't currently valid though as `"a:b1"` isn't a valid export string, right now it needs to be fully qualified. This could be seen as motivation for relaxing those restrictions, but that's balooning the change a good deal so I'm curious if you have thoughts on this.

lukewagner:
Ooh yeah, really good question.  Agreed that the natural first step seems to be to put each nested WIT package into its own nested component.  Then to your question: maybe this is weird but: what if we just didn't export those nested WIT packages at all?  That would imply that when WIT tooling was slurping up a binary WIT package, it would need to recurse through all nested components and populate its internal namespace with all `interfacename`-d type exports of exported component types (rather than only doing so for the root component).  Would that work?

What's neat about this is that it preserves the distinction between:
```wit
package ns:a;
package ns:b { interface i {} }
interface i {}
```
and
```wit
package ns:a { interface i {} }
package ns:b { interface i {} }
```
The former you could say is defining the package `ns:a` (which exports `i`) and bundling `ns:b` as an external dependency whereas the latter is defining an empty unnamed package that bundles external packages `ns:a` and `ns:b`.  I think that distinction makes sense?

alexcrichton:
I think that could work yeah and that seems reasonable to me. For the start I think it'd just walk the root component looking for subcomponents rather than spidering the whole thing, but either probably wouldn't be too hard to implement. I'm thinking though it'd be best to only accept a conservative shape of components to keep our options open in the future to different organizations if we want.

Otherwise though this WIT:

```wit
package ns:a;
package ns:b { interface i {} }
interface i {}
```

currently isn't valid where WIT text files are either `package foo:bar;` mode or exclusively `package foo:bar { ... }` mode. That being said being able to express the distinction there eventually I think is good regardless if we want to support it in the future.

lukewagner:
Ah right, good point; unless and until we add `package { ... }` blocks inside `package { ... }` blocks, you shouldn't need to spider.

macovedj:
So if I'm following correctly then for the scope of this PR, since mixed package declaration styles isn't yet supported, then in Alex's example where we have:

```
package a:b1 { interface foo {} }
package a:b2 { interface foo {} }
```

this would end up encoded with no exports:

```
(component
  (component $a:b1 (; output of `wasm-tools component wit -t for just `package a:b1` ;) )
  (component $a:b2 (; output of `wasm-tools component wit -t for just `package a:b2` ;) )
)
```

Is that right?

Then whenever we end up with mixed declaration styles being supported (maybe I"ll do that as a follow up or if it doesn't seem so bad I may try to include it in this PR) we could see both this and the current day encodings appear side by side in a single encoded wit file.

alexcrichton:
Indeed! So the encoder will basically say if there's one package do what it does today but with multiple packages being encoded they're encoded in this new format. Similarly the decoding process would need to recognize the new multi-package format and decode that in addition to supporting the previous single-package format.

There's still sort of a question of defaults, but I think that's reasonable to tackle at a later date. Specifically I'm wondering for example what the output of `wasm-tools component wit` would be for `wasi:cli`. Whether that by default transitively includes all dependency packages I'm not sure.

macovedj:
So I have implemented these changes locally for encoding multiple wit packages.  When decoding though, I found that some modifications will be have to made [here](https://github.com/bytecodealliance/wasm-tools/blob/main/crates/wit-parser/src/decoding.rs#L148) as the current tooling will try to parse the binary as a component instead of a wit package if no exports are present.

I think a convention needs to be decided on how to discern whether these are nested components in a component implementation vs being used to describe a wit package?  I was thinking that maybe the component names can can provide a mechanism?  If not then we could go back to expecting exports for these cases, and the convention could be established in how they're exported?

macovedj:
After some more thought...  I'll actually probably investigate refactoring [ComponentInfo](https://github.com/bytecodealliance/wasm-tools/blob/main/crates/wit-parser/src/decoding.rs#L47) into `ComponentsInfo`, with multiple packages in it.  Then we could perform the existing check on each package, and it could also provide a clearer route to mixing package declaration styles.

alexcrichton:
That's a good point! I agree though that we can probably update how metadata is tracked and extend the "is empty" check to exclude the case of "no exports but has a bunch of components which all look like WIT packages"

macovedj:
Cool. I think I've essentially accomplished the thrust of what's involved in order to make to that now... As I'm tidying it up I'm evaluating whether or not to include support for mixing package declaration styles or have that be a follow up.

macovedj:
Hmm seems I botched something up switching branches... should be able to look into what went wrong shortly...

macovedj:
I think this should be ready for review now... just investigating why it's struggling to get gcc with yum in CI/CD.

crates/wit-component/src/encoding/wit/v2.rs
```diff
@@ -57,33 +51,85 @@ impl Encoder<'_> {
...
-
-            let world = &self.resolve.worlds[world];
-            let mut wrapper = ComponentType::new();
-            wrapper.ty().component(&component_ty);
-            let pkg = &self.resolve.packages[world.package.unwrap()];
-            wrapper.export(&pkg.name.interface_id(name), ComponentTypeRef::Component(0));
+        let mut names = NameMap::new();
+        for pkg in self.packages {
+            let package = &self.resolve.packages[*pkg];
+            if let PackageKind::Explicit = package.kind {
```
alexcrichton, line 57: I'm a bit confused as to what this new `PackageKind` field is and how it's processed here. I think ideally this would be removed entirely since it seems to be an artifact of the original text encoding and shouldn't affect this binary encoding here.

Instead what I might recommend is:

* If there's only a single package, encode in the old "v2 style".
* If there's multiple packages, encoded in this "explicit" style where a sub-component is created.

I'd also recommend extracing the shared code between these two branches of the `if` since right now it looks like the body of the loop was copy/pasted.

Additionally I think long-term we'll want to move towards the "new v2 style" where all packages, even if it's just one, are encoded with a sub-component. In that sense I think it would make sense to have a few more knobs here like other transitions we've made with the binary format:

* An env var should be added which if specified and is 0 or 1 indicates which encoding branch to use.
* Faililng that a `const` with a doc block explaining what it's controlling.

Eventually in the future we'll invert the value of the `const` to default to the new encoding. Further in the future we'll remove the env var. Further still we'll remove decoding support for the original encoding.

crates/wit-component/tests/interfaces.rs
```diff
@@ -54,43 +54,37 @@ fn run_test(path: &Path, is_dir: bool) -> Result<()> {
...
-        // Finally convert the decoded package to wasm again and make sure it
-        // matches the prior wasm.
-        let wasm2 = wit_component::encode(Some(true), resolve, decoded_package)?;
-        if wasm != wasm2 {
-            let wat2 = wasmprinter::print_bytes(&wasm)?;
-            assert_eq!(wat, wat2, "document did not roundtrip correctly");
-        }
+    // Finally convert the decoded package to wasm again and make sure it
+    // matches the prior wasm.
+    let wasm2 = wit_component::encode(Some(true), resolve, &[decoded_package])?;
```
alexcrichton, line 82: I'm a bit surprised this test is passing because I would expect that the `decoded_package` argument should be the result of `decoded.packages()` now that multiple packages are supported. Could you investigate if a new test is needed to reflect that?

crates/wit-parser/src/decoding.rs
```diff
@@ -81,45 +139,83 @@ impl ComponentInfo {
...
-                Payload::CustomSection(s) if s.name() == PackageMetadata::SECTION_NAME => {
-                    if _package_metadata.is_some() {
-                        bail!("multiple {:?} sections", PackageMetadata::SECTION_NAME);
+                Payload::CustomSection(s) => {
+                    if s.name() == PackageMetadata::SECTION_NAME {
+                        _package_metadata = Some(PackageMetadata::decode(s.data())?);
+                        cur_package.package_metadata = _package_metadata.clone();
+                    } else if s.name() == "component-name" {
+                        if let KnownCustom::ComponentName(reader) = s.as_known() {
+                            let _packages = register_names(reader, &mut cur_package)?;
```
alexcrichton, line 204: I'm a bit hesitant to rely on a custom section for something so significant to the structure of the final WIT, would it be possible to avoid this inference of infer via other means?

crates/wit-parser/src/decoding.rs
```diff
@@ -204,25 +342,23 @@ impl ComponentInfo {
...
         if let Some(package_metadata) = &self.package_metadata {
-            package_metadata.inject(&mut resolve, package)?;
+            package_metadata.inject(&mut resolve, packages[0])?;
         }
-        Ok((resolve, package))
+        Ok((resolve, packages[0]))
     }
 
-    fn decode_wit_v2_package(&self) -> Result<(Resolve, PackageId)> {
+    fn decode_wit_v2_packages(&self) -> Result<(Resolve, Vec<PackageId>)> {
```
alexcrichton, line 352: I'm a bit lost in the changes to this method, can you explain why so many changes were needed? My (probably) naive understanding is that this could work by decoding each sub-component if present in the exact same way they're all decoded today, and then at the end the `Resolve` values are all merged together. This looks like it's lazily inserting packages and/or using various other bits of logic I don't fully understand. In other words I'd expect that there wouldn't be much change in general and instead decoding would decode each sub-component in a loop and then merge everything afterwards.

crates/wit-parser/src/decoding.rs
macovedj: So I actually had similar thoughts when implementing... But I think that the way names of components are handled today is exclusively via custom sections?  Will investigate some more, but if we want something similar to the wat outlined above where the name of the component is tied to the package name, I'm not sure of another way to get the information that informs a package name out of the binary format.

crates/wit-component/tests/interfaces.rs
macovedj: Looks like the assertion was wrong!  accidentally setting `wat2` to be `wasm` instead of `wasm2` when checking the diff.  You're right, when fixing the assertion, the test fails as is, and changing `decoded_package` to meet your expectations fixes it.

crates/wit-parser/src/decoding.rs
alexcrichton: Documentation and `@since`/`@unstable` are definitely all handled with custom sections, but to me those are semi-ok to strip at any time where the name of a package is significant enough that you'd never want it stripped. I'm mostly wondering if this can be inferred from the exported items or something like that rather than having to be listed in the name section. An alternative could be to export the components themselves from the outer component where the export name has the full package name (of sorts at least)

crates/wit-parser/src/decoding.rs
macovedj: So when parsing the binary today, the only metadata captured in `ComponentInfo` is for the "implicit" package at the root, and the decoding is based on the metadata there.  I don't think we have the ability to decode each subcomponent as they're encountered, just based on how `ComponentInfo::from_reader` parses the binary today, I think it has go through the entire binary, and I don't have any great ideas about how to change that.  I didn't realize initially that I had the `merge` method on `Resolve` available to me though.  Since you're comment, I updated to decode each `ExplicitComponentInfo` separately, and then merge each of the resolves together.

I think most of the changes at this point are just clearing out the `PackageFields` between decoding each `ExplicitPackageInfo` , or else interfaces/worlds carry over from when package to the next.  I can look into seeing if there is a way to make this a bit tidier though/potentially create a new `PackageFields` per package rather than reusing and clearing the same one as we iterate through packages.

crates/wit-component/tests/interfaces.rs
macovedj: Turns out when I fixed it fixed the assertion, it came out that the docs are being stored in the wat in a different order depending on if a wit is being parsed vs a binary, in such a way that it doesn't impact how the docs get printed I believe.  Can fix the tests by performing a sort on the metadata fields, but still investigating how to just get encode things in the same order regardless.

crates/wit-component/src/encoding/wit/v2.rs
macovedj: After some thought, and keeping in mind that we'll eventually want to support a wit file that has both an "implicit"/"unscoped" package declaration as well as an "explicit"/"scoped" package declaration, is the number of packages actually what we want to use as the criteria?  If we do things this way instead of capturing the `PackageKind`... then I'm not sure how we can tell the difference between `package a:b {}` and `package a:b` if in the scoped case, it's the only scoped package.  I think we'd end up transforming one package declaration style into the other when performing a roundtrip if this choice is made based on the number of packages that exist, which may be fine, but just wanted to make sure that we're considering this.

crates/wit-parser/src/decoding.rs
macovedj: I guess there's maybe an edge case here where you could have an empty scoped package?  I think we could infer things from exports so long as there are exports, and as far as I know the only case where there wouldn't be is if the package is empty...  Would you be comfortable erroring on empty scoped packages then?  Or yeah, we could also do the alternative and export the components themselves...

crates/wit-component/src/encoding/wit/mod.rs
```diff
@@ -43,12 +43,12 @@ pub fn encode(use_v2: Option<bool>, resolve: &Resolve, package: PackageId) -> Re
...
     resolve: &Resolve,
-    package: PackageId,
+    packages: &[PackageId],
 ) -> Result<ComponentBuilder> {
     if use_v2.unwrap_or_else(use_v2_encoding) {
-        v2::encode_component(resolve, package)
+        v2::encode_component(resolve, packages)
     } else {
-        v1::encode_component(resolve, package)
+        v1::encode_component(resolve, packages[0])
```
alexcrichton, line 51: This should also update to generate an error if `packages.len() > 0` since otherwise it's losing information

crates/wit-component/src/encoding/wit/v2.rs
```diff
@@ -57,33 +51,101 @@ impl Encoder<'_> {
...
+        pkg: &PackageId,
+        mut sub_encoder: Option<Encoder>,
+    ) -> Result<()> {
+        let package = &self.resolve.packages[*pkg];
+        if let Some(ref mut sub) = sub_encoder {
+            let package_metadata = PackageMetadata::extract(self.resolve, *pkg);
+            sub.component.custom_section(&CustomSection {
+                name: PackageMetadata::SECTION_NAME.into(),
+                data: package_metadata.encode()?.into(),
+            });
```
alexcrichton, line 94: To avoid duplicating this with the above could this be extracted to a helper method?

crates/wit-component/src/encoding/wit/v2.rs
```diff
@@ -57,33 +51,101 @@ impl Encoder<'_> {
...
-            let world = &self.resolve.worlds[world];
-            let mut wrapper = ComponentType::new();
-            wrapper.ty().component(&component_ty);
-            let pkg = &self.resolve.packages[world.package.unwrap()];
-            wrapper.export(&pkg.name.interface_id(name), ComponentTypeRef::Component(0));
+    fn encode_package(
+        &mut self,
+        names: &mut NameMap,
+        pkg: &PackageId,
+        mut sub_encoder: Option<Encoder>,
```
alexcrichton, line 86: Instead of passing around `sub_encoder` and being careful to use-it-or-not, could the `self.component` field be `mem::replace`'d to sort of swap out where the encoding is happening within?

alexcrichton: (it would also help the duplication below for each `if let` with the `sub_encoder`)

crates/wit-component/src/encoding/wit/v2.rs
alexcrichton: In the limit of time I would prefer to not spec/support multiple encoding formats for a WIT package in wasm. Right now we support v1/v2 and this PR is adding a third. I'd eventually like to sunset all the prior versions, however.

Given that my goal would be to use this new "v3" unconditionally. I'm proposing to distinguish based off 1-or-more packages purely for backwards compatibility. That way preexisting workflows don't break with older versions since the new format wouldn't be used by default. Once the new format is in enough supporting tooling it would then be enabled by default.

Personally I don't think it's too important to preserve `package a:b { ... }` vs `package a:b; ...` since they're semantically the same thing.

crates/wit-component/tests/interfaces.rs
alexcrichton: Is that a regression from this PR? Or a preexisting issue?

crates/wit-parser/src/decoding.rs
alexcrichton: @lukewagner do you have thoughts on this point specifically? The question being how is the package name recovered when multiple packages are encoded in a component and one package is possibly empty.

crates/wit-parser/src/decoding.rs
alexcrichton: Ah ok that makes sense. I agree the best way to go about this is to modify how `ComponentInfo` is collected and processed.

crates/wit-parser/src/decoding.rs
```diff
@@ -20,6 +26,22 @@ struct ComponentInfo {
...
     externs: Vec<(String, Extern)>,
+    /// Packages
+    explicit: Vec<ExplicitPackageInfo>,
+    /// Decoded package metadata
+    package_metadata: Option<PackageMetadata>,
+}
+
+/// Represents information about a decoded WebAssembly component.
+#[derive(Default)]
+struct ExplicitPackageInfo {
```
alexcrichton, line 37: I'd prefer to avoid the duplication between `ComponentInfo` and `ExplicitPackageInfo` if possible. Could there perhaps be a single `PackageInfo` structure which has `externs` and a list of nested info? For example `externs` would track imports and exports and `explicit` (perhaps modulo a renaming) would store nested components. The parse loop would then recurse for nested components rather than handling everything in a single loop with an if/else on the depth.

crates/wit-parser/src/decoding.rs
```diff
@@ -20,6 +26,22 @@ struct ComponentInfo {
...
+    /// Package name
+    name: Option<PackageName>,
+    /// Wasmparser-defined type information learned after a component is fully
+    /// validated.
+    types: Option<types::Types>,
```
alexcrichton, line 42: It should be ok to only store the types at the "root" since I don't think a snapshot is needed along the way

crates/wit-parser/src/decoding.rs
```diff
@@ -273,22 +410,128 @@ impl ComponentInfo {
...
+            resolve
         } else {
-            bail!("no exported component type found");
+            Resolve::new()
         };
 
-        let (mut resolve, package) = decoder.finish(pkg);
-        if let Some(package_metadata) = &self.package_metadata {
-            package_metadata.inject(&mut resolve, package)?;
+        for explicit in &self.explicit {
```
alexcrichton, line 434: I think this is a copy/paste of the logic above? Can you please use helpers for significant chunks of code like this rather than copy/pasting and adapting?

crates/wit-component/tests/interfaces.rs
macovedj: I think this is a preexisting issue.

```rs
if wasm != wasm2 {
        let wat2 = wasmprinter::print_bytes(&wasm)?;
        assert_eq!(wat, wat2, "document did not roundtrip correctly");
}
```
If I understand correctly, `assert_eq` is receiving two different `wat` samples created from `wasm`, neither represent `wasm2`...  Was spending some time trying to dig into this, and it looks like modifying main also has this issue of doc metadata being reordered if I'm reading the code correctly...

crates/wit-parser/src/decoding.rs
macovedj: I see, at this point though I think I've been able to combine `ComponentInfo` with `ExplicitPackageInfo`, so I think it'll just have a value of `None` when for nested packages.

crates/wit-parser/src/decoding.rs
macovedj: Or yeah... I think I like your idea of exporting the component, seems like the easiest way to deduce the name of the component without relying on the name section...  How the PR handles it at the moment, even if we can tell it's an explicit/scoped package from the way the exports are structured, I think we still need the name section to figured out the actual name of the package that should be printed from a binary (unless there's another mechanism I'm missing).  Also curious to hear @lukewagner's thoughts about if we exported the components.

crates/wit-component/src/encoding/wit/v2.rs
macovedj: Yeah ok that makes sense to me

macovedj: One last thought that struck me...

Is there a use case where the scoped packages are used similarly to how the deps folder is used today?
In this case where we're mixing package declaration styles, the implicit package may be the one being used by guest lang tooling for knowing which world to target, and the others are there as deps.  If we needed to encode in this case, then I think this logic would treat the implicit package like the deps, rather than it being treated as "higher" than the others?  I guess perhaps we could rely on guest lang tooling to keep track of which package is the one that matters for targeting?

alexcrichton:
I'm going to close this as the conclusion of this PR manifested in https://github.com/bytecodealliance/wasm-tools/pull/1700 which is to instead of supporting the feature being added here instead explicitly don't support it. The encoding/printing phases will continue to work on one "primary package".

PR #1610, opened 2024-06-11
add links to playground

bakkot:
These could definitely be improved but are sufficient to be useful, I think, and I'm probably not going to have time to make them fancier in the near future. That said if there's any more improvements you'd like to see, or if playground for any other subcommands would be valuable, let me know. `component wit` seems like it might be worth having a playground for?

I wasn't sure what the best place for these links would be, but `Installation` seems like a reasonable option.


bakkot:
Sounds good to me, done.

alexcrichton (approved):

PR #1611, opened 2024-06-11
Fold `resolve_world_from_name` helper into `select_world`

alexcrichton:
This is a follow-up to #1577 to refactor a bit to have bindings generators continue to use `Resolve::select_world` for figuring out what to generate bindings for.

### DIFF
--- a/crates/wit-parser/src/resolve.rs
+++ b/crates/wit-parser/src/resolve.rs
@@ -293,7 +293,8 @@ impl Resolve {
     /// The second value returned here is the list of paths that were parsed
     /// when generating the return value. This can be useful for build systems
     /// that want to rebuild bindings whenever one of the files change.
-    pub fn push_dir(&mut self, path: &Path) -> Result<(Vec<PackageId>, Vec<PathBuf>)> {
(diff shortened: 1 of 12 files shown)


crates/wit-parser/src/resolve.rs
```diff
@@ -2551,24 +2705,24 @@ mod tests {
...
+            .select_world(&dummy, Some("foo:baz/foo@0.1.0"))
             .is_ok());
         assert!(resolve
-            .select_world(dummy, Some("foo:baz/foo@0.2.0"))
+            .select_world(&dummy, Some("foo:baz/foo@0.2.0"))
             .is_ok());
     }
 
-    fn parse_into(resolve: &mut Resolve, wit: &str) -> PackageId {
+    fn parse_into(resolve: &mut Resolve, wit: &str) -> Vec<PackageId> {
```
pchickey, line 2724: How do you feel about making this a pub method (take `&mut Self`)? We do the same UnresolvedPackageGroup::parse Resolve::append pattern in many places in test code.

pchickey (approved):
Thanks, this solves the problems I was having with the upgrade!

PR #1615, opened 2024-06-12
Use quoted identifiers in `wasmprinter` by default

alexcrichton:
This commit updates `wasmprinter` to use quoted identifiers of the form `$"foo"` when necessary instead of synthesizing identifiers such as `$#func4<foo>`. This helps produce more readable modules by default when names are synthesized since if a name is unique but otherwise has non-identifier characters then the quoted string form can be used.

While here I've additionally changed the way that non-printable characters in strings are printed to using `\u{XXX}` syntax instead of `\NN` syntax. This makes it a bit more obvious in unicode contexts that a single character is present and not multiple.

For a module I was looking at the "before" was:

```wasm
;; ...
    (func $#func73<undefined_weak:thread-local_initialization_routine_for_errno> (@name "undefined_weak:thread-local initialization routine for errno") (;73;) (type 13)
;; ...
    )
    (func $_initialize (;74;) (type 13)
;; ...
    )
    (func $#func75<initialize_char_const*_> (@name "initialize(char const*)") (;75;) (type 0) (param i32) (result i32)
;; ...
```

and the "after" is:

```wasm
;; ...
    (func $"undefined_weak:thread-local initialization routine for errno" (;73;) (type 13)
;; ...
    )
    (func $_initialize (;74;) (type 13)
;; ...
    )
    (func $"initialize(char const*)" (;75;) (type 0) (param i32) (result i32)
;; ...,
```

Notably `$"initialize(char const*)"` is much nicer to raed


fitzgen (approved):
:rocket:

PR #1616, opened 2024-06-13
`wasm-smith`: fix generation of memories with custom page sizes

fitzgen:


crates/wasm-smith/tests/core.rs
```diff
@@ -142,6 +142,25 @@ fn smoke_test_wasm_gc() {
     }
 }
 
+#[test]
+fn smoke_test_wasm_custom_page_sizes() {
```
alexcrichton, line 146: Makes sense to have this in isolation, but mind also [changing this](https://github.com/bytecodealliance/wasm-tools/blob/d6436035fefa4861e64b27d0c416cfe9f309369d/fuzz/src/lib.rs#L22-L30) to enable fuzzers to explore custom-page-sizes as well? (you'll probably want to change [this](https://github.com/bytecodealliance/wasm-tools/blob/d6436035fefa4861e64b27d0c416cfe9f309369d/fuzz/src/lib.rs#L80) as well)

alexcrichton (approved):

PR #1619, opened 2024-06-18
fuzz(mutate): enable the custom-page-sizes proposal as necessary

fitzgen:
If `wasm-smith` enabled it, then we need to enable it for validation of the mutated binaries.


fuzz/src/mutate.rs
```diff
@@ -68,6 +68,10 @@ pub fn run(u: &mut Unstructured<'_>) -> Result<()> {
     features.set(WasmFeatures::MULTI_MEMORY, config.max_memories > 1);
     features.set(WasmFeatures::MEMORY64, config.memory64_enabled);
     features.set(WasmFeatures::THREADS, config.threads_enabled);
+    features.set(
+        WasmFeatures::CUSTOM_PAGE_SIZES,
+        config.custom_page_sizes_enabled,
+    );
```
alexcrichton, line 74: Given the current state of things today should this (and the bits above this) perhaps be replaced by `let features = WasmFeatures::all()`?

PR #1620, opened 2024-06-18
fuzz(no_traps): enable the custom-page-sizes proposal as necessary

fitzgen:
If the `wasm-smith` config enabled it, then we should enable it for validation.

### DIFF
--- a/fuzz/src/no_traps.rs
+++ b/fuzz/src/no_traps.rs
@@ -113,24 +113,7 @@ pub fn run(u: &mut Unstructured<'_>) -> Result<()> {
 
 fn validate_module(config: wasm_smith::Config, wasm_bytes: &Vec<u8>) {
     // Validate the module or component and assert that it passes validation.
-    let mut validator = wasmparser::Validator::new_with_features({
-        let mut features = WasmFeatures::default();
-        features.remove(WasmFeatures::COMPONENT_MODEL);
-        features.set(WasmFeatures::MULTI_VALUE, config.multi_value_enabled);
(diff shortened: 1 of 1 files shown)


fuzz/src/no_traps.rs
```diff
@@ -125,6 +125,10 @@ fn validate_module(config: wasm_smith::Config, wasm_bytes: &Vec<u8>) {
         features.set(WasmFeatures::MEMORY64, config.memory64_enabled);
         features.set(WasmFeatures::THREADS, config.threads_enabled);
         features.set(WasmFeatures::EXCEPTIONS, config.exceptions_enabled);
+        features.set(
+            WasmFeatures::CUSTOM_PAGE_SIZES,
+            config.custom_page_sizes_enabled,
+        );
```
alexcrichton, line 131: Similar to the other PR, it might be best nowadays to use `::all()` here instead?

alexcrichton (approved):

alexcrichton:
The oss-fuzz bug at https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=69749 is still open after this, and I think that the error there is we need to skip wasmtime execution when custom-page-sizes is enabled since the version of wasmtime used here doesn't implement the proposal yet

fitzgen:
> The oss-fuzz bug at https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=69749 is still open after this, and I think that the error there is we need to skip wasmtime execution when custom-page-sizes is enabled since the version of wasmtime used here doesn't implement the proposal yet

https://github.com/bytecodealliance/wasm-tools/pull/1644

PR #1624, opened 2024-06-19
Nested interfaces in wit

macovedj:
This PR enables the ability to describe nested interfaces in wit via the `nest` keyword.

```
package foo:bar;

interface baz {
  nest other:pkg/qux
}
```

This would indicate that that the instance associated with baz would export the instance associated with interface `qux` from package `other:pkg`. 

```
(component 
  (type instance
    (type instance 
    ... definition of "other:pkg/qux" 
    )
   (export "other:pkg/qux" type 0)
  )
  (export "foo:bar/baz" type 0)
)
```

 As I understand it, most use cases most immediately benefit from nesting interfaces from foreign packages, so I started with that, though we could certainly extend this PR or have follow ups to support any combo of locally defined/inlined/anonymous interfaces being nested.

[This issue](https://github.com/WebAssembly/component-model/issues/301) points out that wit currently is not able to express nested instances that can be expressed in wat/binary, so wit-bindgen can't be used by language toolchains to generate that binary.  This syntax can be leveraged so that the unlocked dependency syntax referenced in the linked issue can specify the exports it expects in the imports it's describing.


crates/wit-parser/src/ast.rs
```diff
@@ -545,6 +558,48 @@ enum InterfaceItem<'a> {
...
+    ) -> Result<Self> {
+        tokens.eat(Token::Nest)?;
+        let id = parse_id(tokens)?;
+        tokens.expect(Token::Colon)?;
+        // `foo:bar/baz@1.0`
+        let namespace = id;
+        let pkg_name = parse_id(tokens)?;
+        tokens.expect(Token::Slash)?;
+        let name = parse_id(tokens)?;
+        let version = parse_opt_version(tokens)?;
```
alexcrichton, line 585: I think that this is a duplication of `UseName`'s internal parsing, but could `UseName` be used here instead? That would enable consistently being able to refer to both foreign and local interfaces with `nest`

crates/wit-parser/src/ast/resolve.rs
```diff
@@ -292,6 +302,30 @@ impl<'a> Resolver<'a> {
...
+
+                        AstItem::Interface(self.alloc_interface(nest.name.span))
+                    });
+                    match id {
+                        AstItem::Interface(id) => foreign_interfaces.insert(id),
+                        AstItem::World(_) => unreachable!(),
+                    };
+                    Ok(())
+                })
+                .unwrap();
```
alexcrichton, line 328: I think ideally if nest used a `UsePath` then thiscould go away, along with the `for_each_nest` method?

crates/wit-parser/src/ast/resolve.rs
```diff
@@ -364,6 +399,20 @@ impl<'a> Resolver<'a> {
...
+                                        n.name.span,
+                                        format!("duplicate item named `{}`", n.name.name),
+                                    ))
+                                }
+                                let prev = decl_list_ns.insert(n.name.name, ());
+                                assert!(prev.is_none());
+                                let prev = order.insert(n.name.name, Vec::new());
+                                assert!(prev.is_none());
+                            }
+                        }
```
alexcrichton, line 415: I'm not sure if this is quite right because this seems like it would be an `interface`-level validation rather than a package-level validation which I think this loop is doing.

crates/wit-parser/src/ast/resolve.rs
```diff
@@ -470,22 +540,72 @@ impl<'a> Resolver<'a> {
         let mut iface_id_order = Vec::new();
         let mut world_id_order = Vec::new();
         for name in order {
-            match names.get(name).unwrap() {
-                ast::AstItem::Interface(_) => {
+            match names.get(name) {
+                Some(ast::AstItem::Interface(i)) => {
                     let id = self.alloc_interface(package_items[name]);
                     self.interfaces[id].name = Some(name.to_string());
+                    for item in &i.items {
```
alexcrichton, line 547: I think this loop and processing `nest` items in general would be best done during the elaboration of an interface to uniformly handle it next to other `interface` constructs

crates/wit-parser/src/lib.rs
```diff
@@ -403,6 +403,21 @@ impl WorldItem {
...
+    #[cfg_attr(feature = "serde", serde(serialize_with = "serialize_id"))]
+    pub id: InterfaceId,
+    pub docs: Docs,
+    #[cfg_attr(
+        feature = "serde",
+        serde(skip_serializing_if = "Stability::is_unknown")
+    )]
+    pub stability: Stability,
+    pub package_name: PackageName,
+    pub iface_name: String,
```
alexcrichton, line 418: I think both of these fields can be elided in favor of looking them up through the `id` here

crates/wit-component/src/printing.rs
```diff
@@ -82,6 +82,14 @@ impl WitPrinter {
                 self.output.push_str("interface ");
                 self.print_name(name);
                 self.output.push_str(" {\n");
+                let nested = &resolve.interfaces[*id].nested;
+                for item in nested {
+                    self.print_stability(&item.1.stability);
+                    self.print_docs(&item.1.docs);
+                    self.output.push_str("nest ");
+                    self.print_name(item.0);
```
alexcrichton, line 90: I don't think that this is quite right because this should print `nest a:b/c;` or similar so I think an interface id needs to be printed here rather than just the string name

alexcrichton (commented):
I've done a bit of a cursory review to start off with but I think that this still has a few missing pieces. For example `src/resolve.rs` in the `wit-parser` care will need to be updated to handle this new  features of interfaces in a number of locations.

At a high level though I think it would be best to propose this feature on the component-model repository first. There's a number of high-level questions about how to design this feature that I think should be thought through such as:

* Is the syntax `nest a:b/c;` what everyone agrees on?
* Should `nest a;` be allowed?
* Should `nest a { /* ... */ }` be allowed?
* How will code generators for guest langauges map this?

Those are at least the questions off the top of my head which I think would be best to settle first before finishing up the implementation here. A PR to the component model repository would also help flesh out the wasm-encoding of this WIT construct.

macovedj:
Sounds good!  I can work on opening something up over there.

macovedj:
I think the issues re: encoding seem like they've mostly been hashed out [here](https://github.com/WebAssembly/component-model/pull/372).  While there seems to be a plan for the non-foreign interface nesting, I think it's ok to start with what this PR originally scoped (only `nest a:b/c`)?  There might still be a bit more conversation re: code generation, but I'll probably go ahead and try and address most of the comments here soon.

crates/wit-parser/src/lib.rs
macovedj: So I realized that the `id` is actually not being updated correctly, I believe when [append](https://github.com/bytecodealliance/wasm-tools/blob/main/crates/wit-parser/src/resolve.rs#L461) is being called.  I've spent some time trying to figure out how to update this, but haven't quite worked it out yet.  I thought it might be worth considering whether it would be ok to get rid of the `id` though and use the fields instead.  I see the appeal of using `id`, but I do think that might complicate what it means to update interfaces when combining packages, unless it's clear to you how to simply perform the update.

crates/wit-component/src/printing.rs
macovedj: This got moved around a bit, but I can update once the above comment is resolved re: how nested ids/interface names are captured.  FWIW though, the string used is the interface id at the moment, but I can see how grabbing the interface id would be a bit less clunky.

crates/wit-parser/src/lib.rs
```diff
@@ -419,6 +432,8 @@ pub struct Interface {
     /// This is `None` for inline interfaces in worlds.
     pub name: Option<String>,
 
+    /// The nested interfaces in this interface.
+    pub nested: IndexMap<String, Nest>,
```
alexcrichton, line 436: One thing that might be helpful here (and we should probably do elsewhere but is fine to leave for a separate PR) is to do a `skip_serializing_if` if the map is empty. That'll help reduce the diff in the test files in this PR and I think it makes sense in general to skip serializing empty things too.

crates/wit-parser/src/ast.rs
```diff
@@ -987,6 +997,16 @@ impl<'a> InterfaceItem<'a> {
...
             Some((_span, Token::Use)) => Use::parse(tokens, attributes).map(InterfaceItem::Use),
+            Some((_span, Token::Nest)) => {
+                tokens.eat(Token::Nest)?;
+                let path = UsePath::parse(tokens)?;
+                tokens.expect_semicolon()?;
+                Ok(InterfaceItem::Nest(Nest {
+                    docs,
+                    attributes,
+                    from: path,
+                }))
```
alexcrichton, line 1008: Mind moving this into a `Nest::parse` method? (basically preserving the style/structure of this method where it delegates to other top-level parsers)

crates/wit-parser/src/ast/resolve.rs
```diff
@@ -478,22 +480,23 @@ impl<'a> Resolver<'a> {
         let mut iface_id_order = Vec::new();
         let mut world_id_order = Vec::new();
         for name in order {
-            match names.get(name).unwrap() {
```
alexcrichton, line 481: Was this change still necessary?

crates/wit-parser/src/lib.rs
alexcrichton: Also, thinking more on this, might this perhaps make sense as a `Vec<Nest>` instead of an `IndexMap`? The use case for looking up a specific nest item I think is a little less than it is with functions/types. Another possible key for this map could be `InterfaceId` instead of `String` since the string should be able to be inferred from the id. (that may not play well with json though perhaps...)

crates/wit-parser/src/ast/resolve.rs
```diff
@@ -839,6 +842,56 @@ impl<'a> Resolver<'a> {
...
+                ast::InterfaceItem::Nest(n) => {
+                    if let ast::Nest {
+                        from:
+                            UsePath::Package {
+                                id: package_id,
+                                name,
+                            },
+                        attributes,
+                        docs,
+                    } = n
```
alexcrichton, line 854: Could this block run for all `ast::Nest` items intead of just `UsePath::Package`? I think that means that other `nest` items would be silently ignored here otherwise?

crates/wit-parser/src/ast/resolve.rs
```diff
@@ -839,6 +842,56 @@ impl<'a> Resolver<'a> {
...
+                                version: package_id.clone().version.map(|v| v.1),
+                            })
+                            .unwrap()
+                            .get(&name.name)
+                            .unwrap();
+                        let nested_id = if let AstItem::Interface(id) = nest {
+                            id
+                        } else {
+                            bail!("Expected interface item")
+                        };
```
alexcrichton, line 870: Could this use `resolve_ast_item_path` and `extract_iface_from_item` to handle fallibility and errors?

Also could you be sure to add a test where `nest` is used with a `world` to ensure it returns an error?

crates/wit-parser/src/ast/resolve.rs
```diff
@@ -870,7 +923,7 @@ impl<'a> Resolver<'a> {
                 TypeItem::Use(u) => {
                     self.resolve_use(owner, u)?;
                 }
-                TypeItem::Def(_) => {}
+                _ => {}
```
alexcrichton, line 926: Mind going ahead and listing out the other variants? I personally find it best to avoid `_` as a wildcard most of the time since it makes evolution over time a bit harder (more difficult to audit possible locations that `match`)

crates/wit-parser/src/metadata.rs
```diff
@@ -521,6 +521,7 @@ struct InterfaceMetadata {
         serde(default, skip_serializing_if = "StringMap::is_empty")
     )]
     types: StringMap<TypeMetadata>,
+    nested: StringMap<NestMetadata>,
```
alexcrichton, line 524: Could this include the attribute to skip serializing if this is empty? I think that'll assist with backwards compatibility by not breaking historical decoders that don't have this field when it's not used.

crates/wit-component/src/encoding/wit/v2.rs
```diff
@@ -180,52 +226,40 @@ impl InterfaceEncoder<'_> {
...
+        encode_instance_contents(self, interface)?;
         let iface = &self.resolve.interfaces[interface];
-        let mut type_order = IndexSet::new();
-        for (_, id) in iface.types.iter() {
-            self.encode_valtype(self.resolve, &Type::Id(*id))?;
-            type_order.insert(*id);
+        let mut instance = self.pop_instance();
+        for (name, nest) in &iface.nested {
+            self.encode_nested(name, nest.id, &mut instance)?;
         }
```
alexcrichton, line 234: Since this loop is duplicated with the below, could this perhaps be inlined into `encode_instance_contents`? (making that a self-recursive method)

crates/wit-component/src/encoding/wit/v2.rs
```diff
@@ -41,6 +41,52 @@ pub fn encode_component(resolve: &Resolve, package: PackageId) -> Result<Compone
     Ok(encoder.component)
 }
 
+fn encode_instance_contents(enc: &mut InterfaceEncoder, interface: InterfaceId) -> Result<()> {
```
alexcrichton, line 44: This might be good to keep as a method on `InterfaceEncoder` perhaps?

crates/wit-component/src/encoding/wit/v1.rs
```diff
@@ -170,6 +170,10 @@ impl InterfaceEncoder<'_> {
         let iface = &self.resolve.interfaces[interface];
         let mut type_order = IndexSet::new();
         for (_, id) in iface.types.iter() {
+            let ty = &self.resolve.types[*id];
+            if let TypeOwner::Interface(iface_id) = ty.owner {
+                self.interface = Some(iface_id);
+            }
```
alexcrichton, line 176: Is this still necessary? I ask because the v1 encoding is sort of dead at this point and I should probably work to remove it entirely, so in theory no more updates should be necessary.

crates/wit-parser/src/decoding.rs
```diff
@@ -727,6 +727,111 @@ impl WitPackageDecoder<'_> {
...
+        &mut self,
+        name: &str,
+        ty: &types::ComponentInstanceType,
+    ) -> Result<InterfaceId> {
+        let (is_local, interface) = match self.named_interfaces.get(name) {
+            Some(id) => (true, *id),
+            None => (false, self.extract_dep_interface(name)?),
+        };
+        let owner = TypeOwner::Interface(interface);
+        for (name, ty) in ty.exports.iter() {
```
alexcrichton, line 740: I think this is probably a copy/paste from elsewhere in this file, right? If so could that be refactored into one method that's shared?

crates/wit-component/src/encoding/wit.rs
```diff
@@ -311,14 +315,45 @@ impl InterfaceEncoder<'_> {
...
+                "{}:{}/{}@{}",
+                pkg.name.namespace,
+                pkg.name.name,
+                iface.name.as_ref().unwrap(),
+                v.to_string()
+            )
+        } else {
+            format!("{}/{}", pkg.name.to_string(), iface.name.as_ref().unwrap())
+        }
+    }
```
alexcrichton, line 355: For this I think `Resolve::id_of` might be what you want here? (in that I think it's equivalent to this function)

crates/wit-component/src/encoding/wit.rs
```diff
@@ -272,11 +272,15 @@ impl InterfaceEncoder<'_> {
...
-    fn encode_instance(&mut self, interface: InterfaceId) -> Result<u32> {
+    fn encode_instance_contents(&mut self, interface: InterfaceId) -> Result<InstanceType> {
         self.push_instance();
         let iface = &self.resolve.interfaces[interface];
         let mut type_order = IndexSet::new();
         for (_, id) in iface.types.iter() {
+            let ty = &self.resolve.types[*id];
+            if let TypeOwner::Interface(iface_id) = ty.owner {
+                self.interface = Some(iface_id);
+            }
```
alexcrichton, line 283: I think this might be best done a little differently than updating `self.interface` here. What I'm worried about is that this updates `self.interface` but doesn't restore it what it previously was. 

Looking throughout this code though `push_instance` and `pop_instance` might be a good place to hook into? I think it should be ok to move most modifications of `self.interface` into those functions where push takes an `InterfaceId` and sets `self.interface` and then pop restores the previous value saved during push.

crates/wit-parser/src/ast.rs
```diff
@@ -549,6 +552,30 @@ enum InterfaceItem<'a> {
...
+    from: UsePath<'a>,
+}
+
+impl<'a> Nest<'a> {
+    fn parse(
+        tokens: &mut Tokenizer<'a>,
+        docs: Docs<'a>,
+        attributes: Vec<Attribute<'a>>,
+    ) -> Result<Self> {
+        tokens.eat(Token::Nest)?;
```
alexcrichton, line 570: I think you'll want to `expect` here instead of `eat` (as `eat` has a boolean return)

crates/wit-parser/src/decoding.rs
```diff
@@ -948,35 +991,50 @@ impl WitPackageDecoder<'_> {
...
+                            interface.nested.push(Nest {
+                                id: deep,
+                                docs: Default::default(),
+                                stability: Default::default(),
+                            });
+                        }
+                        _ => bail!("expected interface name"),
+                    }
+                }
+                _ => bail!("instance type export `{name}` is not a type, function or instance"),
```
alexcrichton, line 1037: I think this `name` needs to be updated to `exp_name`

crates/wit-parser/src/decoding.rs
```diff
@@ -948,35 +991,50 @@ impl WitPackageDecoder<'_> {
...
+                        ComponentNameKind::Interface(_) => {
+                            let ty = &self.types[inst];
+                            let deep = self.register_export(&exp_name, &ty)?;
+                            interface.nested.push(Nest {
+                                id: deep,
+                                docs: Default::default(),
+                                stability: Default::default(),
+                            });
+                        }
+                        _ => bail!("expected interface name"),
```
alexcrichton, line 1034: Mind adding in the `exp_name` to the error message here?

crates/wit-parser/src/decoding.rs
```diff
@@ -726,6 +727,40 @@ impl WitPackageDecoder<'_> {
         Ok(())
     }
 
+    fn register_export(
```
alexcrichton, line 730: I think this function may want to get folded into `register_import` perhaps? (maybe coupled with a function rename). The internal structure looks quite similar and I think it's needed for importing an interface with `nest` as well?

alexcrichton:
One thing I'd also recommend for testing this is fuzzing. For example here's a small patch to enable some fuzzing of this commit:

```diff
diff --git a/crates/wit-smith/src/generate.rs b/crates/wit-smith/src/generate.rs
index 41c5c5c1..82db5fdd 100644
--- a/crates/wit-smith/src/generate.rs
+++ b/crates/wit-smith/src/generate.rs
@@ -344,6 +344,7 @@ impl<'a> InterfaceGenerator<'a> {
             Use,
             Type,
             Function,
+            Nest,
         }

         let mut parts = Vec::new();
@@ -374,6 +375,11 @@ impl<'a> InterfaceGenerator<'a> {
                 Generate::Function => {
                     parts.push(self.gen_func(u)?);
                 }
+                Generate::Nest => {
+                    if let Some(part) = self.gen_nest(u)? {
+                        parts.push(part);
+                    }
+                }
             }
         }

@@ -882,6 +888,15 @@ impl<'a> InterfaceGenerator<'a> {
     fn gen_unique_name(&mut self, u: &mut Unstructured<'_>) -> Result<String> {
         gen_unique_name(u, &mut self.unique_names)
     }
+
+    fn gen_nest(&mut self, u: &mut Unstructured<'_>) -> Result<Option<String>> {
+        let mut path = String::new();
+        if self.gen.gen_path(u, self.file, &mut path)?.is_none() {
+            return Ok(None);
+        }
+
+        Ok(Some(format!("nest {path};")))
+    }
 }

 fn gen_unique_name(u: &mut Unstructured<'_>, set: &mut HashSet<String>) -> Result<String> {
```

which is executed via

```
FUZZER=roundtrip_wit cargo +nightly fuzz run -s none run
```

(you can pass `--dev` between `run` and `run` to run in debug mode too)

That'll help weed out issues dealing with round-tripping interfaces and such. For example that quickly finds this locally for me:

```
thread '<unnamed>' panicked at fuzz/src/roundtrip_wit.rs:13:61:
called `Result::unwrap()` on an `Err` value: export name `name:name1/x@2.2.2-alpha.0+1.2.0` conflicts with previous name `name:name1/x@2.2.2-alpha.0+1.2.0` (at offset 0x6a)
```

which I think means two things:

1. `Resolve` validation needs to guarantee that you can't `nest` the same interface twice (maybe this means it should switch to `IndexSet`?)
2. The fuzz generator needs to record what was nested and not generate documents that nest the same interface twice. Either that or the error is handled in the fuzzer and the test case is discarded (but I think here the former is the better route to take)

ydnar:
Why `nest` instead of `include`?

macovedj:
> Why `nest` instead of `include`?

As I recall it, there were a few motivating factors behind this, some of which is covered in the discussion [here](https://github.com/WebAssembly/component-model/pull/372).

1.) `nest` is intended to express something that is distinct from "unioning worlds" like `include`.  Instead, if a component exports an instance, this enables one to specify expectations about the interfaces that that instance itself exports.  This is described a bit [here](https://github.com/WebAssembly/component-model/issues/301).

2.) There is the specific use case of being able to add a dependency on a specific "concrete" component (importing a specific implementation of a wit interface, as opposed to what people do currently which is import an unspecified implementation of an interface).  Having the ability to do this would let you do something like `cargo add` for component implementations (not wit interfaces), without something like a `wac` step.  After chatting quite a bit with @lukewagner about this, this syntax was arrived at (though could still change a bit) as a way to help accomplish this.

Re: the fact that this hasn't been touched in awhile, I wanted to get [this old PR](https://github.com/bytecodealliance/wasm-pkg-tools/pull/124) merged first, so that we can reference a `wkg.lock` when reasoning about the logic, but got a bit distracted, though I intend to come back to all of this.

PR #1625, opened 2024-06-20
wit-component: Allow `cabi_realloc` to be weakly defined

kateinoigakukun:
If a weak "default" visibility symbol is defined in a shared library, wasm-ld emits an import for the symbol for the "env" module for the case where the symbol is overridden by other modules. https://github.com/llvm/llvm-project/blob/5b1cc58b3c72f51b8a1ce27e5fba15783d447d7c/lld/wasm/Writer.cpp#L733-L737

`cabi_realloc` is usually defined weakly by wit-bindgen'ed code, and if the code is compiled with `-fvisibility=default`, it emits an import for "env.cabi_realloc". The import leads re-exporting the symbol to the "env" module, and it conflicts with the special re-exporting logic for `cabi_realloc`.

This commit fixes the duplicate re-exporting issue.


kateinoigakukun:
@dicej Gentle ping :)

dicej (approved):
Sorry for the delayed response.  LGTM!

alexcrichton:
That's a nightly build issue, I'll work on getting that sorted out.

PR #1626, opened 2024-06-20
Initialize a test suite for `wit-encoder`

yoshuawuyts:
Follow-up to #1549 and #1580, this begins adding a test suite for `wit-encoder`. Specifically: this takes the tests from `wit-parser`, and then tries to encode those in individual tests. It fixes a number of bugs, as well as adds support for `include` statements. It also changes the default indentation for WIT documents to `2`, which matches the style that's been used in the WIT spec and WASI documents.

Not all `.wit` files have their own tests yet. But I ran out of time before I could finish up all the tests, but I figured even a partial test suite would already be a useful contribution. Thanks!


yoshuawuyts:
lmao, [what in the line-endings](https://github.com/bytecodealliance/wasm-tools/actions/runs/9598811806/job/26471344027?pr=1626):

```text
---- concrete_types stdout ----
thread 'concrete_types' panicked at crates\wit-encoder\tests\empty.rs:9:5:
assertion failed: `(left == right)`

Diff < left / right > :
 package foo:empty;
```

crates/wit-encoder/tests/functions.rs
```diff
@@ -1,88 +1,72 @@
...
-    f5: func(a: f32, b: f32) -> tuple<u32, u32>;
-    f6: func(a: option<u32>) -> result<u32, f32>;
-    f7: func() -> (u: u32, f: f32);
-    f8: func() -> (u: u32);
-    f9: func() -> result<f32>;
-    f10: func() -> result<_, f32>;
-    f11: func() -> result;
-}
-";
+const PACKAGE: &str = include_str!("./functions.wit");
```
alexcrichton, line 6: I personally find it nice when tests are as self-contained as they can be, e.g. "everything in one file", so could the WIT literal be inlined in these files instead of using `include_str!`?

alexcrichton (commented):
Thanks! I think if the files are included in the source files rather than as separate files that may also fix the line endings issue. Also for landing mind leaving out the WIT files that aren't yet tested?

crates/wit-encoder/tests/functions.rs
yoshuawuyts: Yes, absolutely! Are you okay if I use [indoc](https://github.com/dtolnay/indoc) to make sure the formatting is preserved when inlining?

alexcrichton: sure!

alexcrichton (approved):

Issue #1588, opened 2024-05-30
wasmparser -> wasm-encoder bridge


juntyr:
In my usecase, I perform simple transformations on a WASM module (e.g. adding a few instructions to all existing functions). While walrus with its convenient high-level API has served this usecase in the past, it is falling behind on supporting newer WASM features, and falling back to a low-level API is possible for me as well. Using wasmparser and Wasm-encoder myself seems overkill, however, as it exposes my code to the full complexity of the WASM binary format and the fast-developing APIs of both crates.

As both wasmparser and wasm-encoder are jointly developed in this repo, it would be great if a new crate could be developed to bridge the two. Ideally, this crate would hook into wasmparsers streaming API and recreate the same WASM module using wasm-encoder on the fly. A shallow stream adapter API could then be used to modify items on the fly or to insert new ones, i.e. the API would allow any wasmparser stream to go through an adapter function which takes the iterator and produces and iterator which is then consumed by wasm-encoder. To keep the library minimal and low-level, the user would still need to take care of e.g. renumbering indices if additional imports are added.

TLDR: Adding a bridge crate between wasmparser and wasm-encoder that optionally feeds the parsed streams through user-provided iterator adapters and then into wasm-encoder would be awesome :)

alexcrichton:
Thanks for the report! Nowadays I think this'd definitely fit well in the repository in the sense that with the current release process everything is released as a batch so it's fine to have inter-crate dependencies be deep and public. There's actually already a [`wasmparser` feature](https://github.com/bytecodealliance/wasm-tools/blob/acb410e70ea57ffe93853f37d49e366acb3b4962/crates/wasm-encoder/Cargo.toml#L22-L24) on the `wasm-encoder` crate with a few `From` impls, but I think it'd be reasonable to expand that with more functions and/or more conversions. For example there could be constructors a long the lines of "build this entire section from this wasmparser section" or other things like "initialize this function's locals with this wasmparser function's locals" (or body) or similar.

Basically I think this would be a good idea to do, and I think it'd be best done by adding more code to `wasm-encoder` rather than a new crate on top of the two.

juntyr:
> Basically I think this would be a good idea to do, and I think it'd be best done by adding more code to `wasm-encoder` rather than a new crate on top of the two.

That sounds like a good plan! I had a quick look at the feature and it seems that it provides just a few low-level conversions for now.

[...]

PR #1628, opened 2024-06-21
Implement a wasmparser -> wasm-encoder trait

juntyr:
Progress towards #1588, follow-up to #1627 

Thanks to @alexcrichton for coming up with the design!

This PR provides a new module, `reencoder`, which contains the `WasmParserToWasmEncoder` trait with utility functions to convert between `wasmparser` and `wasm-encoder` types. Every default implementation is also provided as a free function in the `reencoder::utils` module. In this PR, only types that are needed to re-encode core WASM modules are included, but future PRs could extend this.

This PR also does not yet replace similar functionality in other wasm-tools crates, e.g. the `wasm_mutate::mutators::translate::Translator` trait, which could in follow-up PRs be cleaned up to use this general-purpose re-encoder infrastructure instead.

Utility conversion functions added in #1627 are now implemented in terms of the `RoundtripReencoder` all-default implementation of the `WasmParserToWasmEncoder`. While most interfaces stay the same (and #1627 has not been published yet), some methods now return `Result`s.

## Unresolved questions and TODOs

- [x] How should user errors be handled? At the moment, many trait methods cannot error, and the others can only return pre-defined errors. We could add an extra variant to the `reencoder::Error` enum and make every method fallible if that's preferable.
- [x] Should the utility methods be monomorphised or use dynamic dispatch?
- [ ] Add documentation (I'd appreciate some help with this)

### DIFF
--- a/crates/wasm-encoder/src/component/types.rs
+++ b/crates/wasm-encoder/src/component/types.rs
@@ -527,16 +527,6 @@ impl Encode for PrimitiveValType {
     }
 }
 
-#[cfg(feature = "wasmparser")]
-impl From<wasmparser::PrimitiveValType> for PrimitiveValType {
-    fn from(ty: wasmparser::PrimitiveValType) -> Self {
-        crate::reencode::utils::component_primitive_val_type(
-            &mut crate::reencode::RoundtripReencoder,
-            ty,
-        )
-    }
-}
-
 /// Represents a component value type.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum ComponentValType {
--- a/crates/wasm-encoder/src/component/types.rs
+++ b/crates/wasm-encoder/src/component/types.rs
@@ -530,21 +530,10 @@ impl Encode for PrimitiveValType {
 #[cfg(feature = "wasmparser")]
 impl From<wasmparser::PrimitiveValType> for PrimitiveValType {
     fn from(ty: wasmparser::PrimitiveValType) -> Self {
-        match ty {
-            wasmparser::PrimitiveValType::Bool => PrimitiveValType::Bool,
-            wasmparser::PrimitiveValType::S8 => PrimitiveValType::S8,
-            wasmparser::PrimitiveValType::U8 => PrimitiveValType::U8,
-            wasmparser::PrimitiveValType::S16 => PrimitiveValType::S16,
-            wasmparser::PrimitiveValType::U16 => PrimitiveValType::U16,
-            wasmparser::PrimitiveValType::S32 => PrimitiveValType::S32,
-            wasmparser::PrimitiveValType::U32 => PrimitiveValType::U32,
-            wasmparser::PrimitiveValType::S64 => PrimitiveValType::S64,
-            wasmparser::PrimitiveValType::U64 => PrimitiveValType::U64,
-            wasmparser::PrimitiveValType::F32 => PrimitiveValType::F32,
-            wasmparser::PrimitiveValType::F64 => PrimitiveValType::F64,
-            wasmparser::PrimitiveValType::Char => PrimitiveValType::Char,
-            wasmparser::PrimitiveValType::String => PrimitiveValType::String,
-        }
+        crate::reencode::utils::component_primitive_val_type(
+            &mut crate::reencode::RoundtripReencoder,
+            ty,
+        )
     }
 }
 
--- a/crates/wasm-encoder/src/core/code.rs
+++ b/crates/wasm-encoder/src/core/code.rs
@@ -103,33 +103,6 @@ impl CodeSection {
         self.num_added += 1;
         self
     }
-
-    /// Parses the input `section` given from the `wasmparser` crate and adds
-    /// all the code to this section.
-    #[cfg(feature = "wasmparser")]
-    pub fn parse_section(
-        &mut self,
-        section: wasmparser::CodeSectionReader<'_>,
-    ) -> crate::reencode::Result<&mut Self> {
-        crate::reencode::utils::parse_code_section(
-            &mut crate::reencode::RoundtripReencoder,
-            self,
-            section,
-        )
-    }
-
-    /// Parses a single [`wasmparser::Code`] and adds it to this section.
-    #[cfg(feature = "wasmparser")]
-    pub fn parse(
-        &mut self,
-        func: wasmparser::FunctionBody<'_>,
-    ) -> crate::reencode::Result<&mut Self> {
-        crate::reencode::utils::parse_function_body(
-            &mut crate::reencode::RoundtripReencoder,
-            self,
-            func,
-        )
-    }
 }
 
 impl Encode for CodeSection {
@@ -241,16 +214,6 @@ impl Function {
         Function::new(locals_collected)
     }
 
-    /// Create a new [`Function`] by parsing the locals declarations from the
-    /// provided [`wasmparser::FunctionBody`].
-    #[cfg(feature = "wasmparser")]
-    pub fn new_parsed_locals(func: &wasmparser::FunctionBody<'_>) -> crate::reencode::Result<Self> {
-        crate::reencode::utils::new_function_with_parsed_locals(
-            &mut crate::reencode::RoundtripReencoder,
-            func,
-        )
-    }
-
     /// Write an instruction into this function body.
     pub fn instruction(&mut self, instruction: &Instruction) -> &mut Self {
         instruction.encode(&mut self.bytes);
@@ -313,19 +276,6 @@ impl Function {
     pub fn into_raw_body(self) -> Vec<u8> {
         self.bytes
     }
-
-    /// Parses a single instruction from `reader` and adds it to `self`.
-    #[cfg(feature = "wasmparser")]
-    pub fn parse(
-        &mut self,
-        reader: &mut wasmparser::OperatorsReader<'_>,
-    ) -> crate::reencode::Result<&mut Self> {
-        crate::reencode::utils::parse_instruction(
-            &mut crate::reencode::RoundtripReencoder,
-            self,
-            reader,
-        )
-    }
 }
 
 impl Encode for Function {
@@ -365,13 +315,6 @@ impl Encode for MemArg {
     }
 }
 
-#[cfg(feature = "wasmparser")]
-impl From<wasmparser::MemArg> for MemArg {
-    fn from(arg: wasmparser::MemArg) -> MemArg {
-        crate::reencode::utils::mem_arg(&mut crate::reencode::RoundtripReencoder, arg)
-    }
-}
-
 /// The memory ordering for atomic instructions.
 ///
 /// For an in-depth explanation of memory orderings, see the C++ documentation
@@ -400,13 +343,6 @@ impl Encode for Ordering {
     }
 }
 
-#[cfg(feature = "wasmparser")]
-impl From<wasmparser::Ordering> for Ordering {
-    fn from(arg: wasmparser::Ordering) -> Ordering {
-        crate::reencode::utils::ordering(&mut crate::reencode::RoundtripReencoder, arg)
-    }
-}
-
 /// Describe an unchecked SIMD lane index.
 pub type Lane = u8;
 
@@ -428,15 +364,6 @@ impl Encode for BlockType {
             Self::Result(ty) => ty.encode(sink),
             Self::FunctionType(f) => (f as i64).encode(sink),
         }
-    }
-}
-
-#[cfg(feature = "wasmparser")]
-impl TryFrom<wasmparser::BlockType> for BlockType {
-    type Error = crate::reencode::Error;
-
-    fn try_from(arg: wasmparser::BlockType) -> Result<BlockType, Self::Error> {
-        crate::reencode::utils::block_type(&mut crate::reencode::RoundtripReencoder, arg)
     }
 }
 
@@ -3396,15 +3323,6 @@ impl Encode for Instruction<'_> {
     }
 }
 
-#[cfg(feature = "wasmparser")]
-impl<'a> TryFrom<wasmparser::Operator<'a>> for Instruction<'a> {
-    type Error = crate::reencode::Error;
-
-    fn try_from(arg: wasmparser::Operator<'a>) -> Result<Self, Self::Error> {
-        crate::reencode::utils::instruction(&mut crate::reencode::RoundtripReencoder, arg)
-    }
-}
-
 #[derive(Clone, Debug)]
 #[allow(missing_docs)]
 pub enum Catch {
@@ -3436,13 +3354,6 @@ impl Encode for Catch {
                 label.encode(sink);
             }
         }
-    }
-}
-
-#[cfg(feature = "wasmparser")]
-impl From<wasmparser::Catch> for Catch {
-    fn from(arg: wasmparser::Catch) -> Catch {
-        crate::reencode::utils::catch(&mut crate::reencode::RoundtripReencoder, arg)
     }
 }
 
@@ -3607,15 +3518,6 @@ impl Encode for ConstExpr {
     fn encode(&self, sink: &mut Vec<u8>) {
         sink.extend(&self.bytes);
         Instruction::End.encode(sink);
-    }
-}
-
-#[cfg(feature = "wasmparser")]
-impl<'a> TryFrom<wasmparser::ConstExpr<'a>> for ConstExpr {
-    type Error = crate::reencode::Error;
-
-    fn try_from(const_expr: wasmparser::ConstExpr) -> Result<Self, Self::Error> {
-        crate::reencode::utils::const_expr(&mut crate::reencode::RoundtripReencoder, const_expr)
     }
 }
 
--- a/crates/wasm-encoder/src/core/code.rs
+++ b/crates/wasm-encoder/src/core/code.rs
@@ -110,22 +110,25 @@ impl CodeSection {
     pub fn parse_section(
         &mut self,
         section: wasmparser::CodeSectionReader<'_>,
-    ) -> wasmparser::Result<&mut Self> {
-        for code in section {
-            self.parse(code?)?;
-        }
-        Ok(self)
+    ) -> crate::reencode::Result<&mut Self> {
+        crate::reencode::utils::parse_code_section(
+            &mut crate::reencode::RoundtripReencoder,
+            self,
+            section,
+        )
     }
 
     /// Parses a single [`wasmparser::Code`] and adds it to this section.
     #[cfg(feature = "wasmparser")]
-    pub fn parse(&mut self, func: wasmparser::FunctionBody<'_>) -> wasmparser::Result<&mut Self> {
-        let mut f = Function::new_parsed_locals(&func)?;
-        let mut reader = func.get_operators_reader()?;
-        while !reader.eof() {
-            f.parse(&mut reader)?;
-        }
-        Ok(self.function(&f))
+    pub fn parse(
+        &mut self,
+        func: wasmparser::FunctionBody<'_>,
+    ) -> crate::reencode::Result<&mut Self> {
+        crate::reencode::utils::parse_function_body(
+            &mut crate::reencode::RoundtripReencoder,
+            self,
+            func,
+        )
     }
 }
 
@@ -241,13 +244,11 @@ impl Function {
     /// Create a new [`Function`] by parsing the locals declarations from the
     /// provided [`wasmparser::FunctionBody`].
     #[cfg(feature = "wasmparser")]
-    pub fn new_parsed_locals(func: &wasmparser::FunctionBody<'_>) -> wasmparser::Result<Self> {
-        let mut locals = Vec::new();
-        for pair in func.get_locals_reader()? {
-            let (cnt, ty) = pair?;
-            locals.push((cnt, ty.try_into().unwrap()));
-        }
-        Ok(Function::new(locals))
+    pub fn new_parsed_locals(func: &wasmparser::FunctionBody<'_>) -> crate::reencode::Result<Self> {
+        crate::reencode::utils::new_function_with_parsed_locals(
+            &mut crate::reencode::RoundtripReencoder,
+            func,
+        )
     }
 
     /// Write an instruction into this function body.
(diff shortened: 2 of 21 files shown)


crates/wasm-encoder/src/reencode.rs
```diff
@@ -0,0 +1,1308 @@
...
+        match self {
+            Self::ParseError(e) => Some(e),
+            Self::InvalidConstExpr | Self::CanonicalizedHeapTypeReference => None,
+        }
+    }
+}
+
+pub type Result<T, E = Error> = std::result::Result<T, E>;
+
+pub trait WasmParserToWasmEncoder {
```
alexcrichton, line 42: I like the name "reencode" for this module so perhaps this trait could be `Reencode` as well?

crates/wasm-encoder/src/reencode.rs
```diff
@@ -0,0 +1,1308 @@
...
+pub enum Error {
+    /// There was a type reference that was canonicalized and no longer
+    /// references an index into a module's types space, so we cannot encode it
+    /// into a Wasm binary again.
+    CanonicalizedHeapTypeReference,
+    /// The const expression is invalid: not actually constant or something like
+    /// that.
+    InvalidConstExpr,
+    /// There was an error when parsing.
+    ParseError(wasmparser::BinaryReaderError),
```
alexcrichton, line 13: This'll probably want a `From` implementation to avoid all the `.map_err(Error::ParseError)` below too

crates/wasm-encoder/src/component/types.rs
```diff
@@ -530,21 +530,10 @@ impl Encode for PrimitiveValType {
...
-            wasmparser::PrimitiveValType::U64 => PrimitiveValType::U64,
-            wasmparser::PrimitiveValType::F32 => PrimitiveValType::F32,
-            wasmparser::PrimitiveValType::F64 => PrimitiveValType::F64,
-            wasmparser::PrimitiveValType::Char => PrimitiveValType::Char,
-            wasmparser::PrimitiveValType::String => PrimitiveValType::String,
-        }
+        crate::reencode::utils::component_primitive_val_type(
+            &mut crate::reencode::RoundtripReencoder,
+            ty,
+        )
```
alexcrichton, line 536: I feel like I had ambitions for the style of implementation in `wasm-mutate` which may not have quite panned out. Specifically having all the default trait methods call helper functions as opposed to putting the implementation in the trait method itself.

Could the `utils` module instead be not used that much ideally? For example could this be `RoundtripReencoder.component_primitive_val_type(ty)`?

crates/wasm-encoder/src/core/code.rs
```diff
@@ -398,11 +396,8 @@ impl Encode for BlockType {
 impl TryFrom<wasmparser::BlockType> for BlockType {
     type Error = ();
```
alexcrichton, line 397: Given the nice `reencode::Error` I think it'd be reasonable to switch this from `()` to that instead.

alexcrichton: Also, while you're here, it might make sense to also move all the `From`/`TryFrom` impls into the `reencode` module too?

crates/wasm-encoder/src/core/code.rs
```diff
@@ -110,22 +110,25 @@ impl CodeSection {
     pub fn parse_section(
```
alexcrichton, line 110: Question for you, do you think it's worth keeping these methods? They're pretty easy to call as `RoundtripReencoder.parse_code_section(&mut code, section)` so I could see a case being made to removing them in favor of just having the `reencode.rs` implementation

crates/wasm-encoder/src/reencode.rs
```diff
@@ -0,0 +1,1308 @@
+#![allow(missing_docs)] // FIXME
```
alexcrichton, line 1: Mind putting this on the `trait` and then it can be slowly filled in from there? Otherwise though I think it would be good to leave a high-level doc-comment on the module here indicating roughly what it's used for and perhaps a small example.

crates/wasm-encoder/src/reencode.rs
```diff
@@ -0,0 +1,1308 @@
...
+        utils::function_index(self, func)
+    }
+
+    fn tag_index(&mut self, tag: u32) -> u32 {
+        utils::tag_index(self, tag)
+    }
+
+    fn label_index(&mut self, label: u32) -> u32 {
+        utils::label_index(self, label)
+    }
```
alexcrichton, line 80: For functions like this where the `utils` version is blank I think it's ok to remove the `utils` version

alexcrichton: Also sorting-wise could all the indices be moved up to the top of the trait to more clearly call out that it's possible to remap them?

crates/wasm-encoder/src/reencode.rs
```diff
@@ -0,0 +1,1308 @@
...
+                        },
+                        &mut codes,
+                    );
+                }
+                wasmparser::Payload::ElementSection(section) => {
+                    let mut elements = crate::ElementSection::new();
+                    reencoder.parse_element_section(&mut elements, section)?;
+                    module_section_flush_codes(module, &elements, &mut codes);
+                }
+                wasmparser::Payload::DataCountSection { .. } => (),
```
alexcrichton, line 461: I think this'll want to be handled and have a mapped data count section emitted

crates/wasm-encoder/src/reencode.rs
```diff
@@ -0,0 +1,1308 @@
...
+                | wasmparser::Payload::CoreTypeSection(_)
+                | wasmparser::Payload::ComponentSection { .. }
+                | wasmparser::Payload::ComponentInstanceSection(_)
+                | wasmparser::Payload::ComponentAliasSection(_)
+                | wasmparser::Payload::ComponentTypeSection(_)
+                | wasmparser::Payload::ComponentCanonicalSection(_)
+                | wasmparser::Payload::ComponentStartSection { .. }
+                | wasmparser::Payload::ComponentImportSection(_)
+                | wasmparser::Payload::ComponentExportSection(_) => {
+                    unreachable!("wasm component model section is not allowed in core wasm module")
```
alexcrichton, line 487: Could this return an error instead of panicking?

crates/wasm-encoder/src/reencode.rs
```diff
@@ -0,0 +1,1308 @@
...
+                }
+                wasmparser::Payload::CustomSection(section) => {
+                    module_section_flush_codes(
+                        module,
+                        &reencoder.custom_section(section),
+                        &mut codes,
+                    );
+                }
+                wasmparser::Payload::UnknownSection { id, .. } => {
+                    unreachable!("unknown wasm section with id {id}")
```
alexcrichton, line 497: Could this transfer the section as-is without parsing instead of panicking? (not that it's too useful but removing panics seems nice)

crates/wasm-encoder/src/reencode.rs
```diff
@@ -0,0 +1,1308 @@
...
+
+impl WasmParserToWasmEncoder for RoundtripReencoder {}
+
+pub mod utils {
+    use super::{Error, Result, WasmParserToWasmEncoder};
+
+    pub fn core_module<'a, 'b>(
+        reencoder: &mut (impl ?Sized + WasmParserToWasmEncoder),
+        module: &'a mut crate::Module,
+        sections: impl Iterator<Item = Result<wasmparser::Payload<'b>, wasmparser::BinaryReaderError>>,
```
alexcrichton, line 386: Instead of taking `sections` could this instead take `wasmparser::Parser` and `&[u8]`? There's various invariants around the state transitions of parsing the code section which I think want to be asserted on below which aren't guaranteed with an arbitrary iterator but are guaranteeed for a parser/slice combo

crates/wasm-encoder/src/reencode.rs
```diff
@@ -0,0 +1,1308 @@
...
+                    module_section_flush_codes(module, &elements, &mut codes);
+                }
+                wasmparser::Payload::DataCountSection { .. } => (),
+                wasmparser::Payload::DataSection(section) => {
+                    let mut data = crate::DataSection::new();
+                    reencoder.parse_data_section(&mut data, section)?;
+                    module_section_flush_codes(module, &data, &mut codes);
+                }
+                wasmparser::Payload::CodeSectionStart { .. } => {
+                    if let Some(codes) = codes.replace(crate::CodeSection::new()) {
```
alexcrichton, line 468: Here it should be ok to assert that `codes` is `None` and I think that this'll want to fold in the sub-iteration loop to eagerly consume all `CodeSectionEntry` payloads next to create the `CodeSection`. That way the flushing of the code section can be localized to here without having to happen in all other arms as well

crates/wasm-encoder/src/component/types.rs
alexcrichton: Actually thinking a bit more on this, what do you think about removing the `utils` module entirely? Instead having only defaulted trait methods?

alexcrichton (commented):
>  How should user errors be handled? 

I think this'll be best guided by using this module elsewhere. I think it's ok as-is and the introduction of using this in other locations can help figure out how best, if at all, to return custom errors.

> Should the utility methods be monomorphised or use dynamic dispatch?

I think as-is is fine, we can see how bad the blowup is, if at all, later on.

> Add documentation (I'd appreciate some help with this)

Happy to help, if you wouldn't mind moving `missing_docs` to just the trait I think that should be a reasonable balance for now. No need to document all the trait methods.

crates/wasm-encoder/src/component/types.rs
juntyr: My reasoning for having the default dispatch implementations as free functions is that there may be cases where one wants to override one of the methods but just to inspect something and not to change the actual functionality. Instead of then needing to copy-past in the default impl, one can simply call the utils function.

I agree that most of these methods will likely not need that, and we could remove them if we think that the default impl is short enough. My prime usecase would be the reencoding of instructions - if I just need to change a few match arms, I don't want to copy-paste the entire thing (especially since I do want to preserve all recursive reencoding behaviour of indices).

crates/wasm-encoder/src/core/code.rs
juntyr: I put them in the initial commit to minimise the changed API. I think they could probably be removed (and I would remove the mut reference return). If using the RoundtripReencoder becomes common, everybody will need to also have the reencode trait in scope, but that should be a small cost to pay.

crates/wasm-encoder/src/reencode.rs
juntyr: Done

crates/wasm-encoder/src/reencode.rs
juntyr: Done

crates/wasm-encoder/src/core/code.rs
juntyr: Done

crates/wasm-encoder/src/reencode.rs
juntyr: Thanks for catching this - done :)

crates/wasm-encoder/src/reencode.rs
juntyr: Done

crates/wasm-encoder/src/reencode.rs
juntyr: Done

crates/wasm-encoder/src/reencode.rs
juntyr: Done

crates/wasm-encoder/src/reencode.rs
juntyr: Done

crates/wasm-encoder/src/reencode.rs
juntyr: Done

crates/wasm-encoder/src/reencode.rs
juntyr: I've reordered the members, though I'm unsure about the order

juntyr:
@alexcrichton Thanks for your feedback! I've tried to address most of it and made some changes. Feel free to make further tweaks on this branch as I'll now be in my weekend :)

crates/wasm-encoder/src/component/types.rs
alexcrichton: Ok I think that's reasonable yeah 👍

alexcrichton (commented):
I've pushed up a few minor changes, but I'd like to also poke at this a bit more to see if internally wasm-tools can use this in a few places.

crates/wasm-encoder/src/reencode.rs
```diff
@@ -0,0 +1,1755 @@
...
+pub mod utils {
+    use super::{Chunk, Error, Reencode, Result};
+
+    pub fn parse_core_section(
+        reencoder: &mut (impl ?Sized + Reencode),
+        module: &mut crate::Module,
+        parser: &mut wasmparser::Parser,
+        data: &[u8],
+        eof: bool,
+    ) -> Result<Chunk> {
```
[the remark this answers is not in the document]
juntyr: Good point, I mostly included it to give you the option :) I also have the data in memory in my use case

crates/wasm-encoder/src/reencode.rs
```diff
@@ -0,0 +1,1755 @@
...
+    /// `exports` section.
+    pub fn parse_export(
+        reencoder: &mut (impl ?Sized + Reencode),
+        exports: &mut crate::ExportSection,
+        export: wasmparser::Export<'_>,
+    ) {
+        exports.export(
+            export.name,
+            reencoder.export_kind(export.kind),
+            reencoder.export_index(export.index),
```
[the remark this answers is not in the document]
juntyr: Whoops good catch!

crates/wasm-encoder/src/reencode.rs
```diff
@@ -0,0 +1,1755 @@
...
+        utils::function_index(self, func)
+    }
+
+    fn global_index(&mut self, global: u32) -> u32 {
+        utils::global_index(self, global)
+    }
+
+    fn label_index(&mut self, label: u32) -> u32 {
+        utils::label_index(self, label)
+    }
```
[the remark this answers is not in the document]
juntyr: Oh, I didn’t know that, thanks for the explanation

crates/wasm-encoder/src/reencode.rs
```diff
@@ -0,0 +1,1577 @@
...
+
+    /// Parses the single [`wasmparser::Export`] provided and adds it to the
+    /// `exports` section.
+    fn parse_export(&mut self, exports: &mut crate::ExportSection, export: wasmparser::Export<'_>) {
+        utils::parse_export(self, exports, export)
+    }
+
+    /// Parses the input `section` given from the `wasmparser` crate and adds
+    /// all the functions to the `functions` section.
+    fn parse_function_section(
```
juntyr, line 250: Should this also have a singular variant, in case someone wants to e.g. filter out items

crates/wasm-encoder/src/reencode.rs
```diff
@@ -0,0 +1,1577 @@
...
+        &mut self,
+        imports: &mut crate::ImportSection,
+        import: wasmparser::Import<'_>,
+    ) -> Result<()> {
+        utils::parse_import(self, imports, import)
+    }
+
+    /// Parses the input `section` given from the `wasmparser` crate and adds
+    /// all the memories to the `memories` section.
+    fn parse_memory_section(
```
juntyr, line 300: Singular (as above)?

crates/wasm-encoder/src/reencode.rs
```diff
@@ -0,0 +1,1577 @@
...
+        &mut self,
+        tables: &mut crate::TableSection,
+        table: wasmparser::Table<'_>,
+    ) -> Result<()> {
+        utils::parse_table(self, tables, table)
+    }
+
+    /// Parses the input `section` given from the `wasmparser` crate and adds
+    /// all the tags to the `tags` section.
+    fn parse_tag_section(
```
juntyr, line 329: Singular (as above)?

juntyr:
> I've pushed up a few minor changes, but I'd like to also poke at this a bit more to see if internally wasm-tools can use this in a few places.

Thanks for the review and for pushing the fixes!

That sounds good to me and would help with checking if the design works well in practice.

Have a good weekend!

juntyr:
I did some quick testing of how this would work with my usecase and my code is a *lot* cleaner now :)

There are two aspects I'd like to still address:

- [ ] user error handling has now come up for me - I'll submit a PR for that later (I think I'll give the error a generic parameter that is infallible by default)
- [ ] inserting sections when they don't exist in the original - in my case, I always need an import section. While I can detect the case when no import section is present, I don't have access to the module at the point. I'd like to figure out a way for the reencode-a-full-module-helper to support this case. Any ideas?

crates/wasm-encoder/src/reencode.rs
```diff
@@ -0,0 +1,1751 @@
...
+    }
+
+    fn custom_section<'a>(
+        &mut self,
+        section: wasmparser::CustomSectionReader<'a>,
+    ) -> crate::CustomSection<'a> {
+        utils::custom_section(self, section)
+    }
+
+    fn intersperse_section_hook(
```
alexcrichton, line 205: Since this is somewhat subtle mind adding some docs? For example I thought `before` should be just `SectionId` and not `Option` but I see now that it's used to indicate the end of the module with `None`

crates/wasm-encoder/src/reencode.rs
```diff
@@ -0,0 +1,1577 @@
...
+    fn custom_section<'a>(
+        &mut self,
+        section: wasmparser::CustomSectionReader<'a>,
+    ) -> crate::CustomSection<'a> {
+        utils::custom_section(self, section)
+    }
+
+    /// Parses the input `section` given from the `wasmparser` crate and adds
+    /// all the code to the `code` section.
+    fn parse_code_section(
```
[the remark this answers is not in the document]
alexcrichton: I don't think it's used directly at this time but given how it's built on the other primitives I don't think it's too too hard to handle.

crates/wasm-encoder/src/reencode.rs
```diff
@@ -0,0 +1,1577 @@
...
+
+    fn tag_type(&mut self, tag_ty: wasmparser::TagType) -> crate::TagType {
+        utils::tag_type(self, tag_ty)
+    }
+
+    fn val_type(&mut self, val_ty: wasmparser::ValType) -> Result<crate::ValType> {
+        utils::val_type(self, val_ty)
+    }
+
+    fn custom_section<'a>(
```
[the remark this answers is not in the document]
alexcrichton: Perhaps both? The current naming I think makes sense but it might also make sense to have a `parse_custom_section` too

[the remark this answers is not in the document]
alexcrichton: Thinking toward the future, it might make more sense to have `parse_custom_section` which would parse known custom sections and remap indices in them (e.g. the `name` custom section)

crates/wasm-encoder/src/reencode.rs
alexcrichton: Given the simplicity of this and the ones below I'd lean towards "no", but I don't really mind either way.

alexcrichton (commented):
I'm not sure if I'm sold on the user-configurable error here but I'm happy to try it out and see how it works. I was hoping that this might be possible as an adapter like `ReencodeWithError<R: Reencode>` where the outer type would store an error to get returned later (or something like that), but that's a bit wonky too.

crates/wasm-encoder/src/reencode.rs
juntyr: One aspect I'm not set on yet is whether custom sections should be reported as before and after, since one likely does not care whether one is between custom sections but rather wants to know which non-custom section was last and is next

crates/wasm-encoder/src/reencode.rs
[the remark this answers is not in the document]
juntyr: Would that have a different signature?

juntyr:
Well ... I guess ... that's possible too. A user reencoder could store an optional error inside itself and check that afterwards. If we don't go for a user error I'll use that, but it's not my preferred API (though I'm very open to variants of the current one)

crates/wasm-encoder/src/reencode.rs
alexcrichton: Yeah I think it's reasonable to document that one of after or before will be `None` or a non-custom section and not worry about between custom sections.

crates/wasm-encoder/src/reencode.rs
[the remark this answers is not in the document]
alexcrichton: In theory yeah, taking a `&mut Module` perhaps instead of returning a `CustomSection`, but `CustomSection` also can own its contents so maybe not. I don't have a strong preference either way.

alexcrichton:
Yeah sorry I don't mean to say that I don't like `Error<E>`, only that it might be worth considering something else. That being said I agree ergonomically this is probably better so let's stick with `Error<E>`

juntyr:
> Yeah sorry I don't mean to say that I don't like `Error<E>`,

I am very happy that we're discussing alternatives. I must say that I also don't love that the error has to be that invasive (an associated type default would slighty help hide it better), so I'm really interested how the error type now works for you in practice for using it inside wasm-tools, and how we can perhaps improve this first design form it

> only that it might be worth considering something else.

I absolutely agree :D

crates/wasm-encoder/src/reencode.rs
juntyr: Ok, let's keep it as-is then

crates/wasm-encoder/src/reencode.rs
juntyr: Done :)

crates/wasm-encoder/src/reencode.rs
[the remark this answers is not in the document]
juntyr: Thanks for the feedback - I've now added an extra parse method that calls the existing method

juntyr:
@alexcrichton Do you want to explore this PR a bit more to see how existing code can be cleaned up and are there some other changed / improvements that I can make?

alexcrichton:
And thank you for pushing on this and helping to iterate on it!

PR #1630, opened 2024-06-22
wasm-encoder: add method to allow fetching raw Function bytes.

cfallin:
When generating function bodies with `wasm-encoder`, it is sometimes useful to take the function bytecode, cache it, then reuse it later. (For my specific use-case, in weval, I would like to cache weval-specialized function bodies and reuse them when creating new modules.)

Unfortunately the existing API of `wasm-encoder` makes this *almost* but *not quite* easy: `Function` implements `Encode::encode`, but this will produce a function body *with* the length prefixed. However there's no way to use this unmodified with the next level up the entity hierarchy: `CodeSection::raw` wants the function bytes *without* the length prefixed.

This is a fairly annoying if small API gap, and otherwise requires manually stripping the length prefix, a leb128-encoded integer. It's also a small footgun: I naively did not realize this mismatch and tried to do the above, only to get perplexing type errors with locals.

This PR adds one method on `wasm_encoder::Function` to return the inner bytes directly, and the doc-comment contains an example showing the intended use-case.

### DIFF
--- a/crates/wasm-encoder/src/core/code.rs
+++ b/crates/wasm-encoder/src/core/code.rs
@@ -274,6 +274,45 @@ impl Function {
         self.bytes.len()
     }
 
+    /// Unwraps and returns the raw byte encoding of this function.
+    ///
+    /// This encoding doesn't include the variable-width size field
+    /// that `encode` will write before the added bytes. As such, its
+    /// length will match the return value of [`byte_len`].
+    ///
+    /// # Use Case
+    ///
+    /// This raw byte form is suitable for later using with
+    /// [`CodeSection::raw`]. Note that it *differs* from what results
+    /// from [`Function::encode`] precisely due to the *lack* of the
+    /// length prefix; [`CodeSection::raw`] will use this. Using
+    /// [`Function::encode`] instead produces bytes that cannot be fed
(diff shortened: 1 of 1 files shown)


alexcrichton (approved):

alexcrichton:
Happy to add more helpers and/or edit existing helpers as well if needed!

cfallin:
Thanks! I think I have what I need for now (I have caching working in weval).

It looks like the release schedule is somewhat ad-hoc here but that we have the same automation as in Wasmtime -- if it's not too much trouble, would it be possible to do a point release with this? No particular rush if another release is coming soon anyway.

alexcrichton:
It's true yeah there's no release cadence here, but one difference from Wasmtime is that there's not great support for backports and/or release branches. The current patch release process only works if there aren't otherwise breaking changes on `main` relative to the last release, which there currently are. Would it be best to get this out pretty soon? If so I might manually make a release branch for this and backport it

cfallin:
Ah, no worries then, happy to wait for the next major release. To be concrete I don't have any reason to ask for it to be expedited this week in particular, then I'm on PTO for two weeks, so any time in the next month is probably fine :-)

PR #1632, opened 2024-06-24
Implement DWARF support in text-to-binary

alexcrichton:
This commit implements support to emit DWARF debugging information when the WebAssembly text format is converted to binary. Currently there is otherwise no means of translating a binary offset in a WebAssembly file back to the text format that it came from. This support is implemented with a few API knobs here and there as well as two new CLI flags for all the commands that support the text format as input: `-g` and `--generate-dwarf lines|full`.

The original motivation of this commit is that I was curious to learn more about DWARF and it seemed like a neat little feature that wouldn't be too hard to maintain. The inspiration was bytecodealliance/wasmtime#8658 which this does not implement as-is (e.g.  not custom DWARF, just wasm-text-DWARF). Otherwise though I've realized that this can be useful for development in a few situations:

* Wasmtime's backtraces can now mention filenames/line numbers of the WebAssembly text format.
* Validation errors can use `addr2line` to print a filename/line number of a WebAssembly text file.
* Wasmtime doesn't have a native debugger nor does LLDB/GDB know about WebAssembly. Through Wasmtime's DWARF transformation this enables debugging WebAssembly text files.

This commit implements knobs for "full" or "line" support in DWARF. The "line" support should be as complete as it can be (DWARF can only provide information for the code section). The "full" support is somewhat basic still but my hope is that it can be expanded in the future as needed.


fitzgen (approved):
Nice!

mverleg:
Very nice! I'd like the generating part of this from `wasm_encoder` (using my own spans etc as input).

But I believe it's not usable externally, is that true? I see the dwarf data is [inserted here while encoding](https://github.com/bytecodealliance/wasm-tools/blob/5dee4ec15af6a692af0a4db18545c97c021044df/crates/wast/src/core/binary.rs#L186) but it's not public.

alexcrichton:
It's right yeah that it's not externally visible. I'd recommend copying the bits/pieces from https://github.com/bytecodealliance/wasm-tools/blob/5dee4ec15af6a692af0a4db18545c97c021044df/crates/wast/src/core/binary/dwarf.rs and editing as needed. That's pretty specialized to the use case at hand and doesn't support a lot of wasm as well. Most of the heavy lifting is done by `gimli` and those crates, which are all publicly accessible.

PR #1635, opened 2024-06-24
Limit component model flags to 32

alexcrichton:
This commit is an attempt to explore the design space of WebAssembly/component-model#370. This limits, by default, the number of flags in a component model type to 32 by default. The hope of this issue is to be able to ratchet the maximum number of flags to make it easier on bindings generators to not need to work with arbitrary numbers of flags. The secondary hope is that we can ratchet flags straight to 32 instead of 64 due to it being unlikely that more than 32 flags are in use. Once this percolates there can then be a separate feature for enabling 33-64 flags.


crates/wasmparser/src/validator/component.rs
```diff
@@ -2681,14 +2682,23 @@ impl ComponentState {
...
+    ) -> Result<ComponentDefinedType> {
         let mut names_set = IndexSet::default();
         names_set.reserve(names.len());
 
         if names.is_empty() {
             bail!(offset, "flags must have at least one entry");
         }
 
+        if names.len() > 32 && !features.component_model_more_flags() {
+            bail!(offset, "cannot have more than 32 flags");
```
pchickey, line 2699: Add instructions to comment on CM issue and enable the more flags feature here?

pchickey (approved):

PR #1646, opened 2024-06-27
threads: add `shared` composite types

abrown:
This continues the work of https://github.com/bytecodealliance/wasm-tools/pull/1600 to expand the surface area of a WebAssembly module that can be marked `shared`. This is for support of the shared-everything-threads [proposal]. The key change is to convert `CompositeType` in each of the crates from an `enum` to a `struct` in order to add a `shared` boolean field. The original variants (`Func`, `Array`, `Struct`) are moved to a separate `enum` and fill an `inner` field. Propagating this throughout the code base is the bulk of this PR, with occasional, small refactors to avoid larger match patterns.

### DIFF
--- a/crates/wasmparser/src/readers/core/types/matches.rs
+++ b/crates/wasmparser/src/readers/core/types/matches.rs
@@ -104,6 +104,9 @@ impl<'a> Matches for WithRecGroup<&'a SubType> {
 impl<'a> Matches for WithRecGroup<&'a CompositeType> {
     fn matches(types: &TypeList, a: Self, b: Self) -> bool {
         use CompositeInnerType::*;
+        if (*a).shared != (*b).shared {
+            return false;
+        }
         match (&(*a).inner, &(*b).inner) {
             (Func(fa), Func(fb)) => Matches::matches(
                 types,
--- a/crates/wasmparser/src/readers/core/types/matches.rs
+++ b/crates/wasmparser/src/readers/core/types/matches.rs
@@ -14,8 +14,8 @@
 
 use crate::{
     types::{CoreTypeId, RecGroupId, TypeList},
-    ArrayType, CompositeType, FieldType, FuncType, RefType, StorageType, StructType, SubType,
-    ValType,
+    ArrayType, CompositeInnerType, CompositeType, FieldType, FuncType, RefType, StorageType,
+    StructType, SubType, ValType,
 };
 
 /// Wasm type matching.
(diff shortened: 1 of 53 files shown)


abrown:
cc: @fitzgen

crates/wasmparser/src/readers/core/types/matches.rs
```diff
@@ -103,27 +103,28 @@ impl<'a> Matches for WithRecGroup<&'a SubType> {
 
 impl<'a> Matches for WithRecGroup<&'a CompositeType> {
     fn matches(types: &TypeList, a: Self, b: Self) -> bool {
-        match (&*a, &*b) {
-            (CompositeType::Func(fa), CompositeType::Func(fb)) => Matches::matches(
+        use CompositeInnerType::*;
+        match (&(*a).inner, &(*b).inner) {
```
fitzgen, line 107: Shouldn't matching take `shared`ness into account? Does an unshared `T` match a shared `T`? I vaguely remember reading some email about starting conservatively and saying no, but I don't remember...

fitzgen (approved):
Thanks! Can you also add a test where an unshared struct/array/func subtypes a shared struct/array/func and assert that doing so is (in)valid?

crates/wasmparser/src/readers/core/types/matches.rs
abrown: Yeah, that's a good point... let me think about that.

fitzgen: Seems fine to start by only allowing shared things to match other shared things and similarly for unshared. We can land code that does that in `wasm-tools` for now, in the name of incrementality, even if the upstream proposal wants to do something more complicated.

abrown: Ok, I did the simple thing for now and opened: https://github.com/WebAssembly/shared-everything-threads/issues/72.

tlively: Eventually I would like to upstream these related spec tests: 
 - https://github.com/WebAssembly/binaryen/blob/main/test/spec/shared-struct.wast
 - https://github.com/WebAssembly/binaryen/blob/main/test/spec/shared-array.wast
 - https://github.com/WebAssembly/binaryen/blob/main/test/spec/shared-polymorphism.wast

abrown: Nice! I tried to drop those in and I think I'm seeing parsing differences between Binaryen and wasm-tools?!

```
---- "tests/local/shared-everything-threads/array.wast" ----
failed test: tests/local/shared-everything-threads/array.wast

Caused by:
    0: failed to run `json-from-wast` cli subcommand
    1: failed to run "/home/abrown/Code/wasm-tools/target/debug/wasm-tools" "json-from-wast" "tests/local/shared-everything-threads/array.wast" "--pretty" "--wasm-dir" "/tmp/.tmpjHLO2v"
       stdout:
       stderr: error: unexpected token, expected one of: `funcref`, `externref`, `exnref`, `anyref`, `eqref`, `structref`, `arrayref`, `i31ref`, `nullfuncref`, `nullexternref`, `nullexnref`, `nullref`, left paren
            --> tests/local/shared-everything-threads/array.wast:78:8
             |
          78 |   (elem)
             |        ^


---- "tests/local/shared-everything-threads/struct.wast" ----
failed test: tests/local/shared-everything-threads/struct.wast

Caused by:
    0: failed to run `json-from-wast` cli subcommand
    1: failed to run "/home/abrown/Code/wasm-tools/target/debug/wasm-tools" "json-from-wast" "tests/local/shared-everything-threads/struct.wast" "--pretty" "--wasm-dir" "/tmp/.tmpN86fk1"
       stdout:
       stderr: error: expected `(`
            --> tests/local/shared-everything-threads/struct.wast:56:27
             |
          56 |     (type (shared (struct anyref)))
             |                           ^
```

@fitzgen, I'm not the GC expert here... do we need to improve parsing of GC here as well or should we touch these tests up to make them parse?

abrown: I added the tests but modified them just enough to avoid the missing `wast` parsing (see commit message of ef101c3).

abrown:
Not sure what is wrong in CI here but it seems unrelated to this change.

abrown:
Rebased on top of #1651.

fitzgen (approved):

PR #1647, opened 2024-06-28
Added serde support to wit-encoder

MendyBerger:
This PR add serde support to wit-encoder.

Why is wit-parser's serde not enough?
- wit-parser only supports `Serialize` and not `Deserialize`. So we can't recreate it from json.
- wit-parser's representation includes a lot of references by id, and it's hard to keep track of that in a json file.

### DIFF
--- a/crates/wit-encoder/Cargo.toml
+++ b/crates/wit-encoder/Cargo.toml
@@ -10,9 +10,14 @@ version.workspace = true
 [lints]
 workspace = true
 
+[features]
+default = ["serde"]
+serde = ["dep:serde", "semver/serde"]
+
 [dependencies]
 semver = { workspace = true }
 pretty_assertions = { workspace = true }
+serde = { workspace = true, optional = true, features = ["derive"] }
 
 [dev-dependencies]
 indoc = { workspace = true }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -942,9 +942,9 @@ checksum = "78b3ae25bc7c8c38cec158d1f2757ee79e9b3740fbc7ccf0e59e4b08d793fa89"
 
 [[package]]
 name = "log"
-version = "0.4.21"
+version = "0.4.22"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "90ed8c1e510134f979dbc4f070f87d4313098b704861a105fe34231c70a3901c"
+checksum = "a7a70ba024b9dc04c27ea2f0c0548feb474ec5c54bba33a7f72f873a39d07b24"
 
 [[package]]
 name = "logos"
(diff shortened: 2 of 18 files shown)


MendyBerger:
The main reason I need this is to allow adding arbitrary wit to an existing wit package. If there are references by id, that breaks that ability.

I definitely get the problem with having multiple json representations.
How open would you be to only having the json from wit-encoder? I'm working on a tool that converts wit-parser's resolve into a list of wit-encoder packages. That would mean that we can take a resolve, convert it into wint-encoder, and get the json from the wit-encoder.

As for adding `Deserialize` to wit-parser: I don't think that you can build up an arena from json.

Thoughts @alexcrichton?

alexcrichton:
I would prefer to not change the "main" json representation from what it is today because using an id-based encoding makes it much less error-prone to interpret the JSON output. For example it doesn't require all consumers to perform their own name resolution and such. Instead a directly usable form of the AST is processed.

I'd personally prefer to push on adding custom deserialization/etc to be able to deserialize from JSON. That may require changes in `id-arena` but I suspect that would be reasonable to do upstream as well.

alexcrichton:
I don't know the specifics of your use case so it's hard for me to propose a precise solution to it. I agree that modifying the JSON output of `wit-parser` would not be easy. I don't really understand how JSON fits into your workflow because the naive answer I'd have to adding custom WIT would be to add more `*.wit` files adjacent to the original WIT files, but if JSON is in the mix I don't know how that affects things

MendyBerger:
Having separate *.wit files, won't let us augment existing interfaces.

crates/wit-encoder/Cargo.toml
```diff
@@ -10,9 +10,14 @@ version.workspace = true
 [lints]
 workspace = true
 
+[features]
+default = ["serde"]
+serde = ["dep:serde", "semver/serde"]
```
alexcrichton, line 15: I talked with @MendyBerger about this change and this PR and got more context on the rationale. I now think this is reasonable to land in wasm-tools, but I wanted to highlight a change here that @MendyBerger and I also talked about.

Specifically I think that this is a good location to document what this feature is and what it should be used for. Basically explaining that this enables, for example, json serialization/deserialization of `wit-encoder` structures. I think it would be worthwhile to comment here as well that this json format is different from the format of `wit-parser` and additionally serves a different purpose. For example if you're looking to create WIT things from json then this is the crate and feature for you, but if you're parsing WIT and reading the output through json then `wit-parser` is probably the better option.

Another thing I think would be good to mention here as well is that the exact structure of the JSON is probably not going to be super stable over time, so slightly tweaks and variants should be expected as this crate evolves.

MendyBerger: @alexcrichton I've added the notes you requested

alexcrichton (approved):

MendyBerger:
Thanks @alexcrichton for the help and guidance with this!

PR #1649, opened 2024-06-28
threads: add `shared` tables

abrown:
This change spreads `shared` attributes on to tables. As with #1480, this does not yet turn on fuzzing for `shared` things so the only checking is via the in-progress test suite in `tests/local/shared-everything-threads`.

Implementing this raises an issue with the spec: the addition of `shared` in the text format means that inline table expressions (e.g.,
`(table <type> (elem ...))`) are difficult to parse: the parser would potentially have to look past two tokens now, `shared` and `i32|i64` to see if a type appears that indicates we should be parsing an inline format. There are multiple options for what to do here; I opened an issue to figure this out at the spec level first ([https://github.com/bytecodealliance/wasm-tools/pull/71]).

[https://github.com/bytecodealliance/wasm-tools/pull/71]: https://github.com/WebAssembly/shared-everything-threads/issues/71

### DIFF
--- a/tests/local/shared-everything-threads/table.wast
+++ b/tests/local/shared-everything-threads/table.wast
@@ -35,3 +35,9 @@
 (assert_invalid
   (module (table (import "spectest" "table_ref") shared 0 funcref))
   "shared tables must have a shared element type")
+
+(assert_invalid
+  (module
+    (type $t (func))
+    (table shared 0 (ref $t)))
+  "shared tables must have a shared element type")
--- a/tests/local/shared-everything-threads/table.wast
+++ b/tests/local/shared-everything-threads/table.wast
@@ -1,0 +1,37 @@
+;; Check the `shared` attribute on tables.
+
+(module
+  ;; Imported.
+  (table (import "spectest" "table_ref") shared 1 (ref null (shared func)))
+  (table (import "spectest" "table_ref_with_max") shared 1 1 (ref null (shared func)))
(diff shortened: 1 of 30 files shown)


abrown:
__DO NOT MERGE__: this is based on #1646 and will be rebased once that merges.

abrown:
Ok, this is rebased and ready for review.

fitzgen (commented):
Thanks! Can you also add a test that otherwise-valid shared tables are rejected when the shared-everything-proposal is not enabled? See this dir: https://github.com/bytecodealliance/wasm-tools/tree/main/tests/local/missing-features

tests/local/shared-everything-threads/table.wast
```diff
@@ -0,0 +1,37 @@
...
+
+;; The proposal creates too much ambiguity to allow this syntax: the parser
+;; would need to lookahead multiple tokens.
+(assert_malformed
+  (module quote "(table shared i64 (ref null (shared func)) (elem (ref.null (shared func))))")
+  "unexpected token")
+
+(assert_invalid
+  (module (table (import "spectest" "table_ref") shared 0 funcref))
+  "shared tables must have a shared element type")
```
alexcrichton, line 37: Mind adding a negative test for the case of an indexed type too, such as `(ref $t)`?

alexcrichton (approved):

PR #1652, opened 2024-07-03
added eq, partialeq trait for operator equality comparison

thesuhas:
Hi,

I have a use case where I would need to compare different `Operator` enum instances in `wasmparser` to see if they are the same one. It helps me traverse a `.wat` file and identify if a particular `Operator` that I am at is the desired one.

I've added the `Eq` and `PartialEq` trait to `Operator` and all its dependencies to enable this in a clean and easy way.

### DIFF
--- a/crates/wasmparser/src/binary_reader.rs
+++ b/crates/wasmparser/src/binary_reader.rs
@@ -101,7 +101,7 @@ impl BinaryReaderError {
 }
 
 /// A binary reader of the WebAssembly structures and types.
-#[derive(Clone, Debug, Hash, Eq, PartialEq)]
+#[derive(Clone, Debug, Hash)]
 pub struct BinaryReader<'a> {
     buffer: &'a [u8],
     position: usize,
--- a/crates/wasmparser/src/binary_reader.rs
+++ b/crates/wasmparser/src/binary_reader.rs
@@ -101,7 +101,7 @@ impl BinaryReaderError {
 }
 
 /// A binary reader of the WebAssembly structures and types.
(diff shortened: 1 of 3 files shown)


alexcrichton (commented):
Thanks! The one type I'd hesitate on though is the `BinaryReader` type, Could that be removed instead? If that only affects `BrTable` could the `BrTable` type have a manual `Eq` implementation?

thesuhas:
I can do that! To implement `Eq` for `BrTable`, how would I compare the `BinaryReader`? Additionally, I would assume you would not want `PartialEq` derived for the `BinaryReader` as well?

alexcrichton:
Ideally two different readers would be semantically compared, so you'd for example use `.targets()` and equate the two. That being said it'd have to deal with errors and is kinda mess. In lieu of that I'd recommend equating `.remaining_bytes()` or basically the slice contents of the reader

thesuhas:
I've done that! Used `remaining_buffer()` to compare the slices of the contents of the reader. Would this work?

alexcrichton (approved):
Looks good to me, thanks!

thesuhas:
Could this be made into a release so that I can use it?

alexcrichton:
Wanted to comment that I haven't forgotten this, I just wanted to sort out a few other things before a release. There's no easy way to make patch releases at this time so this involves major version bumps which requires a bit more coordination.

alexcrichton:
Ok I've turned the crank over at https://github.com/bytecodealliance/wasm-tools/pull/1663

PR #1654, opened 2024-07-04
wit-parser: Print available worlds in the “multiple worlds” error message

primoly:
Regarding: https://github.com/bytecodealliance/wit-bindgen/issues/996


alexcrichton (approved):
Thanks!

PR #1655, opened 2024-07-08
added conversions for component specific enums

thesuhas:
As per [this](https://github.com/bytecodealliance/wasm-tools/discussions/1639) discussion, I've added some methods in the `Reencode` trait that converts enums specific to components

### DIFF
--- a/crates/wasm-encoder/src/reencode.rs
+++ b/crates/wasm-encoder/src/reencode.rs
@@ -69,6 +69,45 @@ pub trait Reencode {
         utils::component_primitive_val_type(self, ty)
     }
 
+    fn component_export_kind(
+        &mut self,
+        ty: wasmparser::ComponentExternalKind,
+    ) -> crate::component::ComponentExportKind {
+        utils::component_export_kind(self, ty)
+    }
+
+    fn component_outer_alias_kind(
+        &mut self,
+        kind: wasmparser::ComponentOuterAliasKind,
+    ) -> crate::component::ComponentOuterAliasKind {
+        utils::component_outer_alias_kind(self, kind)
+    }
+
+    fn component_val_type(
+        &mut self,
+        ty: wasmparser::ComponentValType,
+    ) -> crate::component::ComponentValType {
+        utils::component_val_type(self, ty)
+    }
+
+    fn type_bounds(&mut self, ty: wasmparser::TypeBounds) -> crate::component::TypeBounds {
+        utils::type_bounds(self, ty)
+    }
+
+    fn canonical_option(
+        &mut self,
+        ty: wasmparser::CanonicalOption,
+    ) -> crate::component::CanonicalOption {
+        utils::canonical_option(self, ty)
+    }
+
+    fn component_type_ref(
+        &mut self,
+        ty: wasmparser::ComponentTypeRef,
+    ) -> crate::component::ComponentTypeRef {
+        utils::component_type_ref(self, ty)
+    }
+
     fn const_expr(
         &mut self,
         const_expr: wasmparser::ConstExpr,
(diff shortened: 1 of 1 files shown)


crates/wasm-encoder/src/reencode.rs
```diff
@@ -806,6 +845,110 @@ pub mod utils {
...
+            }
+        }
+    }
+
+    pub fn component_val_type<T: ?Sized + Reencode>(
+        _reencoder: &mut T,
+        ty: wasmparser::ComponentValType,
+    ) -> crate::component::ComponentValType {
+        match ty {
+            wasmparser::ComponentValType::Type(u) => crate::component::ComponentValType::Type(u),
```
alexcrichton, line 887: Could you add a new method such as `reencoder.component_type_index(u)` here to allow mapping indices like other core wasm indices can be mapped?

crates/wasm-encoder/src/reencode.rs
```diff
@@ -806,6 +845,110 @@ pub mod utils {
...
+            }
+            wasmparser::ComponentTypeRef::Type(bounds) => {
+                crate::component::ComponentTypeRef::Type(crate::component::TypeBounds::from(bounds))
+            }
+            wasmparser::ComponentTypeRef::Instance(u) => {
+                crate::component::ComponentTypeRef::Instance(u)
+            }
+            wasmparser::ComponentTypeRef::Component(u) => {
+                crate::component::ComponentTypeRef::Component(u)
+            }
```
alexcrichton, line 928: Similar to my comment above, this is another location that would be good to allow mapping the indices in addition to just changing the wasmparser/wasm-encoder type structure

crates/wasm-encoder/src/reencode.rs
thesuhas: Can't we just reuse `reencoder.type_index(u)`? It just takes a `u32` and gives out a `u32` which is all we need right?

alexcrichton: Components have separate index spaces for core wasm types and component types, so this will want to add a similar method but with a different name to disambiguate that this is for the component type index space.

thesuhas: That makes sense! Will, add!

thesuhas:
Added the `reencoder.component_type_index()` method!

crates/wasm-encoder/src/reencode.rs
```diff
@@ -806,6 +849,114 @@ pub mod utils {
...
+            wasmparser::CanonicalOption::CompactUTF16 => {
+                crate::component::CanonicalOption::CompactUTF16
+            }
+            wasmparser::CanonicalOption::Memory(u) => crate::component::CanonicalOption::Memory(u),
+            wasmparser::CanonicalOption::Realloc(u) => {
+                crate::component::CanonicalOption::Realloc(u)
+            }
+            wasmparser::CanonicalOption::PostReturn(u) => {
+                crate::component::CanonicalOption::PostReturn(u)
+            }
```
alexcrichton, line 956: Like `component_type_index` above these will want to use the appropriate methods for mapping indices here. For memory that'll be `memory_index` and the others are `func_index`

crates/wasm-encoder/src/reencode.rs
```diff
@@ -806,6 +849,114 @@ pub mod utils {
...
+        ty: wasmparser::TypeBounds,
+    ) -> crate::component::TypeBounds {
+        match ty {
+            wasmparser::TypeBounds::Eq(u) => crate::component::TypeBounds::Eq(u),
+            wasmparser::TypeBounds::SubResource => crate::component::TypeBounds::SubResource,
+        }
+    }
+
+    pub fn component_type_ref<T: ?Sized + Reencode>(
+        _reencoder: &mut T,
```
alexcrichton, line 913: Mind renaming this to `reencoder` now that it's used and no longer needs a leading `_`?

crates/wasm-encoder/src/reencode.rs
```diff
@@ -806,6 +849,114 @@ pub mod utils {
...
+                crate::component::ComponentTypeRef::Func(_reencoder.component_type_index(u))
+            }
+            wasmparser::ComponentTypeRef::Value(valty) => {
+                crate::component::ComponentTypeRef::Value(crate::component::ComponentValType::from(
+                    valty,
+                ))
+            }
+            wasmparser::ComponentTypeRef::Type(bounds) => {
+                crate::component::ComponentTypeRef::Type(crate::component::TypeBounds::from(bounds))
+            }
```
alexcrichton, line 930: Instead of using `from` here these should use methods on `reencoder` to use the state specified there to map indices as necessary

crates/wasm-encoder/src/reencode.rs
```diff
@@ -806,6 +849,114 @@ pub mod utils {
...
+        }
+    }
+
+    pub fn type_bounds<T: ?Sized + Reencode>(
+        _reencoder: &mut T,
+        ty: wasmparser::TypeBounds,
+    ) -> crate::component::TypeBounds {
+        match ty {
+            wasmparser::TypeBounds::Eq(u) => crate::component::TypeBounds::Eq(u),
```
alexcrichton, line 907: This'll need a `component_type_index` mapping here too

crates/wasm-encoder/src/reencode.rs
```diff
@@ -806,6 +849,114 @@ pub mod utils {
...
+                crate::component::ComponentOuterAliasKind::Type
+            }
+            wasmparser::ComponentOuterAliasKind::Component => {
+                crate::ComponentOuterAliasKind::Component
+            }
+        }
+    }
+
+    pub fn component_val_type<T: ?Sized + Reencode>(
+        _reencoder: &mut T,
```
alexcrichton, line 887: Similar to below could the leading `_` be removed here?

thesuhas:
Fixed all the issues you pointed out!

alexcrichton (approved):

PR #1656, opened 2024-07-08
Extend re-encode functionality to support components

itowlson:
This is WIP at the moment (worked in one simple test case but has certainly not been properly tested!).

### DIFF
--- a/crates/wasm-encoder/src/reencode.rs
+++ b/crates/wasm-encoder/src/reencode.rs
@@ -95,6 +95,14 @@ pub trait Reencode {
         utils::function_index(self, func)
     }
 
+    fn component_index(&mut self, func: u32) -> u32 {
+        utils::component_index(self, func)
+    }
+
+    fn value_index(&mut self, func: u32) -> u32 {
+        utils::value_index(self, func)
+    }
+
     fn global_index(&mut self, global: u32) -> u32 {
         utils::global_index(self, global)
     }
@@ -113,6 +121,27 @@ pub trait Reencode {
 
     fn type_index(&mut self, ty: u32) -> u32 {
         utils::type_index(self, ty)
+    }
+
+    fn resource_index(&mut self, ty: u32) -> u32 {
+        utils::resource_index(self, ty)
+    }
+
+    fn instance_index(&mut self, index: u32) -> u32 {
+        utils::instance_index(self, index)
+    }
+
+    fn module_index(&mut self, index: u32) -> u32 {
+        utils::module_index(self, index)
+    }
+
+    fn component_alias_outer_index(
+        &mut self,
+        kind: wasmparser::ComponentOuterAliasKind,
+        count: u32,
+        index: u32,
+    ) -> u32 {
+        utils::component_alias_outer_index(self, kind, count, index)
     }
 
     fn abstract_heap_type(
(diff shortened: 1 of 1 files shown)


crates/wasm-encoder/src/reencode.rs
```diff
@@ -806,6 +1226,127 @@ pub mod utils {
...
+        reencoder: &mut T,
+        alias: wasmparser::ComponentAlias<'a>,
+    ) -> crate::Alias<'a> {
+        match alias {
+            wasmparser::ComponentAlias::CoreInstanceExport {
+                kind,
+                instance_index,
+                name,
+            } => crate::Alias::CoreInstanceExport {
+                instance: instance_index,
```
alexcrichton, line 1297: I had some similar comments on https://github.com/bytecodealliance/wasm-tools/pull/1655, but for indices like this could you also add a methods such as `component_instance_index` which default is a noop but allows the reencode trait to remap the indices if desired?

crates/wasm-encoder/src/reencode.rs
```diff
@@ -528,6 +684,261 @@ impl Reencode for RoundtripReencoder {
...
+                            }
+                            wasmparser::Instance::FromExports(exports) => {
+                                let ee = exports
+                                    .iter()
+                                    .map(|e| (e.name, reencoder.export_kind(e.kind), e.index));
+                                is.export_items(ee);
+                            }
+                        };
+                    }
+                    component.section(&is);
```
alexcrichton, line 785: Mind splitting these section handlers into components such as:

* A `parse_instance_section` method which takes the wasmparser-based reader
* A `From` and conversion method such as `parse_instance` for each item in the section

That'll help keep these sorts of impls a bit neater and additionally enable implementors of `Reencode` to be able to customize the smallest portion available (and or call the "biggest" portion they want)

crates/wasm-encoder/src/reencode.rs
itowlson: How sophisticated should this be?  For example, in the case of `wasmparser::ComponentAlias::Outer`, the index seems like it refers to a different type of item depending on `kind` (and, per `count`, in a different scope, if that would affect the transformation) - should I therefore pass `kind` and `count` into the remapping function?  Sorry, I'm flying a bit blind here in terms of familiarity with the underlying concepts...!

Another example is cases such as `TypeBounds` where we are dealing with enums of indexes.  E.g. for TypeBounds the reencode logic is:

```
        match bounds {
            wasmparser::TypeBounds::Eq(index) => crate::TypeBounds::Eq(index),
            wasmparser::TypeBounds::SubResource => crate::TypeBounds::SubResource,
        }
```

Do I need to offer a separate hook to remap the index in cases like this, or would a consumer override the entire logic?

juntyr:
Thank you for working on this!

I feel like the module and trait are becoming a bit big. Perhaps the core module and component functionalities in the untold module could be split into two submodules? If all methods go into the same trait, having a consistent prefix for core vs component functions might be useful. If they could be split, would a subtrait for component encoding make sense?

itowlson:
Thanks @juntyr - there may be something to be said for that.  Right now, however, I'm not sure if I'm going to be pushing this to completion, as I'm now looking at a different approach to the problem I was trying to solve.  I'm not yet ready to close it for definite, though, and if I do pick it up again I'll explore that suggestion for sure!

alexcrichton:
I'm attempting to help push this over the finish line in https://github.com/bytecodealliance/wasm-tools/pull/1722

Issue #1653, opened 2024-07-04
Maybe bring back the deprecated "try-catch-catch_all" opcodes in wasmparser::validator after the "rust-feature" flag?


trzeciak:
I would like to use exceptions in wasm code in the wild. But I'm afraid that the new approach `exnref` is not yet well supported by my clients side, so I would like to use legacy exception proposal (for a while).

I also want to use sentry as crash-tracker, but I have a problem with my wasm file on sentry backend (https://github.com/getsentry/symbolic/issues/850), because getsentry/symbolic project use wasmparser::validator, but latest wasmparser are already deprecated those opcodes (https://github.com/bytecodealliance/wasm-tools/pull/1333).

I was wondering if it would be possible to handle this on the symbolic side, but now it occurred to me that maybe it could be brought back in this project and hidden behind a rust-flag?

As a test, I restored this code and confirm that it solve this problem: https://github.com/trzeciak/wasm-tools/commit/871dbcba98ad05dc3ec55d1ed91a948a2b0f9c35

**The question:**

If I prepared a PR with restoring this code in the validator, behind a flag (say `"legacy-exception-handling"`), would this be acceptable?

alexcrichton:
Thanks for the report! I think it would be reasonable to bring back these opcodes as well if you're interested in it. Your commit looks good to me, and some comment I might have would be:

* The old opcodes should be moved under a new `WasmFeatures` like you mention which would be likely permanently off-by-default
* Could you add back some of the tests to ensure that the validator works?

One caveat is that I'd like to personally keep the s-expression-forms removed since they were somewhat cumbersome to maintain. Is that ok with you? It means that the text-to-binary transform wouldn't support `(try (do ..))` for example, you'd have to use `try ... catch_all ... end` (or something like that I forget the exact opcodes, basically you couldn't use parens in the text format)

trzeciak:
Hi, I agree with the comments, I will include them in the PR draft.  

I'm not sure I understand the second part of your statement xD (I think):
* I don't need conversion from wasm-binary-format to textual-representation, or even from textual to binary. My code with business logic is in C++, I use emscripten to compile, it uses llvm and wasm-binary-format is created.
* I pass the wasm-binary file (with debug symbols) to sentry. If I'm not mistaken, the [getsentry/symbolic](https://github.com/getsentry/symbolic) code doesn't use binary to textual conversion either (I checked briefly, but I don't see a logical reason to do this).

[...]

PR #1661, opened 2024-07-11
Bring back legacy exception handling opcodes

trzeciak:
In this MR I bring back legacy exception handling opcodes, in reference to this report: https://github.com/bytecodealliance/wasm-tools/issues/1653.

@alexcrichton Two things:
- [x] I have prepared the simplest wasm file possible which uses all the deprecated opcodes (see `check_legacy_exception_handling_opcodes` test function),
could you please help me run the validator with this data?

- [x] Since this is behind the `legacy-exception-handling` feature, tests should be run with the `cargo test --all-features` flag, or I can add this feature to default list.
Given this, how should I do it to make sense?


crates/wasmparser/Cargo.toml
```diff
@@ -69,3 +70,8 @@ validate = [
 # Enable Serialize/Deserialize implementations for types in
 # `wasmparser::collections`
 serde = ['dep:serde', 'indexmap/serde', 'hashbrown/serde']
+
+# A feature that enables legacy exception handling proposal (phase 1)
+# support this feature as long as all leading browsers also support it
+# https://github.com/WebAssembly/exception-handling/blob/main/proposals/exception-handling/legacy/Exceptions.md
+legacy-exception-handling = []
```
alexcrichton, line 77: Ah I apologize I fear I may have misled you. Currently wasmparser doesn't have other Cargo features for wasm proposals so there's not a lot of infrastructure to support this and as you're discovered there's bugs and things to shake out. Given all that it's ok to remove this and have the support always be compiled-in-by-default.

What I moreso meant is that we'll want a new feature gate in `WasmFeatures` which is turned off at runtime by default but support for it is compiled in by default.

crates/wasmparser/src/validator/operators.rs
```diff
@@ -121,6 +121,28 @@ pub enum FrameKind {
...
     /// This belongs to the Wasm exception handling proposal.
     TryTable,
+
+    #[cfg(feature = "legacy-exception-handling")]
+    /// A Wasm `deprecated-try` control block.
+    ///
+    /// # Note
+    ///
+    /// This belongs to the now deprecated Wasm exception handling proposal from Phase 1.
+    DeprecatedTry,
```
alexcrichton, line 131: Mind calling this "Legacy" instead of "Deprecated"?

Also mind updating the docs here to modernize what it points to?

crates/wasmparser/tests/legacy-exception.rs
```diff
@@ -0,0 +1,26 @@
...
+              end
+            delegate 0 (;@2;)
+          catch 0
+          end
+        )
+        (tag (;0;) (type 0))
+      )";
+
+    // TODO: Use `_binary_data` or `_text_format_data`
+}
```
alexcrichton, line 26: I'd recommend adding this tests as `tests/local/legacy-exceptions.wast` where you could add a variety of modules as necessary.

crates/wasmparser/Cargo.toml
trzeciak: Updated ✅

crates/wasmparser/src/validator/operators.rs
trzeciak: Updated ✅

crates/wasmparser/tests/legacy-exception.rs
trzeciak: I don't fully understand this, I will take a closer look at it and let you know or ask for help.

trzeciak: Updated ✅

trzeciak:
@alexcrichton Okay, I changed 'cargo feature' to WasmFeatures and I added simple wat test file. I also changed some names and description, I think it's ready for review.

crates/wasmparser/src/validator/operators.rs
```diff
@@ -4124,6 +4128,115 @@ where
         self.pop_operand(Some(ValType::Ref(RefType::I31REF)))?;
         self.push_operand(ValType::I32)
     }
+    fn visit_try(&mut self, mut ty: BlockType) -> Self::Output {
+        if !self.features.legacy_exceptions() {
```
alexcrichton, line 4132: Oh this is automatically checked via the macro above so you can skip the check on `self.features` here and below

alexcrichton (approved):

crates/wasmparser/src/validator/operators.rs
trzeciak: Hm.. then these tests fail:

[test-01.log](https://github.com/user-attachments/files/16195427/test-01.log)

trzeciak: Maybe it is because now, to parse legacy exceptions we need to set all or two flags:
```sh
❯ cargo build
❯ ./target/release/wasm-tools validate -fexception-handling,legacy-exceptions tests/local/legacy-exceptions.wat
```

alexcrichton: Ah for that you'll need to update [this exception list](https://github.com/bytecodealliance/wasm-tools/blob/021389c149dc7fa0c0bac97c0bd84ffe4986d889/tests/roundtrip.rs#L646-L663)

alexcrichton: You additionally might need to [remove legacy-exceptions from the default list](https://github.com/bytecodealliance/wasm-tools/blob/021389c149dc7fa0c0bac97c0bd84ffe4986d889/tests/roundtrip.rs#L578-L582) and then add a [special case](https://github.com/bytecodealliance/wasm-tools/blob/021389c149dc7fa0c0bac97c0bd84ffe4986d889/tests/roundtrip.rs#L627-L629) where it is enabled.

trzeciak: Should I also update https://github.com/bytecodealliance/wasm-tools/blob/021389c149dc7fa0c0bac97c0bd84ffe4986d889/crates/wasmparser/src/lib.rs#L161-L165 to this:
```
// Deprecated old instructions from the exceptions proposal
@legacy_exceptions Try { blockty: $crate::BlockType } => visit_try
@legacy_exceptions Catch { tag_index: u32 } => visit_catch
@legacy_exceptions Rethrow { relative_depth: u32 } => visit_rethrow
@legacy_exceptions Delegate { relative_depth: u32 } => visit_delegate
@legacy_exceptions CatchAll => visit_catch_all
```
?

alexcrichton: Oh sorry yes definitely!

trzeciak: After add `"legacy-exceptions.wat" => features.insert(WasmFeatures::LEGACY_EXCEPTIONS),` and other fixes, some tests still failed:
[test-02.log](https://github.com/user-attachments/files/16195973/test-02.log)

trzeciak: To be clear, I pushed all changes now (tests not pass locally).

alexcrichton: That's [this comment](https://github.com/bytecodealliance/wasm-tools/pull/1661#discussion_r1676024130), notably updating [this list](https://github.com/bytecodealliance/wasm-tools/blob/021389c149dc7fa0c0bac97c0bd84ffe4986d889/tests/roundtrip.rs#L646-L663)

alexcrichton (approved):

PR #1662, opened 2024-07-11
added from floating point to ieee trait

thesuhas:
I see that in `wasmparser`, you can convert from `Ieee32` to `f32` using the `From` trait. And in `wasm_encoder`, you can encode `f32`.
I have a use case where I'm maintaining an Intermediate Representation between the parser and encoder in `wasmparser` types and specifically need to create `wasmparser::Operator`. This trait enables me to convert floating point numbers to `wasmparser::Operator` floating point constants.


alexcrichton (approved):
Thanks!

PR #1664, opened 2024-07-12
threads: add remaining instructions

abrown:
With #1649 merged, we can now implement the remaining shared-everything-threads [instructions]. Many of the tests are generated, so there may be tweaks coming in the future; this change just gets the initial encoding, decoding, and validation pieces in place.

[instructions]: https://github.com/WebAssembly/shared-everything-threads/blob/main/proposals/shared-everything-threads/Overview.md#instructions

### DIFF
--- a/crates/wasmparser/src/validator/operators.rs
+++ b/crates/wasmparser/src/validator/operators.rs
@@ -28,7 +28,7 @@ use crate::{
     Result, StorageType, StructType, SubType, TableType, TryTable, UnpackedIndex, ValType,
     VisitOperator, WasmFeatures, WasmModuleResources, V128,
 };
-use crate::{prelude::*, CompositeInnerType};
+use crate::{prelude::*, CompositeInnerType, Ordering};
 use core::ops::{Deref, DerefMut};
 
 pub(crate) struct OperatorValidator {
@@ -740,17 +740,6 @@ where
         Ok(index_ty)
     }
 
-    /// Validates that the `table` is valid and returns the type it points to.
-    fn check_table_index(&self, table: u32) -> Result<TableType> {
-        match self.resources.table_at(table) {
-            Some(ty) => Ok(ty),
-            None => bail!(
-                self.offset,
-                "unknown table {table}: table index out of bounds"
-            ),
-        }
-    }
-
     fn check_floats_enabled(&self) -> Result<()> {
         if !self.features.floats() {
             bail!(self.offset, "floating-point instruction disallowed");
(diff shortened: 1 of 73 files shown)


crates/wasmparser/src/validator/operators.rs
```diff
@@ -1155,6 +1144,40 @@ where
...
+    /// Common helper for checking the types of arrays accessed with atomic RMW
+    /// instructions, which only allow `i32` and `i64`.
+    fn check_atomic_array_rmw_ty(&self, type_index: u32) -> Result<ValType> {
+        let elem_ty = self.array_type_at(type_index)?.0.element_type;
+        match elem_ty {
+            StorageType::Val(ValType::I32) => Ok(ValType::I32),
+            StorageType::Val(ValType::I64) => Ok(ValType::I64),
+            _ => bail!(
+                self.offset,
+                "invalid type: `struct.atomic.rmw.*` only allows `i32` and `i64`"
```
alexcrichton, line 1176: copy/paste with `struct.atomic.*` here, but perhaps this and the above method could share more implementation details?

crates/wasmparser/src/validator/operators.rs
```diff
@@ -3757,6 +3902,149 @@ where
...
+        _ordering: Ordering,
+        struct_type_index: u32,
+        field_index: u32,
+    ) -> Self::Output {
+        let field_ty = self.check_atomic_struct_rmw_ty(struct_type_index, field_index)?;
+        self.pop_operand(Some(field_ty))?;
+        self.pop_concrete_ref(true, struct_type_index)?;
+        self.push_operand(field_ty)?;
+        Ok(())
+    }
```
alexcrichton, line 3990: Could these method bodies be refactored into a shared function?

crates/wasmparser/src/validator/operators.rs
```diff
@@ -3989,6 +4333,99 @@ where
...
+    fn visit_array_atomic_rmw_xor(&mut self, _ordering: Ordering, type_index: u32) -> Self::Output {
+        let elem_ty = self.check_atomic_array_rmw_ty(type_index)?;
+        self.pop_operand(Some(elem_ty))?;
+        self.pop_operand(Some(ValType::I32))?;
+        self.pop_concrete_ref(true, type_index)?;
+        self.push_operand(elem_ty)?;
+        Ok(())
+    }
```
alexcrichton, line 4375: Like structs could these be moved into a shared function?

crates/wasmparser/src/validator/operators.rs
abrown: Fixed the typo and looked at merging them: the only part that is the same is the i32/i64 check. The other parts--type access and push/pop logic--ends up being different.

PR #1665, opened 2024-07-12
Handle multiple versions during `wasm-tools component wit`

karthik2804:
This commits makes sure that if a component depends on multiple versions of a package that it is handled correctly. It fixes the behavior by appending the version string in case there are multiple packages with the same name and namespaces specifically when there is only one namespace for the given package name.


src/bin/wasm-tools/component.rs
```diff
@@ -648,11 +648,20 @@ impl WitOpts {
...
                     let stem = if packages_with_same_name.len() == 1 {
-                        pkg.name.name.clone()
+                        if packages_with_same_namespace == 1 {
+                            pkg.name.name.clone()
+                        } else {
+                            let version_suffix = pkg
+                                .name
+                                .version
+                                .as_ref()
+                                .map(|ver| format!("@{}", ver))
```
rylev, line 660: Nit: we're formatting here and below. If there's a way to avoid the extra format! call that would be nice.

rylev (approved):

src/bin/wasm-tools/component.rs
karthik2804: updated to format only if version exists.

alexcrichton:
Thanks! Could you add at test for this as well? I think the test could look [like this](https://github.com/bytecodealliance/wasm-tools/blob/main/tests/cli/wit-directory-output-valid.wit) with multiple packages in a single WIT file and if you need a component itself to reproduce then [this test](https://github.com/bytecodealliance/wasm-tools/blob/main/tests/cli/wit-directory-output-in-deps-folder.wit) could one to copy.

karthik2804:
@alexcrichton I have added the test.

alexcrichton (approved):

PR #1668, opened 2024-07-12
Refactor handling of `--features` on the CLI

alexcrichton:
This commit uses the `bitflags::Flags` trait to refactor the parsing of the `--features` CLI flag to avoid needing to manually list out wasm features. This is something I at least always personally forget during reviews and such so it should make the CLI support more robust.

### DIFF
--- a/src/bin/wasm-tools/validate.rs
+++ b/src/bin/wasm-tools/validate.rs
@@ -1,5 +1,6 @@
 use addr2line::LookupResult;
 use anyhow::{anyhow, Context, Result};
+use bitflags::Flags;
 use rayon::prelude::*;
 use std::fmt::Write;
 use std::mem;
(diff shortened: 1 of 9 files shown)


src/bin/wasm-tools/validate.rs
```diff
@@ -211,28 +184,27 @@ fn parse_features(arg: &str) -> Result<WasmFeatures> {
         };
         match part {
             "all" => {
-                for (name, feature) in FEATURES {
-                    // don't count this under "all" for now.
-                    if *name == "deterministic" {
-                        continue;
-                    }
```
abrown, line 218: You don't need to keep this deterministic exclusion?

alexcrichton: Ah yeah this was an old feature of wasmparser that was removed awhile back and this was forgotten to be removed at the time.

PR #1675, opened 2024-07-15
[wit-parser] select default world among multiple packages

dicej:
Previously, `Resolve::select_world` would return an error if multiple packages were provided and no world was specified.  Now, we return `Ok` in this case as long as exactly one package has exactly one world.

### DIFF
--- a/crates/wit-parser/src/resolve.rs
+++ b/crates/wit-parser/src/resolve.rs
@@ -1015,24 +1015,34 @@ impl Resolve {
             }
             None => match packages {
                 [] => bail!("no packages were specified nor is a world specified"),
-                [one] => {
-                    let pkg = &self.packages[*one];
-                    match pkg.worlds.len() {
-                        0 => bail!("no worlds found in package `{}`", pkg.name),
-                        1 => return Ok(*pkg.worlds.values().next().unwrap()),
-                        _ => bail!(
-                            "multiple worlds found in package `{}`, one must be explicitly chosen:{}",
-                            pkg.name,
-                            pkg.worlds.keys().map(|name| format!("\n  {name}")).collect::<String>()
-                        ),
-                    }
-                }
                 [..] => {
-                    bail!(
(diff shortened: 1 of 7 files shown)


crates/wit-parser/src/resolve.rs
```diff
@@ -1015,24 +1015,34 @@ impl Resolve {
...
-                    bail!(
-                        "the supplied WIT source files describe multiple packages; \
-                         please provide a fully-qualified world-specifier select \
-                         a world amongst these packages"
-                    )
+                    let mut world = None;
+                    for id in packages {
+                        let pkg = &self.packages[*id];
+                        match pkg.worlds.len() {
+                            0 => bail!("no worlds found in package `{}`", pkg.name),
```
alexcrichton, line 1023: Should this perhaps `continue` rather than returning an error?

crates/wit-parser/src/resolve.rs
```diff
@@ -1015,24 +1015,34 @@ impl Resolve {
...
+                                     containing at least one world; \
+                                     please provide a fully-qualified world-specifier to select \
+                                     a world amongst these packages"
+                                );
+                            }}
+                            _ => bail!(
+                                "multiple worlds found in package `{}`, one must be explicitly chosen:{}",
+                                pkg.name,
+                                pkg.worlds.keys().map(|name| format!("\n  {name}")).collect::<String>()
+                            ),
```
alexcrichton, line 1038: Could this collect all worlds in all packages and recommend their full syntax of `foo:bar/baz` instead of only recommending the names from one package?

crates/wit-parser/src/resolve.rs
dicej: yup, good catch

dicej:
I've added a happy path test.  I'll add a couple of sad path ones, too.

dicej:
Ugh, forgot to check in a file, sorry.

alexcrichton (approved):
Thanks!

PR #1686, opened 2024-07-18
Rewrite `Resolve::merge_worlds`

### LINKED ISSUE #1685: `merge_worlds` on `wit_parser::resolve::Resolve` fails when worlds contain the same export
While trying to merge multiple worlds, I noticed worlds with the same exports could not be merged (i.e.) if 2 worlds export wasi:http/incoming-handler@0.2.0, the worlds cannot be merged. 
```
export `wasi:http/incoming-handler@0.2.0` conflicts with the previous name of `wasi:http/incoming-handler@0.2.0`
```

The expected behavior would be for the worlds to merge successfully as long as the types of exports are the same.

Here is a [repo](https://github.com/karthik2804/wit-parser-merge-issue) with an example of it

alexcrichton:
This commit updates and rewrites the `Resolve::merge_worlds` helper which is one of the workhorses of `wit-bindgen` and `wit-component`. The previous implementation is largely thrown out and replaced with:

* Any new import can always be added.
* Any preexisting import must "match" the one being added. This is done by updating `MergeMap` with some more functionality (but not everything) and handling all kinds of imports.
* Any preexisting export must "match" in the same way as imports.
* New exports can only be added if the meaning of previous exports don't change. This means that if a new export is a transitive dependency of an existing export then it can't be added.

This fixes a number of issues with the previous implementation which was written during a different time with a different structure of WIT.

Closes #1685

### DIFF
--- a/crates/wit-parser/src/resolve.rs
+++ b/crates/wit-parser/src/resolve.rs
@@ -531,7 +531,7 @@ impl Resolve {
             self.packages.len()
         );
 
-        let mut map = MergeMap::new(&resolve, &self)?;
+        let mut map = MergeMap::new(&resolve, &self);
         map.build()?;
         let MergeMap {
             package_map,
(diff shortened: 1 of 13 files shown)


pchickey (approved):
Excellent, thanks for all the comments and tests.

PR #1687, opened 2024-07-18
Implement `@deprecated` in WIT

### LINKED ISSUE #1683: Implement support for `@deprecated` gates in `wit-parser`
For the WASI 0.2 timeframe (early October), it would be great if we could land support for `@deprecated` gates (https://github.com/WebAssembly/component-model/pull/377) in WIT, so we can begin closing out issues like https://github.com/WebAssembly/wasi-clocks/issues/69 and https://github.com/WebAssembly/wasi-http/issues/107.

The syntax we're proposing for this builds on the existing gate notation we have for `@since` and `@unstable`, extending it with a new `@deprecated` notation with a required `version` field and an optional `message` field:

```ebnf
gate ::= unstable-gate
       | since-gate
       | deprecated-gate

unstable-gate ::= '@unstable' '(' feature-field ')'
since-gate ::= '@since' '(' version-field ( ',' feature-field )? ')'
deprecated-gate ::= '@deprecated' '(' version-field ( ',' message-field )? ')'

feature-field ::= 'feature' '=' id
version-field ::= 'version' '=' <valid semver>
message-field ::= 'message' '=' <core:string>
```

I believe that https://github.com/bytecodealliance/wasm-tools/pull/1508 probably did most of the heavy lifting required to implement gates, introducing the concept of extensible metadata to the parser. I'm hoping that adding this feature will be made a lot simpler because of that; perhaps as easy as extending the parser and adding a new entry in the metadata.

On the component model side we're again waiting for feedback from at least one implementation before landing the extension. It would be great if this project could again be that implementation. Thanks!

alexcrichton:
This is intended to correspond with WebAssembly/component-model#377.

Closes #1683

### DIFF
--- a/crates/wit-component/src/printing.rs
+++ b/crates/wit-component/src/printing.rs
@@ -903,7 +903,11 @@ impl WitPrinter {
     fn print_stability(&mut self, stability: &Stability) {
         match stability {
             Stability::Unknown => {}
-            Stability::Stable { since, feature } => {
+            Stability::Stable {
+                since,
+                feature,
+                deprecated,
+            } => {
                 self.output.push_str("@since(version = ");
                 self.output.push_str(&since.to_string());
                 if let Some(feature) = feature {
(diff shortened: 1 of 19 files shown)


pchickey (approved):

yoshuawuyts:
I believe we've figured out everything we need to in https://github.com/WebAssembly/component-model/pull/377 with respect to the design, which I believe matches the current implementation here!

PR #1689, opened 2024-07-18
fix(wit-parser): improve checking for stable feature gates

vados-cosmonic:
This commit introduces some extra checking (and panicking) for feature gates that are stable (i.e. `Stability::Stable`) AKA `@since` with a version and/or features specified.

There are two primary changes:
- Make referring to a *future* version of the package an error (for now)
- Ensure that if a feature is specified that it is checked before inclusion

In the past `Stability::Stable` feature gates were simply treated as `Unknown`, which hides the information necessary for downstream crates/consumers to use (and created the question around whether referring to future package versions is valid).

See also the [relevant Zulip discussion](https://bytecodealliance.zulipchat.com/#narrow/stream/223391-wasm/topic/Missing.20impl.20for.20.40since.20in.20.60Resolve.3A.3Ainclude_stability.28.29.60/near/452175637)

### DIFF
--- a/crates/wit-parser/src/resolve.rs
+++ b/crates/wit-parser/src/resolve.rs
@@ -1,3 +1,13 @@
+use std::collections::{BTreeMap, HashMap, HashSet};
+use std::mem;
+use std::path::{Path, PathBuf};
+
+use anyhow::{anyhow, bail, ensure, Context, Result};
+use id_arena::{Arena, Id};
+use indexmap::{IndexMap, IndexSet};
+#[cfg(feature = "serde")]
+use serde_derive::Serialize;
+
 use crate::ast::lex::Span;
 use crate::ast::{parse_use_path, ParsedUsePath};
 #[cfg(feature = "serde")]
@@ -8,14 +18,6 @@ use crate::{
     TypeOwner, UnresolvedPackage, UnresolvedPackageGroup, World, WorldId, WorldItem, WorldKey,
     WorldSpan,
 };
-use anyhow::{anyhow, bail, Context, Result};
-use id_arena::{Arena, Id};
-use indexmap::{IndexMap, IndexSet};
-#[cfg(feature = "serde")]
-use serde_derive::Serialize;
-use std::collections::{BTreeMap, HashMap, HashSet};
-use std::mem;
-use std::path::{Path, PathBuf};
 
 /// Representation of a fully resolved set of WIT packages.
 ///
@@ -1408,13 +1410,37 @@ impl Resolve {
         }
     }
 
-    fn include_stability(&self, stability: &Stability) -> bool {
-        match stability {
-            Stability::Stable { .. } | Stability::Unknown => true,
+    fn include_stability(&self, stability: &Stability, pkg_id: &PackageId) -> Result<bool> {
+        Ok(match stability {
+            Stability::Unknown => true,
+            // NOTE: deprecations are intentionally omitted -- an existing `@since` takes precedence over `@deprecated`
+            Stability::Stable { since, .. } => {
+                let Some(p) = self.packages.get(*pkg_id) else {
+                    // We can't check much without a package (possibly dealing with an item in an `UnresolvedPackage`),
+                    // @since version & deprecations can't be checked because there's no package version to compare to.
+                    //
+                    // Feature requirements on stabilized features are ignored in resolved packages, so we do the same here.
+                    return Ok(true);
+                };
+
+                // Use of feature gating with version specifiers inside a package that is not versioned is not allowed
+                let package_version = p.name.version.as_ref().with_context(|| format!("package [{}] contains a feature gate with a version specifier, so it must have a version", p.name))?;
+
+                // If the version on the feature gate is:
+                // - released, then we can include it
+                // - unreleased, then we must check the feature (if present)
+                ensure!(
+                    since <= package_version,
+                    "feature gate cannot reference unreleased version {since} of package [{}] (current version {package_version})",
+                    p.name
+                );
+
+                true
+            }
             Stability::Unstable { feature, .. } => {
                 self.features.contains(feature) || self.all_features
             }
-        }
+        })
     }
 }
 
@@ -1503,7 +1529,16 @@ impl Remap {
             .zip(&unresolved.type_spans)
             .skip(foreign_types)
         {
-            if !resolve.include_stability(&ty.stability) {
+            if !resolve
+                .include_stability(&ty.stability, &pkgid)
+                .with_context(|| {
+                    format!(
+                        "failed to process feature gate for type [{}] in package [{}]",
+                        ty.name.as_ref().map(String::as_str).unwrap_or("<unknown>"),
+                        resolve.packages[pkgid].name,
+                    )
+                })?
+            {
                 self.types.push(None);
                 continue;
             }
@@ -1544,13 +1579,26 @@ impl Remap {
             .zip(&unresolved.interface_spans)
             .skip(foreign_interfaces)
         {
-            if !resolve.include_stability(&iface.stability) {
+            if !resolve
+                .include_stability(&iface.stability, &pkgid)
+                .with_context(|| {
+                    format!(
+                        "failed to process feature gate for interface [{}] in package [{}]",
+                        iface
+                            .name
+                            .as_ref()
+                            .map(String::as_str)
+                            .unwrap_or("<unknown>"),
+                        resolve.packages[pkgid].name,
+                    )
+                })?
+            {
                 self.interfaces.push(None);
                 continue;
             }
-            self.update_interface(resolve, &mut iface, Some(span))?;
             assert!(iface.package.is_none());
             iface.package = Some(pkgid);
+            self.update_interface(resolve, &mut iface, Some(span))?;
             let new_id = resolve.interfaces.alloc(iface);
             assert_eq!(self.interfaces.len(), id.index());
             self.interfaces.push(Some(new_id));
@@ -1589,11 +1637,19 @@ impl Remap {
             .zip(unresolved.world_spans)
             .skip(foreign_worlds)
         {
-            if !resolve.include_stability(&world.stability) {
+            if !resolve
+                .include_stability(&world.stability, &pkgid)
+                .with_context(|| {
+                    format!(
+                        "failed to process feature gate for world [{}] in package [{}]",
+                        world.name, resolve.packages[pkgid].name,
+                    )
+                })?
+            {
                 self.worlds.push(None);
                 continue;
             }
-            self.update_world(&mut world, resolve, &span)?;
+            self.update_world(&mut world, resolve, &pkgid, &span)?;
 
             let new_id = resolve.worlds.alloc(world);
             assert_eq!(self.worlds.len(), id.index());
@@ -1945,6 +2001,16 @@ impl Remap {
         spans: Option<&InterfaceSpan>,
     ) -> Result<()> {
         iface.types.retain(|_, ty| self.types[ty.index()].is_some());
+        let iface_pkg_id = iface.package.as_ref().unwrap_or_else(|| {
+            panic!(
+                "unexpectedly missing package on interface [{}]",
+                iface
+                    .name
+                    .as_ref()
+                    .map(String::as_str)
+                    .unwrap_or("<unknown>"),
+            )
+        });
 
         // NB: note that `iface.doc` is not updated here since interfaces
         // haven't been mapped yet and that's done in a separate step.
@@ -1954,17 +2020,31 @@ impl Remap {
         if let Some(spans) = spans {
             assert_eq!(iface.functions.len(), spans.funcs.len());
         }
-        for (i, (_, func)) in iface.functions.iter_mut().enumerate() {
-            if !resolve.include_stability(&func.stability) {
+        for (i, (func_name, func)) in iface.functions.iter_mut().enumerate() {
+            if !resolve
+                .include_stability(&func.stability, iface_pkg_id)
+                .with_context(|| {
+                    format!(
+                        "failed to process feature gate for function [{func_name}] in package [{}]",
+                        resolve.packages[*iface_pkg_id].name,
+                    )
+                })?
+            {
                 continue;
             }
             let span = spans.map(|s| s.funcs[i]);
             self.update_function(resolve, func, span)
                 .with_context(|| format!("failed to update function `{}`", func.name))?;
         }
-        iface
-            .functions
-            .retain(|_, f| resolve.include_stability(&f.stability));
+
+        // Filter out all of the existing functions in interface which fail the
+        // `include_stability()` check, as they shouldn't be available.
+        for (name, func) in mem::take(&mut iface.functions) {
+            if resolve.include_stability(&func.stability, iface_pkg_id)? {
+                iface.functions.insert(name, func);
+            }
+        }
+
         Ok(())
     }
 
@@ -2017,6 +2097,7 @@ impl Remap {
         &mut self,
         world: &mut World,
         resolve: &mut Resolve,
+        pkg_id: &PackageId,
         spans: &WorldSpan,
     ) -> Result<()> {
         // NB: this function is more more complicated than the prior versions
@@ -2047,7 +2128,16 @@ impl Remap {
                 *id = self.map_type(*id, Some(*span))?;
             }
             let stability = item.stability(resolve);
-            if !resolve.include_stability(stability) {
+            if !resolve
+                .include_stability(stability, pkg_id)
+                .with_context(|| {
+                    format!(
+                        "failed to process imported world item type [{}] in package [{}]",
+                        resolve.name_world_key(&name),
+                        resolve.packages[*pkg_id].name,
+                    )
+                })?
+            {
                 continue;
             }
             self.update_world_key(&mut name, Some(*span))?;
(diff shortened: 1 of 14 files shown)


crates/wit-parser/src/resolve.rs
```diff
@@ -1329,9 +1329,22 @@ impl Resolve {
         }
     }
 
-    fn include_stability(&self, stability: &Stability) -> bool {
+    fn include_stability(&self, stability: &Stability, pkg_id: Option<&PackageId>) -> bool {
```
alexcrichton, line 1332: It's ok to take a `PackageId` argument here directly since the optional cases below using `as_ref` should all be filled in earlier during resolve so I believe it should be safe to `.unwrap()`

crates/wit-parser/src/resolve.rs
```diff
@@ -1329,9 +1329,22 @@ impl Resolve {
...
         match stability {
-            Stability::Stable { .. } | Stability::Unknown => true,
+            Stability::Unknown => true,
+            Stability::Stable { since, feature } => {
+                if pkg_id
+                    .and_then(|id| self.packages.get(*id))
+                    .map(|p| &p.name.version)
+                    .is_some_and(|v| v.as_ref().is_some_and(|v| since > v))
+                {
+                    panic!("referring to an unreleased (future) package version in a feature gate is not allowed");
```
alexcrichton, line 1341: This'll want to use `bail!` instead to return an error to get bubbled up eventually to the original API call

crates/wit-parser/src/resolve.rs
```diff
@@ -1329,9 +1329,22 @@ impl Resolve {
...
+                }
+
+                feature.is_none()
+                    || self.all_features
+                    || feature.as_ref().is_some_and(|f| self.features.contains(f))
```
alexcrichton, line 1346: This'll want to be omitted because `since` means that it's stable so regardless of whether or not the feature is active it should be enabled. The `feature` listing is mostly just for historical reference

alexcrichton (commented):
Thanks! Could you be sure to add a test for this too? Something in `tests/cli/*` should suffice with a `// FAIL: ...` header

crates/wit-parser/src/resolve.rs
vados-cosmonic: Thanks for pointing this out -- saw some other `panic`s so naively thought it was preferred! Much prefer to `bail!`.

- [ ] remove `panic` and return `Result` from stability fn

crates/wit-parser/src/resolve.rs
vados-cosmonic: So this I'm a bit surprised by -- I was thinking of the features kind of like Rust features, as being important gates no matter the version.

So if I'm understanding right now, the progression would be something like:

- New functionality starts out as `@unstable(feature = f)`
- Once new functionality is stabilized it becomes `@since(version = x.y.z, feature = f)`
- At some point (?) the functionality just becomes `@since(version = x.y.z)` ?

Maybe it's worth making a follow up addition to the WIT docs to note the regular use cases/progression, I can do that after I'm done fixing up this PR if it's worth doing.

crates/wit-parser/src/resolve.rs
vados-cosmonic: This required more changes than expected -- signatures of some of the update fns needed to change

crates/wit-parser/src/resolve.rs
```diff
@@ -1329,11 +1340,23 @@ impl Resolve {
...
-            Stability::Stable { .. } | Stability::Unknown => true,
+    fn include_stability(&self, stability: &Stability, pkg_id: &PackageId) -> Result<bool> {
+        Ok(match stability {
+            Stability::Unknown => true,
+            Stability::Stable { since, .. } => {
+                match self.packages.get(*pkg_id) {
+                    // If the `@since` version specified on this stability is ahead of the version on the
+                    // package that it belongs to, we throw an explicit error as that is not allowed
+                    Some(p) if p.name.version.as_ref().is_some_and(|v| since > v) => {
+                        bail!("referring to an unreleased (future) package version in a feature gate is not allowed");
```
alexcrichton, line 1351: Could this error message include the two versions in question as well? (since there's no span information here so this isn't pointing to WIT files themselves).

Also, should this return an error if the package's version is `None` perhaps?

crates/wit-parser/src/resolve.rs
```diff
@@ -600,7 +600,18 @@ impl Resolve {
...
+                        format!(
+                            "unexpectedly missing package on interface [{}] (idx: {})",
+                            iface
+                                .name
+                                .as_ref()
+                                .map(String::as_str)
+                                .unwrap_or("<unknown>"),
+                            id.index(),
+                        )
+                    })?;
```
alexcrichton, line 613: For these cases the `.with_context()` here and below is pretty wordy and I think it should appropriately be `.unwrap()` because a failure here isn't a user error but rather an unrecoverable error/bug in the tool itself. Given that I think it's more appropriate to panic here if the assumption is wrong rather than returning an error

crates/wit-parser/src/resolve.rs
```diff
@@ -2040,7 +2124,21 @@ impl Remap {
...
         {
-            if !resolve.include_stability(&stability) {
+            if !resolve
+                .include_stability(&stability, pkg_id)
+                .with_context(|| {
+                    format!(
+                        "failed to process feature gate for included world [{}] in package [{}]",
+                        resolve
+                            .worlds
+                            .get(include_world)
```
alexcrichton, line 2134: It's ok to not use `.get` here since the index of `include_world` should always be valid, so indexing `[...]` syntax can be used

crates/wit-parser/src/resolve.rs
```diff
@@ -2007,7 +2082,16 @@ impl Remap {
             .zip(&spans.exports)
         {
             let stability = item.stability(resolve);
-            if !resolve.include_stability(stability) {
+            if !resolve
+                .include_stability(stability, pkg_id)
+                .with_context(|| {
+                    format!(
+                        "failed to process feature gate for exported item [{}] in package [{}]",
+                        item.name(resolve).unwrap_or("<unknown>"),
```
alexcrichton, line 2090: To avoid adding a `name` function for `WorldItem` could this use the `name` that's in scope here instead of the name on the item itself? The `name` should be sufficient for pinpointing the particular item. (e.g. there's `resolve.name_world_key` for this)

crates/wit-parser/src/resolve.rs
```diff
@@ -1873,17 +1928,27 @@ impl Remap {
...
                 .with_context(|| format!("failed to update function `{}`", func.name))?;
         }
-        iface
-            .functions
-            .retain(|_, f| resolve.include_stability(&f.stability));
+        iface.functions.retain(|_, f| {
+            resolve
+                .include_stability(&f.stability, iface_pkg_id)
+                .is_ok_and(|v| v)
+        });
```
alexcrichton, line 1951: Can this be rewritten to not swallow the error? Otherwise it looks like this is not reporting any errors from `include_stability`

crates/wit-parser/src/resolve.rs
```diff
@@ -1861,6 +1915,7 @@ impl Remap {
         &mut self,
         resolve: &mut Resolve,
         iface: &mut Interface,
+        iface_pkg_id: &PackageId,
```
alexcrichton, line 1918: Could this be inferred from `iface.pkg.unwrap()` instead of being passed in perhaps?

crates/wit-parser/src/resolve.rs
vados-cosmonic: Thanks for the quick review! So we can definitely include the two versions -- I've updated the message a bit.

> Also, should this return an error if the package's version is None perhaps?

And nope -- some packages that come in may not have versions specified at all (which we don't yet consider a proper bug). 

In particular the following tests have this use case:
- https://github.com/bytecodealliance/wasm-tools/blob/main/tests/cli/wit-stability-in-binary-format.wit
- https://github.com/bytecodealliance/wasm-tools/blob/main/tests/cli/wit-stability-inherited.wit

Do we want this to be an error? Effectively encoding that people *must* use top-level versioned-package WITs if they want to use `@since`? Obviously `@unstable` does not really require the same.

crates/wit-parser/src/resolve.rs
vados-cosmonic: Ahhh yeah I figured it might be -- will change all these to `expect`s and let the panics flow. Thanks for laying out the rationale here

vados-cosmonic: Ah also a note -- by "and below" do you want to change *all* the `include_stability` calls as well? 

Here's an example of *just* the one being changed (which I hope is what you meant): 
https://github.com/bytecodealliance/wasm-tools/pull/1689/commits/626e77927b320cc408d16f99a61ad3c014055cc2

crates/wit-parser/src/resolve.rs
vados-cosmonic: Done!

crates/wit-parser/src/resolve.rs
vados-cosmonic: Ahh thanks, didn't know that -- will replace

crates/wit-parser/src/resolve.rs
vados-cosmonic: So I could have sworn here that I ran into an issue where the `iface` that came in *did not* have the right package on it... But now that I look back, they always seem to. Don't know what happened here.

Ah, so after rechecking, I run into this *inside* the `append()` code, at precisely here:

https://github.com/bytecodealliance/wasm-tools/blob/c811041f3549dd93a889bd99ebb80ef33757152c/crates/wit-parser/src/resolve.rs#L1549

So on the very next line `iface.package` is *supposed* to be `None`, but it turns out it's *also* `None` on the line *before* that -- I assumed this was because this fn was called while going through`UnresolvedPackage`s.

It *seems* save to simply update the package first (as the code does a few lines down), so I changed it to do that instead -- please let me know if that violates some assumption that `update_interface` was making previously (I can't see one).

https://github.com/bytecodealliance/wasm-tools/pull/1689/commits/cb4dbf9b3577600a97eef2c75171ec544e32f280

crates/wit-parser/src/resolve.rs
vados-cosmonic: Did it -- it did get pretty wild though, I'm open to any simpler refactors if you have a suggestion:

https://github.com/bytecodealliance/wasm-tools/pull/1689/commits/667de50bb72a576adf79dabec25a56fd9a398772

crates/wit-parser/src/resolve.rs
alexcrichton: Ah sorry forgot to answer this earlier. This might come down to the interpretation of `@since`. In my mind the `feature = f` directive is purely optional and informational and isn't a step on the path towards stabilization, so `@since` means that the API is unconditionally included, not conditionally included.

If you've got a different impression though it might be worth clarifying this upstream in the component-model WIT.md file?

crates/wit-parser/src/resolve.rs
alexcrichton: My thinking is that if we're going to so far as to check version ranges that we should probably check that versions are present at all. I agree some tests will need an update, but that's ok to just update them and tag some versions.

crates/wit-parser/src/resolve.rs
alexcrichton: Oh `include_stability` contexts are fine since those are possibly going to be shown to developers so it's definitely good to stuff more context there.

crates/wit-parser/src/resolve.rs
```diff
@@ -1873,17 +1929,48 @@ impl Remap {
...
+                    iface
+                        .name
+                        .as_ref()
+                        .map(String::as_str)
+                        .unwrap_or("<unknown>"),
+                )
+            })?
+            .into_iter()
+            .filter_map(|(name, func, include)| include.then_some((name, func)))
+            .collect::<IndexMap<String, Function>>();
```
alexcrichton, line 1972: This whole combinator might be a bit easier to read as an imperative loop perhaps?

```rust
for (name, func) in mem::take(&mut iface.functions) {
    // include_stability + with_context
    iface.functions.insert(name, func);
}
```

crates/wit-parser/src/resolve.rs
vados-cosmonic: No problem! OK I see your viewpoint here -- this is probably better discussed (and basically recorded either way) as a change to component model's WIT design!

I've updated the code to ignore the version and we can revisit if the design changes are desired!

crates/wit-parser/src/resolve.rs
vados-cosmonic: OK will use that -- agree that it would be a bit easier to read that way

- [ ] rewrite iface fn update code

crates/wit-parser/src/resolve.rs
vados-cosmonic: Thanks! That's what I'll do -- will make those proper bugs. This will mean that specifying a version on an interface will *require* a version specified on the package otherwise it will fail to parse.

- [ ] Make versioned `@since` in unversioned packages an error

vados-cosmonic:
Hey @alexcrichton so looking back at the WIT design doc (trying to change it):

> Together, these gates support a development flow in which new features start with an @unstable gate while the details are still being hashed out. Then, once the feature is stable (and, in a WASI context, voted upon), the @unstable gate is switched to a @since gate. To enable a smooth transition (during which producer toolchains are targeting a version earlier than the @since-specified version), the @since gate contains an optional feature field that, when present, says to enable the feature when either the target version is greator-or-equal or the feature name is explicitly enabled by the developer. Thus, c is enabled if the version is 0.2.2 or newer or the fancy-foo feature is explicitly enabled by the developer. The feature field can be removed once producer toolchains have updated their default version to enable the feature by default.

It looks like there is an expectation for `feature` *to be used* to decide when to process a `@since`... Maybe this PR should reflect that, rather than changing it and then updating the docs upstream?

alexcrichton:
Ah I had missed that as well, good catch! That seems reasonable to me to implement yeah, want to do so as part of this PR?

vados-cosmonic:
Done! 

I needed to add some changes (and update some tests), would appreciate some extra close eyes on [this change set](https://github.com/bytecodealliance/wasm-tools/pull/1689/commits/4b8069a553507cc3d1628b28eb5b7ad010f42e87) (and of course the rest of the PR). 

Requiring the feature be specified is making a bunch of interfaces be missing, along with the deprecated flag (which wasn't being taken into account before), just want to make sure I'm reading the test output right and it passes a human spot check that isn't mine.

vados-cosmonic:
Actually, some more thinking here @alexcrichton ... The way that the note is phrased, we actually *must* allow `@since` with versions in the future right?

> Thus, c is enabled if the version is 0.2.2 or newer or the fancy-foo feature is explicitly enabled by the developer. The feature field can be removed once producer toolchains have updated their default version to enable the feature by default.

Given this sentence it assumes it is possible for `c` to be at a version that is *after* the package version -- otherwise it's just a feature check (because right now we consider future versions a bug).

I think to match this expected functionality we have to allow `@since(version = x)` where `x` is a future version.

I'm updating the code now to match

alexcrichton:
Hm I think you're right on your interpretation of the text, but now I'm having second thoughts about supporting it exactly as-is. How about reverting back to:

* With `@since` the `feature` is ignored and is purely informational in the WIT itself
* The `@since` version must be <= the current package version
* Using `@since` requires the package to have a version
* Mixing `@deprecated` with `@since` has no effect on whether it's included or not

That to me feels like a more straightforward way to interpret these attributes and if you agree I can work on sending a PR upstream as well.

vados-cosmonic:
Hey so after some discussion at the JS meeting with @guybedford  we actually realized another issue -- the AND vs OR combination of `version` and `feature` makes a bunch of things easier w/ respect to figuring out dependencies (i.e. if you know some world imports `a`, and `a` has `@since(version = v.v.v, feature = f)` you can intuit the required package and enabled features of the relevant package).

There's also the benefit (possibly anti-benefit?) of being able to span a feature over multiple versions (changing how a certain item works over two different versions). An example of this that might make sense is something like:

`@since(0.2.0, feature = async)`
`@since(0.3.0, feature = async)`

I'm I think there might be a bunch of surprise at the ORing of those, does it make sense to actually hammer this out/give everyone a chance to chat about it in an issue on `component-model`?

alexcrichton:
I'm not sure I quite follow what you mean by AND vs OR? But regardless it seems fine to discuss some more of the finer details on the component-model repository as well.

vados-cosmonic:
Ah so before I make the other issue -- basically the idea of whether `version` and `feature` should be AND or OR'd -- 

The logic right now is describe as an OR (valid version *or* feature enabled), but it might make sense/be unsurprising to have an AND -- i.e. valid version *and* feature enabled to see the import at all

alexcrichton:
Ah ok I see. Personally I don't think an AND is viable given how features/since/unstable are planned to be used in WASI. I also don't think that the OR requirement as stated in WIT.md necessarily makes sense because it doesn't really make sense for `since` to list a version in the future, but if it's always <= the current package version then the OR-ness is like it's OR-ing with true which doesn't affect too much.

In any case I agree with your 3 options, and probably good to discuss upstream!

crates/wit-parser/src/resolve.rs
```diff
@@ -1408,13 +1410,55 @@ impl Resolve {
...
+                    .get(*pkg_id)
+                    .and_then(|p| p.name.version.as_ref())
+                    .is_some_and(|pkg_version| {
+                        deprecated
+                            .as_ref()
+                            .is_some_and(|deprecation_version| deprecation_version <= pkg_version)
+                    })
+                {
+                    return Ok(false);
+                }
```
alexcrichton, line 1457: I'm a bit surprised by this clause because I was thinking that deprecation is just a compiler warning, not affecting semantics like whether the API is included or not?

alexcrichton (commented):
Thanks! One last comment here and I think this is good to go

crates/wit-parser/src/resolve.rs
vados-cosmonic: Ah you're right! This should absolutely be only a warning for tooling to use! Fixed

vados-cosmonic:
Seems like actions/upload-artifact is having a bit of an outage, hopefully this that step can just be re-run and pass soon!

PR #1700, opened 2024-07-25
Require root packages when defining nested packages in wit

macovedj:
This PR is an initial implementation of the wit functionality described [here](https://github.com/WebAssembly/component-model/pull/380)

Now, whenever defining packages as `package {...}`, it will be required that some file has `package namespace:name;` as a header somewhere in the project.  Thus there is a single root package, and the other package declarations are nested.  It also supports intermixing interface and world definitions between nested package definitions across files, and the packages in a deps folder can also nest packages, which are usable by the root package definition and its nested packages


tests/cli/wit-directory-output-valid.wit
```diff
@@ -1,14 +0,0 @@
-// RUN: component wit % --out-dir %tmpdir | component wit %tmpdir
-
-package a:b {
```
alexcrichton, line 3: Mind transforming this test to be valid with a `package foo:bar;` at the top?

tests/cli/multiple-packages-multiple-worlds.wit.stderr
```diff
@@ -1,3 +1 @@
-error: multiple worlds found; one must be explicitly chosen:
-  test:foo2/some-other-world
-  test:hola/hello
+error: none of the specified packages contain a world
```
alexcrichton, line 1: Could this error be made a bit more descriptive in that it's just the main package which contains no worlds?

tests/cli/multiple-packages-one-world.wit
```diff
@@ -1,5 +1,5 @@
 // RUN: component embed --dummy --wat % | component wit
-
+package foo:root;
 package test:foo1 {
```
alexcrichton, line 3: This test feels a bit off, now that there's the concept of a "main package" what this test was previously testing isn't quite as applicable any more and it should fail saying the main package has no worlds in it. The test could also be updated to pass a `--world` argument perhaps.

tests/cli/wit-directory-output-in-deps-folder-with-multiple-version.wit
```diff
@@ -1,16 +0,0 @@
-// RUN: component wit % --out-dir %tmpdir | component wit %tmpdir
```
alexcrichton, line 1: Could this test be preserved with a new package header at the top?

crates/wit-component/src/encoding.rs
```diff
@@ -2155,7 +2155,7 @@ world test {
 "#,
             )
             .unwrap();
-        let world = resolve.select_world(&pkgs, None).unwrap();
+        let world = resolve.select_world(&[pkg], None).unwrap();
```
alexcrichton, line 2158: You might already be planning to get to this, but `select_world` can be updated to taking a `PackageId` now instead of a list which will have a ripple effect throughout this repository.

alexcrichton (commented):
Some initial thoughts jotted down, but I'll hold off on a full review until this is ready.

tests/cli/wit-directory-output-in-deps-folder-with-multiple-version.wit
macovedj: Yeah I've been trying to decide what to do about this test, as well as what the desired behavior even is for this command.

Right now `DecodedWasm` doesn't differentiate between whether packages are from a deps folder or nested.  We could add this differentiation.
 
As I see it at the moment, there are a few options available:

1.) Capture if a package is nested or a deps folder package when we resolve, so we know how the package should be printed.  I think this would also require storing this info in binaries?  Which feels like something we may not want to do?

2.) Every package that isn't the root is nested... no deps folder is printed.

3.) Every package that isn't the root is printed in a deps folder... no nested packages (though packages in the original deps folder can have nested package definitions, so those would be nested still)

4.) Flags determine behavior

Maybe I'm missing an obvious alternative...

tests/cli/wit-directory-output-valid.wit
macovedj: This test has the same uncertainties as the one below testing the `--out-dir` flag.

tests/cli/wit-directory-output-in-deps-folder-with-multiple-version.wit
alexcrichton: Ah good points, but I think it's ok to create a `deps` folder when there wasn't one before. Organization of WIT might vary but the tooling here splatting out a single way to represent things I think makes sense. This is mostly just a test that whatever `--out-dir` produces is readable by `component wit`.

Implementation-wise I think it's fine to either shove everything into a `deps/*.wit` file or even just use `package { ... }` for all deps in the root folder. The whole `--out-dir` argument dosn't make a ton of sense any more nowadays that there's multi-package-forms-in-one-file but I figure  it's still good to not break it in the meantime.

crates/wasm-wave/src/value/wit.rs
```diff
@@ -155,10 +155,7 @@ mod tests {
 
     #[test]
     fn resolve_wit_type_smoke_test() {
-        let UnresolvedPackageGroup {
-            mut packages,
-            source_map,
-        } = UnresolvedPackageGroup::parse(
+        let UnresolvedPackageGroup { mut main, .. } = UnresolvedPackageGroup::parse(
```
alexcrichton, line 158: Mind updating these tests to using `resolve.push_str` instead of using `UnresolvedPackageGroup`?

crates/wit-component/src/printing.rs
```diff
@@ -51,61 +51,65 @@ impl WitPrinter {
...
     /// Print a set of one or more WIT packages into a string.
-    pub fn print(
-        &mut self,
-        resolve: &Resolve,
-        pkg_ids: &[PackageId],
-        force_print_package_in_curlies: bool,
-    ) -> Result<String> {
-        let print_package_in_curlies = force_print_package_in_curlies || pkg_ids.len() > 1;
-        for (i, pkg_id) in pkg_ids.into_iter().enumerate() {
+    pub fn print(&mut self, resolve: &Resolve, pkg_ids: &[PackageId]) -> Result<String> {
```
alexcrichton, line 54: Could this change to taking a single `PackageId` instead of a list? I think it's a bit confusing to have the "main" one be at the end and it would help emphasize that this only prints a single package at a time. 

I'm imagining that internally it would use `LiveTypes` to calculate the set of packages that this package depends on and then all of those packages would be printed in curly braces at the end. That way the semantic meaning of the output is the same but this is a way to print a whole package as a self-contained entity.

As an optional part the "print all deps too" logic could be controlled by a `bool` parameter, but that's ok to defer to later.

crates/wit-parser/src/resolve.rs
```diff
@@ -1474,22 +1490,26 @@ impl Remap {
     fn append(
         &mut self,
         resolve: &mut Resolve,
-        unresolved: UnresolvedPackage,
+        unresolved: &mut UnresolvedPackage,
```
alexcrichton, line 1493: How come this changed from `UnresolvedPackage` to `&mut`? The previous signature helped reflect that ownership was transferred and I think it'd be best to keep it if possible.

crates/wit-parser/src/resolve.rs
```diff
@@ -1474,22 +1490,26 @@ impl Remap {
...
+        } else {
+            let pkgid = resolve.packages.alloc(Package {
+                name: unresolved.name.clone(),
+                docs: unresolved.docs.clone(),
+                interfaces: Default::default(),
+                worlds: Default::default(),
+            });
+            resolve.package_names.insert(unresolved.name.clone(), pkgid);
+            pkgid
+        };
```
alexcrichton, line 1512: As I read over this I don't think that this is quite right, but not because of this PR. This is actually a preexisting issue today. Basically this isn't quite right because below the packages need to be merged rather than appended to otherwise duplicates might show up.

That being said I'll try to fix this in a follow up since it's a preexisting issue independent of this PR.

crates/wit-parser/src/resolve.rs
```diff
@@ -293,32 +309,64 @@ impl Resolve {
...
     }
 
-    fn parse_deps_dir(&mut self, path: &Path) -> Result<Vec<UnresolvedPackageGroup>> {
-        let mut unresolved_deps = Vec::new();
-        // If there's no `deps` dir, then there's no deps, so return the
-        // empty set.
+    fn parse_deps_dir(
+        &mut self,
+        path: &Path,
+    ) -> Result<BTreeMap<PackageName, UnresolvedPackageGroup>> {
```
alexcrichton, line 366: Could this preserve the `Vec` return value? This looks like it's returning an error on duplicate packages but most of the rest of WIT processing ends up merging things together so that may be best to do here instead.

crates/wit-parser/src/resolve.rs
```diff
@@ -293,32 +309,64 @@ impl Resolve {
...
+        let mut visiting = HashSet::new();
+        for pkg in &top_pkg.nested {
+            visit_group(&pkg, &deps, &mut order, &mut visiting, &pkg.source_map)?;
+        }
+        for pkg in deps.values().chain([&top_pkg]) {
+            for pkg in &pkg.nested {
+                visit_group(&pkg, &deps, &mut order, &mut visiting, &pkg.source_map)?;
+            }
+            visit_group(&pkg.main, &deps, &mut order, &mut visiting, &pkg.source_map)?;
+        }
```
alexcrichton, line 332: Is it possibly to rely on `sort_unresolved_packages` here instead of duplicating some parts of it? Ideally there'd be just one "sort this soup" method but I also realize that there's a lot of entry points here and may make that complicated to write.

crates/wit-parser/src/lib.rs
```diff
@@ -111,14 +111,22 @@ pub struct UnresolvedPackage {
...
+}
+
 /// Tracks a set of packages, all pulled from the same group of WIT source files.
-#[derive(Default)]
+#[derive(Clone)]
 pub struct UnresolvedPackageGroup {
+    pub main: UnresolvedPackage,
     /// A set of packages that share source file(s).
-    pub packages: Vec<UnresolvedPackage>,
+    pub nested: Vec<UnresolvedPackage>,
```
alexcrichton, line 129: Mind adding documentation to these two fields? (adding to `main` and updating the ones on `nested`)

crates/wit-parser/src/ast.rs
```diff
@@ -82,60 +82,162 @@ impl<'a> DeclList<'a> {
...
+            if tokens.eat(Token::Package)? {
+                let package_id = PackageName::parse(tokens, &mut docs)?;
+                if let Some(main) = main_pkg_decl {
+                    if main.package_name() == package_id.package_name() {
+                        bail!(
+                            "Nested package name already used by main package declaration: {}",
+                            main.package_name()
+                        );
+                    }
+                }
```
alexcrichton, line 102: Could this error get deferred to somewhere in `ast/resolve.rs` perhaps? Ideally the errors in this file are just parse errors rather than semantic errors.

crates/wit-parser/src/ast.rs
```diff
@@ -82,60 +82,162 @@ impl<'a> DeclList<'a> {
...
         }
-        Ok(DeclList { items })
+        Ok((DeclList { items }, nested))
     }
 }
 
-impl<'a> Ast<'a> {
-    fn parse(tokens: &mut Tokenizer<'a>) -> Result<Self> {
+impl<'a> SourceAst<'a> {
+    fn parse(
```
alexcrichton, line 139: The previous split of parsing implicit/explicit items doesn't quite make sense any more with this and I find that the structure here is a bit confusing. Could this instead be modelled as:

1. Parse the `package foo:bar;` at the top of the file, this is unconditionally required to come first.
2. Parse one-item-at-a-time after that which is either a nested package or an ast item.
3. When empty, return

I think that would help simplify a good chunk of the parsing here and in `parse_main_package_items` perhaps? For example an `AstItem` could be expanded to hold `NestedPackage(...)` or something like that. That would AST-wise allow arbitrarily nested packages but the AST validator could reject that saying that nested packages are only supported at the top level.

alexcrichton:
I'll do a final review of this tomorrow and afterwards start to turn the cranks again.

alexcrichton (approved):

PR #1703, opened 2024-07-29
converter from wit-parser Resolve to wit-encoder packages

MendyBerger:
Also improves the interface `use` api


alexcrichton (approved):
Nice!

alexcrichton:
I think the test might be due to line endings, so perhaps having it as an in-file literal instead?

MendyBerger:
@alexcrichton can confirm that line endings are indeed the problem here.
How do you feel about a manual `.replace("\r\n", "\n")`? I don't think that `wit-parser` can just read text without a file.

alexcrichton:
Would [`Resolve::push_str`](https://docs.rs/wit-parser/latest/wit_parser/struct.Resolve.html#method.push_str) work for that?

MendyBerger:
Missed that one. Should work!

Issue #1637, opened 2024-06-26
Relicensing `wasm-tools` and this repository's crates


alexcrichton:
# Relicensing `wasm-tools`

Hello everyone! This issue is here to serve as a checklist for relicensing the `wasm-tools` repository and all crates that it contains.

* **Before**: `Apache-2.0 WITH LLVM-exception`
* **After**: `Apache-2.0 WITH LLVM-exception OR Apache-2.0 OR MIT`.

In other words this issue is adding two other licensing options to the crate, the Rust standard `Apache-2.0 OR MIT` option. Everyone who has contributed to this repository is cc'd below with a checkbox tracking their approval status.

If you approve of this license change it's requested you leave a comment along the lines of "I approve this license change" and your checkbox will be filled in by a maintainer.

## Why change licenses?

The original choice of `Apache-2.0 WITH LLVM-exception` for this repository matches the licensing decision of Wasmtime itself and was intended to be compatible with it. Some crates in this repository, however, would like to be integrated into upstream Rust itself. For example the `wasmparser` and `wasm-encoder` crates are already part of the Rust compiler. Currently the Rust compiler [has an exception][exception] which allows these crates despite not having the Rust-standard `Apache-2.0 OR MIT` license. The goal of this issue is to make integration in the upstream Rust compiler easier by ensuring that the license is fully compatible with Rust's existing license.

[exception]: https://github.com/rust-lang/rust/blob/d929a42a664c026167800801b26d734db925314f/src/tools/tidy/src/deps.rs#L97-L98

## Contributors

Contributors to `wasm-tools` are generated with `git shortlog --summary
--numbered --email` as of [28e96f680752c999f296e5f9d7d31a6be66112f5](https://github.com/bytecodealliance/wasm-tools/commit/28e96f680752c999f296e5f9d7d31a6be66112f5).

* [x] @abrown ([commits](https://github.com/bytecodealliance/wasm-tools/commits?author=abrown))
* [x] @adambratschikaye ([commits](https://github.com/bytecodealliance/wasm-tools/commits?author=adambratschikaye))
* [x] @alexcrichton ([commits](https://github.com/bytecodealliance/wasm-tools/commits?author=alexcrichton))
* [x] @andylizi ([commits](https://github.com/bytecodealliance/wasm-tools/commits?author=andylizi))
* [x] @azaslavsky ([commits](https://github.com/bytecodealliance/wasm-tools/commits?author=azaslavsky))
[...]

PR #1707, opened 2024-07-31
Relicense the C API to match the repository

alexcrichton:
We have received final sign-off in #1637 for the C API so this updates it to have the same license as the rest of the repository to avoid a one-off special case.


pchickey (approved):
🎉

PR #1710, opened 2024-08-01
chore(wit-component): enable all features for docsrs

vados-cosmonic:
This commit enables all features for docsrs generation so that `wit-component` docs will contain information (e.g. for the `dummy` feature).

### DIFF
--- a/crates/wit-component/Cargo.toml
+++ b/crates/wit-component/Cargo.toml
@@ -16,6 +16,9 @@ rust-version.workspace = true
 [lints]
 workspace = true
 
+[package.metadata.docs.rs]
+all-features = true
+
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks! Mind also adding [this line](https://github.com/bytecodealliance/wasm-tools/blob/06e7673a4a070510d41595904c0c3a4ab45ed317/crates/wasmparser/src/lib.rs#L31) to `src/lib.rs` to have the required features show up in the documentation as well?

vados-cosmonic:
Ah thanks for the catch! Always forget to add that as well!

alexcrichton (approved):

PR #1711, opened 2024-08-03
Calculate sizes, alignment and offsets for both wasm32 and wasm64

cpetig:
This enables wit-bindgen to emit code like `let l2 = *ptr0.add(core::mem::size_of::<*const u8>()).cast::<usize>();` which automatically calculates the right offset of 4 for wasm32 and 8 for wasm64 e.g. for string or list lowering.

The more complex cases combine fixed byte offsets and pointer sized offsets.

~~This also adds a compatibility layer which falls back to 32bit for the classical `SizeAlign` usage.~~

### DIFF
--- a/crates/wit-parser/src/sizealign.rs
+++ b/crates/wit-parser/src/sizealign.rs
@@ -18,6 +18,15 @@ impl Default for Alignment {
     }
 }
 
+impl std::fmt::Display for Alignment {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Alignment::Pointer => f.write_str("ptr"),
+            Alignment::Bytes(b) => f.write_fmt(format_args!("{}", b.get())),
+        }
+    }
+}
+
 impl Ord for Alignment {
     fn cmp(&self, other: &Self) -> std::cmp::Ordering {
         match (self, other) {
@@ -81,6 +90,27 @@ impl From<Alignment> for ArchitectureSize {
                 bytes: 4,
                 add_for_64bit: 4,
             },
+        }
+    }
+}
+
+impl std::fmt::Display for ArchitectureSize {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        if self.add_for_64bit != 0 {
+            if self.bytes > self.add_for_64bit {
+                // both
+                f.write_fmt(format_args!(
+                    "{}+{}*ptrsz",
+                    self.constant_bytes(),
+                    self.usize_to_add()
+                ))
+            } else {
+                // only pointer
+                f.write_fmt(format_args!("{}*ptrsz", self.usize_to_add()))
+            }
+        } else {
+            // only bytes
+            f.write_fmt(format_args!("{}", self.constant_bytes()))
         }
     }
 }
--- a/crates/wit-parser/src/sizealign.rs
+++ b/crates/wit-parser/src/sizealign.rs
@@ -50,6 +50,16 @@ impl Ord for Alignment {
     }
 }
 
+impl Alignment {
+    /// for easy migration this gives you the value for wasm32
+    pub fn align_wasm32(&self) -> usize {
+        match self {
+            Alignment::Pointer => 4,
+            Alignment::Bytes(bytes) => bytes.get(),
+        }
+    }
+}
+
 /// Architecture specific measurement of position,
 /// the combined amount in bytes is
 /// `bytes + if 4 < std::sizeof::<usize> { add_for_64bit } else { 0 }`
@@ -116,11 +126,12 @@ impl std::fmt::Display for ArchitectureSize {
 }
 
 impl ArchitectureSize {
-    fn max(&self, other: &Self) -> Self {
-        let new_bytes = self.bytes.max(other.bytes);
+    pub fn max<B: std::borrow::Borrow<Self>>(&self, other: B) -> Self {
+        let new_bytes = self.bytes.max(other.borrow().bytes);
         Self {
             bytes: new_bytes,
-            add_for_64bit: (self.bytes + self.add_for_64bit).max(other.bytes + other.add_for_64bit)
+            add_for_64bit: (self.bytes + self.add_for_64bit)
+                .max(other.borrow().bytes + other.borrow().add_for_64bit)
                 - new_bytes,
         }
     }
@@ -145,6 +156,11 @@ impl ArchitectureSize {
     /// Shortcut for compatibility with previous versions
     pub fn size_wasm32(&self) -> usize {
         self.bytes
+    }
+
+    /// prefer this over >0
+    pub fn is_empty(&self) -> bool {
+        self.bytes == 0
     }
 }
 
--- a/crates/wit-parser/src/sizealign.rs
+++ b/crates/wit-parser/src/sizealign.rs
@@ -95,6 +95,13 @@ impl ArchitectureSize {
         }
     }
 
+    pub fn add_bytes(&self, b: usize) -> Self {
+        Self {
+            bytes: self.bytes + b,
+            add_for_64bit: self.add_for_64bit,
+        }
+    }
+
     /// The effective offset/size is
     /// `constant_bytes() + std::sizeof::<usize> * usize_to_add()`
     pub fn constant_bytes(&self) -> usize {
--- a/crates/wit-parser/src/sizealign.rs
+++ b/crates/wit-parser/src/sizealign.rs
@@ -70,6 +70,21 @@ impl AddAssign<ArchitectureSize> for ArchitectureSize {
     }
 }
 
+impl From<Alignment> for ArchitectureSize {
+    fn from(align: Alignment) -> Self {
+        match align {
+            Alignment::Bytes(bytes) => ArchitectureSize {
+                bytes: bytes.get(),
+                add_for_64bit: 0,
+            },
+            Alignment::Pointer => ArchitectureSize {
+                bytes: 4,
+                add_for_64bit: 4,
+            },
+        }
+    }
+}
+
 impl ArchitectureSize {
     fn max(&self, other: &Self) -> Self {
         let new_bytes = self.bytes.max(other.bytes);
@@ -105,21 +120,9 @@ pub struct ElementInfo {
 
 impl From<Alignment> for ElementInfo {
     fn from(align: Alignment) -> Self {
-        match align {
-            Alignment::Bytes(bytes) => ElementInfo {
-                size: ArchitectureSize {
-                    bytes: bytes.get(),
-                    add_for_64bit: 0,
-                },
-                align,
-            },
-            Alignment::Pointer => ElementInfo {
-                size: ArchitectureSize {
-                    bytes: 4,
-                    add_for_64bit: 4,
-                },
-                align,
-            },
+        ElementInfo {
+            size: align.into(),
+            align,
         }
     }
 }
--- a/crates/wit-parser/src/sizealign.rs
+++ b/crates/wit-parser/src/sizealign.rs
@@ -106,22 +106,7 @@ impl From<Alignment> for ArchitectureSize {
 
 impl std::fmt::Display for ArchitectureSize {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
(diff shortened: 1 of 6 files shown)


cpetig:
This is a generalization of #1446

alexcrichton:
Thanks for working on this! I've gotten a chance to review this now (thanks for the patience). I don't really understand how this works at a low-level and it feels like this form of generically handling size/align for 32/64 seems somewhat error-prone. It does seem like a nice property though to have where it would make code generation easier in the wit-bindgen and source languages.

To help evaluate this I whipped up a small [differential fuzzer](https://github.com/bytecodealliance/wasm-tools/commit/6d7a233cfa71cfddf56893a511e5c26a6e3f29c7) which generates arbitrary WITs and then asserts that the sizes reported by this branch are the same as the sizes reported on the `main` branch. That currently fails relatively quickly where `ArchitectureSize` looks like it's holding negative values as unsigned numbers which means that overflow is panicking on addition. Is that intended? If not that might be a bug?

Once that's sorted out I'd also like to extend the fuzzer a bit to call `field_offsets` and `payload_offset` for types (if it matches) to assert that those return values are the same as well.

For running the fuzzer I was locally doing:

```
FUZZER=wit64 cargo +nightly fuzz run -s none run
```

cpetig:
Thank you for taking the time to fuzz it. I wouldn't fully trust code of this complexity either.

I found the difference: Resources are considered very large, but the new code considers them (very large)*2. Honestly, I had no idea about the correct way to handle this type, but perhaps it should use add_for_64bit: 0.

The code giving the problem:
```
TypeDefKind::Resource => ElementInfo::new(
                ArchitectureSize::new(usize::MAX, usize::MAX),
                Alignment::Bytes(NonZeroUsize::new(usize::MAX).unwrap()),
            ),
```

alexcrichton:
I think resources should be unconditionally 32-bit values (always 4 bytes) regardless of wasm32 vs wasm64?

cpetig:
I agree that resource **handles** should be 4+0, but this code calculates the size of a resource itself, which doesn't make much sense, see here https://github.com/bytecodealliance/wasm-tools/blob/main/crates/wit-parser/src/sizealign.rs#L61

I was simply at loss on which number to return in this unrealistic case.

alexcrichton:
Aha I see! That looks like it's just something to update in the fuzzer then and skip `TypeDef::Resource` declarations within interfaces (makes sense in retrospect). Would you be up for developing the fuzzer a bit further to see if this PR's approach is viable? If the differential fuzzer is clean that would make me more confident in this approach personally and I can do a deeper review.

cpetig:
I planned to use intmax+0 which would make the old and new values identical again, then add the field and payload offsets, hiding size calculation of resources would be unnecessary with the first change.

PS: I will for sure take inspiration from your fuzzer in testing other projects. Now I know the real purpose of wit-smith ;-)

alexcrichton:
Ah it's ok to do whatever for `TypeDefKind::Resource`, it's an edge case needed for completeness but shouldn't show up practically in theory. This is just a fuzzer issue really and I think the way you map it right now is fine.

cpetig:
Oh, after some time the fuzzer found another one: `fuzz/artifacts/run/crash-84dc68e18872aad7a98cfcf676848973ef8aa1af`

```
fuzz/src/wit64.rs:45:9:
assertion `left == right` failed
  left: 24
 right: 30
```

cpetig:
See https://github.com/cpetig/wasm-tools/tree/fuzz64 for an updated variant of the fuzzer checking field offsets and payload offset. I assume that the types contained within other types are already part of the all types list, so no need to check the individual fields.

Now I will debug the failed case, it found another one `fuzz/artifacts/run/crash-e4149b42d86d41bf7a4230735e54a801f5be70ef`

`xx: func(name: option<option<option<option<option<option<result<result<result<list<s32>, bool>, result<result<result<result<list<s32>, option<option<option<option<option<option<option<option<option<bool>>>>>>>>>>, bool>, bool>, bool>>, bool>>>>>>>, name1: bool, name2: bool, name3: bool, name4: bool);`

cpetig:
I found the problem (in my 64bit combination reasoning), pushed and fuzzer running again

cpetig:
I stopped the fuzzer after several hours (3.4M tests, if I read correctly) without any difference detected.

crates/wit-parser/src/sizealign.rs
```diff
@@ -1,26 +1,210 @@
...
+                if b.get() > 4 {
+                    std::cmp::Ordering::Greater
+                } else {
+                    std::cmp::Ordering::Less
+                }
+            }
+            (Alignment::Bytes(a), Alignment::Bytes(b)) => a.cmp(b),
+        }
+    }
+}
```
alexcrichton, line 51: Could you have some code comments for what ordering this is achieving? It's not immediately clear to this myself but I think some comments would help out indicating what order is being achieved.

crates/wit-parser/src/sizealign.rs
```diff
@@ -1,26 +1,210 @@
...
-pub enum AddressSize {
-    #[default]
-    Wasm32,
-    Wasm64,
+/// Architecture specific alignment
+#[derive(Eq, PartialEq, PartialOrd, Clone, Copy, Debug)]
+pub enum Alignment {
+    Pointer,
+    Bytes(NonZeroUsize),
 }
```
alexcrichton, line 13: Mind expanding the documentation here explaining the representation? Basically saying that this is intended to be a pointer-width-agnostic representation of alignment.

crates/wit-parser/src/sizealign.rs
```diff
@@ -1,26 +1,210 @@
...
+/// Architecture specific measurement of position,
+/// the combined amount in bytes is
+/// `bytes + if 4 < core::mem::size_of::<*const u8>() { add_for_64bit } else { 0 }`
+#[derive(Default, Clone, Copy, Eq, PartialEq, Debug)]
+pub struct ArchitectureSize {
+    /// exact value for 32-bit pointers
+    pub bytes: usize,
+    /// amount of bytes to add for 64-bit architecture
+    pub add_for_64bit: usize,
+}
```
alexcrichton, line 79: Would it be possible to represent this as something like `bytes` and `pointers`? That way the size could be `bytes + pointers * sizeof(pointer)`. That I think may also make methods like `pointers_to_add` a bit more clear how to be implemented and/or unnecessary since it's just a field then.

crates/wit-parser/src/sizealign.rs
```diff
@@ -116,64 +292,249 @@ impl SizeAlign {
...
+        let info = self.0.record(types);
+        (info.size.size_wasm32(), info.align.align_wasm32())
+    }
+    pub fn params<'a>(&self, types: impl IntoIterator<Item = &'a Type>) -> (usize, usize) {
+        self.record(types.into_iter())
+    }
+}
+
+/// compatibility alias
+pub type SizeAlign = SizeAlign32;
```
alexcrichton, line 424: Would it be possible to go ahead and remove these typedefs/aliases? I think it'd be best to only have the new `SizeAlign64` (probably renamed to `SizeAlign` if it's the only one) which would force users to update one way or another. There's enough conveniences on the above types that not using 64-bit support I think would be reasonable.

alexcrichton (approved):
Ok thanks for fuzzing! That makes me more confident in this approach so I've done a bit of a deeper dive into this. 

Do you think it would be possible to preserve the fuzzer? Basically having a separate "relatively simple" algorithm for calculating sizes with no frills and ensuring that this result continues to be the same?

crates/wit-parser/src/sizealign.rs
cpetig: The answer is easy to state (I will add some comments):

 Bytes(1) < Bytes(2) < Bytes(4) <= Pointer <= Bytes(8)
 
 This is needed to calculate max alignment.

crates/wit-parser/src/sizealign.rs
cpetig: I tried this once, encountered some difficulties and changed it to the current form. I will try again … I guess the current representation avoids multiplications and divisions and makes alignment more straightforward.

crates/wit-parser/src/sizealign.rs
cpetig: I implemented this fully compatible drop-in replacement type to enable migrating wit-bindgen (and perhaps other users of `SizeAlign`) more gradually to the new API - to prevent locking a new version of each crate to each other. Once this MR merges wit-bindgen would require many code changes directly when updating its wasm-tools dependency.

But as you seem willing to take the risk (and are the maintainer of both repositories) … requiring migration clearly is the more clean solution in the long run.

cpetig:
> Do you think it would be possible to preserve the fuzzer? Basically having a separate "relatively simple" algorithm for calculating sizes with no frills and ensuring that this result continues to be the same?

I think the fuzzer could simply live in the new code and compare the new version with an older released version (flipping the crate alias roles).

crates/wit-parser/src/sizealign.rs
alexcrichton: Should the `<=` there be `<`?

crates/wit-parser/src/sizealign.rs
alexcrichton: In theory it should be possible to skip mul/div by only having shifts left-and-right by storing the log2 of the pointer size (or something like that), but I'm mostly curious about this from an understandability point of view.

crates/wit-parser/src/sizealign.rs
alexcrichton: Personally I think it's ok to go ahead and replace the old `SizeAlign`. If necessary `wit-bindgen-core` can contain the `SizeAlign32` helper you have written here and then the code generators can be temporarily updated to using that.

alexcrichton:
Oh excellent idea on fuzzing, I like that!

crates/wit-parser/src/sizealign.rs
cpetig: For Ord it is `<`. Conceptually it is more `<=`. So I documented `<` here.

On wasm32 `Pointer` and `Bytes(4)` would be equivalent, but still you would consider Pointer as the (potentially) larger one, for Bytes(8) it is the other way round.

cpetig:
> Oh excellent idea on fuzzing, I like that!

I feel we might want to replace `wit-smith` in the fuzzer with the old version soon ... agreed?

crates/wit-parser/src/sizealign.rs
cpetig: Oh, while implementing a more obscure test case I realized that I made a conceptual oversight:

I you take the max of `10 bytes` and `2 pointers` you get `4 bytes + 1.5 pointers` - and I didn't take fractional pointer amounts into account, yet. It still can work if I multiply the number of pointers by 4 and handle fractional amounts properly. This wasn't visible in the fuzzing because of the additional alignment applied to the result before outputting the size.

cpetig: Previously this example resulted in `10 bytes + 6 bytes_on_64_bit` - if you start with 9 bytes you even get a quarter pointer amount (9 + 7).

cpetig: Probably you don't want to express/represent partial pointer sizes for objects containing pointers, so you would want to align in the max function and return (8+1).

cpetig: I was able to make it work, but align is a bit more complex than I would like it to be and max now uses the easy way out (calculate 32bit and 64bit then max and then separate again). Max can probably be made more efficient by expanding the 32bit/64bit calculation and then simplifying terms, but as it is mostly used for variants it should be less performance sensitive.

cpetig:
@alexcrichton Do you want me to mark these threads as resolved or do you want to check first and mark then?

alexcrichton (approved):
Looks great to me, thanks again for working on this! I'll admit I still don't really understand how this works but I'm willing to trust the combo of you and the fuzzer, so let's merge and see how it goes.

alexcrichton:
Oh and either way works for me on threads, thanks!

crates/wit-parser/src/sizealign.rs
alexcrichton: Thanks for pushing on this! I think at least from a code generator perspective it'll be easier to reason about size + pointers so I'm glad this was able to work out.

PR #1713, opened 2024-08-06
Increase limits of some component types

alexcrichton:
Increase some limits from 1_000 to 10_000 to handle bytecodealliance/wit-bindgen#1019


pchickey (approved):
Didn't we clamp flags to 32 or 64 elsewhere?

alexcrichton:
Good point, I'll leave that limit as-is. The validation limit per-definition for some of these is smaller as well.

PR #1715, opened 2024-08-12
threads: round out `missing-features` tests

abrown:
This change attempts to add tests that check that
shared-everything-threads features indeed fail when the proposal is not enabled. In doing this, I realized that a few features fail on other features first (GC, reftypes), which meant the tests couldn't be as exhaustive as I would have liked. This commit also changes the wording of one of the failures to be consistent across the proposal.


alexcrichton (approved):
Thanks!

> I realized that a few features fail on other features first (GC, reftypes)

One way I've handled this in the past is that you can exploit the way features are calculated by:

* First the test is in the normal `missing-features` folder, disabling all features
* Next the test is in a folder named `gc`, which triggers logic to enable the `gc` proposal
* So on and so forth for other proposals

That way you can sort of reset to a baseline set of features with `missing-features` and then by layering more folders you can activate more features. This might need some frobbing in the loop that handles features in the `roundtrip.rs` test as well.

PR #1719, opened 2024-08-12
threads: improve tests

abrown:
These changes improve shared-everything-threads tests; see commit messages for more details.


abrown:
I would like to add two more Binaryen tests here but can't yet:
- for [`shared-polymorphism.wasm`](https://github.com/WebAssembly/binaryen/blob/main/test/spec/shared-polymorphism.wast), wasm-tools needs some easier to validate both the shared and unshared version of a type; e.g., `self.pop_operand(Some(...), can_be_shared)?`
- for [`shared-i31.wast`](https://github.com/WebAssembly/binaryen/blob/main/test/spec/shared-i31.wast), I just need to figure out what should be brought over (all? parts?)

crates/wasmparser/src/validator/operators.rs
```diff
@@ -2715,8 +2715,27 @@ where
...
+                        shared_eqref
+                    );
+                }
+                if a != b {
+                    bail!(
+                        self.offset,
+                        "type mismatch: expected `ref.eq` types to match"
+                    );
+                }
+            }
```
abrown, line 2736: Both of these checks are incorrect: the first should be a subtype check and I now am wondering as well if `a` and `b` can actually be different types.

alexcrichton: If it's common enough it's probably worth having a helper for, but as a one-off here what about:

```rust
if self.top_of_stack_is_shared_ref() {
    // two pops with EQ.shared()
} else {
    // two pops with EQ
}
```

abrown: Looks like I'll also need it for `i31.get_s`, `i31.get_u`, `array.len`, `any.convert_extern`, and `extern.convert_any`. I've been staring at `pop_operand`, `_pop_operand`, and `pop_ref` but I don't see a nice way to incorporate this logic there. Should I just do what you suggested in all the spots?

alexcrichton: Perhaps a version of `pop_ref` which takes a `enum Shared { Yes, No, Maybe }` argument or similar? Given the use in other instructions a helper does seem in order

abrown: Ok, take a look at b7d1e6f. There are certainly other ways to refactor things here so feel free to suggest alternate routes; I tried to match what was already there but it might be time for a more wholesale change.

crates/wasmparser/src/validator/operators.rs
```diff
@@ -625,6 +634,29 @@ where
         }
     }
 
+    /// Pop a reference type from the operand stack, checking if it matches
+    /// `expected` or the shared version of `expected`. This function will panic
+    /// if `expected` is shared or a concrete type.
+    fn pop_maybe_shared_ref(&mut self, expected: RefType) -> Result<Option<RefType>> {
+        assert!(!self.resources.is_shared_ref_type(expected));
```
alexcrichton, line 641: C

crates/wasmparser/src/validator/operators.rs
```diff
@@ -625,6 +634,29 @@ where
...
+        let expected = if self.resources.is_shared_ref_type(actual_ref) {
+            expected
+                .shared()
+                .expect("this only expects abstract heap types")
+        } else {
+            expected
+        };
+        self._check_operand_type(expected.into(), actual)?;
+        Ok(Some(actual_ref))
+    }
```
alexcrichton, line 658: This method feels like roughly the right signature but I'm not sure it requires the other refactorings in this PR. Some changes I might recommend:

* Don't add `is_shared_ref_type` and instead `debug_assert!` here that `expected` is an abstract heap type with the `shared` flag unset
* Use `pop_ref` internally here which doesn't take an `expected` argument
* Given a popped ref test if it's a subtype of the given `expected` either with the `shared` flag set or unset. 
* Return `Result<Option<(RefType, bool)>>` so callers don't have to recalculate `shared`-vs-not.

I think that should satisfy most use cases and avoid the need for larger refactorings?

crates/wasmparser/src/validator/operators.rs
```diff
@@ -2715,8 +2747,18 @@ where
         Ok(())
     }
     fn visit_ref_eq(&mut self) -> Self::Output {
-        self.pop_operand(Some(RefType::EQ.nullable().into()))?;
-        self.pop_operand(Some(RefType::EQ.nullable().into()))?;
+        let a = self
+            .pop_maybe_shared_ref(RefType::EQ.nullable())?
+            .expect("an actual ref type");
```
alexcrichton, line 2752: Note that `.expect` here isn't valid because that means that this will panic when `ref.eq` is found after unreachable code. The two results here need to be matched and if they're both `Some` then the shared-ness needs to match. If one is `None` or both are `None` then we're in unreachable code and they can be assumed to match (IIRC)

crates/wasmparser/src/validator/operators.rs
```diff
@@ -4398,24 +4440,34 @@ where
...
-            .as_type()
-            .map_or(false, |ty| ty.as_reference_type().unwrap().is_nullable());
+        let extern_ref = self.pop_maybe_shared_ref(RefType::EXTERNREF)?;
+        let (is_nullable, shared) = if let Some(extern_ref) = extern_ref {
+            (
+                extern_ref.is_nullable(),
+                self.resources.is_shared_ref_type(extern_ref),
+            )
+        } else {
+            (false, false)
```
alexcrichton, line 4450: This is something that might be worth clarifying in the upstream shared-everything-threads proposal. Specifically once unreachable code is being validated wha's the value of `shared` to use here? The nullable flag works out since it's easy to pick a "top type" that everything else is a subtype, but that's not true for `shared`. In theory this is invalid because later instructions might be expecting `shared` and that should work given the high-level description of unreachable code (IIRC) but won't work in the implementation of this algorithm.

crates/wasmparser/src/validator/operators.rs
abrown: Sure, let's try that. I think we need to keep `is_shared_ref_type` or something like it, though, because we need need to be able to figure out the sharedness of an `actual` concrete type. We can `debug_assert!` away the checks for `expected` but, because `actual` can actually be concrete, we need to be able to resolve the index to the underlying type somehow to inspect its sharedness.

abrown: A side note: at some point in earlier PRs I had a `RefType::is_shared(&self) -> Option<bool>` that would check sharedness only for the abstract types but I made do without it until now &mdash; even if I re-added it, it wouldn't be enough here because we still have to check sharedness for concrete types.

abrown:
Ok, I refactored and rewrote parts of this to address the feedback. I kept the `self.resources.is_shared_ref_type` since I didn't have any other way to check shared-ness of concrete types (though we could definitely rename it `is_shared` if that's preferable &mdash; the type should be self-evident). I didn't return a shared-ness `bool` from `pop_maybe_shared_ref` but I could; it just felt like if we have `self.resources.is_shared_ref_type` we might as well use it and keep `pop_maybe_shared_ref` that much simpler. I also ended up having to rebase to bring in #1724; let's see if that gets rid of the CI errors.

crates/wasmparser/src/validator/operators.rs
abrown: https://github.com/WebAssembly/shared-everything-threads/issues/80

crates/wasmparser/src/validator/operators.rs
```diff
@@ -625,6 +617,37 @@ where
         }
     }
 
+    /// Pop a reference type from the operand stack, checking if it is a subtype
+    /// of `expected` or the shared version of `expected`. This function will
+    /// panic if `expected` is shared or a concrete type.
+    fn pop_maybe_shared_ref(&mut self, expected: RefType) -> Result<Option<RefType>> {
```
alexcrichton, line 623: To avoid redudant calls to `is_shared_ref_type` could this return `Result<Option<(RefType, bool)>>` where the `bool` is whether it's shared or not?

crates/wasmparser/src/validator/operators.rs
alexcrichton: Thanks! Mind leaving a TODO here and in `extern_convert_any` pointing to that issue?

alexcrichton (commented):
Looks good to me! I see now as well yeah how `is_shared_ref_type` is needed, but I think we can cut down on the usage of it a bit by returning a `bool` from `pop_maybe_shared_ref`

alexcrichton (approved):

PR #1725, opened 2024-08-15
Add a new `wasm-tools component unbundle` command

alexcrichton:
This commit adds a new subcommand to `wasm-tools component` named `unbundle` which is tasked with extracting core wasm modules out of a component. This is intended to showcase an example implementation of doing this and the main motivation is to extract core wasm modules out of a component for runtimes such as Wasmtime. This enables sharing machine code for WebAssembly modules between components that use the same core wasm module.

This new subcommand is primarily built on `wasm_encoder::reencode` with a simple size-based heuristic for which modules to extract.


pchickey (approved):
Wow! This is really exciting, we have been talking about this for years and its finally real.

Is there any way for the tests to check the contents of the unbundled modules?

alexcrichton:
Good point yeah, added a few small tests to print the output in `%tmpdir`

PR #1728, opened 2024-08-20
threads: add tests for `ref.i31_shared`

abrown:
This change brings in Binaryen's existing [`shared-i31.wast`] test, minus the commented-out parts that Binaryen does not yet support. To get this to pass, we add the ability to parse `ref.i31_shared` as a constant expression in WAST.

[`shared-i31.wast`]: https://github.com/WebAssembly/binaryen/blob/main/test/spec/shared-i31.wast

### DIFF
--- a/crates/wasmparser/src/validator/core.rs
+++ b/crates/wasmparser/src/validator/core.rs
@@ -479,6 +479,10 @@ impl ModuleState {
                 $self.validate_gc("ref.i31")?;
                 $self.validator().visit_ref_i31()
             }};
+            (@visit $self:ident visit_ref_i31_shared) => {{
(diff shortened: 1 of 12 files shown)


crates/wasmparser/src/validator/core.rs
```diff
@@ -479,6 +479,10 @@ impl ModuleState {
                 $self.validate_gc("ref.i31")?;
                 $self.validator().visit_ref_i31()
             }};
+            (@visit $self:ident visit_ref_i31_shared) => {{
+                $self.validate_gc("ref.i31_shared")?;
```
alexcrichton, line 483: Should this perhaps be `validate_shared_everything_threads`?

PR #1729, opened 2024-08-20
threads: check mutability of GC instructions

abrown:
This adds logic and tests to check that `struct|array.atomic.rmw.*` instructions cannot modify GC objects unless they're marked as mutable.


crates/wasmparser/src/validator/operators.rs
```diff
@@ -3963,7 +3975,10 @@ where
     fn visit_struct_set(&mut self, struct_type_index: u32, field_index: u32) -> Self::Output {
         let field_ty = self.struct_field_at(struct_type_index, field_index)?;
         if !field_ty.mutable {
-            bail!(self.offset, "invalid struct.set: struct field is immutable")
+            bail!(
+                self.offset,
+                "invalid struct `set`: struct field is immutable"
```
alexcrichton, line 3980: How come the `struct.set` changed to `struct set`?

alexcrichton (commented):
Thanks! Could these checks perhaps be encapsulated in helpers? Something like `self.mutable_struct_field_at(...)` instead of `self.struct_field_at(...)` plus a check?

crates/wasmparser/src/validator/operators.rs
abrown: We're reusing this `bail!` for both `struct.set` and `struct.atomic.set`; failing with a `struct.set` message when the instruction is actually `struct.atomic.set` seemed confusing.

alexcrichton (approved):

PR #1730, opened 2024-08-20
Refactor type checks when popping refs

alexcrichton:
This commit refactors users of `pop_ref` during validation to be able to pass in an expected type. This removes the need for a few manual `is_subtype` checks and helps centralize type-checking in one location.

### DIFF
--- a/crates/wasmparser/src/validator/operators.rs
+++ b/crates/wasmparser/src/validator/operators.rs
@@ -605,8 +605,8 @@ where
     }
 
     /// Pop a reference type from the operand stack.
-    fn pop_ref(&mut self) -> Result<Option<RefType>> {
-        match self.pop_operand(None)? {
+    fn pop_ref(&mut self, expected: Option<RefType>) -> Result<Option<RefType>> {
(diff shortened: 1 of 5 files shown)


crates/wasmparser/src/validator/operators.rs
```diff
@@ -1130,51 +1120,31 @@ where
...
-            bail!(
-                self.offset,
-                "{inst_name}'s heap type must be a sub type of the type on the stack: \
-                 {sub_ty} is not a sub type of {sup_ty}"
-            );
-        }
-
+        self.pop_ref(Some(sup_ty))?;
         Ok(sub_ty)
     }
```
alexcrichton, line 1135: @fitzgen asking for your review on this PR, and in particular for the changes in this function. I don't believe I'm preserving the precise shape of what was here previously but I think I'm maintaining the spec still, but wanted to confirm.

fitzgen: Yeah, this looks good to me.

fitzgen (approved):

PR #1733, opened 2024-08-21
Support optimizing feature checks at compile time

alexcrichton:
Something I've long wanted for `wasmparser` is the ability to constant-propagate decisions about feature selection instead of requiring runtime checks to do so. This should be a performance win for proposals as they become standardized and more and more of wasm is gated behind some feature or another. The goal with this is to enable a mode of `wasmparser` which doesn't require runtime checks of booleans/bits to see whether a proposal is enabled or not.

In lieu of not redesigning everything too much (e.g. adding `F: WasmFeatures` everywhere) this commit adds a new compile time Cargo feature to the crate called `features` which, when disabled, makes `WasmFeatures` a zero-sized type that represent the default active features for `wasmparser`. This enables const-propagating decisions about features throughout the codebase and should help optimize the validator/decoder in niche situations.

### DIFF
--- a/crates/wasmparser/src/features.rs
+++ b/crates/wasmparser/src/features.rs
@@ -131,7 +131,7 @@ define_wasm_features! {
     /// [`BinaryReader::set_features`](crate::BinaryReader::set_features) can be
     /// used.
     ///
-    /// This crate additionally has a compile-time Cargo features called
+    /// This crate additionally has a compile-time Cargo feature called
     /// `features` which can be used to enable/disable most of this type at
     /// compile time. This crate feature is turned on by default and enables
     /// this bitflags-representation of this structure. When the `features`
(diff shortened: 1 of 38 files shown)


crates/wasmparser/src/features.rs
```diff
@@ -82,11 +98,46 @@ macro_rules! define_wasm_features {
...
+    /// Enabled/disabled features can affect both parsing and validation at this
+    /// time. When decoding a WebAssembly binary it's generally recommended if
+    /// possible to enable all features, as is the default with
+    /// [`BinaryReader::new`](crate::BinaryReader::new). If strict conformance
+    /// with historical versions of the specification are required then
+    /// [`BinaryReader::new_features`](crate::BinaryReader::new_features) or
+    /// [`BinaryReader::set_features`](crate::BinaryReader::set_features) can be
+    /// used.
+    ///
+    /// This crate additionally has a compile-time Cargo features called
```
fitzgen, line 134:
```suggestion
    /// This crate additionally has a compile-time Cargo feature called
```

fitzgen (approved):

PR #1734, opened 2024-08-21
Update handling of unreachable code and heap types

alexcrichton:
This commit updates validation of wasm modules with unreachable code using gc/heap types. It notably fixes cases with the shared-everything-threads proposal where existing instructions are polymorphic over `shared` and not and wasn't supported before. Specifically code was refactored to use `MaybeType` a bit more to propagate the bottom-ness and the `MaybeType::HeapBot` variant has grown a new `AbstractHeapType` payload for various instructions to use such as `any.convert_extern`.

### DIFF
--- a/crates/wasmparser/src/readers/core/types.rs
+++ b/crates/wasmparser/src/readers/core/types.rs
@@ -1464,7 +1464,7 @@ pub enum AbstractHeapType {
 }
 
 impl AbstractHeapType {
-    const fn as_str(&self, nullable: bool) -> &str {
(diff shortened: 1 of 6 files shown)


alexcrichton:
I should also mention this is a product of the discussion on https://github.com/WebAssembly/shared-everything-threads/issues/80

fitzgen (approved):

abrown (approved):
Discussed elsewhere: swapping out `Bot`, `HeapBot` and `Type` with `Unknown`, `PartialRef`, and `Known` could clarify the use of `MaybeType` a bit.

PR #1736, opened 2024-08-22
threads: add validation for shared calls

abrown:
The shared-everything-threads [proposal] prevents `shared` contexts (e.g., in a `shared` function) from accessing unshared objects. This change adds validation to prevent `shared`-to-unshared calls.

### DIFF
--- a/crates/wasmparser/src/resources.rs
+++ b/crates/wasmparser/src/resources.rs
@@ -50,14 +50,11 @@ pub trait WasmModuleResources {
     /// The sub type must be canonicalized.
     fn sub_type_at(&self, type_index: u32) -> Option<&SubType>;
 
-    /// Returns the type id associated with the given function
-    /// index.
+    /// Returns the type ID associated with the given function index.
     fn type_id_of_function(&self, func_idx: u32) -> Option<CoreTypeId>;
 
-    /// Returns the `FuncType` associated with the given function index.
-    ///
-    /// The function type must be canonicalized.
-    fn type_of_function(&self, func_idx: u32) -> Option<(&FuncType, bool)>;
+    /// Returns the type index associated with the given function index.
+    fn type_index_of_function(&self, func_index: u32) -> Option<u32>;
 
     /// Returns the element type at the given index.
     ///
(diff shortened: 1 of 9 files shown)


abrown:
cc: @alexcrichton, I have another branch that adds `shared` to `FuncType` but that doesn't seem quite right &mdash; no other composite type carries that along with it (but it would make things look nicer!).

alexcrichton:
Hm how about making `CompositeType` an `enum` again? Basically move the `shared: bool` field into each of the enum arms. Then have something `fn shared(&self) -> bool` on `CompositeType` which accesses each variant?

abrown:
> Hm how about making `CompositeType` an `enum` again? Basically move the `shared: bool` field into each of the enum arms. Then have something `fn shared(&self) -> bool` on `CompositeType` which accesses each variant?

Well, (1) that complicates parsing because we would have to pass down shared-ness into the variant parsing and (2) I think there's some argument to be made to keeping the wasm-tools AST close to the spec AST.

alexcrichton:
To come back to this Andrew and I talked bit on Zulip about this and concluded the way to go might be to refactor `WasmResources` to only be able to return `&CompositeType` for a certain index. Existing func/struct/array accessors would all be removed. Then in the operator validator there would be helpers that would downcast this further to `&FuncType` and would auto-check for shared-ness which would avoid the need to thread around a boolean and make sure everything is checked everywhere.

abrown:
In working on this, I was able to reuse the `*_type_at` helpers to add shareability checks but this only works when we retrieve by sub-type index; for some kinds of calls, we actually return the type by function index. If we want to get rid of the weird signature on `WasmModuleResources::type_of_function`, we might be able to just remove that entirely, call `WasmModuleResources::type_id_of_function` and use that ID with `func_type_at` (does that work?).

alexcrichton:
Ah yeah I see, could that perhaps return just the `u32` type index of where the function's type is located in the module? (or `CoreTypeId` if that works, either way)

alexcrichton (approved):

PR #1737, opened 2024-08-22
Rename `MaybeType` variants, tweak documentation

abrown:
This tweaks #1734 to use terms I consider a bit more clear. It also updates the documentation to explain some more of the context that I currently have paged in but may not later.

### DIFF
--- a/crates/wasmparser/src/validator/operators.rs
+++ b/crates/wasmparser/src/validator/operators.rs
@@ -160,23 +160,31 @@ pub struct OperatorValidatorAllocations {
 
 /// Type storage within the validator.
 ///
-/// This is used to manage the operand stack and notably isn't just `ValType`
-/// to handle unreachable code and the "bottom" type.
+/// When managing the operand stack in unreachable code, the validator may not
+/// fully know an operand's type. this `Unknown` state is known as the `bottom` type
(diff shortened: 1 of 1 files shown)


fitzgen (commented):
Love the comments, not sold on the renaming of the variants, especially since the current names have rough correspondence with concepts from the spec. What do you think of just doing the docs updates and not the variant renaming?

abrown:
> Love the comments, not sold on the renaming of the variants, especially since the current names have rough correspondence with concepts from the spec. What do you think of just doing the docs updates and not the variant renaming?

I don't think `HeapBot` does? But I do see a bit of a point in the `Bot` naming since that is mentioned in the spec. I'm open to leaving this as `Bottom -> PartialRef -> Known` if that makes more sense but I kind of like the clarity of the progression `Unknown -> PartialRef -> Known` since that is really what is happening here.

fitzgen:
> I'm open to leaving this as `Bottom -> PartialRef -> Known`

I like this best, over what I've been able to bikeshed so far. Maybe `PartialRef` should be `UnknownRef`? I don't know, don't want to bikeshed tooooooo much...

alexcrichton (approved):

PR #1738, opened 2024-08-22
Remove `Display for {Alignment,ArchitectureSize}`

alexcrichton:
These display representations are easy to get confused with the previous return values of `u32` which do implement `Display`, but the `Display` here isn't as applicable as it used to be. As a stopgap remove the `Display` implementations entirely to help identify locations that need to be updated to explicitly use an accessor or otherwise to calculate the size/align information.

### DIFF
--- a/crates/wit-parser/src/sizealign.rs
+++ b/crates/wit-parser/src/sizealign.rs
@@ -6,7 +6,7 @@ use std::{
 use crate::{FlagsRepr, Int, Resolve, Type, TypeDef, TypeDefKind};
 
 /// Architecture specific alignment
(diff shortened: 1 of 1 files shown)


cpetig:
Oh, yeah, sorry for forgetting about Display. I also disabled Display to fix code generation, on the other hand quite a lot of places in wit-bindgen needed Display. So it was trading quickly being able to recompile against correctness.

I feel turning Display off is the more safe choice in the long run.

cpetig (approved):

fitzgen (approved):

PR #1741, opened 2024-08-26
refactor: Remove optional feature attribute from stable feature gates

vados-cosmonic:
With an eye towards making [the discussion in `component-model`](https://github.com/WebAssembly/component-model/issues/382) a reality, the commits in this PR update the `wasm-tools` to remove

Note that the change to `wit-parser` is a semver breaking change, as we're removing a piece of the `Stability` enum.

### DIFF
--- a/crates/wit-component/src/printing.rs
+++ b/crates/wit-component/src/printing.rs
@@ -913,17 +913,9 @@ impl WitPrinter {
     fn print_stability(&mut self, stability: &Stability) {
         match stability {
             Stability::Unknown => {}
-            Stability::Stable {
-                since,
-                feature,
-                deprecated,
-            } => {
+            Stability::Stable { since, deprecated } => {
                 self.output.push_str("@since(version = ");
                 self.output.push_str(&since.to_string());
-                if let Some(feature) = feature {
-                    self.output.push_str(", feature = ");
-                    self.output.push_str(feature);
-                }
                 self.output.push_str(")\n");
                 if let Some(version) = deprecated {
                     self.output.push_str("@deprecated(version = ");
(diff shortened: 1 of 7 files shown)


alexcrichton:
Could this perhaps just be removed from `enum Stability` for now but left in `enum Attribute`? I'm wary to remove a feature cold turkey lest projects be broken, so having a smoother path to "this does nothing for awhile" before it's removed feels a bit better. Unfortunately there's not a great vector at this time to provide warnings about WIT files so we don't have the option of warning on deprecation of using WIT features at this time.

pchickey:
For whatever its worth, I think for brand new features like this one, its OK to break this corner case. In a few months I'd be more careful, but I think its pretty unlikely anyone will encounter this.

vados-cosmonic:
I can definitely see it both ways, and with regards to:

> so having a smoother path to "this does nothing for awhile" before it's removed feels a bit better. 

One thing I thought of might be to just make sure it the code never does anything useful (i.e. always puts `None`) and possibly outputs a warning.

So putting out the options at least:

- 1. Make `feature` always None (least breaking -- not a breaking change)
- 2. Remove from `Stability`, keep in `Attribute` (medium breaking)
- 3. Remove from both

Could go either way here, but if @alexcrichton you're comfortable with (2) I can go with that, or if you want to go with what @pchickey suggested (which is 3), I can do that as well.

alexcrichton:
API-wise it's ok to have breaking changes whenever, that's what the semver-major-bumps are for in this crates on regular releases. Interface-wise-breaking though is more worrisome (e.g. changing WIT syntax). That's because it's expected that WIT with 215.* is compatible with 216.* and vice-versa (where possible). Changing WIT syntax by removing `feature = ...` breaks the both API and WIT compatibilty, but we can ignore the API parts.

That's basically a longer-winded way of saying, yes, I was adovcating for (2) and Pat for (3). That being said I don't disagree that this is new enough we can probably go with (3), so seems reasonable to just remove it outright (but that will require updating some tests it looks like)

alexcrichton (approved):

PR #1745, opened 2024-08-28
threads: no shared function expansion

abrown:
If I understand `wast`'s `Expander` correctly (which I may not!), there is no need for it to know how to create type definitions for shared functions because there is no way to write WAT that would mark a function shared in an inline way. Since the only way to create a shared function is with an explicit type ID (e.g., `(func (type $ft))`), we can set `shared = false` when a `TypeDef` must be created.


alexcrichton (approved):
Mind opening an issue for this as a feature request on the shared-everything-threads repo? I've found it quite convenient historically to be able to write functions in the past without having to write a type ahead of time, so in the limit of time I think this'd be a neat feature to have.

PR #1751, opened 2024-08-31
Fix build of `wast` with randomized layouts

### LINKED ISSUE #1750: wast lexer size assert not compatible with randomized struct layouts
https://github.com/bytecodealliance/wasm-tools/blob/d68b3828f0ebbc46ce0b4fff9c4d5d881487114a/crates/wast/src/lexer.rs#L64-L66

breaks when compiling with [layout randomization](https://github.com/rust-lang/rust/issues/106764)

If the assert is necessary for correctness then `repr(C)` should be used to get a deterministic layout.
If it's just an optimization then it's probably better to just check it in tests/CI.

Found this while trying to build the rust compiler with randomized layouts. https://github.com/rust-lang/rust/pull/101339

alexcrichton:
Closes #1750

### DIFF
--- a/crates/wast/src/lexer.rs
+++ b/crates/wast/src/lexer.rs
@@ -61,9 +61,10 @@ pub struct Token {
     pub len: u32,
 }
 
-const _: () = {
+#[test]
+fn token_is_not_too_big() {
     assert!(std::mem::size_of::<Token>() <= std::mem::size_of::<u64>() * 2);
-};
+}
 
 /// Classification of what was parsed from the input stream.
 ///


crates/wast/src/lexer.rs
```diff
@@ -46,13 +46,14 @@ pub struct Lexer<'a> {
 
 /// A single token parsed from a `Lexer`.
 #[derive(Copy, Clone, Debug, PartialEq)]
+#[repr(C)] // ensure the assertion below always works
```
the8472, line 49: This attribute doesn't affect the inner structs. Since those would get randomized to they can also affect layout.

You could do a bunch of runs with -Zrandomize-layout and different `-Zlayout-seed=` values.

alexcrichton: Do you know if the [`"randomize"` feature](https://github.com/rust-lang/rust/blob/9649706eada1b2c68cf6504356efb058f68ad739/compiler/rustc_abi/src/layout.rs#L981-L992) is enabled for nightly compilers? If I compile with `-Zrandomize-layout` before/after this PR both builds work, so it seems like "no" but I could also be getting unlucky with seeds

the8472: It's part of the default features, so it should be available:

https://github.com/rust-lang/rust/blob/9649706eada1b2c68cf6504356efb058f68ad739/compiler/rustc_abi/Cargo.toml#L20

the8472: the error is pointing at 215, so it's not the latest, but it doesn't look like the file changed since then.

```
   Compiling wast v215.0.0
error[E0080]: evaluation of constant value failed
    --> /cargo/registry/src/index.crates.io-6f17d22bba15001f/wast-215.0.0/src/lexer.rs:65:5
     |
65   |     assert!(std::mem::size_of::<Token>() <= std::mem::size_of::<u64>() * 2);
     |     |
     |     |
     |     the evaluated program panicked at 'assertion failed: std::mem::size_of::<Token>() <= std::mem::size_of::<u64>() * 2', /cargo/registry/src/index.crates.io-6f17d22bba15001f/wast-215.0.0/src/lexer.rs:65:5
     |
    ::: /checkout/library/core/src/macros/mod.rs:1596:5
```

the8472: tried a local checkout, it failed on the first try for me, I didn't even have to iterate through seeds.

```
$ RUSTFLAGS="-Zrandomize-layout" cargo +nightly check
    Checking wast v216.0.0 (/home/the8472/git/wasm-tools/crates/wast)
error[E0080]: evaluation of constant value failed
  --> crates/wast/src/lexer.rs:65:5
   |
65 |     assert!(std::mem::size_of::<Token>() <= std::mem::size_of::<u64>() * 2);
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the evaluated program panicked at 'assertion failed: std::mem::size_of::<Token>() <= std::mem::size_of::<u64>() * 2', crates/wast/src/lexer.rs:65:5
   |
   = note: this error originates in the macro `assert` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0080`.
error: could not compile `wast` (lib) due to 1 previous error
```

alexcrichton: Ah ok yeah it took a few tries but I was able to get the crate to fail the assertion before this commit as well. After this commit though I can't get it to fail yet, I'll probably just move this to a runtime test in the meantime.

the8472 (approved):
CI test is good, then it won't break downstream users.

abrown (approved):

PR #1756, opened 2024-09-03
threads: fuzz `shared-everything-threads`

abrown:
This adds initial support for fuzzing the WebAssembly side of the shared-everything-threads [proposal]. Eventual addition of the CM builtins will require extensions to this.

### DIFF
--- a/crates/wasm-smith/src/core.rs
+++ b/crates/wasm-smith/src/core.rs
@@ -1,4 +1,4 @@
-//! Generating arbitary core Wasm modules.
+//! Generating arbitrary core Wasm modules.
 
 mod code_builder;
 pub(crate) mod encode;
@@ -516,7 +516,7 @@ impl Module {
 
             (HT::Concrete(a), HT::Abstract { shared, ty }) => {
                 let a_ty = &self.ty(a).composite_type;
-                if a_ty.shared == shared {
+                if a_ty.shared != shared {
                     return false;
                 }
                 match ty {
@@ -530,7 +530,7 @@ impl Module {
 
             (HT::Abstract { shared, ty }, HT::Concrete(b)) => {
                 let b_ty = &self.ty(b).composite_type;
-                if shared == b_ty.shared {
+                if shared != b_ty.shared {
                     return false;
                 }
                 match ty {
@@ -572,6 +572,10 @@ impl Module {
         let index = u32::try_from(self.types.len()).unwrap();
 
         if let Some(supertype) = ty.supertype {
+            debug_assert!(
+                !self.is_shared_type(supertype)
+                    || (self.is_shared_type(supertype) && ty.composite_type.shared)
+            );
             self.super_to_sub_types
                 .entry(supertype)
                 .or_default()
@@ -658,12 +662,12 @@ impl Module {
 
     fn clone_rec_group(&mut self, u: &mut Unstructured, kind: AllowEmptyRecGroup) -> Result<()> {
         // NB: this does *not* guarantee that the cloned rec group will
-        // canonicalize the same as the original rec group and be
-        // deduplicated. That would reqiure a second pass over the cloned types
-        // to rewrite references within the original rec group to be references
-        // into the new rec group. That might make sense to do one day, but for
-        // now we don't do it. That also means that we can't mark the new types
-        // as "subtypes" of the old types and vice versa.
+        // canonicalize the same as the original rec group and be deduplicated.
+        // That would require a second pass over the cloned types to rewrite
+        // references within the original rec group to be references into the
+        // new rec group. That might make sense to do one day, but for now we
+        // don't do it. That also means that we can't mark the new types as
+        // "subtypes" of the old types and vice versa.
         let candidates: Vec<_> = self.clonable_rec_groups(kind).collect();
         let group = u.choose(&candidates)?.clone();
         let new_rec_group_start = self.types.len();
@@ -678,9 +682,10 @@ impl Module {
 
     fn arbitrary_sub_type(&mut self, u: &mut Unstructured) -> Result<SubType> {
         if !self.config.gc_enabled {
+            let shared = self.arbitrary_shared(u)?;
             let composite_type = CompositeType {
-                inner: CompositeInnerType::Func(self.arbitrary_func_type(u)?),
-                shared: false,
+                inner: CompositeInnerType::Func(self.arbitrary_func_type(u, shared)?),
+                shared,
             };
             return Ok(SubType {
                 is_final: true,
@@ -713,7 +718,7 @@ impl Module {
                 *f = self.arbitrary_matching_func_type(u, f)?;
             }
             CompositeInnerType::Struct(s) => {
-                *s = self.arbitrary_matching_struct_type(u, s)?;
+                *s = self.arbitrary_matching_struct_type(u, s, composite_type.shared)?;
             }
         }
         Ok(SubType {
@@ -727,6 +732,7 @@ impl Module {
         &mut self,
         u: &mut Unstructured,
         ty: &StructType,
+        must_share: bool,
     ) -> Result<StructType> {
         let len_extra_fields = u.int_in_range(0..=5)?;
         let mut fields = Vec::with_capacity(ty.fields.len() + len_extra_fields);
@@ -734,7 +740,7 @@ impl Module {
             fields.push(self.arbitrary_matching_field_type(u, *field)?);
         }
         for _ in 0..len_extra_fields {
-            fields.push(self.arbitrary_field_type(u)?);
+            fields.push(self.arbitrary_field_type(u, must_share)?);
         }
         Ok(StructType {
             fields: fields.into_boxed_slice(),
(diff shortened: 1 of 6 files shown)


abrown:
At this stage (after df11990) the fuzzer will trigger an `unimplemented!` failure [here](https://github.com/bytecodealliance/wasm-tools/blob/main/crates/wasm-mutate/src/module.rs#L46):

```console
$ cargo +nightly fuzz run run
...
thread '<unnamed>' panicked at crates/wasm-mutate/src/module.rs:46:18:
not implemented
```

It is unclear to me why this hasn't been triggered previously: shouldn't occasionally enabling GC have triggered this failure due to

https://github.com/bytecodealliance/wasm-tools/blob/8c3346e72d3f5100ed80db2e44e326873a9d2443/crates/wasm-smith/src/core.rs#L1052-L1058

crates/wasm-smith/src/core.rs
```diff
@@ -2494,15 +2497,21 @@ pub(crate) fn configured_valtypes(config: &Config) -> Vec<ValType> {
...
+            for &shared in shareability {
+                use AbstractHeapType::*;
+                for ty in [
+                    Any, Eq, I31, Array, Struct, None, Func, NoFunc, Extern, NoExtern,
+                ] {
+                    valtypes.push(ValType::Ref(RefType {
+                        nullable,
+                        heap_type: HeapType::Abstract { shared, ty },
+                    }));
+                }
```
fitzgen, line 2514: This could be an `extend` and will generally be a little faster since the space can be reserved one time up front.

fitzgen (approved):
Looks great modulo some very minor nitpicks -- thanks!

abrown:
@fitzgen, as I revisited this PR, I realized that I wanted to add several missing bits as well as split several of the commits into separate PRs. Since I've substantially changed the PR, let's forget about the previous review, though I do believe I've addressed the comments you made.

This PR is currently in an intermediate state and I could use some help figuring out what to do: I could (a) just continue hacking away at this until fuzzing "fully works" but that might take a while due to the large amount of reverse validation I have to add to wasm-smith (i.e., many instructions generate invalid parameters in a shared function context). Another option is to (b) just cordon off the ability to create shared function bodies for now; this would allow generating all the types and what not and for that to be fuzzed while I work on (a). Opinions?

fitzgen:
> This PR is currently in an intermediate state and I could use some help figuring out what to do: I could (a) just continue hacking away at this until fuzzing "fully works" but that might take a while due to the large amount of reverse validation I have to add to wasm-smith (i.e., many instructions generate invalid parameters in a shared function context). Another option is to (b) just cordon off the ability to create shared function bodies for now; this would allow generating all the types and what not and for that to be fuzzed while I work on (a). Opinions?

I'm all for landing incremental progress, so long as the parts that are landing are working. We don't need to take an all-or-nothing approach to adding support for generating new wasm proposals in `wasm-smith`.

crates/wasm-smith/src/core.rs
```diff
@@ -2549,6 +2635,16 @@ impl Module {
             }
         }
     }
+
+    fn arbitrary_shared(&self, u: &mut Unstructured) -> Result<bool> {
+        Ok(self.config.shared_everything_threads_enabled && u.arbitrary()?)
```
abrown, line 2640: We might want to dial down the frequency of shared things compared to unshared ones? This would be a good place to do that.

fuzz/src/lib.rs
```diff
@@ -27,6 +27,12 @@ pub fn generate_valid_module(
     config.canonicalize_nans = u.arbitrary()?;
     config.custom_page_sizes_enabled = u.arbitrary()?;
     config.wide_arithmetic_enabled = u.arbitrary()?;
+    config.shared_everything_threads_enabled = u.arbitrary()?;
+    if config.shared_everything_threads_enabled {
+        config.threads_enabled = true;
+        config.gc_enabled = true;
+    }
+    config.reference_types_enabled = config.reference_types_enabled || config.gc_enabled;
```
alexcrichton, line 35: Mind putting this in `Config::sanitize`? 

Ideally as well shared-everything-threads wouldn't require `gc_enabled` since that would help test just the shared functions/globals/etc side of things.

For the `reference_types_enabled` change here, I'm surprised that was needed due to `Config::sanitize`. Did you run into an issue where this line was required? If so we might be forgetting to sanitize something somewhere?

crates/wasm-smith/src/core.rs
```diff
@@ -1001,69 +1019,90 @@ impl Module {
...
+            element_type: self.arbitrary_storage_type(u, must_share)?,
             mutable: u.arbitrary()?,
         })
     }
 
-    fn arbitrary_storage_type(&mut self, u: &mut Unstructured) -> Result<StorageType> {
+    fn arbitrary_storage_type(
+        &mut self,
+        u: &mut Unstructured,
+        must_share: bool,
```
alexcrichton, line 1077: One thing we've done in the past historically is to avoid threading through parameters like this and storing them on `Self`. Would that perhaps make sense for this `must_share` field? That'd help many callers I think since ideally the shared-ness would, in a perfect world, integrate in a few locations but wouldn't permeate throughout.

crates/wasm-smith/src/core.rs
alexcrichton: Yeah I think that'd be reasonable. Something like `u.ratio(1, 10)?` or similar perhaps? 

This might also be a good place to use the field I mentioned above on `Self` to force shared-ness since that could be used to make this method always return `true` for example.

alexcrichton (commented):
> The latest fuzz failure happens in the wasm-mutate: ... it may be time to let oss-fuzz start finding them?

Failures are fine yeah, and 200k+ is a good start. I'd recommend maybe ~1h of local fuzz time in the background and if that turns up nothing then it's probably ready for oss-fuzz.

For this failure specifically it might be best to just reject modules with shared types for now in `wasm-mutate`. There's a number of mutators in wasm-mutate and many of them are written in such a way to be pretty generic over the input module. My guess is that `ConstExpressionMutator` is probably what's doing this and either updating that to just not mutate these expressions with `shared` or adding support there is reasonable.

> One thing I've started to worry about in wasm-smith is generation speed. I'm not sure what effect this PR has on the number of fuzz executions per second, but I don't like all the added ...map().filter()... logic to get at the shared or unshared sides of the world. 

To confirm, I only saw one of those in this PR, is that right?

You're right that generation speed matters, and it's one we've had to optimize for in the past. What you mention of having lists of shared/unshared things I think makes sense. What I'd recommend doing is that whenever an operation needs a linear search to build up a cache for that beforehand and consult the cache instead. The `CodeBuilder` class has an `*Allocations` type (IIRC) for this where it builds up a bunch of caches for function body generation. You can do that sooner as well though if you need the caches for generation of the module skeleton itself.

Overall the basic rule of thumb is to avoid quadratic behavior. That'll time out on oss-fuzz and happens whenever a linear search is done for an operation such as per-opcode selection. That means that the one loop I found in this PR as-is, a linear search at the top-level of generating functions, is no issue. That's just a one-time processing of the input so there's no problem with that.

crates/wasm-smith/src/core.rs
abrown: I don't think we can: if we set it on `Self` it would be a module-level decision but we actually need a type-by-type decision. It boils down to: "we just generated a `shared` object and we need to ensure that any types inside it are also `shared`" &mdash; thus, `must_share = true`. The unshared case is simpler but we also want to generate it: "we just generated an unshared object and it doesn't matter if the types inside are shared or not" &mdash; `must_share = false`, which allows us to choose later.

There is another way to do this: we could "bubble up" the shared-ness from down below instead of pushing it down from above. I asked @fitzgen what he preferred and he picked the "push down" approach... so here we are.

abrown: And to clarify: if it's a module-level flag we would have to set it before calling down and then unset it on the way back up. I figured threading it is more clear.

fuzz/src/lib.rs
abrown: > Mind putting this in Config::sanitize?

Ok, but I'm going to use the inverted form to follow what's already there (right?).

> Ideally as well shared-everything-threads wouldn't require gc_enabled since that would help test just the shared functions/globals/etc side of things.

Ok, I'll try it without that.

> For the reference_types_enabled change here, I'm surprised that was needed due to Config::sanitize. Did you run into an issue where this line was required? If so we might be forgetting to sanitize something somewhere?

This is code from months ago and I see you just added or modified `Config::sanitize` about a month ago. I think I might have just brought something that was necessary before that change. I'll try without it...

alexcrichton: Ah yeah it'll be inverted where shared-everything-threads will be disabled if threads is disabled (for example)

Also for reference-types aha makes sense! I think it may be the case it's no longer necessary, but would be good to double-check.

crates/wasm-smith/src/core.rs
alexcrichton: Ah yes sorry I meant for this to be a you-set-it-then-unset-it kind of field where it's not global-per-module, just stored on the side to avoid threading everywhere. This pattern is used for fields like `max_type_limit` I believe so I think this'll fit reasonably well into that pattern. 

I agree that passing arguments is more explicit, but it's a fair bit of noise much of the time since it's just being forwarded around. The integration with `arbitrary_shared` I think might also make the most sense by putting it into `Module` itself.

abrown: Take a look at f8b9edb. After that change I ran it for more than an hour without failure racking up ~14m iterations. (Hope we're still emitting shared stuff!).

crates/wasm-smith/src/core.rs
```diff
@@ -1055,12 +1083,20 @@ impl Module {
...
         if !self.config.reference_types_enabled {
-            return Ok(RefType::FUNCREF);
+            // Create a `RefType::FUNCREF` but with variable sharedness.
+            Ok(RefType {
+                nullable: true,
+                heap_type: HeapType::Abstract {
+                    shared: self.arbitrary_shared(u)?,
+                    ty: AbstractHeapType::Func,
+                },
+            })
```
alexcrichton, line 1093: Given that shared-everything-threads now currently requires reference-types the changes here might be reasonable to undo

crates/wasm-smith/src/core.rs
```diff
@@ -1073,7 +1109,16 @@ impl Module {
...
+            // If the caller is demanding a shared heap type but the concrete
+            // type we found is not in fact shared, we skip down below to use an
+            // abstract heap type instead. If the caller is not demanding a
+            // shared type, though, we can use either a shared or unshared
+            // concrete type.
+            if let Some(ty) = self.types.get(idx as usize) {
+                if !(self.must_share && !ty.composite_type.shared) {
+                    return Ok(HeapType::Concrete(idx));
+                }
+            }
```
alexcrichton, line 1121: This might be good to haev as a TODO where in the future this consults the list of shared types instead of "all types" to avoid this arm entirely if there are no shared types.

crates/wasm-smith/src/core.rs
```diff
@@ -1328,10 +1373,12 @@ impl Module {
                             .collect(),
                     });
                     index_store.replace(new_index as u32);
+                    let shared =
+                        self.config.shared_everything_threads_enabled && u.arbitrary().ok()?;
```
alexcrichton, line 1377: `self.arbitrary_shared(u)` perhaps?

alexcrichton (approved):

PR #1764, opened 2024-09-05
Allow parsing `rectypes` in components

abrown:
This change supports [#392], which adds a way to use GC's `rectypes` as type definitions in components. Previously, only function types were supported and there was no way express array and struct types. This keeps the previous function decoding support based on peeking the function type `0x60` prefix but adds support for encoding `rectypes` with a new `0x00` prefix.

[#392]: https://github.com/WebAssembly/component-model/pull/392


crates/wasm-encoder/src/reencode/component.rs
```diff
@@ -892,29 +886,11 @@ pub mod component_utils {
 
     pub fn parse_component_sub_type<T: ?Sized + ReencodeComponent>(
```
alexcrichton, line 887: Since the trait method above was removed, mind removing this as well?

alexcrichton (commented):
For added context this is an implementation of https://github.com/WebAssembly/component-model/pull/392 but will wait on the exact path forward there to be settled on before landing here to avoid conflicting implementation/spec.

crates/wasmparser/src/validator/types.rs
```diff
@@ -465,7 +465,7 @@ define_wrapper_id! {
     pub enum ComponentCoreTypeId {
         #[unwrap = unwrap_sub]
         /// A core type.
-        Sub(CoreTypeId),
+        Rec(CoreTypeId),
```
alexcrichton, line 468: This'll actually stay as `Sub` unlike some of the other refactorings I've since realized. (it's not quite obvious given the current structure of the code but https://github.com/bytecodealliance/wasm-tools/pull/1765 should make it more clear as to why)

crates/wasmparser/src/validator/types.rs
```diff
@@ -257,7 +257,7 @@ pub enum CoreType {
 }
 
 /// Represents a unique identifier for a core type type known to a
-/// [`crate::Validator`]
+/// [`crate::Validator`]. This should be used with the type arenas....
```
alexcrichton, line 260: leftover comment

crates/wast/src/component/binary.rs
```diff
@@ -884,15 +880,17 @@ impl From<&ModuleType<'_>> for wasm_encoder::ModuleType {
...
-                        f.params.iter().map(|(_, _, ty)| (*ty).into()),
-                        f.results.iter().copied().map(Into::into),
-                    ),
-                    core::InnerTypeKind::Struct(_) | core::InnerTypeKind::Array(_) => {
-                        todo!("encoding of GC proposal types not yet implemented")
-                    }
-                },
+                ModuleTypeDecl::Type(t) => {
+                    let sub_type = SubType {
+                        is_final: t.final_type.unwrap_or(true), // TODO: is this correct?
```
alexcrichton, line 885: I think it should be safe to remove this TODO now

crates/wast/src/component/binary.rs
```diff
@@ -884,15 +880,17 @@ impl From<&ModuleType<'_>> for wasm_encoder::ModuleType {
...
+                        supertype_idx: t.parent.map(u32::from), // TODO: resolve?
```
alexcrichton, line 886: For this TODO, do you mean resolving `$t` to an index? If so that happens in prior phases so I think it's ok to remove this TODO

crates/wast/src/core/types.rs
```diff
@@ -505,6 +505,17 @@ impl<'a> Parse<'a> for StorageType<'a> {
     }
 }
 
+impl From<StorageType<'_>> for wasm_encoder::StorageType {
```
alexcrichton, line 508: Mind moving these new `From` impls to `core/binary.rs`? That matches the style of `component/binary.rs` and as we transition to wasm-encoder over time for core bits I'm thinking that more things will pop up there

alexcrichton:
Oh oops apologies but this as of now has a merge conflict in `tests/roundtrip.rs`, but I suspect that's not too hard to resolve. I've put https://github.com/bytecodealliance/wasm-tools/pull/1765 in the merge queue though which, if merged by when you read this, should probably also cause some conflicts. Let me know though if anything is non-obvious how to resolve.

alexcrichton (approved):

PR #1770, opened 2024-09-09
Fix wasm-smith's page_size calculation

alexcrichton:
Pointed out in [this comment] this should help `wasm-smith` generate offsets/data segments with a wider range of valid values in wasm-smith rather than accidentally thinking the page size is much smaller than it actually is.

[this comment]: https://github.com/bytecodealliance/wasm-tools/commit/fe1307be46d0a30b5332ae12b219dd90cc90fbe4#r146408758


fitzgen (approved):
Whoops!

PR #1773, opened 2024-09-10
Recommend `--locked` for `cargo install`

samestep:
This PR changes `README.md` so it recommends users pass the [`--locked`](https://doc.rust-lang.org/cargo/commands/cargo-install.html#dealing-with-the-lockfile) flag to `cargo install`, which should prevent some surprises. This should also make the behavior more similar to the `cargo binstall` example command right below.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ platform](https://www.rust-lang.org/tools/install) and then use the included
 Cargo package manager to install:
 
 ```
-$ cargo install wasm-tools
+$ cargo install --locked wasm-tools
 ```
 
 Alternatively if you use [`cargo


pchickey (approved):

alexcrichton:
Thanks for the PR! Out of curiosity was there a concrete breaking thing that happened as a result of missing `--locked`?

samestep:
@alexcrichton there was not, so this might not actually matter at all; in that case I apologize for the noise!

alexcrichton:
Oh no worries! I don't think that this is an unreasonable suggestion, I just personally try to stick to Cargo defaults where possible. I understand though that Cargo's defaults in this regard are somewhat contentious and there's definitely breakage that's possible without `--locked`. Additionally given the breadth of the full dependency graph with all subcommands enabled it seems like having this is probably a good idea to ensure that breakage never happens in the future.

PR #1775, opened 2024-09-10
json-from-wast: Include binary form of text files where possible

### LINKED ISSUE #1771: JsonFromWast issue with the latest testsuite
If you run jsonfromwast on the latest testsuite comments.wast it will generate:

- comments.0.wasm
- comments.1.wasm
- comments.2.wasm
- comments.3.wasm
- comments.4.wat

and an entry in command.json

```json
        {
            "type": "module",
            "line": 83,
            "filename": "comments.4.wat"
        },
```

I'm migrating over from WABT and I found there was never a module command previously that can take a wat file as input. WABTs output for this same scenario is the following:

```json
         {
            "type": "module",
            "line": 0,
            "filename": "comments.4.wasm"
        },
```

alexcrichton:
This commit is aimed at addressing #1771 by including a new `binary_filename` field in the JSON which is populated when the test itself is `module quote`, or a textual module, but the module parse successfully. This should help avoid the need for a text parser for external users using the tool while still reflecting the test case itself where it's text-based.

Closes #1771

### DIFF
--- a/src/bin/wasm-tools/json_from_wast.rs
+++ b/src/bin/wasm-tools/json_from_wast.rs
@@ -131,19 +131,19 @@ impl<'a> JsonBuilder<'a> {
         let line = self.lineno(directive.span());
         let command = match directive {
             WastDirective::Module(module) => {
-                let (name, _module_type, filename) = self.emit_file(module)?;
+                let (name, file) = self.emit_file(module, false)?;
                 json::Command::Module {
                     line,
                     name: name.map(|s| self.module_name(s)),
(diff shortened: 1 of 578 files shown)


alexcrichton:
Looks like Windows was stack overflowing due to an invalid module being encoded. The invalid module stack overflows in debug mode but presents a first-class error in release mode. Instead of going too deep into fixing that I've updated this to skip `encode()` for `assert_malformed` modules

fitzgen (approved):

PR #1777, opened 2024-09-11
Correct float docs for `wasmparser::BinaryReader`

samestep:
This PR corrects the docstrings for `wasmparser::BinaryReader::read_f32` and `wasmparser::BinaryReader::read_f64`, which were previously identical to each other and both incorrect.

### DIFF
--- a/crates/wasmparser/src/binary_reader.rs
+++ b/crates/wasmparser/src/binary_reader.rs
@@ -696,21 +696,19 @@ impl<'a> BinaryReader<'a> {
         Ok((result << ashift) >> ashift)
     }
 
-    /// Advances the `BinaryReader` up to four bytes to parse a variable
-    /// length integer as a 32 bit floating point integer, returned as `Ieee32`.
+    /// Advances the `BinaryReader` four bytes to parse a 32 bit floating point
+    /// number, returned as `Ieee32`.
     /// # Errors
-    /// If `BinaryReader` has less than one or up to four bytes remaining, or
-    /// the integer is larger than 32 bits.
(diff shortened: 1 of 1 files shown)


samestep:
I'm guessing [this build error](https://github.com/bytecodealliance/wasm-tools/actions/runs/10812110855/job/29993017594) is unrelated?

```
error: missing documentation for a constant
  --> crates/wast/src/lexer.rs:65:1
   |
64 |   #[test]
   |   ------- in this procedural macro expansion
65 | / fn token_is_not_too_big() {
66 | |     assert!(std::mem::size_of::<Token>() <= std::mem::size_of::<u64>() * 2);
67 | | }
   | |_^
   |
note: the lint level is defined here
  --> crates/wast/src/lib.rs:53:9
   |
53 | #![deny(missing_docs, rustdoc::broken_intra_doc_links)]
   |         ^^^^^^^^^^^^
   = note: this error originates in the attribute macro `test` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `wast` (lib test) due to 1 previous error
```

alexcrichton:
Ah that looks like a nightly issue/regression, I'll work on getting that fixed, will let you know!

samestep:
@alexcrichton done! (And the same effect would have been achieved even if I had just pushed a dummy commit instead of actually merging/rebasing onto `main`, since the GitHub Actions `pull_request` event automatically creates a synthetic merge commit with the target branch before running CI.)

alexcrichton (approved):

PR #1781, opened 2024-09-11
Remove v1 encoding support for WIT documents/worlds

alexcrichton:
This commit removes support for the "v1" encoding of WIT packages into WebAssembly. This was originally migrated to v2 in #1252 and v2 was enabled by default in #1274. This finishes the transition by removing support for creating the old encoding.


crates/wit-component/src/encoding/wit.rs
```diff
@@ -57,11 +157,44 @@ impl Encoder<'_> {
         // decoding process where everyone's view of a foreign document agrees
         // notably on the order that types are defined in to assist with
         // roundtripping.
-        let mut interfaces = IndexSet::new();
-        for (_, id) in self.resolve.packages[self.package].interfaces.iter() {
-            self.add_live_interfaces(&mut interfaces, *id);
+        for (name, &id) in self.resolve.packages[self.package].interfaces.iter() {
```
dicej, line 160: Should the comment above this code be reworded given these changes?  It seems odd that the exact same comment is duplicated below in `encode_interface`; it seems like a better fit there than here.

dicej (approved):

PR #1784, opened 2024-09-12
Add importize method to Resolve

karthik2804:
A new method `importize` is added to the Resolve. This is to allow to mutate the Resolve to state where it would resemble what a consuming component would expect to see during composition.


alexcrichton:
I've rebased this on https://github.com/bytecodealliance/wasm-tools/pull/1785 and https://github.com/bytecodealliance/wasm-tools/pull/1782 to get better assertions as well as better testing utilities. With that I've added an "Updates to importize" commit which updates the method of doing this slightly and adds a few more tests as well. My hope is that it's a bit more robust to conditionally preserve imports instead of throwing them all away and recreating them. I've then also hooked this up to fuzzing.

@karthik2804 one thing I'm not certain of is that the design currently is to mutate the world in-place. I updated this to rename the world by adding `-importized` as a suffix, but that probably isn't great. Do you think that the name matters for the tooling integration you were envisioning? For example would it be better to create a new world and insert it into a provided package?

karthik2804:
I do not have strong opinions here. I do not think the name of the world should not matter.

alexcrichton:
Ok in that case let's leave this as-is and see how it works out.

Given that this is co-authored by me and @karthik2804 I'd prefer to not approve-and-merge, so @dicej I'm going to tag you in for a third set of eyes on this to make sure nothing looks too awry.

dicej (approved):

PR #1787, opened 2024-09-13
Don't panic on duplicate imports when creating components

### LINKED ISSUE #1786: wit-component: Component validation panics on duplicate imports
https://github.com/bytecodealliance/wasm-tools/blob/8f247b9e4accfe23166dca6f54ad3e6ce9458339/crates/wit-component/src/validation.rs#L177

This is validating correctly but since it can be triggered by real (albeit technically malformed) input wasm it should presumably be a regular error, not a panic.

alexcrichton:
Return an error instead since that's more appropriate in this context.

Closes #1786

### DIFF
--- a/crates/wit-component/src/validation.rs
+++ b/crates/wit-component/src/validation.rs
@@ -174,7 +174,13 @@ pub fn validate_module<'a>(
                                 Entry::Vacant(e) => e.insert(IndexMap::new()),
                             };
 
-                            assert!(map.insert(import.name, ty).is_none());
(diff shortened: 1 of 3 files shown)


dicej (approved):
Are there any tools we can recommend to the user to merge those duplicates?

alexcrichton:
Unfortunately not at this time. Sometimes it can be handled like in this tests's case where the imported item is the same, but it's also valid to import the same name as a func and then a global for example, and that can't necessarily be handled automatically and probably require source toolchain changes.

PR #1788, opened 2024-09-16
wasmprinter: print function bodies with code section

keithw:
wasmprinter has traditionally read the code section first and then printed each function's body as soon as it reaches its entry in the function section.

For printing instructions in folded form, it's going to be inconvenient to print code before having read the whole function section (or any of the tag section), because it's necessary to have parsed these sections to know the arities of the `call` and `throw` instructions.

This PR reorders the wasmprinter output to print function bodies with the code section.

(PR 1/3 of adding a folding option to wasmprinter)


alexcrichton (approved):
Sounds like a good idea to me, thanks! IIRC the motivation for doing it this way was historically wasm-tools tried to match wabt's output exactly and this is the order wabt printed things. Over time though the exact match test requirement was removed and so there's no longer any need to preserve this exact output, so I think it's reasonable to change.

For the test failures there's a few more files that  need to be "blessed", and you should be able to do it with:

```
BLESS=1 cargo test --workspace
```

(or `BLESS=1 cargo test -p wasm-compose` for just the exact test that's failing on CI, but I suspect at least the `wit-component` crate may also have some differences)

PR #1789, opened 2024-09-16
Annotate for_each_operator with op types or arities; simplify validator

keithw:
This PR adds annotations to the operators in the `for_each_operator!` macro. For most operators, the annotation specifies the exact type, but for the rest (which are the most interesting ones), the annotation only specifies the type arities. This is a very rough analog of WABT's [opcode.def](https://github.com/WebAssembly/wabt/blob/main/include/wabt/opcode.def) + [GetExprArity](https://github.com/WebAssembly/wabt/blob/main/src/ir-util.cc#L98-L271).

For the "exact type" annotations, the validator can use the annotation directly instead of needing a per-operator visitor. This allows removal of about 1,530 lines of code from crates/wasmparser/src/validator/operators.rs .

The PR includes a `operator_arity!` macro to compute the operator arities from the annotation. This will be useful when printing instructions in folded form. In a build with debug_assertions, the PR checks that this macro's output agrees with the validator's actual behavior (i.e. the number of operands popped from and pushed to the operand stack).

(This is independent of #1788, but PR 2/3 of adding a folding option to wasmprinter)


alexcrichton:
Up front it's definitely clear you've delved quite a bit into Rust macros in developing this, so thank you for taking the time to do that! I suspect it wasn't easy... Also, as I'm sure you've no doubt had suspicions of, macros aren't always the easiest to work with from an end-user perspective. Given that if you're ok with it I'd like to take a second to think about the high-level API design of this arity support in wasmparser itself. 

In the abstract if wasmparser were to only support an API of "give me the arity of this wasm instruction" then I think that might best be done with something like:

```rust
impl Operator {
   pub fn arity(&self, module: impl ModuleArity) -> Option<(u32, u32)>;
}

impl BinaryReader<'_> {
    pub fn operator_arity(&self, module: impl ModuleArity) -> Result<Option<(u32, u32)>>;
}
```

where my idea is that the public API wouldn't have to have the `operator_arity!` macro in it, it would just be an internal implementation detail of `wasmparser` itself. I like this in the sense that from an end-user perspective it's easier to use than macros typically.

For the `wasmprinter` use case I think this should be sufficient where wasmprinter would do something like:

```rust
while !reader.is_end_then_eof() {
    let arity = reader.operator_arity(&my_module_arity)?;
    reader.visit_operator(&mut printer_that_prints_to_temp_buffer)?;
    self.handle_folded_instruction(printer_that_prints_to_temp_buffer.buffer, arity)?;
}
```

(or something like that)

where the macro bits wouldn't necessarily need to leak into wasmprinter in per se.

---

Now otherwise though a major change in this PR is an update to the validator itself to use this new classification of instructions through the arity annotation. That would be much more difficult to separate from the macro-based interface of arity here because it's pretty heavily intertwined that more than just param/result counts are needed. Personally though I actually prefer the previous iteration of the validator where it's less macro-based. I realize that it's quite long line-wise and repetitive betwen ops that have similar type signatures, but I also like how it's easy to jump in to any individual opcode and hack on the validation easily. With an extra abstraction layer through macros it becomes more difficult to do this (still possible of course). I'd also be worried that for future contributors adding new instructions they'd have to not only understand the instructions and such themselves but also the macros and how to thread through the abstraction there to the validation logic needed (assuming the instruction doesn't fit in a preexisting shape).

So with all that put together, I'd like to ask about your thoughts on moving this in a different direction. What do you think about deferring the validator changes to a possibly future PR but preserving the arity changes? I realize you've probably already put quite a bit of work into this PR series itself and I realize you may not be keen on refactoring it in a large way like I'm proposing, so I'd also like to offer to help out where I can too. Additionally I'd want to confirm with you that a direction like this would still achieve the goal of folded instructions support for wasmprinter, because if it doesn't actually work for that then there's no use pursuing an alternative.

If you're ok deferring the validator bits to a future PR then there's still some other details to sort out too, but I don't want to go too far into that just yet (for example I like the debug-mode validation of the arity calculation done in the validator and I think it'd be worthwhile to still preserve that)

keithw:
Heya, I don't have a strong view and happy to adapt this to your preference (or if you want to take a crack at it, please feel free), but I think there's a few pieces that relate to one another from my point of view. Let me try to pull some of them apart:

1. One issue is about the API that exposes arity, e.g. a public `operator_arity!` macro (that can be called from an operator visitor) vs. a method on `BinaryReader`. I agree that the method would be nicer, but one challenge is that the consumer typically wants to get the arity information **in** the operator visit function, not later.

   E.g. it would be nice to keep the validator confirming that it pops and pushes the same number of operands as the claimed arity, to make sure this info doesn't bitrot. It seemed more convenient to do that in the visit function vs. trying to save something in longer-lived state and check it later (and the visit function is part of the validator's public API and is called various places, so there would be a risk the callsite never supplies the arity info and the debugging safeguard doesn't happen). The wasmprinter also probably wants to know the arity in its visit function; it *could* push the instruction to a temporary place and deal with it later when it learns the arity, but that seems less pleasant.
   
   Maybe this could be achieved if the signature of the operator visit functions were extended to provide something that could provide the arity (given a ModuleArity), along with the payload args? Not sure how feasible that would be.
   
2. Another issue is about how the arity information is represented in the source code. It seemed to make sense to add it as annotations in the `for_each_operator!` macro, since that's where there's already the list of all instructions, but it could be somewhere else (at the cost of duplicating the big list of instructions?). The format/DSL for the annotations themselves is also something that could be changed.

3. And then a third issue is, given that in many cases the type annotation is enough to fully specify the behavior of the validator, is it worth removing the 1,500 lines of duplicate-ish code? I agree this is orthogonal to anything having to do with arity or folding; if you prefer it the old way, I think it's pretty easy to go back. My weakly held personal view (as somebody who is new to Rust and this codebase) is that it's probably easier to understand the code when things are sort of done consistently across all Wasm instructions in a category, and it reduces the chance of bugs, vs. having a lot of similar functions that handle different variations of the same instruction. But it's easy to back out.

alexcrichton:
I do agree that wasmprinter would be a little awkward if this were a method-based approach. What I'm imagining it that it would "parse" the opcode once to get the arity and then again to print the actual opcode. One hope though is that the new folding instruction visitor wouldn't need to be a visitor at all but instead could reuse the non-folded opcode visitor. That way its version of "parse this instruction" would be to (a) acquire the arity and (b) do the normal print followed by (c) use those two results to manage the data structures you've added here.

I also think it's ok if the arity is only checked in the validator during normal validation, I don't think it's necessary to check arity in all usage of the validator. It's mostly just ensuring that the annotations are correct as opposed to ensuring that every instance of it is used correctly.

For how this is represented, I definitely agree it's best to keep "list all the instructions" in one location. We already do a pretty bad job of that (instructions are listed in a number of locations) and it's something I'd like to do better at, so I like your idea of putting arity in the `for_each_operator!` macro for now. I've got an idea of how to deduplicate more bits related to instructions, but that's orthogonal to this PR.

And finally overall I do agree that it's nice to see so much code removed from the validatory, but what I'm concerned about is the tradeoff in the complexity of the implementation. Rust macros aren't always the easiest to learn for folks coming into a project and the validator is one of the primary locations modified over time here as well (e.g. new proposals, new optimizations, etc). That makes me think that the tradeoff in reduction of lines here isn't worth the increase in complexity of the layer of macros on top. To be clear there's no technical reason that the macro system here won't work, and in fact this clearly shows it does work! Even for future proposals I can imagine it always working. Personally I feel that repeated code is easier to read and while you're right that it's easier to get wrong that's also why we have relatively extensive fuzzing.

alexcrichton:
Ok I've done a bit of work on this and https://github.com/alexcrichton/wasm-tools/commits/folding, or the last two commits there, are the shape of what I was thinking. I realize it's probably not quite as optimal in terms of runtime at parsing/printing but I think that's probably acceptable in this use case. I also haven't implemented the verification of arity annotations in the validator just yet, but wanted to share this to see what you think first.

alexcrichton:
I've been thinking a bit more about this, but here's another possible somewhat radical idea: since printing folder operators only works anyway if the wasm is basically valid one other way to implement this would be to just go ahead and run the validator. That would make it easy to track the arity by just recording the operand stack height before/after opcodes and wouldn't require otherwise manually annotating opcodes and verifying the annotations are correct. 

Do you think though that having a standalone arity calculation is otherwise useful to have? Or do you think this would help simplify the folding printing?

keithw:
I'm on board with your approach at https://github.com/alexcrichton/wasm-tools/commits/folding ! I went ahead and re-added the confirmation that the validator matches the arities and pushed it here: https://github.com/alexcrichton/wasm-tools/commits/folding

What's the preferred etiquette etiquette -- do you want me to force-push that here (and maybe close one of these two PRs) or make a merge commit, or...?

Re: running the validator in wasmprinter, I went down that road a little when I was originally writing this, but I didn't think you'd like the end result. The downsides seem to be:

1. Every build of wasmprinter would depend on a `validate` build of wasmparser
2. It makes it impossible (or hard) to print out an invalid module, whereas otherwise it's possible, even if the way the instructions are folded may not be very helpful.
3. It's annoying to implement the WasmModuleResources trait from the information maintained by a PrintOperator.
4. The current validator sometimes does extra pops and matching pushes (e.g. the current implementations of `visit_br_table` and `visit_try_table` before my last commit), so unless you have an outside human-readable reference for the "intended" arities of the instruction, I think it's possible to get bugs where (if you just go by what the validator does to the stack) an instruction can appear to have more params and results than it really does, even when the validator is otherwise 100% correct.

alexcrichton:
> What's the preferred style

How about closing this PR and updating https://github.com/bytecodealliance/wasm-tools/pull/1790 with the contents of the new branch? I have no strong preference though, so happy to defer to whatever feels most natural to you.

As for the validator, alas! Sounds like it wouldn't work without some extra heroics which makes it less appealing. In that case I'd say let's stick with the macro for now and merge this with folding, and then revisit what to do about the validator.

keithw:
Sounds good, closing this one and let's use #1790.

PR #1790, opened 2024-09-16
wasmprinter: add option to print instructions in folded form

keithw:
~~This is a draft PR that depends on #1788 and #1789.~~

This PR adds an option to wasmprinter to print instructions in [folded form](https://webassembly.github.io/spec/core/text/instructions.html#folded-instructions), which some people find more comprehensible. It uses the arity information ~~provided by #1789~~ (which is only computable once the earlier sections have been read, as in #1788) to group instructions with their operands.

(PR 3/3 of adding support for printing instructions in folded form)


tests/roundtrip.rs
```diff
@@ -172,6 +173,29 @@ impl TestState {
...
+            folding_printer.fold_instructions(true);
+            folding_printer
+                .print(contents, &mut PrintFmtWrite(&mut folded_string))
+                .context("failed to print wasm in folded form")?;
+            self.bump_ntests();
+            // Snapshot the folded WAT as well.
+            self.snapshot("print-folded", test, &folded_string)
+                .context("failed to validate the `print-folded` snapshot")?;
+            self.bump_ntests();
+        }
```
alexcrichton, line 197: Question for you, how useful was this during development? Do you think it's worthwhile to have both the normal and folded forms checked in under `snapshots`? I occasionally second-guess having `snapshots` for all the spec tests since they're so large and this is definitely increasing the size, but I wanted to see if you found this useful during development and whether it helped uncover bugs.

The alternative to this would be to add tests under `tests/cli/*.wat` for various forms of folded instructions (e.g. the `fold-multi.wat` added above). That has much less test coverage though.

Another possible alternative though, which I actually think is probably a good idea regardless, is that it'd be possible to hook things up to the fuzzers where `fuzz/src/print.rs` could get the ability to toggle the folded option and `fuzz/src/validate_valid_module.rs` could perform a number of the same assertion checks that this test module does over there as well.

Well in any case no need to handle all the fuzzing bits here, I can also tackle that later. For now though I'm mostly curious how useful you found these were during development.

crates/wasmprinter/src/operator.rs
```diff
@@ -1,36 +1,81 @@
...
+            label_indices: Vec::new(),
+            sep,
+        }
+    }
+}
+
+pub struct PrintOperator<'printer, 'state, 'a, 'b> {
+    pub(super) printer: &'printer mut Printer<'a, 'b>,
+    state: &'state mut State,
+    operator_state: &'printer mut OperatorState,
```
alexcrichton, line 32: If you'd like I think it'd make sense to store this as `OperatorState` instead of `&mut OperatorState` since the state itself won't outlive the print operation (unlike `State` which does outlive). Having a pointer here works fine though, so mostly a stylistic thing.

crates/wasmprinter/src/operator.rs
```diff
@@ -1286,3 +1341,450 @@ impl<'a> VisitOperator<'a> for PrintOperator<'_, '_, '_, '_> {
...
+    (handle TryTable  $self:ident $ty:ident $inst:ident $params:ident $results:ident) => {$self.push_block($ty, FrameKind::TryTable, $inst)};
+    (handle If        $self:ident $ty:ident $inst:ident $params:ident $results:ident) => {$self.push_if($ty, $inst)};
+
+    (handle Else      $self:ident $ty:ident $inst:ident $params:ident $results:ident) => {$self.handle_else()};
+    (handle End       $self:ident $ty:ident $inst:ident $params:ident $results:ident) => {$self.handle_end($results)};
+
+    (handle Try       $self:ident $ty:ident $inst:ident $params:ident $results:ident) => {bail!("legacy-exceptions not supported")};
+    (handle Catch     $self:ident $ty:ident $inst:ident $params:ident $results:ident) => {bail!("legacy-exceptions not supported")};
+    (handle CatchAll  $self:ident $ty:ident $inst:ident $params:ident $results:ident) => {bail!("legacy-exceptions not supported")};
+    (handle Delegate  $self:ident $ty:ident $inst:ident $params:ident $results:ident) => {bail!("legacy-exceptions not supported")};
```
alexcrichton, line 1493: If you'd like to implement support for this you're more than welcome to do so, but if you'd prefer to leave this out that's also totally fine. Regardless I agree that as-is is ok where an error is returned.

crates/wasmprinter/src/operator.rs
```diff
@@ -1286,3 +1341,450 @@ impl<'a> VisitOperator<'a> for PrintOperator<'_, '_, '_, '_> {
...
+                    results,
+                    offset: self.operator_state.op_offset,
+                };
+                if let Some(hint) = self.branch_hint.take() {
+                    inst.folded.push(hint);
+                }
+                stack.folded.push(inst);
+            }
+            _ => bail!("instruction without enclosing block"),
+        }
```
alexcrichton, line 1590: As as a stylistic note one thing I like to do in situations like this to help reduce indentation is:

```rust
let stack = match self.control.last_mut() {
    Some(stack) => stack,
    None => bail!("..."),
};
// ...
```

that way it helps to avoid the nested indentation becoming too much.

Also, I'd generally recommend where possible to explicitly list out the "other cases" in a match rather than using `_`, for example here using `None` instead of `_`. For `Option<T>` it doesn't matter a whole lot because that'll never get another variant, but in general it helps with refactorings where if the match expression changes it's a good double-check to ensure each arm is updated as well

crates/wasmprinter/src/operator.rs
```diff
@@ -1286,3 +1341,450 @@ impl<'a> VisitOperator<'a> for PrintOperator<'_, '_, '_, '_> {
...
+            OperatorSeparator::None => (),
+            OperatorSeparator::NoneThenSpace => *sep = OperatorSeparator::Space,
+            OperatorSeparator::Space => printer.result.write_str(" ")?,
+        }
+
+        printer.result.write_str("(")?;
+        printer.result.write_str(&inst.plain)?;
+        if inst.folded.is_empty() && inst.plain.contains(";;") {
+            printer.newline(inst.offset)?;
+        }
```
alexcrichton, line 1625: Mind leaving a small comment as to what this is handling?

crates/wasmprinter/src/operator.rs
```diff
@@ -1286,3 +1341,450 @@ impl<'a> VisitOperator<'a> for PrintOperator<'_, '_, '_, '_> {
...
+						 code_section_hints: Vec::new() };
+
+	    let mut op_printer = PrintOperator::new(&mut internal_printer, self.state, self.operator_state);
+	    define_visit!(before_op op_printer $op);
+
+	    op_printer.push_str(define_visit!(name $op))?;
+            $(
+                define_visit!(payload op_printer $op $($arg)*);
+            )?
+            define_visit!(after_op op_printer $op);
```
alexcrichton, line 1466: I think this is a duplication of the `visit_*` methods of `PrintOperator`, right? If so would it be possible to instead do `op_printer.$visit($($arg),*)?;` here instead? (or something like that)

crates/wasmprinter/src/operator.rs
```diff
@@ -1286,3 +1341,450 @@ impl<'a> VisitOperator<'a> for PrintOperator<'_, '_, '_, '_> {
...
+					   name_unnamed: self.printer.config.name_unnamed,
+					   fold_instructions: false };
+	    let mut internal_printer = Printer { config: &internal_config,
+						 result: if use_color { &mut buf_color } else { &mut buf_nocolor },
+						 nesting: self.printer.nesting,
+						 line: self.printer.line,
+						 group_lines: Vec::new(),
+						 code_section_hints: Vec::new() };
+
+	    let mut op_printer = PrintOperator::new(&mut internal_printer, self.state, self.operator_state);
```
alexcrichton, line 1459: As you've probably grown accustomed to at this point I find that code in macros is best minimized where possible, and the above bits look like they might be ripe for extraction into a separate function. Perhaps something like:

```rust
fn with_op_printer<R>(&mut self, f: impl FnOnce(PrintOperator<'some, 'lifetimes, Here, Unsure>) -> R) -> R {
    // ..
}
```

which should help keep the amount of stuff in this macro pretty small hopefully

alexcrichton (commented):
Thank you again so much for working on this, I'm quite happy to see this implemented and I know I'll definitely find it very useful day-to-day myself :)

crates/wasmprinter/src/lib.rs
```diff
@@ -68,8 +70,10 @@ struct Printer<'cfg, 'env> {
 struct CoreState {
     types: Vec<Option<SubType>>,
     funcs: u32,
+    func_to_type: Vec<Option<u32>>,
```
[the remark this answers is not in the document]
keithw: > To help prevent some divergence of sources of truth, could funcs be removed here and replaced with a fn funcs(&self) -> u32 { self.func_to_type.len() as u32 } method?

It's not totally trivial, because `CoreState::funcs` is incremented when reading each `Payload::CodeSectionEntry` to keep track of the current `func_idx` (whereas by that point, `func_to_type` has already learned all the function types in the module). But there's probably a simpler way to represent this somehow... let me work on it.

crates/wasmprinter/src/lib.rs
```diff
@@ -1262,6 +1280,10 @@ impl Printer<'_, '_> {
         let func_idx = state.core.funcs;
         self.print_name(&state.core.func_names, func_idx)?;
         self.result.write_str(" ")?;
+        let ty = match state.core.func_to_type.get(func_idx as usize) {
+            Some(Some(x)) => *x,
+            _ => bail!("invalid function type"),
```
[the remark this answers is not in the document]
keithw: Done.

tests/roundtrip.rs
keithw: No strong feelings. It was helpful when developing the feature, since I could see if I was changing anything (or what I was changing)... but not sure it needs to be committed upstream for all time, vs. a smaller selection of folding examples as you propose. Maybe the nice thing would be to be able to do a regression test privately, where you give two Git hashes and it tells you what if anything has changed across them (running the whole testsuite), without checking it all in to the repo.

I do think it's very nice to be checking the binary roundtrippability of the folded text format across the whole testsuite (even if it doesn't check the folded text against a snapshot).

crates/wasmprinter/src/operator.rs
keithw: It's a little tricky because the `PrintOperatorFolded` also constructs a temporary `PrintOperator` in its visitor function and wants to let it borrow (but not consume) its `OperatorState`.

crates/wasmprinter/src/operator.rs
keithw: wasm-tools doesn't parse the folded text format for legacy-exceptions, so it seemed like a pain to implement it and I wasn't sure I'd get it right without tests. :-)

crates/wasmprinter/src/operator.rs
keithw: Done.

crates/wasmprinter/src/operator.rs
keithw: Done.

```rust
// Wasm line comment (e.g. label annotation) shouldn't comment out the closing parenthesis
```

crates/wasmprinter/src/operator.rs
keithw: Done!

crates/wasmprinter/src/operator.rs
keithw: Ack -- will take a stab at it.

crates/wasmprinter/src/lib.rs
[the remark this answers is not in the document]
alexcrichton: Ah ok if it's not easy to do it's ok to leave this as-is. It's not the end of the world to ensure that both are mutated at the same time. I can always try to refactor things as well after this lands.

tests/roundtrip.rs
alexcrichton: Ok how about this:

* Enhance `roundtrip.rs` to ensure that round-tripping through the text format tests both the folded and non-folded forms.
* Run the "is whitespace pretty" check for both folded and non-folded forms.
* Add selective tests to `tests/cli/*.wat` as necessary.
* Leave out folded versions of `snapshot` tests for now.
* (future PR) add support to the fuzzer for testing round-tripping with folded forms

That way we get a strong level of baseline support for correctness and it's just stylistic bits that may fall through the cracks?

crates/wasmprinter/src/operator.rs
alexcrichton: Ah ok it's fine to leave as-is in that case 👍

keithw:
(Let me rebase to support stack switching and then I'll mark this ready for review.)

tests/roundtrip.rs
keithw: Done (minus the fuzzing support).

crates/wasmprinter/src/operator.rs
keithw: I took your commit from ebd6de2232b786aa364df0319ce33f4803774c99.

alexcrichton (approved):
Looks great to me, thank you again so much for spearheading the implementation of this!

PR #1792, opened 2024-09-17
 Maintain more internal consistency checks when merging worlds 

alexcrichton:
This commit fixes two issues, both preexisting, which represented
violated invariants of a `Resolve` after worlds were merged. World
merging is used during component generation and might be used for
bindings generation so it's important to maintain the various dynamic
invariants of `Resolve` to ensure that surrounding tooling works. An
example of this is that printing a merged world doesn't produce
parseable WIT today because some of the internal invariants aren't
respected.

This commit beefs up the `Resolve::assert_valid` function and then
applies necessary fixes to resolve these assertion failures on
preexisting test cases. The main difference is that anonymous
interfaces/types are now "cloned" when they move from one world to
another. This is required to ensure that all the various links between
these items are consistent and point to the right place. This cloning
infrastructure may end up being more generalized in the future if the
need arises, but for now it's quite limited.

### DIFF
--- a/crates/wit-parser/src/resolve.rs
+++ b/crates/wit-parser/src/resolve.rs
@@ -19,6 +19,8 @@ use crate::{
     WorldKey, WorldSpan,
 };
 
+mod clone;
+
 /// Representation of a fully resolved set of WIT packages.
 ///
 /// This structure contains a graph of WIT packages and all of their contents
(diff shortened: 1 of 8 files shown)


alexcrichton:
I'll note that this is stacked on https://github.com/bytecodealliance/wasm-tools/pull/1791 so the first commit can be ignored.

PR #1793, opened 2024-09-17
Update some fuzzing defaults and infrastructure

alexcrichton:
* Update `wasm_smith::Config` to default-enable some stage4+ proposals: `exceptions`, `gc`, `reference_types`, `relaxed_simd`, `simd`, `tail_call`, `threads`. These can still all be disabled via configuration and CLI flags.
* All stage4+ proposals are now swarm-enabled through `Arbitrary for Config`
* Default generation of modules in wasm-tools's own fuzzing no longer special-cases these proposals since they're all already handled.
* The `WasmFeatures` used for validating fuzz-generated modules now starts with a minimal baseline set of features to ensure that all proposals are disabled in the validator if the corresponding wasm-smith configuration flag is disabled.
* The `wasm-mutate` crate was updated to return errors instead of panicking for unsupported wasm proposals. All wasm proposals are now enabled when passing to `wasm-mutate`.

The primary motivation for this commit was this last point where I'm seeing panics on OSS-Fuzz for Wasmtime using `wasm-mutate` as a mutation hook because `wasm-mutate` is panicking on some GC types. When fixing that I noticed other fuzz-related things I wanted to clean up while I was here.

### DIFF
--- a/crates/fuzz-stats/src/bin/failed-instantiations.rs
+++ b/crates/fuzz-stats/src/bin/failed-instantiations.rs
@@ -99,6 +99,9 @@ impl State {
         let mut config = wasm_smith::Config::arbitrary(&mut u)?;
         config.allow_start_export = false;
 
+        config.exceptions_enabled = false; // Not implemented by Wasmtime
+        config.threads_enabled = false; // not enabled by default in Wasmtime
+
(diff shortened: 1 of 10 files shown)


fitzgen (approved):
Nice!

PR #1794, opened 2024-09-17
Start removing custom translation in wasm-mutate

alexcrichton:
Now that `wasm_encoder::reencode` exists that's the better option to use for reencoding modules. This commit removes one of the major users of the equivalent `Translator` trait in `wasm_mutate` to start down the path of removing that trait entirely and instead relying on the support in `wasm_encoder`.


fitzgen (approved):
Nice!

PR #1796, opened 2024-09-18
Remove custom translation framework from wasm-mutate

alexcrichton:
This commit removes the final users of the `Translator` trait and migrates them to `wasm_encoder::reencode`. Additionally when fuzzing this some minor issues have additionally been fixed in wasm-mutate.


fitzgen (approved):
Very nice

Issue #1774, opened 2024-09-10
Deduplicate interfaces by version in wit-component instead of importing multiple versions


alexcrichton:
For example if `wasi:io/poll@0.2.0` and `wasi:io/poll@0.2.1` are both imported only import the latter in the final component and use that to satisfy the imports of the core wasm module.

Motivated by [this discussion on Zulip](https://bytecodealliance.zulipchat.com/#narrow/stream/219900-wasi/topic/PSA.3A.20wasi.3Aio.400.2E2.2E0.20and.20wasi.3Aio.400.2E2.2E1.20don't.20mix.20well/near/469159187)

PR #1800, opened 2024-09-18
 Extract "world elaboration" to a separate function 

alexcrichton:
This commit extracts the process of elaborating a worlds imports/exports
to a dedicated function. This function ensures the transitive closure of
all imports/exports are listed in the world in proper topograhical
order. This additionally validates that the world has a coherent
definition given WIT today, notably that imports don't accidentally try
to use exported types.

This functionality was all present previously during the process of
taking a WIT document and creating `Resolve`. This moves the logic
around and refactors it slightly given its new surroundings. This
reorders a few imports/exports in worlds from what was previously
present but the underlying meaning of worlds should be the same.

The goal of this commit is to enable `Resolve`-mutating transformations
to not need to preserve all these invariants each time a world is
modified. Instead a world can be modified and then this function can be
called to "clean up" the world afterwards.


alexcrichton:
The impact of this new function shows up a bit in the second commit here where `importize` is simplified with this operation, but my main motivation is tackling https://github.com/bytecodealliance/wasm-tools/issues/1774 as my work-in-progress to implement that is getting pretty unwieldy trying to maintain these invariants.

dicej (approved):

PR #1802, opened 2024-09-18
Add support for the stack switching proposal

dhil:
This patch implements the [stack switching proposal](https://github.com/WebAssembly/stack-switching) (which at the time of writing is a [phase 2 proposal](https://github.com/WebAssembly/proposals)). 

The proposal adds a new composite type:
* `cont $ft` indexed by a function type, which is the type of a continuation.

Two new abstract continuation heap types and their subtyping hierarchy are added:
* `cont` the top type;
* and `nocont` the bottom type, i.e. `nocont <: cont`.

Six new instructions:
* `cont.new $ct`, parameterised by a continuation type index, which creates a new continuation.
* `cont.bind $ct-0 $ct-1`, parameterised by two continuation type indices, which partially applies a continuation of type `$ct-0`, resulting in a continuation of type `$ct-1`.
* `suspend $tag`, parameterised by a tag, which suspends the current context.
* `resume $ct (on $tag $label)*`, parameterised by a continuation type index and a jump table which maps tags to labels (similar to how `(catch $tag $label)` works in the exception handling proposal); this instruction resumes a continuation.
* `resume_throw $ct $tag (on $tag-1 $label)*`, parameterised by both a continuation type index and a tag index (in addition to the jump table). This instruction injects the exception tag `$tag` into the continuation argument.
* `switch $ct $tag`, parameterised by a continuation type index and a tag index. This instruction performs a direct context switch to a given continuation (i.e. a symmetric transfer of control).

The wellformedness condition for tags is relaxed, such that they are allowed to have a non-empty result type. This result type is used to validate the types of return values of `suspend $tag`.

The features of the proposal are gated under a new feature flag `stack-switching` (off by default). Most of the additions are test artifacts. The implementation of the proposal itself is modest.

<hr>

There are a few things in this patch that I am not quite sure about:

1. I have used a `UnpackedIndex` in the representation of the new indexed continuation type. In the validator I sometimes have to work with `CoreTypeId` and other times with raw `u32` indices. I am not convinced this is the best approach.
2. In validating `cont.bind` and the jump tables (aka resume tables) on `resume` and `resume_throw`, I have to construct and manipulate the underlying function types of continuations, in particular, I have to check subtyping relations between parts of these function types (c.f. `cont.bind` where I have to drop a prefix of the parameter list). So I have had to essentially reimplement function subtype checking. I have done this via a function `is_subtype_many` that applies the `is_subtype` function pointwise to two equally sized collections. I am not particularly fond of this approach.
3. I have not extended the `TagKind` structure. Though, an argument can be made for adding a further kind, lets call it `Control`, to differentiate exception and control tags internally. But I am not really sure whether the distinction is helpful.


crates/wasmparser/src/validator/operators.rs
```diff
@@ -1488,6 +1559,83 @@ where
...
+    }
+
+    /// Applies `is_subtype` pointwise two equally sized collections
+    /// (i.e. equally sized after skipped elements).
+    fn is_subtype_many(
+        &mut self,
+        skipcnt1: usize,
+        ts1: &[ValType],
+        skipcnt2: usize,
+        ts2: &[ValType],
```
alexcrichton, line 1631: Instead of taking a `skipcntN` could this only take two slices? It should be possible to chop off the front of the slices in callers explicitly if necessary

crates/wasmparser/src/resources.rs
```diff
@@ -50,6 +50,9 @@ pub trait WasmModuleResources {
     /// The sub type must be canonicalized.
     fn sub_type_at(&self, type_index: u32) -> Option<&SubType>;
 
+    /// Returns the `SubType` associated with the given core type id.
+    fn sub_type_at_id(&self, id: CoreTypeId) -> Option<&SubType>;
```
alexcrichton, line 54: Unlike the other method in this trait which are called with possibly-arbitrary indices we know that `CoreTypeId` is always trusted so it's ok to return `&SubType` from this instead of `Option<_>`

crates/wasmparser/src/validator/operators.rs
```diff
@@ -1416,6 +1428,65 @@ where
...
+            bail!(
+                self.offset,
+                "non-continuation type {}",
+                match key {
+                    Either::A(at) => at as usize,
+                    Either::B(id) => TypeIdentifier::index(&id),
+                }
+            )
+        }
+    }
```
alexcrichton, line 1457: I might recommend splitting this into `cont_type_at` which takes a `u32` and `const_type_at_id` which takes a `CoreTypeId`. I think that might be a bit easier to work with instead of working with `Either` perhaps? Functions like `func_type_of_cont_type_at` below could take `&ContType` instead of `TypeKey` in that case to let the caller call the appropriate `const_type_at*` function and then pass the result into later parts of validation.

crates/wasmparser/src/validator/operators.rs
```diff
@@ -1416,6 +1428,65 @@ where
         }
     }
 
+    fn cont_type_at(&self, key: TypeKey) -> Result<&ContType> {
+        let sub_ty = match key {
+            Either::A(at) => self.sub_type_at(at)?,
+            Either::B(id) => match self.resources.sub_type_at_id(id) {
+                None => bail!(self.offset, "unknown type {}", TypeIdentifier::index(&id)),
```
alexcrichton, line 1435: Here we won't want to surface `TypeIdentifier::index` since that's just a purely internal index during validation. In these cases it's ok to jus say "unknown type" or "non-continuation type" or something like that without mentioning the index.

crates/wasmparser/src/validator/operators.rs
```diff
@@ -1416,6 +1428,65 @@ where
...
+                        bail!(
+                            self.offset,
+                            "non-function type {}",
+                            TypeIdentifier::index(&func_id)
+                        )
+                    }
+                }
+            },
+        }
+    }
```
alexcrichton, line 1488: This function, if it takes `&ContType`, I believe can be infallible (e.g. unwraps/expects/etc). The validation that `ContType` always points to a function should have already been validated when the type was added to the module, so here's it's safe to assume that invariant is always upheld.

alexcrichton (commented):
Thanks so much again for working on this here and being willing to rebase your work while it was in development, it's much appreciated! I've left an initial round of comments below but I'll want to dig in more later to some of your concerns around duplication of function type subtyping as well after they're handled. Feel free to push back on anything I'm saying below as well that doesn't make sense too

crates/wasmparser/src/readers/core/operators.rs
```diff
@@ -477,3 +477,54 @@ impl<'a> FromReader<'a> for Catch {
...
+    fn from_reader(reader: &mut BinaryReader<'a>) -> Result<Self> {
+        Ok(match reader.read_u8()? {
+            0x00 => Handle::OnLabel {
+                tag: reader.read_var_u32()?,
+                label: reader.read_var_u32()?,
+            },
+            0x01 => Handle::OnSwitch {
+                tag: reader.read_var_u32()?,
+            },
+            x => return reader.invalid_leading_byte(x, "on"),
```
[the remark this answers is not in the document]
dhil: How about "on clause"?

crates/wasm-encoder/src/core/types.rs
```diff
@@ -611,6 +625,13 @@ impl<'a> CoreTypeEncoder<'a> {
         }
     }
 
+    fn encode_cont(&mut self, ty: &ContType) {
+        self.bytes.push(0x5d);
+        // Note that this is encoded as a signed type rather than unsigned
+        // as it's decoded as an s33
+        i64::from(ty.0).encode(self.bytes);
```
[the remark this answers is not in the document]
dhil: I think, I copy-pasted it from my downstream implementation, where I think I copy pasted it from how `HeapType::Concrete` is encoded. Actually, I think I modelled a lot of `Cont` on how `Concrete` works for heap types. But maybe it fine to encode and decode continuation types as `u32`s?

crates/wasmparser/src/validator/core/canonical.rs
```diff
@@ -246,6 +246,35 @@ pub(crate) trait InternRecGroup {
...
+                    UnpackedIndex::Module(idx) => self.type_id_at(idx, offset)?,
+                    UnpackedIndex::RecGroup(_) => unreachable!(),
+                };
+                match types[id].composite_type.inner {
+                    CompositeInnerType::Func(_) => (),
+                    _ => {
+                        return Err(BinaryReaderError::new(
+                            crate::validator::format!("non-function type {}", id.index()),
+                            offset,
+                        ))
```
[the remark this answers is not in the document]
dhil: Yes, absolutely! I'll fix that.

crates/wasmparser/src/validator/core/canonical.rs
```diff
@@ -246,6 +246,35 @@ pub(crate) trait InternRecGroup {
...
+                if !features.gc_types() {
+                    bail!(
+                        offset,
+                        "cannot define continuation types when gc types are disabled",
+                    );
+                }
+                // Check that the type index points to a valid function type.
+                let id = match t.0 {
+                    UnpackedIndex::Id(id) => id,
+                    UnpackedIndex::Module(idx) => self.type_id_at(idx, offset)?,
```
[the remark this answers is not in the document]
dhil: Good point! Thanks!

crates/wasmparser/src/validator/core.rs
```diff
@@ -946,7 +946,7 @@ impl Module {
             ));
         }
         let ty = self.func_type_at(ty.func_type_idx, types, offset)?;
-        if !ty.results().is_empty() {
+        if !ty.results().is_empty() && !features.stack_switching() {
```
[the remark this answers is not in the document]
dhil: Yes, good point! That's an oversight from me.

crates/wasmprinter/src/operator.rs
```diff
@@ -667,6 +691,35 @@ macro_rules! define_visit {
...
+        $self.push_str(" ")?;
+        $self.printer.print_idx(&$self.state.core.type_names, $type_index)?;
+        $self.tag_index($tag_index)?;
+        $self.resume_table($table)?;
+    );
+    (payload $self:ident Switch $type_index:ident $tag_index:ident) => (
+        $self.push_str(" ")?;
+        $self.printer.print_idx(&$self.state.core.type_names, $type_index)?;
+        $self.tag_index($tag_index)?;
+    );
```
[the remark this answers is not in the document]
dhil: Yes, I'll fix that.

crates/wasmparser/src/validator/operators.rs
dhil: Ah yes, it is much prettier. Thanks!

crates/wasmparser/src/readers/core/types.rs
```diff
@@ -776,6 +797,27 @@ pub struct StructType {
     pub fields: Box<[FieldType]>,
 }
 
+/// Represents a type of a continuation in a WebAssembly module.
+#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
+pub struct ContType(pub UnpackedIndex);
```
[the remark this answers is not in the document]
dhil: Excellent suggestion! I like this much better. Thanks!

crates/wasmparser/src/validator/operators.rs
dhil: Great suggestion, thanks! After changing the index representation on `Cont` to `PackedIndex` everything else kind of falls into place. It is much nicer.

crates/wasm-encoder/src/core/types.rs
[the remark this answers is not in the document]
dhil: Actually, I retract the above. I looked at the reference implementation. There we encode it as a `s33`. But we have written in the explainer/informal spec that it is supposed to be encoded as a `typeidx`, which I believe per the spec should be encoded as an `u32`. So, there seems to be an inconsistency here. I will chat with @rossberg about it.

dhil:
Thanks for your feedback! I think the key change was to represent the index on continuation types as `PackedIndex`. It makes everything much nicer.

crates/wasm-encoder/src/core/types.rs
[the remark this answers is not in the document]
alexcrichton: I believe `s33` is used in binary encoding locations where a positive value indicates "this is a type index" and a negative value indicates "this is actually a construct which isn't a type index but it's a structure to parse". It's needed to make the binary representation more compact by not needing a prefix byte for differentiating between the two. In the context of `Cont`, though, I don't think it's necessary because the prefix byte of `Cont`, 0x5d, has already been pushed so there's no ambiguity of what the next item is. That's what makes me think that this may want to be a u32 instead of an s33, but I wanted to confirm to be sure

crates/wasmparser/src/readers/core/types.rs
```diff
@@ -776,6 +797,24 @@ pub struct StructType {
...
+    pub(crate) fn remap_indices(
+        &mut self,
+        map: &mut dyn FnMut(&mut PackedIndex) -> Result<()>,
+    ) -> Result<()> {
+        let mut idx = self.0;
+        map(&mut idx)?;
+        *self = ContType(idx);
+        Ok(())
+    }
+}
```
alexcrichton, line 816: If you'd like it's probably ok to omit this entirely and replace the caller with `map(&mut ty.0)?;`

crates/wasmparser/src/validator/core/canonical.rs
```diff
@@ -246,6 +246,29 @@ pub(crate) trait InternRecGroup {
...
+                    bail!(
+                        offset,
+                        "cannot define continuation types when gc types are disabled",
+                    );
+                }
+                // Check that the type index points to a valid function type.
+                let id = match t.0.as_core_type_id() {
+                    None => bail!(offset, "invalid continuation type index"),
+                    Some(id) => id,
+                };
```
alexcrichton, line 266: Here it's ok to `.unwrap()` the result since a failure here means a bug in wasmparser itself rather than something the user can fix with the input wasm module

crates/wasmparser/src/validator/operators.rs
```diff
@@ -1416,12 +1416,46 @@ where
...
+            Ok(cont_ty)
+        } else {
+            bail!(self.offset, "non-continuation type {at}",)
+        }
+    }
+
+    fn func_type_of_cont_type(&self, cont_ty: &ContType) -> Result<&'resources FuncType> {
+        let func_id = cont_ty.0.as_core_type_id().expect("valid core type id");
+        Ok(self.resources.sub_type_at_id(func_id).unwrap_func())
+    }
```
alexcrichton, line 1437: This looks like it should be possible to remove the `Result` return value and just return `&'resources FuncType` here

crates/wasmparser/src/readers/core/types/matches.rs
```diff
@@ -252,6 +259,40 @@ impl Matches for WithRecGroup<StorageType> {
...
+                        ida
+                    } else {
+                        types
+                            .at_canonicalized_unpacked_index(
+                                WithRecGroup::rec_group(a),
+                                ca.unpack(),
+                                usize::MAX,
+                            )
+                            .expect("type references are checked during canonicalization")
+                    };
```
alexcrichton, line 277: This looks like it may be a duplicate of `let core_type_id = |...| ...;` in validator/types.rs, would you be up for extracting that to a helper function on `TypeList` to help share the implementation between these locations?

alexcrichton: Alternatively what you could also perhaps do here is to manufacture a `RefType`  and then call `reftype_is_subtype_impl` perhaps?

crates/wast/src/core/binary.rs
```diff
@@ -342,6 +342,18 @@ impl From<&ArrayType<'_>> for wasm_encoder::ArrayType {
     }
 }
 
+impl Encode for ContType<'_> {
+    fn encode(&self, e: &mut Vec<u8>) {
+        self.0.encode(e);
+    }
+}
```
alexcrichton, line 349: If you'd like it should be ok to remove this now as it's no longer needed (just a minor thing though)

alexcrichton (commented):
Ok a few small comments here and there, but otherwise this looks good to go to me 👍 . Thanks again!

alexcrichton:
And no rush of course! Just wanted to check in

crates/wasm-encoder/src/core/types.rs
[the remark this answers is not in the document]
dhil: I think you are correct. I spoke with Andreas about this. We decided to go with the `s33` encoding as it may be more forwards compatible as it gives us some extension point for adding future attributes to continuation types (should we ever want that). So I have proposed this change on the spec repo https://github.com/WebAssembly/stack-switching/pull/89.

crates/wasmparser/src/readers/core/types/matches.rs
dhil: Yes! I opted for the second approach.

dhil:
Thanks Alex! I have implemented your feedback now. The issue of whether we will encode the index on continuation types as an `u32` or a `s33` is yet to be resolved on the spec repo. In the meantime I have reverted back to the `s33` encoding to stay in sync with the reference interpreter.

dhil:
Thanks for your help!

PR #1803, opened 2024-09-19
Transition more encoding in wast to using wasm-encoder

alexcrichton:
The only big section remaining is the code section with expressions. That'll require some more finesse so it's deferred to a future PR.


crates/wasm-encoder/src/core/branch_hints.rs
```diff
@@ -0,0 +1,72 @@
...
+
+/// Helper structure to encode the `metadata.code.branch_hint` custom section.
+///
+/// This section was defined in the branch-hinting proposal for WebAssembly:
+/// <https://github.com/WebAssembly/branch-hinting>.
+#[derive(Default, Debug)]
+pub struct BranchHints {
+    bytes: Vec<u8>,
+    num_hints: u32,
+}
```
fitzgen, line 12: Can you add an example to this doc comment, like the rest of the `wasm-encoder` sections have?

fitzgen (approved):
So good to see this finally happen!

PR #1804, opened 2024-09-19
Added a `len` method to `wasm_encoder::Module`

thesuhas:
Hi,

I have a use case where I would require to know the number of bytes of a module that has been encoded so far so I've added an API that does this.


alexcrichton (approved):
Thanks!

PR #1813, opened 2024-09-20
Flesh out more text support for GC types in components

### LINKED ISSUE #1808: Fill out support for GC types in the component model text format
With Andrew's recent refactorings and updates to binary encoding this should be possible to support now, so this should be filled out and implemented where necessary. I'm not sure if this will require too much work or if it's "just connect the wires", that'll be part of the investigation.

alexcrichton:
This commit builds on the work of #1764 to support the text format for GC types in components more than before. This notably bring support for:

* `(sub ...)` types in components and module types
* `(rec ...)` groups in components and module types
* types can refer to themselves like with core wasm

The main consequence of this work is that unlike most other `$foo` identifiers in the component model the identifiers found in types will not automatically inject outer aliases to refer to outer types. For example this will not parse:

    (component $C
      (type $t (struct))
      (component
        (type (array (ref $t)))
      )
    )

The reason for this is that automatic injection of an outer alias requires that types are resolved and then their names are registered. The resolution process queues up aliases to inject which are accounted for during registration when indices are assigned. Here though because types can refer to themselves (or future types in `rec` groups) the registration process has to happen first before resolution. This means that if resolution were to inject more type indices then that would mess up the indexes already assigned.

This is hopefully relatively minor in terms of how often this'll bite someone. For now various changes have been made to the name resolution pass of components to handle this and some tests have been added too for both positive and negative situations.

Closes #1808


fitzgen (approved):
Nice!

PR #1817, opened 2024-09-23
wasm-wave: Accept nullary func calls

### LINKED ISSUE #1816: wave: Can't parse function calls with empty arguments
Using [`UntypedFuncCall`](https://docs.rs/wasm-wave/latest/wasm_wave/untyped/struct.UntypedFuncCall.html) wave can't currently parse `f()` and yields the error `empty tuple at 1..3`. I think this is due to [function arguments being parsed as a tuple](https://github.com/bytecodealliance/wasm-tools/blob/6fc5d68b30931db5e9639a5cecf107fdf1bef894/crates/wasm-wave/src/parser.rs#L54) and tuples in general cannot be empty.

cc @lann

lann:
Fixes #1816


alexcrichton (approved):
Thanks!

PR #1818, opened 2024-09-24
wasm-smith: Add config option to disallow floats

Scott-Guest:
- Add finer-grained categories `InstructionKinds::NumericNonFloat`, `InstructionKinds::VectorNonFloat` and `InstructionKinds::MemoryNonFloat`
    - https://github.com/bytecodealliance/wasm-tools/pull/400#issuecomment-983114142
- Add a `wasm_smith::Config` option `allow_floats` toggling whether floating point types and instructions are generated.


crates/wasm-smith/src/core.rs
```diff
@@ -2813,15 +2836,18 @@ flags! {
...
-        Vector,
-        Reference,
-        Parametric,
-        Variable,
-        Table,
-        Memory,
-        Control,
-        Aggregate,
+        NumericNonFloat = 1 << 0,
+        Numeric = (1 << 1) | (1 << 0),
```
fitzgen, line 2840: Since we don't have a way to express subsets or instructions, can we make these options be `NumericFloat` and `NumericNonFloat`? I think that would make it clearer what is required to enable the generation of certain instructions. Otherwise people might think that enabling `Numeric` was enough to get integer instructions, but if I understand correctly, it would only enable the float instructions and not the integer instructions.

fitzgen (commented):
Thanks! Looks good to me, but one thing I'd like to change before landing.

crates/wasm-smith/src/core.rs
alexcrichton: Another possible solution would be to not change `InstructionKind` here but instead add a new condition guarding all float-related instructions on `Config::disallow_floats` perhaps?

crates/wasm-smith/src/config.rs
```diff
@@ -253,13 +253,24 @@ define_config! {
...
         /// `allowed_instruction()` does not include vector instructions, the
         /// generated programs will not include these instructions but could
         /// contain vector types.
         pub allowed_instructions: InstructionKinds = InstructionKinds::all(),
 
+        /// Indicates whether we should abstain from generating floating point
+        /// instructions and types.
+        ///
+        /// Defaults to `false`.
+        pub disallow_floats: bool = false,
```
alexcrichton, line 272: Would you be ok inverting this to `enable_floats`? (or `allow_floats` or similar). I find "positive" config features are generally easier to understand and rationalize how they compose rather than "negative" features. A minor thing though so doesn't matter too much either way

crates/wasm-smith/src/core.rs
Scott-Guest: > Otherwise people might think that enabling `Numeric` was enough to get integer instructions, but if I understand correctly, it would only enable the float instructions and not the integer instructions.

I might have overlooked something, but the intention is that `Numeric` allows all numeric instructions to be generated including both integers and floats. I tested a small example with `Numeric` as the only `allowed_instructions`, and it seems to work properly.

If the bits in the `FlagSet` are labeled `... b2 b1 b0`, then the flags are defined to set the following bits
- `Numeric`: `b1 b0`, `NumericNonFloat`: `b0`
- `Vector`: `b3 b2`, `VectorNonFloat`: `b2`
- `Memory`: `b9 b8`, `MemoryNonFloat`: `b8`
In each case then, setting the `Foo` flag also sets the `FooNonFloat` flag, and we have
```
assert!(InstructionKinds::new(&[InstructionKind::Numeric]).contains(InstructionKind::NumericNonFloat));
assert!(InstructionKinds::new(&[InstructionKind::Vector]).contains(InstructionKind::VectorNonFloat));
assert!(InstructionKinds::new(&[InstructionKind::Memory]).contains(InstructionKind::MemoryNonFloat));
```

crates/wasm-smith/src/config.rs
Scott-Guest: Sure, I switched it to `allow_floats`

crates/wasm-smith/src/core.rs
gtrepta: `Integral` instead of `NonFloat` might be a better name here.

alexcrichton: I think the overlapping-bits representation here would probably work (in contrast to @fitzgen's observation where `Numeric` as-written should generate integer instructions). As for the original intention here, the hope was to basically help fuzzing generate interesting modules by only giving it access to certain classes of instructions. In that sense the exact set of classes here and the exact definitions don't matter too too much.

Personally I'd advocate for not changing `InstructionKind` though, but only for the reason that I think it could be confusing to have `allow_floats: false` and then have float instructions generated because you might have also forgotten to disable float-related instruction kinds here. Or, @Scott-Guest, do you have a use case for disabling the f32/f64 type but enabling float-related instructions? If not then I'd personally prefer that `allow_floats: false` additionally disables all float-related instructions without having to additionally modify `InstructionKind` classes.

If we keep the `InstructionKind` changes, though, then I'd agree that we should probably split it into `Numeric{Int,Float}` or similar names.

crates/wasm-smith/src/core.rs
```diff
@@ -1472,25 +1472,28 @@ impl Module {
...
-            }
-            ValTypeClass::Ref => {
-                if self.config.reference_types_enabled {
-                    Ok(ValType::Ref(self.arbitrary_ref_type(u)?))
-                } else {
-                    Ok(ValType::I32)
-                }
-            }
+            ValTypeClass::V128 => Ok(ValType::V128),
+            ValTypeClass::Ref => Ok(ValType::Ref(self.arbitrary_ref_type(u)?)),
```
gtrepta, line 1496: `reference_types_enabled: false` no longer seems to work in my testing, and this looks related.

crates/wasm-smith/src/core/code_builder.rs
```diff
@@ -1239,7 +1239,11 @@ impl CodeBuilder<'_> {
         module: &Module,
     ) -> Result<Vec<Instruction>> {
         let max_instructions = module.config.max_instructions;
-        let allowed_instructions = module.config.allowed_instructions;
+        let allowed_instructions = if module.config.allow_floats {
+            module.config.allowed_instructions
+        } else {
+            module.config.allowed_instructions.without_floats()
+        };
```
Scott-Guest, line 1246: @alexcrichton If floats aren't allowed, we automatically replace each `InstructionKind::Foo`  with `InstructionKind::FooNonFloat`.

crates/wasm-smith/src/core.rs
Scott-Guest: > Or, @Scott-Guest, do you have a use case for disabling the f32/f64 type but enabling float-related instructions?

Nope - only use case I have is fully disabling anything related to floats. Changing the `InstructionKinds` was just an easy way to implement this.

> I think it could be confusing to have allow_floats: false and then have float instructions generated because you might have also forgotten to disable float-related instruction kinds here.

We won't generate float instructions if `allow_floats: false` regardless of the instruction kinds - see the other comment I tagged you in. 

> If we keep the InstructionKind changes, though, then I'd agree that we should probably split it into Numeric{Int,Float} or similar names.

I agree, but this is starting to feel pretty messy and non-orthogonal. There are really two concerns here:
- The general categories of instructions we generate
- The types of values that we generate

Let me try implementing a `allowed_value_types: ValTypes` config option instead - I'll open a separate PR.

crates/wasm-smith/src/core.rs
Scott-Guest: I believe this is fixed - try the latest commit.

crates/wasm-smith/src/core.rs
alexcrichton: Oh it's ok to not go to too much effort I think, you raise a good point that the `InstructionKind` enum here is already a pretty handy way to implement this today and works relatively well. Given that I'm pretty convinced that this is reasonable. Maybe modulo a small rename from `NumericNonFloat` to `NumericInt` perhaps? 

Otherwise though if @fitzgen agrees I think it'd be reasonable to merge.

fitzgen: SGTM :+1:

crates/wasm-smith/src/core.rs
gtrepta: I'm still seeing reference types.

I'm running this:
```
wasm-tools smith --config <(echo '{ "reference_types_enabled": false }') -t <(head -c 50 /dev/urandom)
```

Scott-Guest: Debugging the actual `config` passed to the `Module` shows that it still has `reference_types_enabled: true`, and seems to be completely ignoring whatever you pass to `--config`. Not sure why - this is unrelated to my changes and repros on main (@fitzgen).

crates/wasm-smith/src/core.rs
Scott-Guest: Sounds good, I renamed `NonFloat` to `Int`. Feel free to merge!

crates/wasm-smith/src/core.rs
alexcrichton: @gtrepta looks like you accidentally ran into https://github.com/bytecodealliance/wasm-tools/pull/1824 where the field name is `reference-types-enabled`, with `-` instead of `_`.

I'll note though that with that I'm occasionally seeing panics in `wasm-tools smith ...` on main (not released yet) so I'll look to track that down.

alexcrichton (approved):

crates/wasm-smith/src/core.rs
alexcrichton: The panic should be fixed in https://github.com/bytecodealliance/wasm-tools/pull/1825

PR #1821, opened 2024-09-25
Fix a race condition in some tests

### LINKED ISSUE #1820: cargo test sometimes fails with `Broken pipe (os error 32)` under aarch64
Hello!

This bug is not always reproducible, but in my environment happens around ~60% of the runs. One example of the failure:

```shell-session
❯ cargo test -p wasm-tools --all

...

test "tests/cli/semver-check-add-imports.wit"                                  ... ok
test "tests/cli/add-metadata.wat"                                              ... ok
test "tests/cli/world-merging-export-conflict.wit"                             ... ok

failures:

---- "tests/cli/unbundle-print-module1.wat" ----
failed test "tests/cli/unbundle-print-module1.wat"

Caused by:
    Broken pipe (os error 32)


failures:
    "tests/cli/unbundle-print-module1.wat"

test result: FAILED. 105 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.14s

error: test failed, to rerun pass `-p wasm-tools --test cli`
```

Sometimes it's this test, sometimes it's others. Another example: https://github.com/NixOS/nixpkgs/pull/341059#issuecomment-2343590484

What all failing tests have in common though, is that they use the tmpdir and pipe feature of the tests.

I can confirm that I could not reproduce the issue after setting `--test-threads=1` during the `cargo test` execution, like:

```
$ cargo test -p wasm-tools --all -- --test-threads=1
```

alexcrichton:
Some tests were modeled as "do a thing piped into another thing" but the second command didn't actually take any input. That meant that writing the stdout of the previous command into the next command might sometimes fail if the second command finishes before the write finishes. This commit fixes the various affected tests with the now-added ability to run multiple separate commands in a single file. These failures were detected by inserting a `sleep` before writing stdin and then fixing all tests.

Closes #1820


ereslibre:
I can confirm I have not been able to reproduce any issue reported in https://github.com/bytecodealliance/wasm-tools/issues/1820 with this PR. Thanks!

abrown (approved):

PR #1828, opened 2024-09-28
Implement support for the standard mangling scheme in the component model

alexcrichton:
This PR is a series of commits leading up to the conclusion of supporting the name mangling scheme described in https://github.com/WebAssembly/component-model/pull/378. This is intended to be "the future" for how core wasm modules are shaped in terms of their imports and exports. The highlights of this support are:

* Both the old and the new schemes are supported simultaneously. They've got no overlap and this shouldn't cause issues. This also means that this PR should break no one.
* Utilities and some helpers were added for bindings generators. It's expected that generators will "simply switch" to the new scheme and will not retain support for the "legacy" naming scheme.
* Lots of refactoring was needed internally in wit-component to further decouple baked-in assumptions about what names correspond to what.
* The `wasm-tools component embed --dummy` flag now uses the new naming scheme by default
* A new `--dummy-names legacy` flag can be used to enable the old names.
* A new `wasm-tools component new --reject-legacy-names` flag was added to help test code generators and make sure they're entirely using the "new scheme".

The current idea I have for a rollout plan of this is to get this into wasm-tools and places like `wasm-component-ld` and others. Once that's propagated for a bit start changing guest bindings generators to using the new mangling scheme. Once that's done, and quite some time has passed, consider removing support for the old mangling scheme or enabling `--reject-legacy-names` by default.


alexcrichton:
Also I'll preemptively apologize @dicej because this will further conflict with your async work rebase. I'm in no rush to merge this though so happy to rebase around you too.

dicej (approved):
LGTM.

BTW, I don't think BuildTargets.md says anything yet about name mangling for async imports, exports, callback functions, stream/future built-ins, etc.  When I rebase my async code onto this, should I just make something (plausible) up?  Or else just use `todo!` in the async parts of `impl NameMangling for Standard`, meaning you have to use `Legacy` if you want to experiment with async?

alexcrichton:
I think that's reasonable yeah, maybe not `todo!()` but `None`-returning things yeah. I think it's probably best to sketch out with "legacy" and get "standard" working later on once it's more official in the component model itself.

PR #1831, opened 2024-09-30
Hook up folded instructions to fuzzing

alexcrichton:
Extend the `validate_valid_module` fuzzer to print both in folded and non-folded form to ensure that the output binary is the same when re-parsed by `wast`.

While here also tweak some APIs and documentation.


fitzgen (approved):
Nice

Issue #1826, opened 2024-09-25
Release notes for this repository


alexcrichton:
Currently this repository does not have any release notes. I don't think that's great, however! This repository feels "big enough" to warrant release notes and releases are also frequent enough and packed enough that I think it's worthwhile to have an overview. I'd like to open an issue to discuss what it might be like to add release notes to this repository.

In the abstract the question of "should repository X have release notes" is "yes, of course!" but the problem is that writing release notes takes time and often specialized knowledge. Release notes are also (at least empirically in my experience) very easy to forget. This leads to a few drawbacks which I'd ideally like to have solved, but I'm not sure have an easy solution:

* Releases in this repository are currently on no strict cadence, and sometimes it's somewhat urgent that a release is pushed out. Ideally writing release notes wouldn't unduly block releases.
* If someone's interested in a release for a bugfix or feature, I don't think it's reasonable to ask that person to write release notes if they're not otherwise written.
* Personally I don't necessarily have the time to write release notes all the time. I'd find this easier if release notes weren't "urgent" in the sense of blocking a release. For example for Wasmtime there's a 2 week window after a branch is created until a release is made, and that's a nice buffer time to write release notes.
* Releases, however, are best paired with notes to accompany them. If someone gets a notification about a new release they're probably interested in a digest of what's new in that release.

Personally I've found that automatic release note aggregation tools generally aren't great. Most are based on commit messages which for a repository with external contributions is generally a drag to require everyone have the exact same format of commit messages and making sure everything is perfect. Starting with automated notes is great, but I find they almost always benefit from editing by a human.

Writing release notes after-the-fact though often requires specialized knowledge of PRs that were merged or changes made. This means that someone who wasn't involved probably can't be easily called up and expected to write release notes either.

---

Putting all of that together I'm left with the dilemma of I'm probably the best (if not only reasonable) candidate for writing release notes. The biggest problem for me is there's not a "grace period" where release notes can be written at a relaxed pace that doesn't block anyone's workflows otherwise.

I wanted to at least write up my thoughts on this to at least explain why I'm not currently writing release notes. If others have thoughts though on this I'd be happy to hear them! (maybe now's the time to create a formal release cadence for this repo...)

PR #1832, opened 2024-09-30
Auto-generate release notes for releases

alexcrichton:
While #1826 is figured out this is probably better than nothing, so set this flag to get GitHub's auto-generated release notes.


alexcrichton:
For an example of this see [1.217.0](https://github.com/bytecodealliance/wasm-tools/releases/tag/v1.217.0) vs [1.218.0](https://github.com/bytecodealliance/wasm-tools/releases/tag/v1.218.0) (I manually hit the button on 218 to generate release notes)

fitzgen (approved):

PR #1834, opened 2024-10-01
Fix an infinite loop in wasm-smith

alexcrichton:
This fixes a possible infinite loop in wasm-smith when the gc proposal is enabled. When the GC proposal is enabled and a minimum number of types were specified it could get wasm-smith stuck in an infinite loop where it would either always create an empty rec group or it would always try to clone an empty rec group. The fix here is to thread a boolean which toggles whether an empty rec group is allowed based on whether we're generating the minimum number of types or any extra types.

This is a relatively old bug so I'm not sure why this hasn't been discovered prior to this. Local fuzzing in Wasmtime found this quite quickly, so I may have just gotten unlucky. We also tend to not look at timeouts on OSS-Fuzz that closely as well, so perhaps this is a reminder to actually do that.

### DIFF
--- a/crates/wasm-smith/src/core.rs
+++ b/crates/wasm-smith/src/core.rs
@@ -175,6 +175,12 @@ pub(crate) enum DuplicateImportsBehavior {
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum AllowEmptyRecGroup {
+    Yes,
+    No,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 enum MaxTypeLimit {
     ModuleTypes,
     Num(u32),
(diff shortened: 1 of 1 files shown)


crates/wasm-smith/src/core.rs
```diff
@@ -586,20 +586,21 @@ impl Module {
         index
     }
 
-    fn arbitrary_rec_group(&mut self, u: &mut Unstructured) -> Result<()> {
+    fn arbitrary_rec_group(&mut self, u: &mut Unstructured, allow_empty: bool) -> Result<()> {
```
fitzgen, line 589: Can we create an `enum` with better names? The bool params are always hard to read at call sites.

fitzgen (approved):

crates/wasm-smith/src/core.rs
alexcrichton: How would you feel `let allow_empty = false;` at callsites? I also feel like it's a pretty big burden to create a brand new enum every time a `bool` is passed...

fitzgen: This one is passed around in a bunch of places so, while I definitely understand the drawback, it feels like it is amortized in this case. Not in the public api so shouldn't need docs either.

```rust
enum AllowEmptyRecGroup {
    Yes,
    No,
}
```

PR #1837, opened 2024-10-01
Remove duplicate API on `Types`

alexcrichton:
This commit removes the duplicate API of `Types` and `TypesRef` in the `wasmparser` crate by removing the API on `Types`. The source of truth is the one on `TypesRef` and maintaining this duplication is a bit of a burden as it's a lot of copy/pasted code. It's hoped that users can instead use `.as_ref()` "early on" which will enable the same usage as before. Ideally this would be a `Deref` impl but due to how `Deref` works that's not possible right now.


fitzgen (approved):
Good clean up

PR #1840, opened 2024-10-02
wasmprinter folding: change some panics to errors or successful roundtrips

### LINKED ISSUE #1838: Panic printing `if` with folded instructions where parameter is from block param

keithw:
Fixes #1838

### DIFF
--- a/crates/wasmprinter/src/operator.rs
+++ b/crates/wasmprinter/src/operator.rs
@@ -1699,13 +1699,16 @@ impl<'printer, 'state, 'a, 'b> PrintOperatorFolded<'printer, 'state, 'a, 'b> {
     }
 
     fn push_if(&mut self, ty: BlockType, plain: String) -> Result<()> {
-        let mut predicate = vec![self
+        let mut predicate = Vec::new();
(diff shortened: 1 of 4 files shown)


alexcrichton:
Thanks for this! Mind adding a test as well? Something under `tests/cli/*.wat` should be fine, probably with a `FAIL: print %` annotation.

alexcrichton (approved):
Perfect! Thanks!

PR #1842, opened 2024-10-02
Add WABT command aliases

fitzgen:
This adds a few aliases for WABT commands to make it easier to use `wasm-tools` for people that are used to the WABT tools. The aliases are:

* `wat2wasm` -> `parse`
* `wasm2wat` -> `print`
* `wast2json` -> `json-from-wast`

This allows usage like

```
$ wasm-tools wasm2wat foo.wasm
```

which prints the WAT disassembly of `foo.wasm` to `stdout`.


alexcrichton:
Nice! Mind adding a test for this too? Probably just one of the commands should be good enough

alexcrichton (approved):

PR #1847, opened 2024-10-07
Gate wasm-smith component support behind a feature

alexcrichton:
This feature is additionally off-by-default because the component support of `wasm-smith` has fallen quite far behind where components are at today.


fitzgen (approved):
Maybe we should just remove this, since we are fuzzing components with a different approach in wasmtime? AFAIK, no one is using this?

alexcrichton:
Probably yeah, it's not too much of a burden to have though and might be a good starting point in the future. I think though we need to rearchitect things along the lines of the "top down" instruction generation idea from wasm-smith where we generate things "top down" rather than bottom-up (or both). I leave this as a problem for a future-someone though.

PR #1862, opened 2024-10-11
Add support for `extended-const` to `wasm-smith`

alexcrichton:
This commit adds support for the `extended-const` proposal to the `wasm-smith` generator crate. This is gated behind a new `extended_const_enabled` field in `wasm_smith::Config`

This commit additionally refactors the `wasm-smith` test suite and fuzzers in this repository to share the same source for generating a `WasmFeatures` from a `Config`.

Finally this also fixes a few minor issues with wide arithmetic, namely it was enabled in fuzzing for components by accident and needed to be added to the `Config`-to-`WasmFeatures` conversion.


fitzgen (approved):
Thanks!

PR #1864, opened 2024-10-13
Update unicode-width to 0.2.0

printfn:
This updates the unicode-width crate to 0.2.0. The library now uses different widths for some control characters, but it doesn't look like the changes affect wast.


alexcrichton (approved):
Thank you for this!

PR #1868, opened 2024-10-16
wit-parser: Group sources used for resolve by package ids

noise64:
This PR is a request for changing the format of the returned source paths for `Resolve::push_path` and `Resolve::push_dir`, from `Vec<PathBuf>` to `BTreeMap<PackageId, Vec<PathBuf>>`.

Our (Golem Cloud) use case is about using a resolved wit dir and then apply further transformations to it, e.g. managing dependencies between related components. Currently we use a workaround which partially re-parses the wit sources using `UnresolvedPackageGroup`, and then maps the sources to the resolved package ids. With this change the double parsing could be avoided, as the already available mapping info would not lost, and I believe there can be other tooling related use cases for this information for others too. The original intention for the sources (triggering builds on change) should also work, with the caller side change of collapsing the sources into a unique set.

In case changing this return type is not okay or preferable, I would be also grateful to get pointers on how to approach this problem differently.

Thanks in advance!


alexcrichton:
Thanks for the PR! I think this is a reasonable feature to support, and it's ok to change the return types here. I'm wondering though if it might be good to have a dedicated newtype here as the return value with various methods and accessors perhaps. For example one of the main uses I know of this is for bindings generators to generate a list of files to depend on but simple iteration over this map will now produce duplicate listings of files if a WIT package is defined with other inline WIT packages as well.

What do you think about something like:

```rust
struct PackageSourceMap { /* ... */ }

impl PackageSourceMap {
    // deduplicates paths so each one only shows up once in the return value
    fn paths(&self) -> impl Iterator<Item = &Path> { /* ... */ }

    // basically BTreeMap::get
    fn package_paths(&self, id: PackageId) -> impl Iterator<Item = &Path> { /* ... */ }
}
```

Does that seem reasonable?

noise64:
@alexcrichton Thank you for the response! It it much better that way, I will update the PR soon.

noise64:
@alexcrichton Updated the PR with the suggested `PackageSourceMap`.

Issue #1701, opened 2024-07-26
`wasmparser`: Performance regressions since `v0.100.2`


Robbepop:
Due to technical reasons I had to use the (by now) ancient `wasmparser 0.100.2` in Wasmi and was using a fork called `wasmparser-nostd`. The fork had no significant impact on the performance.

Recently I updated Wasmi to use the newest `wasmparser v0.214.0` and unfortunately saw massive performance regressions, especially for Wasm validation. Many Wasmi benchmarks regressed by 15-35% and some even by 45% compared to the Wasmi that uses the ancient `wasmparser v0.100.2`.

The Wasmi upgrade PR is here: https://github.com/wasmi-labs/wasmi/pull/1141

Maybe I did some mistakes when upgrading which explain the performance regressions.

Since Wasmi is very focused on speedy translation times those performance regressions unfortunately are a show stopper for using the new `wasmparser` versions.

alexcrichton:
Do you have a link to the module with the largest regression? That'd probably be the easiest to start with in terms of evaluating validation performance on that module and seeing what sticks out in perf

Robbepop:
Do you mean Wasm module?

From what I saw `tiny_keccak` is a good contender. But there were others.
To benchmark `tiny_keccak` translation test cases in Wasmi, use:

```
cargo bench --bench benches translate/tiny_keccak/
```

The `tiny_keccak` benchmark can be found here:
https://github.com/wasmi-labs/rust-benchmarks/tree/main/cases/tiny_keccak

alexcrichton:
Ok I've been poking at this a bit and so far what I've found is that #701, the first commit after 0.100.0, led to a precipitous drop in performance. That was known at the time however. There are other smaller regressions here and there but they're looking to be more difficult to pin down.

Profiling the code itself does not show any unexpected hot spots. That means that it's just sort of spread out a lot.

Doing various tweaks seems to help here and there but nothing is a major improvement. For example outlining string-based errors to their own `#[cold]` function sometimes helps and sometimes doesn't.

One improvement that has shown promise is skipping initialization checking for locals entirely. This was first added in the function-references proposal and later proposals like gc rely on it as well. This yields a ~5% gain (ish) but requires completely removing the check. That means that only way to achieve this would be to add compile-time features to wasmparser.

[...]

PR #1870, opened 2024-10-17
`wasmparser`: Create abstraction for local initialization (simplify future optimizations)

Robbepop:
This implements the `first_non_default` index optimization in [this comment](https://github.com/bytecodealliance/wasm-tools/issues/1701#issuecomment-2256665560).

Note that the issue also suggests using a bitmap instead of a `Vec<bool>`. This is _not_ implemented in this PR as I was unsure if we should implement our own bitvec or use one from crates.io. Furthermore it seems that they are using a `bitvec` with 32-bit words and 16 words inline storage to avoid heap allocations and indirections for up to 512 locals, thus complicating the matter.

## Benchmarks

Minor improvements, not sure if within noise.

image


Robbepop:
How best to handle the `hint::likely` unused warnings? They are only used when the `validator` feature is used, thus we see the warning when `validator` is disabled. Should we put it under the `validator` module or put `cfg`s everywhere since it is not actually a validator specific utility.

alexcrichton:
Thanks for this! Does the `likely` function help performance enough to keep it? If so I think it's fine to put it in the validator module for now.

Robbepop:
> Thanks for this! Does the `likely` function help performance enough to keep it? If so I think it's fine to put it in the validator module for now.

When i benchmarked it unfortunately did
 But you are free to run benchmarks on your CPU to test if this is just a local artifact. I only have a Macbook M2 machine for local benchmarks.

alexcrichton:
Oh sure yeah that's no problem, just double-checking. Since it's only used in the validator I think it's ok to move it under there to avoid dead code warnings.

Robbepop:
> Oh sure yeah that's no problem, just double-checking. Since it's only used in the validator I think it's ok to move it under there to avoid dead code warnings.

Okay I will fix this later.
Note though that the perf gains are modest since most validator regressions are in the type section validation as presented here: https://github.com/bytecodealliance/wasm-tools/issues/1701#issuecomment-2423804444

Robbepop:
@alexcrichton I moved `hint.rs` into the `validator` sub-module.

However, I conducted another benchmark run and mostly saw no improvements anymore compared to `main`.
Still, I think this PR is decent since it will simplify future optimization attempts due to the introduction of the `LocalInits` abstraction. For example, we could test next to swap out the `Vec<bool>` with a bit-vec without only internal changes within `LocalInits`.

Robbepop:
@alexcrichton I just removed the `hint` submodule altogether since I no longer see any wins with it.

alexcrichton (approved):

PR #1872, opened 2024-10-17
threads: add `rec` group test

abrown:
It should be invalid to create a sub-type in a recursive group that does not match the sharedness of the super-type. This test adds a check for this; feel free to suggest further cases.


fitzgen (approved):
Thanks!

PR #1873, opened 2024-10-18
wit-encoder: interface identifier and getter setter fixes

noise64:
Changes:
- Add some more missing getters and setters which were left out from https://github.com/bytecodealliance/wasm-tools/pull/1867.
- Interface identifiers were missing the package parts in case they were from another package

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2321,6 +2321,7 @@ dependencies = [
 name = "wit-encoder"
 version = "0.219.1"
 dependencies = [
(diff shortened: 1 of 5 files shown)


alexcrichton:
Thanks! Happy to merge when you feel this is ready

noise64:
I would like to test a few more cases (indirectly, in our repos), and also add tests here, once those are done I will mark this as ready for review.

alexcrichton (approved):

PR #1875, opened 2024-10-21
wit-parser package not found: explain error better

pchickey:
Hopefully this helps people understand why a package is not found, e.g. did you mismatch the version string? did you put it in the wrong directory structure so its not found that way?

The downcast rewrite is ugly, perhaps there's a better way to do that but I did it the dumbest way possible. The alternative is to make this a separate constructor of Error.


alexcrichton (approved):
Nice!

PR #1877, opened 2024-10-22
Remove a copy being made of wasm modules

alexcrichton:
This commit removes a copy performed when the input file is WebAssembly when parsing input files on the CLI. This happened due to the implementation of `Cow::to_owned` but isn't necessary since we have the original owned copy around to return that instead.

### DIFF
--- a/crates/wat/src/lib.rs
+++ b/crates/wat/src/lib.rs
@@ -235,6 +235,12 @@ impl Parser {
             }),
         })?;
         match self.parse_bytes(Some(file), &contents) {
+            // If the result here is borrowed then that means that the input
+            // `&contents` was itself already a wasm module. We've already got
+            // an owned copy of that so return `contents` directly after
+            // double-checking it is indeed the same as the `bytes` return value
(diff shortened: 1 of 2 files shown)


crates/wat/src/lib.rs
```diff
@@ -235,7 +235,12 @@ impl Parser {
             }),
         })?;
         match self.parse_bytes(Some(file), &contents) {
-            Ok(bytes) => Ok(bytes.into_owned()),
+            Ok(Cow::Borrowed(bytes)) => {
```
pchickey, line 238:
```suggestion
            // removes a copy performed when the input file is WebAssembly when parsing input files on the CLI. This happened due to the implementation of Cow::to_owned but isn't necessary since we have the original owned copy around to return that instead.
            Ok(Cow::Borrowed(bytes)) => {
```

pchickey (approved):
Looks good, but since this is a little surprising the source should probably include a comment along the lines of your PR description.

PR #1879, opened 2024-10-22
feature: Output wasm custom section

lxfontes:
Intended to be used as part of an experiment in [Go bindgen tooling](https://github.com/bytecodealliance/wasm-tools-go)

The current TinyGo experience is as following:
- codegen using wit-bindgen-go ( need wit + target world )
- tinygo build wasm32 module ( also need wit + target world )
  - run `wasm-tools component embed`
  - run `wasm-tools component new`

We are investigating the addition of a language feature ( `//go:wasmsection` ) so TinyGo ( and potentially Big Go ) are not required to know about wit-worlds. The experience would become:
- codegen using wit-bindgen-go ( need wit + target world )
- tinygo build wasm32 module <- no longer needs wit + target world visibility
- User Initiated / Other Tools
  - run `wasm-tools component embed`
  - run `wasm-tools component new`

The following code is analogous to `wasm-tools component embed -w demo -o core_embedded.wasm ./wit core.wasm`

```go
// Creates a custom section 'component-type:demo' with contents from 'demo.bin'
package demo

//go:wasmsection component-type:demo
//go:embed demo.bin
var wasmSection []byte
```

This PR introduces `wasm-tools component section`, which will be responsible for generating the `demo.bin` described above.

During wit-2-go codegen, `wasm-tools` will be executed once per world as needed. Ex for wasi http:
```
wasm-tools component section -w wasi:http/proxy -o http.bin .
```

The contents of `http.bin` will be referenced in `//go:wasmsection component-type:wasi-http http.bin`, resulting in a custom section named `component-type:wasi-http` with content from `http.bin`.

From there, wasm module to component conversion happens as usual with `wasm-tools component new`.

## Alternative
Initially I had this behavior under `wasm-tools component embed --custom` ( flag `--custom` ), which would look like
```
wasm-tools component embed -w example ./wit --dummy  --custom -o /tmp/custom_section.wasm
```
Decided towards another sub-command as the output is significantly different ( a single section vs a whole module ).

cc @ydnar

### DIFF
--- a/src/bin/wasm-tools/component.rs
+++ b/src/bin/wasm-tools/component.rs
@@ -26,7 +26,6 @@ pub enum Opts {
     Embed(EmbedOpts),
     Targets(TargetsOpts),
     Link(LinkOpts),
-    Section(SectionOpts),
     SemverCheck(SemverCheckOpts),
     Unbundle(UnbundleOpts),
 }
@@ -39,7 +38,6 @@ impl Opts {
             Opts::Embed(embed) => embed.run(),
             Opts::Targets(targets) => targets.run(),
             Opts::Link(link) => link.run(),
-            Opts::Section(section) => section.run(),
             Opts::SemverCheck(s) => s.run(),
             Opts::Unbundle(s) => s.run(),
         }
@@ -52,7 +50,6 @@ impl Opts {
             Opts::Embed(embed) => embed.general_opts(),
             Opts::Targets(targets) => targets.general_opts(),
             Opts::Link(link) => link.general_opts(),
-            Opts::Section(section) => section.general_opts(),
             Opts::SemverCheck(s) => s.general_opts(),
             Opts::Unbundle(s) => s.general_opts(),
         }
@@ -312,6 +309,10 @@ pub struct EmbedOpts {
     /// Print the output in the WebAssembly text format instead of binary.
     #[clap(long, short = 't')]
     wat: bool,
+
+    /// Print the wasm custom section only.
+    #[clap(long, short = 's')]
+    only_custom: bool,
 }
 
 impl EmbedOpts {
@@ -322,7 +323,21 @@ impl EmbedOpts {
     /// Executes the application.
     fn run(self) -> Result<()> {
         let (resolve, pkg_id) = self.resolve.load()?;
+
         let world = resolve.select_world(pkg_id, self.world.as_deref())?;
+
+        if self.only_custom {
+            let encoded = metadata::encode(
+                &resolve,
+                world,
+                self.encoding.unwrap_or(StringEncoding::UTF8),
+                None,
+            )?;
+
+            self.io.output_wasm(&encoded, self.wat)?;
+            return Ok(());
+        }
+
         let mut wasm = if self.dummy {
             wit_component::dummy_module(&resolve, world, Mangling::Standard32)
         } else if let Some(mangling) = self.dummy_names {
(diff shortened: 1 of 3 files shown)


ydnar:
Thanks! Is the `--world` argument optional?

aykevl:
While I might have some nits about the specific Go syntax, the overall approach seems fine by me!

ydnar:
Is there a path to evolving the format of this section?

Eg textual WIT, or JSON, etc?

alexcrichton:
Not concretely, no, but it's already evolved a few times in the past. It depends on the precise changes being made though I think. A completely different format might, for example, have a new section prefix name.

src/bin/wasm-tools/component.rs
```diff
@@ -308,6 +309,10 @@ pub struct EmbedOpts {
     /// Print the output in the WebAssembly text format instead of binary.
     #[clap(long, short = 't')]
     wat: bool,
+
+    /// Print the wasm custom section only.
+    #[clap(long, short = 's')]
+    only_custom: bool,
```
alexcrichton, line 315: I'm a bit wary to too eagerly reserve short flags so mind leaving this as just `long`?

src/bin/wasm-tools/component.rs
```diff
@@ -318,7 +323,21 @@ impl EmbedOpts {
...
+
+        if self.only_custom {
+            let encoded = metadata::encode(
+                &resolve,
+                world,
+                self.encoding.unwrap_or(StringEncoding::UTF8),
+                None,
+            )?;
+
+            self.io.output_wasm(&encoded, self.wat)?;
```
alexcrichton, line 337: Since this isn't wasm I think that `false` should be passed here instead of `self.wat`?

src/bin/wasm-tools/component.rs
alexcrichton: Also mind adding some `conflicts_with` annotations to indicate that it doesn't work with flags such as `--wat`, `--dummy`, etc?

alexcrichton (commented):
Thanks! Mind adding a test for this too? The custom section is itself a component (IIRC) so you I think should be able to have a test with `component embed % | print` and that should print the text of a component.

ydnar:
If the content is just a component, is this new command necessary?

E.g. can the content of the custom section be created with `component new` from a single WIT file?

alexcrichton:
Yeah I think we'll want this new flag on the `embed` command still. There's no command that produces the component that's embedded within the custom section and `embed` is the closest equivalent, it just always stuffs it in a custom section rather than being able to emit the section itself (which this is adding)

Which is to say: no there's no preexisting command, but this PR is adding that, so I think it's a good idea to have.

lxfontes:
@alexcrichton thanks for the review! got the updates in 🙌

alexcrichton (approved):
Thanks!

ydnar:
Verified that we can add multiple custom sections in `wasm-tools component embed` that `component new` consumes and seems to merge into a single world.

ydnar:
I started integrating this into `wit-bindgen-go` last night, and quickly learned that `wasm-tools component embed --only-custom` does *not* read WIT from stdin:

```console
$ cat wasi-cli.wit | wasm-tools component embed --only-custom
error: the following required arguments were not provided:
  <WIT>

Usage: wasm-tools component embed --only-custom <WIT> [INPUT]
```

This makes tooling integration a bit tricker—have to write a temp file first, then process the file with `wasm-tools`. In addition, the usage text is incorrect—there is no `[INPUT]` for `--only-custom`.

I still think a separate command would provide a cleaner UX, e.g. `wasm-tools component type [-o output] [INPUT]`, where it reads from stdin in the absence of `INPUT`. This allows upstream tools like `wit-bindgen-go` to create the wasm file with the appropriately-named custom section(s).

alexcrichton:
Good points! Sounds reasonable to me to use a separate command 👍

alexcrichton:
One significant difference with `wit --wasm` is that using `wit --wasm` will encode all worlds where `component embed` only embeds a single world. That's currently crucial for the decoding process where if it found multiple worlds encoded it wouldn't know which to use.

ydnar:
Could we add a --world argument?

ydnar:
Makes sense. 

FWIW, the WIT text passed to this command will already be tree-shaken and contain only a single synthetic world that maps 1:1 to the intersection of an explicit world and a single WIT interface.

PR #1881, opened 2024-10-23
Support `i8x16.shuffle` in `wasm-mutate`

alexcrichton:
Support another SIMD instruction in `wasm-mutate` to assist with shrinking functions which might contain this instruction.


crates/wasm-mutate/src/mutators/peephole/eggsy/lang.rs
```diff
@@ -714,6 +714,7 @@ lang! {
         F64x2ReplaceLane(u8, [Id; 2]) = "f64x2.replace_lane",
 
         I8x16Swizzle([Id; 2]) = "i8x16.swizzle",
+        I8x16Shuffle(Shuffle, [Id; 2]) = "i8x16.swizzle",
```
abrown, line 717:
```suggestion
        I8x16Shuffle(Shuffle, [Id; 2]) = "i8x16.shuffle",
```

fitzgen (approved):

PR #1888, opened 2024-10-31
Improve the comments for hexadecimal float parsing.

sunfishcode:
Replace the comments requesting better comments with better comments, and tidy up the code.


alexcrichton (approved):
Thanks for this!

PR #1889, opened 2024-11-01
Split out `MVP` from `WASM1` feature

alexcrichton:
I recently learned that these were actually two distinct points in time, so while the points-in-time are all listed out may as well try to be accurate along the way.

### DIFF
--- a/crates/wasmparser/src/features.rs
+++ b/crates/wasmparser/src/features.rs
@@ -233,19 +233,34 @@ define_wasm_features! {
 }
 
 impl WasmFeatures {
-    /// The feature set associated with the 1.0 version of the
-    /// WebAssembly specification or the "MVP" feature set.
+    /// The feature set associated with the MVP release of WebAssembly (its
+    /// first release).
+    //
+    // Note that the features listed here are the wasmparser-specific built-in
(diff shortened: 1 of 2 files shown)


crates/wasmparser/src/features.rs
```diff
@@ -233,19 +233,34 @@ define_wasm_features! {
...
+    pub const MVP: WasmFeatures = WasmFeatures::FLOATS.union(WasmFeatures::GC_TYPES);
+
     /// The feature set associated with the 1.0 version of the
-    /// WebAssembly specification or the "MVP" feature set.
+    /// WebAssembly specification circa 2017.
+    ///
+    /// <https://webassembly.github.io/spec/versions/core/WebAssembly-1.0.pdf>
     #[cfg(feature = "features")]
-    pub const WASM1: WasmFeatures = WasmFeatures::FLOATS.union(WasmFeatures::GC_TYPES);
+    pub const WASM1: WasmFeatures = WasmFeatures::MVP.union(WasmFeatures::MUTABLE_GLOBAL);
```
fitzgen, line 253: Surprised that saturating conversions and sign extensions weren't in 1.0, I had thought they were.

fitzgen (approved):

PR #1890, opened 2024-11-02
Fix `PrintTermcolor` description in `wasmprinter` crate

PizzasBear:
Changed the docs to be more correct and less confusing


alexcrichton (approved):
Oops, thanks!

PR #1891, opened 2024-11-02
`wast`: Add `#[non_exhaustive]` to `WastArg` and `WastRet`

Robbepop:
This attribute is required because depending on `#[cfg(feature = "component-model")]` both `WastArg` and `WastRet` can be `enum`s with a single variant. Since crate features are resolved together this might cause problems in a crate due to feature usage of another crate or dependency, e.g. in the same workspace, with respect to irrefutable patterns.

# Examples

1.

```rust
let arg = WastArg::Core(..);
let WastArg::Core(arg) = arg; // ok if `component-model` is disabled, error if enabled
```

2.

```rust
let arg = WastArg::Core(..);
let arg = match arg {
    WastArg::Core(arg) = arg,
    _ => panic!(), // ok if `component-model` is enabled, unreachable code warning if enabled
};
```

### DIFF
--- a/crates/wast/src/wast.rs
+++ b/crates/wast/src/wast.rs
@@ -498,6 +498,7 @@ pub enum QuoteWatTest {
 
 #[derive(Debug)]
 #[allow(missing_docs)]
+#[non_exhaustive]
 pub enum WastArg<'a> {
     Core(WastArgCore<'a>),
     // TODO: technically this isn't cargo-compliant since it means that this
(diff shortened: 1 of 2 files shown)


alexcrichton:
Blast I was [hoping no one would run into this](https://github.com/bytecodealliance/wasm-tools/blob/5dee4ec15af6a692af0a4db18545c97c021044df/crates/wast/src/wast.rs#L503-L506). 

I was hoping to avoid `#[non_exhaustive]` but I agree it's the easiest solution available at this time. Want to delete that comment? Other than that looks good to me 👍

Robbepop:
@alexcrichton thanks for your reply. I removed the now outdated comment.

alexcrichton (approved):

PR #1893, opened 2024-11-04
Improve `json-from-wast -h` output

alexcrichton:
Don't render a long variable name placeholder as
`ALLOW_CONFUSING_UNICODE` and instead render it just as `ALLOW`.

### DIFF
--- a/src/bin/wasm-tools/json_from_wast.rs
+++ b/src/bin/wasm-tools/json_from_wast.rs
@@ -54,7 +54,7 @@ pub struct Opts {
     ///
     /// This is defaulted to `true` to enable parsing all upstream spec tests
     /// but can be disabled if desired too.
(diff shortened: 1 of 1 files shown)


src/bin/wasm-tools/json_from_wast.rs
```diff
@@ -54,7 +54,7 @@ pub struct Opts {
     ///
     /// This is defaulted to `true` to enable parsing all upstream spec tests
     /// but can be disabled if desired too.
-    #[clap(long)]
+    #[clap(long, value_name = "ALLOW")]
```
fitzgen, line 57: Alternatively:

```suggestion
    #[clap(long, value_name = "true|false")]
```

fitzgen (approved):

PR #1895, opened 2024-11-05
Add support for async ABI, futures, streams, and errors

dicej:
This adds support for encoding and parsing components which use the [Async ABI](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Async.md) and associated canonical options and functions, along with the [`stream`, `future`, and `error`](https://github.com/WebAssembly/component-model/pull/405) types.

~~Note that the `error` type was recently (about 30 minutes ago) renamed to `error-context` in Luke's spec PR.  I haven't updated this implementation to reflect that yet, but will do so in a follow-up commit.  That should allow us to avoid conflicts with existing WIT files that use `error` as a type and/or interface name.~~ EDIT: done

This does not yet include any new tests; I'll also add those in a follow-up commit.

See https://github.com/bytecodealliance/rfcs/pull/38 for more context.

This is ready for review, but I'll leave it in draft mode until the following are complete:

- [x] Rename `error` to `error-context`
- [x] Add tests to `tests/local/component-model` and `tests/local/missing-features` to exercise new `wasm-encoder`, `wasmparser`, and `wasmprinter` features
- [x] Add tests to `crates/wit-component/tests` to exercise its new features
- [x] Add tests to `crates/wit-parser/tests` to exercise its new features


dicej:
Sorry about all the churn in `wasm-compose`; I'm currently using it for end-to-end testing, but I could switch to WAC for that and revert most of the `wasm-compose` changes in this PR if desired.

alexcrichton:
Thanks again for this! My current plan is to start tackling this the week-after-next (after WasmCon) but if you've got any high-level points you want me to confirm ahead of time feel free to let me know

crates/wasm-compose/src/composer.rs
```diff
@@ -531,7 +554,8 @@ impl<'a> ComponentComposer<'a> {
     /// Returns the bytes of the composed component.
     pub fn compose(&self) -> Result<Vec<u8>> {
         let (root_instance, graph) =
-            CompositionGraphBuilder::new(self.component, self.config)?.build()?;
+            CompositionGraphBuilder::new(self.component, self.config, self.features)?
+                .build(self.features)?;
```
alexcrichton, line 558: To avoid threading `features` around throughout here, WDYT of using `WasmFeatures::all()` as the validation side of things? Personally I find it useful enough to defer specific features/flags/etc to the main validate step in an engine or similar.

crates/wasmparser/src/validator/component.rs
```diff
@@ -1048,6 +1086,896 @@ impl ComponentState {
...
+                SubType {
+                    is_final: true,
+                    supertype_idx: None,
+                    composite_type: CompositeType {
+                        shared: false,
+                        inner: CompositeInnerType::Func(FuncType::new([ValType::I32], [])),
+                    },
+                },
+            ),
+        );
```
alexcrichton, line 1115: You might be able to tighten this up a bit with this pattern perhaps?

https://github.com/bytecodealliance/wasm-tools/blob/00091b2aa1cbb7d700c0d16ee00eddd453f02260/crates/wasmparser/src/validator/component.rs#L1016-L1018

crates/wasmparser/src/validator/component.rs
```diff
@@ -1048,6 +1086,896 @@ impl ComponentState {
...
+    ) -> Result<()> {
+        if !features.component_model_async() {
+            bail!(
+                offset,
+                "`task.return` requires the component model async feature"
+            )
+        }
+
+        let id = self.type_id_at(type_index, offset)?;
+        self.core_funcs.push(id);
```
alexcrichton, line 1135: Should this validate that `id` points to a core function type? (as opposed to an array/struct/etc)

Also this may be elsewhere but I would otherwise have expected some sort of validation of the type here as well (as opposed to allowing any signature)

crates/wasmparser/src/validator/component.rs
alexcrichton: Alternatively this pattern is used in a number of locations throughout this file (both added in this PR and preexisting), so if you'd like having a helper as well for this I think would also be reasonable.

crates/wasmparser/src/validator/component_types.rs
```diff
@@ -1062,7 +1106,13 @@ impl ComponentDefinedType {
                 .any(|case| case.ty.map(|ty| ty.contains_ptr(types)).unwrap_or(false)),
             Self::List(_) => true,
             Self::Tuple(t) => t.types.iter().any(|ty| ty.contains_ptr(types)),
-            Self::Flags(_) | Self::Enum(_) | Self::Own(_) | Self::Borrow(_) => false,
+            Self::Flags(_)
+            | Self::Enum(_)
+            | Self::Own(_)
+            | Self::Borrow(_)
+            | Self::Future(_)
+            | Self::Stream(_)
```
alexcrichton, line 1114: I'm a bit surprised by this, how come `contains_ptr` isn't transitive through these types?

crates/wasmprinter/src/component.rs
```diff
@@ -252,6 +274,9 @@ impl Printer<'_, '_> {
                 self.print_idx(&state.component.type_names, *idx)?;
                 self.end_group()?;
             }
+            ComponentDefinedType::Future(ty) => self.print_future_type(state, *ty)?,
+            ComponentDefinedType::Stream(ty) => self.print_stream_type(state, *ty)?,
+            ComponentDefinedType::ErrorContext => self.result.write_str("error-context")?,
```
alexcrichton, line 279: Mind using `print_type_keyword` here to colorize `error-context`?

crates/wasmprinter/src/component.rs
```diff
@@ -885,6 +916,307 @@ impl Printer<'_, '_> {
...
+                CanonicalFunction::ErrorContextDrop => {
+                    self.start_group("core func ")?;
+                    self.print_name(&state.core.func_names, state.core.funcs)?;
+                    self.result.write_str(" ")?;
+                    self.start_group("canon error-context.drop")?;
+                    self.end_group()?;
+                    self.end_group()?;
+                    debug_assert_eq!(state.core.func_to_type.len(), state.core.funcs as usize);
+                    state.core.funcs += 1;
+                    state.core.func_to_type.push(None);
```
alexcrichton, line 1218: If you'd like there's a lot of duplication between these cases around the boilerplate of printing a core func (e.g. the initial group, the debug assert, the increment, etc). That might be good to extract into a helper perhaps to have each case here focus on just the printing-the-intrinsic part

crates/wast/src/component/func.rs
```diff
@@ -460,6 +556,602 @@ impl Default for CanonThreadHwConcurrency {
...
+}
+
+impl<'a> Parse<'a> for CanonTaskWait<'a> {
+    fn parse(parser: Parser<'a>) -> Result<Self> {
+        parser.parse::<kw::task_wait>()?;
+        let mut l = parser.lookahead1();
+        let async_ = l.peek::<kw::r#async>()?;
+        if async_ {
+            parser.parse::<kw::r#async>()?;
+        }
```
alexcrichton, line 601: I think this can be simplified as `let async_ = parser.parse::<Option<kw::r#async>>()?.is_some();`

crates/wast/src/component/func.rs
```diff
@@ -460,6 +556,602 @@ impl Default for CanonThreadHwConcurrency {
...
+        let memory = parser.parens(|parser| {
+            let span = parser.parse::<kw::memory>()?.0;
+            parse_trailing_item_ref(kw::memory(span), parser)
+        })?;
+
+        Ok(Self { async_, memory })
+    }
+}
+
+impl Default for CanonTaskWait<'_> {
```
alexcrichton, line 611: There's a fair amount of boilerplate with `Default for ...` for all these new intrinsics, and if you'd like feel free to refactor the `expand.rs` step above to avoid `mem::take` and probably `mem::replace` at the front of the function with some dummy variant to avoid needing `Default` on all the intrinsics.

Happy to see this refactoring deferred to a future PR though

crates/wast/src/component/func.rs
alexcrichton: (and in a few more places down below too)

crates/wast/src/lib.rs
```diff
@@ -556,8 +556,35 @@ pub mod kw {
...
+    custom_keyword!(future_cancel_read = "future.cancel-read");
+    custom_keyword!(future_cancel_write = "future.cancel-write");
+    custom_keyword!(future_close_readable = "future.close-readable");
+    custom_keyword!(future_close_writable = "future.close-writable");
+    custom_keyword!(error_context_new = "error-context.new");
+    custom_keyword!(error_context_debug_message = "error-context.debug-message");
+    custom_keyword!(error_context_drop = "error-context.drop");
     custom_keyword!(wait);
     custom_keyword!(definition);
+    custom_keyword!(r#async = "async");
```
alexcrichton, line 584: Personally I find raw keyword `r#...` annoying and confusing to work with, so if you'd prefer as well I think it's reasonable to call this `async_ = "async"` to avoid the need for `r#`.

This might also be an inconsistent suggestion if `r#` is already used above in which case ignore me.

crates/wit-parser/src/lib.rs
```diff
@@ -991,6 +986,82 @@ impl Function {
...
+        }
+        for ty in self.results.iter_types() {
+            find_futures_and_streams(resolve, *ty, &mut results);
+        }
+        results
+    }
+}
+
+fn find_futures_and_streams(resolve: &Resolve, ty: Type, results: &mut Vec<TypeId>) {
+    if let Type::Id(id) = ty {
```
alexcrichton, line 1017: Mind doing:

```rust
let Type::Id(id) = ty else { return; };
```

to de-indent most of this one level?

crates/wit-parser/src/decoding.rs
```diff
@@ -46,8 +47,8 @@ enum WitEncodingVersion {
 impl ComponentInfo {
     /// Creates a new component info by parsing the given WebAssembly component bytes.
 
-    fn from_reader(mut reader: impl Read) -> Result<Self> {
-        let mut validator = Validator::new();
+    fn from_reader(mut reader: impl Read, features: WasmFeatures) -> Result<Self> {
+        let mut validator = Validator::new_with_features(features);
```
alexcrichton, line 51: Similar to `wasm-compose` I think it's fine to use `WasmFeatures::all()` here to avoid needing to thread features around

crates/wit-parser/src/ast.rs
```diff
@@ -1403,29 +1403,21 @@ impl<'a> Type<'a> {
...
+                tokens.expect(Token::LessThan)?;
+                let ty = Type::parse(tokens)?;
+                tokens.expect(Token::GreaterThan)?;
+                Ok(Type::Stream(Stream {
+                    span,
+                    ty: Box::new(ty),
+                }))
             }
 
+            // error
```
alexcrichton, line 1418: s/error/error-context/

crates/wit-parser/src/ast.rs
```diff
@@ -1403,29 +1403,21 @@ impl<'a> Type<'a> {
                 Ok(Type::Future(Future { span, ty }))
             }
 
-            // stream<T, Z>
-            // stream<_, Z>
             // stream<T>
             // stream
```
alexcrichton, line 1407: mind deleting this line? (only `stream<T>` is parsed here I think)

crates/wit-component/tests/components.rs
```diff
@@ -153,12 +154,25 @@ fn run_test(path: &Path) -> Result<()> {
         }
     };
 
+    let features =
+        WasmFeatures::WASM2 | WasmFeatures::COMPONENT_MODEL | WasmFeatures::COMPONENT_MODEL_ASYNC;
+    Validator::new_with_features(features)
```
alexcrichton, line 159: Similar to a few other locations I think it's fine to use `WasmFeatures::all()` here

crates/wit-component/src/printing.rs
```diff
@@ -987,6 +997,7 @@ fn is_keyword(name: &str) -> bool {
             | "with"
             | "include"
             | "constructor"
+            | "error"
```
alexcrichton, line 1000: s/error/error-context/

alexcrichton (commented):
Everything's looking great here, thanks again for your work here! I've left a number of mostly stylistic things here and there, nothing major. I still haven't gotten to most of `wit-component` though which I'll tackle tomorrow.

One thing though I'd like to ask for, if you're up for it here, is to integrate some of this into fuzzing. The main way we fuzz components right now is via the `wit-smith` crate and the `roundtrip_wit` fuzzer.  Could you extend `wit-smith` with support for the new types of stream/future/error-context? (I think this should be relatively easy). Afterwards if you run the `roundtrip_wit` fuzzer for a bit that should hopefully have no issues.

After that though there's a perhaps slightly more substantial part, which is can you add async support to `wasm-tools component embed --dummy`? I'm imagining something like `--async` which assumes that all imports/exports are async (or maybe `--async-callback` or `--async-stackful` as well?). That would then import all the various intrinsics. I ask this because it's useful for two primary reason:

1. Mostly it enables more fuzzing. This helps exercise everything in `wit-component` to ensure that core modules that do async things can always be turned into a component.
2. Otherwise it helps for exploring things at a low level. For example if you're curious to remember what the async core signatures look like you can run `--dummy --async-callbacks` (or w/e the flags are) to see the core wasm module shape that needs to be produced.

crates/wasmparser/src/validator/component.rs
dicej: Regarding the first point: yes, I'll add a check to make sure it's a core function type.

Regarding the second point: We actually do need to allow arbitrary signatures here, since the signature will vary depending on which component-level exported function the module is calling `task.return` for, and we can't statically determine that in general, so we need to delay the type check until runtime (at which point we've presumably interned the expected and actual type, so it's just a quick integer comparison).

crates/wasmparser/src/validator/component_types.rs
dicej: A `Future` or `Stream` handle is just a u32, like a `Own` resource handle.  The payload type doesn't affect that; the (non)existence of pointers in the payload type only affects how we treat reads to or writes from the stream/future, but has no significance when we're just passing handles around.

crates/wast/src/lib.rs
dicej: Yeah, we're already using it for `r#try`, `r#struct` and a bunch of other keywords, so I'm following the existing pattern here.  I agree that `r#` is annoying in general, though.

dicej:
Thanks for the thorough review as always, @alexcrichton .  I believe I've addressed all your inline comments with either code changes or explanations.  I'll work on the fuzzing and `embed --dummy` items tomorrow.

crates/wit-component/src/encoding/world.rs
```diff
@@ -414,22 +414,27 @@ impl<'a> ComponentWorld<'a> {
...
     fn add_func(&mut self, required: &Required<'_>, resolve: &Resolve, func: &Function) {
-        match required.interface_funcs.get(&self.interface) {
-            Some(set) if set.contains(func.name.as_str()) => {}
+        let abi = match required.interface_funcs.get(&self.interface) {
+            Some(set) if set.contains(&(func.name.as_str(), AbiVariant::GuestImport)) => {
+                AbiVariant::GuestImport
+            }
+            Some(set) if set.contains(&(func.name.as_str(), AbiVariant::GuestImportAsync)) => {
+                AbiVariant::GuestImportAsync
+            }
```
alexcrichton, line 429: This looks like it might not handle the case that a function is imported both sync and async?

crates/wit-component/src/encoding.rs
```diff
@@ -493,6 +505,10 @@ impl<'a> EncodingState<'a> {
         let ty = encoder.ty;
         // Don't encode empty instance types since they're not
         // meaningful to the runtime of the component anyway.
+        //
+        // TODO: Is this correct? What if another imported interface needs to
+        // alias a type exported by this interface but can't because we skipped
+        // encoding the import?
```
alexcrichton, line 511: In theory this is handled by a "liveness pass" somewhere where if only an interface's types are needed those are included but the functions are pruned. That being said there's been bugs around this in the past.

crates/wit-component/src/encoding.rs
```diff
@@ -1667,6 +2092,27 @@ enum ShimKind<'a> {
...
+    },
+    ErrorContextNew {
+        for_module: CustomModule<'a>,
+        encoding: StringEncoding,
+    },
+    ErrorContextDebugMessage {
+        for_module: CustomModule<'a>,
+        encoding: StringEncoding,
+        realloc: &'a str,
+    },
```
alexcrichton, line 2115: Mind adding some comments here explaining what the variants are?

crates/wit-component/src/encoding.rs
```diff
@@ -1243,6 +1422,33 @@ impl<'a> EncodingState<'a> {
         Ok(())
     }
 
+    fn payload_type_index(&mut self, ty: TypeId, imported: bool) -> Result<u32> {
```
alexcrichton, line 1425: Mind adding a comment here explaining what this function is doing? I think a comment on `owner` might also be reasonable to explain why it's traversing the way it is

crates/wit-component/src/validation.rs
```diff
@@ -114,6 +117,26 @@ pub enum ImportInstance {
...
 
+#[derive(Debug, Eq, PartialEq, Clone)]
+pub struct PayloadInfo {
+    pub name: String,
+    pub ty: TypeId,
+    pub function: Function,
+    pub key: WorldKey,
+    pub interface: Option<InterfaceId>,
+    pub imported: bool,
+}
```
alexcrichton, line 128: Mind adding some comments to this? I'm this far through wit-component is and I'm still not entirely sure what a "payload" is other than that it's async-related

crates/wit-component/src/validation.rs
```diff
@@ -114,6 +117,26 @@ pub enum ImportInstance {
...
+
+impl Hash for PayloadInfo {
+    fn hash<H: Hasher>(&self, state: &mut H) {
```
alexcrichton, line 131: Mind leaving a comment here explaining why this isn't `derive` and why `self.function` is omitted?

crates/wit-component/src/validation.rs
```diff
@@ -172,29 +195,107 @@ pub enum Import {
...
+    },
 
     /// An arbitrary item from either the main module or an adapter is being
     /// imported.
     ///
     /// (should probably subsume `MainModule*` and maybe `AdapterExport` above
     /// one day.
     Item(Item),
+
+    ExportedTaskReturn(Function),
```
alexcrichton, line 210: Mind adding some comments to these variants as well? (nothing huge, just something indicating what it is like the above variants)

crates/wasmparser/src/validator/component.rs
```diff
@@ -1041,13 +1074,573 @@ impl ComponentState {
...
+                offset,
+                "`task.return` requires the component model async feature"
+            )
+        }
+
+        let id = self.type_id_at(type_index, offset)?;
+        let Some(SubType {
+            composite_type:
+                CompositeType {
+                    inner: CompositeInnerType::Func(_),
```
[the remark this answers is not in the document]
dicej: Yeah, I've struggled with this myself, and indeed my first draft (written before a spec existed) used a component function type instead of a core function type.  After all, you can always derive the latter from the former.  Luke and I discussed it and landed on the current scheme, but I don't remember the details as to why.  Anyway, yes, please open an issue upstream about it.

crates/wasmparser/src/features.rs
```diff
@@ -45,6 +45,12 @@ macro_rules! define_wasm_features {
         }
 
         impl WasmFeatures {
+            /// Construct a `WasmFeatures` which includes all features.
+            #[cfg(not(feature = "features"))]
+            pub fn all() -> Self {
+                Self { _priv: () }
+            }
```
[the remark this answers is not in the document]
dicej: This came up after I addressed some of your earlier feedback about not threading `WasmFeatures` through a bunch of functions:
```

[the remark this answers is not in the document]
dicej: Update: nevermind my last comment -- I understand what you're suggesting now and will push an update shortly.

crates/wasmparser/src/validator/component.rs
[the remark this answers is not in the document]
dicej: I chatted with Luke about this, and he'll update the spec to use the component-level return type instead of the core function type for `task.return` declarations, then I'll update the implementation to match.

crates/wit-component/src/encoding/world.rs
dicej: Good catch.  I've just added a test for that, and it is indeed failing.  Will work on fixing it.

dicej: Fixed!

dicej:
I didn't have time today to do the fuzzing and `embed --dummy` stuff, but I'll aim to do that and the `task.return` change next week.

alexcrichton:
Oh I thought I wrote this down last Friday but I appear to have forgotten, sorry about that! In any case if you'd like I think it's reasonable to merge this as-is and do the dummy/fuzzing bits as follow-ups.

dicej:
> Oh I thought I wrote this down last Friday but I appear to have forgotten, sorry about that! In any case if you'd like I think it's reasonable to merge this as-is and do the dummy/fuzzing bits as follow-ups.

Either way works for me.  There are three to-do items at this point:
- fuzzing
- `embed --dummy`
- updating `task.return` to match https://github.com/WebAssembly/component-model/pull/431

If you think all those are fine to leave as follow-ups, feel free to merge.  Either way, I'll aim to address those items ASAP.

alexcrichton (approved):
Ok I'll go ahead and flag this for merge to prevent any possible future merge conflicts, and sounds good to me 👍

PR #1898, opened 2024-11-07
threads: add validation for constant expressions

abrown:
Previously, it was possible to use an unshared global to initialize a shared global, table element. This change makes that invalid by propagating sharedness in to the constant expression validator.


crates/wasmparser/src/validator/core.rs
```diff
@@ -256,8 +255,18 @@ impl ModuleState {
             }
             crate::ElementItems::Expressions(ty, reader) => {
                 validate_count(reader.count())?;
+                let shared = match ty.heap_type() {
+                    HeapType::Abstract { shared, .. } => shared,
+                    HeapType::Concrete(unpacked_index) => {
+                        if let Some(id) = unpacked_index.as_core_type_id() {
+                            types[id].composite_type.shared
+                        } else {
+                            todo!()
```
abrown, line 264: Not sure what to do here &mdash; is this actually unreachable?

alexcrichton: I think so yeah, but I think this can also use the `types.reftype_is_shared(ty)` helper?

PR #1901, opened 2024-11-10
Vendor install-rust CI action from Wasmtime

alexcrichton:
Copy the action over to this repository and edit it to suit this repository's needs such as not installing wasm targets and dealing with MSRV differently. This fixes a CI error from the latest nightly where `wasm32-wasi` isn't available. The Wasmtime version of the action is no longer used since the MSRV in this repository doesn't have `wasm32-wasip1` and otherwise the latest nightly doesn't have `wasm32-wasi`, so it's easier to just edit it here.

This has the additional effect of handling Rust versions in CI slightly different. Previously the Wasmtime behavior of MSRV+2 was chosen for most runs but that was actually 1.78 as opposed to the latest stable 1.82, so the default now is "stable" and only the one builder for "msrv" explicitly reads `Cargo.toml` to run that.

For now this repository has more "floating" CI than the Wasmtime repository where the Rust versions are automatically updated. This will (and has) cause breakage when Rust versions update but otherwise I'm pretty sure I'll always forget to update the CI so it seems like a more reasonable balance for now. If contributions on this repository increase and a disruption of CI is significant it's probably best to fix the Rust versions used in CI

### DIFF
--- a/.github/actions/install-rust/action.yml
+++ b/.github/actions/install-rust/action.yml
@@ -1,0 +1,43 @@
+name: 'Install Rust toolchain'
+description: 'Install a rust toolchain'
+
+inputs:
+  toolchain:
+    description: 'Default toolchan to install'
+    required: false
+    default: 'stable'
+
+runs:
+  using: composite
+  steps:
+    - name: Install Rust
+      shell: bash
+      id: select
+      run: |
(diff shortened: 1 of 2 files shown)


alexcrichton:
I've updated the description of the PR to accurately reflect the current status.

fitzgen (approved):
Sounds reasonable

PR #1902, opened 2024-11-11
`wasmparser`: fix crate feature propagation

Robbepop:
No longer enable `indexmap` and `hashbrown` when `std` feature is enabled without `hash-collections` feature.

### DIFF
--- a/crates/wasm-metadata/Cargo.toml
+++ b/crates/wasm-metadata/Cargo.toml
@@ -13,7 +13,7 @@ workspace = true
 [dependencies]
 clap = { workspace = true, optional = true }
 anyhow = { workspace = true }
-wasmparser = { workspace = true, features = ['component-model'] }
+wasmparser = { workspace = true, features = ['component-model', 'hash-collections'] }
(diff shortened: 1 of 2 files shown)


Robbepop:
The `fuzz` job fails with
```
error: toolchain 'nightly-x86_64-unknown-linux-gnu' does not support target 'wasm32-wasi'; did you mean 'wasm32-wasip1'?
```
Seems to be unrelated to this PR.

alexcrichton:
Thanks! Mind rebasing to pick up the fix from https://github.com/bytecodealliance/wasm-tools/pull/1901?

Robbepop:
> Thanks! Mind rebasing to pick up the fix from #1901?

Done :)

alexcrichton (approved):

Robbepop:
@alexcrichton It seems `wasm-metadata` strictly requires the `hash-collections` crate feature from the `wasmparser` crate. I had to enable it to make `wasm-metadata` compile again.

Issue #1899, opened 2024-11-09
`wasmparser`: Should we put 128-bit SIMD behind a crate feature?


Robbepop:
With the recent movement to put some major Wasm proposals behind crate features, e.g. `component-model` and `gc`, I was wondering if this was also practical or useful for the very large 128-bit `simd` (and thus also `relaxed-simd`) proposal with its tons of op-codes. (`simd`: 235, `flexible-simd`: 19)

I see plenty of Wasm runtimes not (yet) supporting `simd` and it isn't even clear to me whether `simd` is supposed to become an optional component of Wasm in the future due to its massive size and the outlook of the more elaborate `flexbile-vectors` on the (far) horizont. For Wasmi I am starting to lean towards supporting Wasm `simd` proposal soonish but behind a crate feature that is disabled by default since it is just way too massive.

The hope is that putting the `simd` and `flexible-simd` feature behind a single `simd` crate feature for `wasmparser` would improve compile times, artifact sizes and maybe (or hopefully) also parse and validation performance since certain checks can be performed at compile time instead.

Ideas, thoughts and discussion welcome. :)

alexcrichton:
One point perhaps worth clarifying is that the `gc` feature is for Wasmtime, not wasm-tools. I bring this up because while wasmparser has a `component-model` feature one of the main things that made it possible is that it's purely an optional layer on top of core wasm. That made it quite easy to have structs "just disappear" from the API for component-model support and from an end-user perspective it's easy to work with the crate both with/without component-model support.

This is a bit different for core wasm proposals because everything is pervasively present throughout the wasm type system and structures exposed by this crate. For example the `Visitor` trait and `Operator` enum would be entirely different with-and-without the `gc` or `simd` features. To me this would be a major sticking point of adding a feature gate here because the common workarounds of default trait methods or `#[non_exhaustive]` I don't think are great here because it removes the exhaustiveness checking done by rustc.

One possible option perhaps is something like:

```rust
trait Visitor {
    fn simd_visitor(&mut self) -> Option<&mut dyn SimdVisitor> { None }
}

trait SimdVisitor {
    // ... all simd methods 
}
```

for the trait side of things and maybe `enum Operator { Simd(SimdOperator), ... }` for `Operator`, but there's still a fair bit of design work to do there to determine whether that'd work.

---

[...]

PR #1903, opened 2024-11-12
`wasmparser`: feature gate Wasm `simd` support

Robbepop:
Closes https://github.com/bytecodealliance/wasm-tools/issues/1899.

This allows to feature gate parsing and validation of all 268 `simd` and `relaxed-simd` Wasm operators.

**Goals:** improved compile times and reduced artifact size.

# Summary

- Removes all `simd` and `relaxed-simd` definitions from `macro_rules! for_each_operator`.
- Adds `macro_rules! for_each_simd_operator`.
- Adds `SimdOperator` enum.
- Adds `VisitSimdOperator` trait.
- Adds `VisitOperator::simd_visitor` trait method with default implementation.
- Adds `simd` crate feature, gating `simd` and `relaxed-simd` Wasm proposals support.
- Adds `Operator::Simd(SimdOperator)` variant.
- Puts `#[non_exhaustive]` on `Operator` enum.

# Notes

- There might be a performance regression since visiting SIMD operators in `BinaryReader` now is done via dynamic dispatch. Although it is likely that LLVM can optimize this. Benchmarks needed.
- Usage of `wasmparser` as library is a bit more complex since we now have 2 generator macros instead of just one which is a bit ugly in my honest opinion.

# Benchmarks

Command: `cargo build -p wasmparser --no-default-features --features validate,features`
Machine: Macbook M2 Pro

| Compile Time | `!simd` | `simd` |
|:--|--:|--:|
| `debug` | 2.64s | 2.90s |
| `release` | 3.27s | 3.64s |
| | |
| Artifact Size | | |
| `debug` | 19MB | 21MB |
| `release` | 8.4MB | 9.6MB |

# ToDo's

- [x] Fix other crates in the `wasm-tools` workspace.
    - [x] `wasm-encoder`
    - [x] `wasmprinter`
    - [x] `wasm-mutate`
- [x] Properly feature gate using the new `simd` crate feature where possible.
- [x] Move some `simd` definitions into a single file to speed-up conditional compilation gating.
    - [x] Move operator validator SIMD related functionality into a separate file.
        - This also helps with reviewability of the changes.
    - [x] Move `for_each_simd_operator` macro into its own separate file.
    - [x] Move `BinaryReader::visit_0xfd_operator` into its own separate file.
- [x] Write docs for new `VisitSimdOperator` trait and `SimdOperator` enum.
    - [x] `macro_rules! for_each_simd_operator`
    - [x] `enum SimdOperator` enum
    - [x] `trait VisitSimdOperator`
    - [x] `VisitOperator::simd_visitor`
- [ ] ~~Remove some duplicate code around macro expansions if possible.~~
    - Unfortunately this is not entirely preventable. Maybe @alexcrichton has an idea here and there.


alexcrichton:
Thanks again for pushing on this! Overall this is roughly what I was thinking, and while I'm ok with the current state of it I think it's worth reviewing the final end-state as well when all the CI is passing and everything is hooked up (e.g. #[cfg]s are in place and compile-time-wins are remeasured and such)

> OperatorFactory now panics when trying to visit a SIMD operator while simd_visitor returns None.

Can this be fixed by marking `enum Operator` as `#[non_exhaustive]` and then conditionally including the `SimdOperator` enum?

> There might be a performance regression since visiting SIMD operators in BinaryReader now is done via dynamic dispatch. Although it is likely that LLVM can optimize this. Benchmarks needed.

I'm not necessarily overly worried about this since simd instructions are likely a very small fraction of a larger module. Might be worth still trying to confirm one way or another, but I don't think it'd be the end of the world either way. When this might perhaps be an issue is with the GC proposal where if a similar strategy is taken for GC then that could have a significant impact since modules are probably all-GC or mostly-all-MVP for example.

> Usage of wasmparser as library is a bit more complex since we now have 2 generator macros instead of just one which is a bit ugly in my honest opinion. Maybe it is possible to shrink to a single macro again but I don't know how at the moment.

I think it's pretty reasonable as is personally. There's two macros to invoke but they're both passed the same macro since they have the same syntax so while it's a bit more boilerplate that's sort of just par for the course with these sorts of refactorings and is a cost to acknowledge being placed on users where the benefit is being able to conditionally disable simd/etc.

Robbepop:
@alexcrichton Thank you for the preliminary review of the WIP PR. :)

> Thanks again for pushing on this! Overall this is roughly what I was thinking, and while I'm ok with the current state of it I think it's worth reviewing the final end-state as well when all the CI is passing and everything is hooked up (e.g. #[cfg]s are in place and compile-time-wins are remeasured and such)

I will be working on that now. I just wanted to know if the current structure is as you imagined which seems to be the case.

> Can this be fixed by marking `enum` Operator as `#[non_exhaustive]` and then conditionally including the `SimdOperator` `enum`?

Should be possible indeed.

> When this might perhaps be an issue is with the GC proposal where if a similar strategy is taken for GC then that could have a significant impact since modules are probably all-GC or mostly-all-MVP for example.

We only do all this acrobatics because there are so many `simd` operators. From what I know, no other Wasm proposal has nearly as many operators as `simd`, not even the `gc` proposal. So the first question is if we actually need to conditionally compile `gc` support. So far I do not think so.

> I think it's pretty reasonable as is personally. There's two macros to invoke but they're both passed the same macro since they have the same syntax so while it's a bit more boilerplate that's sort of just par for the course with these sorts of refactorings and is a cost to acknowledge being placed on users where the benefit is being able to conditionally disable simd/etc.

Okay that you for your view on this. I agree!

alexcrichton:
I did a small analysis of opcodes in wasmparser recently and this is what I got:

wasm opcodes

basically while simd is big it's only ~1/3. Roughly 1/4 of all opcodes are mvp and threads+shared-everything-threads are becoming a relatively large slice of the pie

alexcrichton:
Oh not necessarily, and shared-everything-threads is still under development itself. While simd is a relatively obvious choice of "can split this off" I'm not sure if there are other reasonable things to split off in chunks at this time.

Robbepop:
I finished all TODOs for the `wasmparser` crate and performed some compilation and artifact size benchmarks:

| Compile Time | `!simd` | `simd` |
|:--|--:|--:|
| `debug` | 2.64s | 2.90s |
| `release` | 3.27s | 3.64s |

| Artifact Size | `!simd` | `simd` |
|:--|--:|--:|
| `debug` | 19MB | 21MB |
| `release` | 8.4MB | 9.6MB |

crates/wasmparser/src/readers/core/operators.rs
```diff
@@ -423,16 +446,55 @@ pub trait VisitOperator<'a> {
...
             }
-
         }
         for_each_operator!(visit_operator)
     }
 
+    #[cfg(feature = "simd")]
+    fn simd_visitor(&mut self) -> Option<&mut dyn VisitSimdOperator<'a, Output = Self::Output>> {
+        None
+    }
```
alexcrichton, line 465: Could you add some documentation to this method as well? Mostly indicating what it is and how it's intended to be used. I think ideally it'd be good to have an example which shows the return-`Some(self)` trick

alexcrichton: Also, I'm not sure if it's possible, but one possible route of exploration would be to have an associated `type SimdVisitor;` for this trait where the return value is `Option<&mut Self::SimdVisitor>`. Typically users would write `type SimdVisitor = Self;`. The only slightly tricky part would be the default value of the associated type but I think that may be possible with something like `enum NonexistentVisitor {}` which can't be constructed.

alexcrichton (commented):
Ah alas I was hoping for slightly more compile-time wins in the end but if this works for you I think it's reasonable to land.

crates/wasmparser/src/readers/core/operators.rs
Robbepop: > Also, I'm not sure if it's possible, but one possible route of exploration would be to have an associated `type SimdVisitor;` for this trait where the return value is `Option<&mut Self::SimdVisitor>`. Typically users would write `type SimdVisitor = Self;`. The only slightly tricky part would be the default value of the associated type but I think that may be possible with something like `enum NonexistentVisitor {}` which can't be constructed.

This is how I would have probably designed this in the first place. However, your design idea has the crucial elegance of requiring no further boilerplate code for the default `NonExistentVisitor` which could be massive and diminish the compile time wins in the worst case.

Robbepop:
> Ah alas I was hoping for slightly more compile-time wins in the end but if this works for you I think it's reasonable to land.

Indeed, I was also looking for more. Maybe we can gain some more wins by some further refactorings. At the moment I do not see any trivial ones though.

Robbepop:
@alexcrichton this PR is ready for a proper review now.

crates/wasm-encoder/Cargo.toml
```diff
@@ -33,8 +33,10 @@ wasmparser = { path = "../wasmparser" }
...
 
 [features]
-default = ['component-model']
+default = ['component-model', 'simd']
 
 # On-by-default: conditional support for emitting components in addition to
 # core modules.
 component-model = ['wasmparser?/component-model']
+# On-by-default: conditional support for emitting SIMD wasm operators.
+simd = ['wasmparser?/simd']
```
alexcrichton, line 42: Given that this feature only gates the wasmparser-to-wasm-encoder conversion, as opposed to gating the instructions themselves, I think it might be better here to just unconditionally enable the `simd` feature of wasmparser. That can be left as future work to implement full conditional gating in `wasm-encoder`

crates/wasm-mutate/src/mutators/modify_const_exprs.rs
```diff
@@ -76,7 +76,7 @@ impl<'cfg, 'wasm> Reencode for InitTranslator<'cfg, 'wasm> {
             O::RefNull { .. } | O::I32Const { value: 0 | 1 } | O::I64Const { value: 0 | 1 } => true,
             O::F32Const { value } => value.bits() == 0,
             O::F64Const { value } => value.bits() == 0,
-            O::V128Const { value } => value.i128() == 0,
+            O::Simd(SimdOperator::V128Const { value }) => value.i128() == 0,
```
alexcrichton, line 79: Looking at this again, what do you think of throwing all the simd ops back into `Operator`? It's already `#[non_exhaustive]` and would be nice to have a bit less churn as well in theory

Cargo.toml
```diff
@@ -239,6 +240,11 @@ component = [
   'dep:wasmparser',
   'dep:serde_json',
 ]
+simd = [
+  'wasmparser/simd',
+  'wasmprinter/simd',
+  'wasm-encoder/simd',
+]
```
alexcrichton, line 247: I think for this, sort of like wasm-encoder below, let's go ahead and just unconditionally enable simd for these dependencies at the CLI level perhaps?

crates/wasmprinter/Cargo.toml
```diff
@@ -25,5 +25,6 @@ termcolor = { workspace = true }
 wat = { path = "../wat" }
 
 [features]
-default = ['component-model']
+default = ['component-model', 'simd']
 component-model = ['wasmparser/component-model']
+simd = ['wasmparser/simd']
```
alexcrichton, line 30: Similar to wasm-encoder, what do you think about unconditionally enabling simd here? That would avoid the need for more `#[cfg]` and I don't think there's as compelling of a use case to making this conditional on this crate as there is for wasmparser

crates/wasm-encoder/Cargo.toml
Robbepop: I agree. I did this just because I wanted to mic the behavior for the component-model. Enabling sims by default should also decrease the diff of this PR which I welcome.

crates/wasm-mutate/src/mutators/modify_const_exprs.rs
Robbepop: I am not yet sure if this works as intended due to the split into 2 macros and the problem that you cannot have macro invokation at enum variant definition position but I ll see how it goes. A downside to this approach is that this would result in tons of cfgs.

alexcrichton (commented):
Thanks for continuing to push on the compile-time improvements with the macro!

This new structure is currently used to define `for_each_operator` and `for_each_simd_operator`, but the end goal was to remove the `Operator::Simd` variant and `SimdOperator` entirely. Would you be up for sketching that out here? That I think will be pretty helpful to avoid lots of churn in downstream users currently using the `Operator` enum.

What I might envision as the next transition step might be:

* `for_each_visit_operator!` - used with `trait VisitOperator` (no simd)
* `for_each_simd_visit_operator!` - used with `trait VisitSimdOperator` (only simd)
* `for_each_operator!` - repurpose this to be used with `Operator` (both simd + others)

A possible future change would be to update the syntax of each macro possible where `for_each_operator` may not want the `visit_*` methods as arguments and the others may not want the camel case struct names as arguments. That's fine to defer to later though.

Robbepop:
> What I might envision as the next transition step might be:
> 
>     * `for_each_visit_operator!` - used with `trait VisitOperator` (no simd)
> 
>     * `for_each_simd_visit_operator!` - used with `trait VisitSimdOperator` (only simd)
> 
>     * `for_each_operator!` - repurpose this to be used with `Operator` (both simd + others)

Associating the new macros with their respective intended use is a brilliant idea. I was already struggeling how to best expose and name those new macros but this is a solution I am very willing to put forward.

> A possible future change would be to update the syntax of each macro possible where `for_each_operator` may not want the `visit_*` methods as arguments and the others may not want the camel case struct names as arguments. That's fine to defer to later though.

Yes, this indeed might be a good idea. I agree that a follow-up PR is a better place for those improvements.

alexcrichton (approved):

PR #1905, opened 2024-11-13
`wasmparser`: Update `hashbrown` to `v0.15.1` and use its `default-hasher` to replace `ahash` usage

Robbepop:
- Updates to `hashbrown` v0.15.1.
- Also remove `ahash` and use hashbrown's `default-hasher` crate feature as replacement. This allows to drop yet another `wasmparser` dependency.

### Note

- `hashbrown` started using [`foldhash`](https://github.com/orlp/foldhash) as its default-hasher starting in `v0.15.0`. From what I know it should be an improvement over `ahash`.
- `hashbrown` by default is using `foldhash-f` and not `foldhash-q` where `foldhash-f` is the "fast" version and `foldhash-q` is the hash-DOS resistant version. It seems that `foldhash-f` is more DOS-resistant than `fxhash` but less than both `foldhash-q` and `ahash`. Maybe this is a reason to _not_ use `hashbrown`'s default hasher. For Wasmi my stance is that if you want DOS-resistance, you simply do not use hash-tables at all by disabling `hash-collections` feature.

Q: Are there any other workspace crates that need adjustment?

Background: [I did the same in Wasmi](https://github.com/wasmi-labs/wasmi/pull/1305) and found it to be a good solution so far.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -746,11 +746,12 @@ dependencies = [
 
 [[package]]
 name = "hashbrown"
-version = "0.15.0"
+version = "0.15.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1e087f84d4f86bf4b218b927129862374b72199ae7d8657835f1e89000eea4fb"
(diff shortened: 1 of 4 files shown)


alexcrichton (approved):
Thanks!

PR #1906, opened 2024-11-13
`wasmparser`: optimize type section valiadation by specializing over Wasm feature subset

Robbepop:
Closes https://github.com/bytecodealliance/wasm-tools/issues/1701.

cc @alexcrichton : Are there any other Wasm proposals that might conflict?

This implements a lightweight type section validation for a subset of Wasm proposals.

## Benchmarks

I compared `main` against PR and changed the used `WasmFeatures` set for both:

```rust
Validator::new_with_features(
    WasmFeatures::WASM2
        .union(WasmFeatures::CUSTOM_PAGE_SIZES)
        .union(WasmFeatures::EXTENDED_CONST)
        .union(WasmFeatures::MEMORY64)
        .union(WasmFeatures::MULTI_MEMORY)
        .union(WasmFeatures::RELAXED_SIMD)
        .union(WasmFeatures::TAIL_CALL)
        .union(WasmFeatures::THREADS)
        .union(WasmFeatures::WIDE_ARITHMETIC);
)
```

image

### DIFF
--- a/crates/wasmparser/src/validator/core.rs
+++ b/crates/wasmparser/src/validator/core.rs
@@ -569,7 +569,7 @@ impl Module {
                 offset,
             )?;
         }
-        if !(features.function_references() || features.gc() || features.component_model()) {
+        if !(features.function_references() || features.gc() || features.component_model() || features.stack_switching()) {
             // Without `function-references`, `gc` and the `component-model` proposals
             // we can use a special type validation that is simpler and more efficient.
             if rec_group.is_explicit_rec_group() {
--- a/crates/wasmparser/src/validator/core.rs
+++ b/crates/wasmparser/src/validator/core.rs
@@ -571,7 +571,11 @@ impl Module {
         }
         /// The subset of `WasmFeatures` for which we know that the
         /// fast type section validation can be safely applied.
-        const ALLOWED_FEATURES: WasmFeatures = WasmFeatures::WASM2
+        /// 
+        /// Fast type section validation does not have to canonicalize
+        /// (deduplicate) types and does not have to perform sub-typing
+        /// checks.
+        const FAST_VALIDATION_FEATURES: WasmFeatures = WasmFeatures::WASM2
             .union(WasmFeatures::CUSTOM_PAGE_SIZES)
             .union(WasmFeatures::EXTENDED_CONST)
             .union(WasmFeatures::MEMORY64)
@@ -580,9 +584,7 @@ impl Module {
             .union(WasmFeatures::TAIL_CALL)
             .union(WasmFeatures::THREADS)
             .union(WasmFeatures::WIDE_ARITHMETIC);
-        if ALLOWED_FEATURES.contains(*features) {
-            // Note: without the proposals guarded above we can use a special
-            //       type validation that is both simpler and more efficient.
+        if FAST_VALIDATION_FEATURES.contains(*features) {
             if rec_group.is_explicit_rec_group() {
                 bail!(offset, "requires `gc` proposal to be enabled")
             }
--- a/crates/wasmparser/src/validator/core.rs
+++ b/crates/wasmparser/src/validator/core.rs
@@ -569,11 +569,18 @@ impl Module {
                 offset,
             )?;
         }
-        if !(features.function_references()
-            || features.gc()
-            || features.component_model()
-            || features.stack_switching())
-        {
+        /// The subset of `WasmFeatures` for which we know that the
+        /// fast type section validation can be safely applied.
+        const ALLOWED_FEATURES: WasmFeatures = WasmFeatures::WASM2
+            .union(WasmFeatures::CUSTOM_PAGE_SIZES)
+            .union(WasmFeatures::EXTENDED_CONST)
+            .union(WasmFeatures::MULTI_MEMORY)
+            .union(WasmFeatures::WIDE_ARITHMETIC)
+            .union(WasmFeatures::MEMORY64)
+            .union(WasmFeatures::TAIL_CALL)
+            .union(WasmFeatures::RELAXED_SIMD)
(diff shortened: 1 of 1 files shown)


Robbepop:
I also ran the `wam-tools` Wasm testsuite for validation and disabled the `function-references` and `gc` proposals to trigger the fast-validation path. As expected, the testsuite did not pass with the following failures:

```
failures:
    "tests/local/component-model/gc.wast"
    "tests/local/exnref/try-table.wast"
    "tests/local/folding/array_fill.wast"
    "tests/local/folding/array_init_data.wast"
    "tests/local/folding/array_copy.wast"
    "tests/local/folding/array_init_elem.wast"
    "tests/local/folding/array.wast"
    "tests/local/function-reference-structural-matching.wat"
    "tests/local/folding/struct.wast"
    "tests/local/folding/fold-stack-switching-gc.wast"
    "tests/local/folding/fold-stack-switching.wast"
    "tests/local/instance.wast"
    "tests/local/invalid-init-expr.wast"
    "tests/local/shared-everything-threads/polymorphism.wast"
    "tests/local/shared-everything-threads/gc-unreachable.wast"
    "tests/local/shared-everything-threads/builtins.wast"
    "tests/local/shared-everything-threads/funcs.wast"
    "tests/local/shared-everything-threads/rec-groups.wast"
    "tests/local/shared-everything-threads/i31.wast"
    "tests/local/shared-everything-threads/tables.wast"
    "tests/local/shared-everything-threads/ref-equivalence.wast"
    "tests/local/stack-switching/cont_bind.wast"
    "tests/local/stack-switching/actor-lwt.wast"
    "tests/local/shared-everything-threads/globals.wast"
    "tests/local/stack-switching/actor.wast"
    "tests/local/stack-switching/cont_canonicalisation.wast"
    "tests/local/stack-switching/async-await.wast"
    "tests/local/stack-switching/cont_new.wast"
    "tests/local/shared-everything-threads/structs.wast"
    "tests/local/stack-switching/cont_suspend.wast"
    "tests/local/stack-switching/cont_resume.wast"
    "tests/local/shared-everything-threads/types.wast"
    "tests/local/stack-switching/fun-state.wast"
    "tests/local/stack-switching/generators.wast"
    "tests/local/stack-switching/pipes.wast"
    "tests/local/stack-switching/fun-actor-lwt.wast"
    "tests/local/stack-switching/fun-lwt.wast"
    "tests/local/stack-switching/control-lwt.wast"
    "tests/local/stack-switching/lwt.wast"
    "tests/local/shared-everything-threads/arrays.wast"
    "tests/local/stack-switching/static-lwt.wast"
    "tests/local/stack-switching/cont.wast"
    "tests/local/stack-switching/validation_gc.wast"
    "tests/local/stack-switching/validation.wast"
    "tests/local/component-model/types.wast"
```

I have looked through a bunch of the reported errors and they seem to be expected from the turning-off of features. However, it would certainly be nicer if we had an automated way of testing this fast-validation path.

Robbepop:
I think is possible to simplify Wasm feature checking by making `wasmparser` aware of Wasm feature dependencies. For example, always enable `function-references` when the `gc` proposal is enabled because `gc` depends on `function-references`, etc..
Or are there problems that I have overlooked? 🤔

Robbepop:
> Thanks! Do you think it would be possible to optimize the processing of types in-place though rather than conditionally gating on features? From a high-level I don't think we have any other logic where we significantly differ on features for what path to take at runtime (except to return an error). I'm worried that this will become a relatively poorly tested path due to not being the default and being relatively nontrivial. I haven't dug in much though to try to understand what the slow parts are here though so it may not be possible.

I agree that this solution introduces a testing blind spot. For me this is just a temporary solution to finally make `wasmparser` be usable by Wasmi again. Unfortunately I was not able to come up with a more elegant solution since I was kinda confused by the canonicalization and internment of types which is also what is making type section validation so slow. Do you have an idea?

crates/wasmparser/src/validator/core.rs
```diff
@@ -569,6 +569,23 @@ impl Module {
                 offset,
             )?;
         }
+        if !(features.function_references()
+            || features.gc()
+            || features.component_model()
+            || features.stack_switching())
```
Robbepop, line 575: One improvement that I just see here is that instead of disallow-listing certain proposals, we could allow-list all the proposals of which we know this fast validation is applied correctly. I think this is a more future proof solution. We could construct a `const FAST_VALIDATE_FEATURES: WasmFeatures = ...` and then just check via a single
```rust
FAST_VALIDATE_FEATURES.contains(features)
```

Robbepop: I just implemented this and am convinced that this is a much better approach for maintainability and future proofing of continuous feature development.

Robbepop:
I have experimented with an implemented for the idea presented in [this comment](https://github.com/bytecodealliance/wasm-tools/pull/1906#issuecomment-2477526745). However, problems arised quickly due to certain invariants of other structures in the validator's greater scheme. For example resetting the `TypeAlloc`. Furthermore, compared to the solution presented in this PR, there was some overhead, even for the fast case.

All in all, I think a proper, more elegant approach would take a lot of more work and consideration. This PR implements a very fast type section validation at nearly no runtime cost (one bit-flags check). The feature gated allow list should be fairly maintainble since development of new features won't casually break the type validator. Once we figure out a more elegant and inline approach for fast type section validation, this solution can easily be replaced since it is neatly isolated and contained in a single spot.

alexcrichton (approved):
Thanks for investigating! I think that's a reasonable approach myself and the benefits seem worth the theoretical risk here, so I'm happy to land this. It's also not like this has zero testing, the "feature frobbing" done in Wasmtime and wasm-tools fuzzing should still cover this case.

PR #1908, opened 2024-11-15
Add an assert to wit-parser's `assert_valid`.

sunfishcode:
Indexmap's [`MutableKeys` trait] allows one to mutate the keys in an `IndexMap`, however it doesn't prevent bugs where the new keys have different hash values from the old keys. Since I recently made this mistake, add an assert to help catch such bugs. It doesn't always catch the bug, because whether or not the bug is observable depends on hashmap internals, but it's better than nothing.

[`MutableKeys` trait]: https://docs.rs/indexmap/latest/indexmap/map/trait.MutableKeys.html


alexcrichton (approved):
Whoa I had no idea mutation was possible, that's useful!

PR #1909, opened 2024-11-16
Prepare to update spec test suite

alexcrichton:
Some minor updates in preparation for WebAssembly/testsuite#110:

* Don't limit the size of tables in the validator and instead defer such maximal validation to runtimes themselves.
* Parse the `(pagesize N)` syntax on "inline memories"
* Always parse table/memory limits as 64-bit integers
* Stop matching spec test suite error messages

This last point is something I've tried to hold off on doing for a long time but as the giant `error_matches` function continues to grow and become more unwieldy it has become less and less tenable to do this. Overall it doesn't seem too beneficial to keep maintaining this especially in the face of numerous proposals, so instead basically remove it entirely.

### DIFF
--- a/crates/wasmparser/src/validator/core.rs
+++ b/crates/wasmparser/src/validator/core.rs
@@ -761,12 +761,6 @@ impl Module {
         }
 
         self.check_limits(ty.initial, ty.maximum, offset)?;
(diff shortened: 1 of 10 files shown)


pchickey (approved):
Maybe change the name of `table-too-big` test to just `table-big`? because its no longer too big...

PR #1911, opened 2024-11-22
use error_in_core for BinaryReaderError

12101111:


pchickey:
Thanks! This rust feature got stabilized in 1.82. We will have to wait a little longer because of the MSRV policy of this project is to be at least 2 behind stable: https://github.com/bytecodealliance/wasm-tools/blob/main/Cargo.toml#L64-L66

So, once 1.84 is out, we can bump the MSRV in Cargo.toml and land this.

alexcrichton:
For wasm-tools the minimum version lags a bit behind Wasmtime due to the broader usage in other projects as well, but in the meantime @12101111 would you be up for adding detection in the build script? One option for doing this is to detect the rustc version in a build script and if it's at-or-above the stabilization of `core::error` then we can unconditionally include this impl (so either `feature = "std"` or new enough rustc)

Issue #1922, opened 2024-11-29
[wasm-metadata] support OCI annotations


yoshuawuyts:
## Description

During a BA Packaging SG meeting earlier this year we discussed adopting support for [OCI Annotations Spec](https://specs.opencontainers.org/image-spec/annotations/)-compatible metadata for Wasm Components. The idea is that language-tools like `cargo-component` can read metadata from e.g. `Cargo.toml` and encode it as custom sections in Wasm. And then tools like `wkg` can read those custom sections, and encode them into the OCI images in the correct locations. This makes it so that adding component metadata into registries is something that just happens automatically.

Currently `wasm-metadata` includes support for something called `RegistryMetadata`. This is encoded in Wasm binaries as a section named `registry-metadata` and stored as JSON. In the Packaging SG meeting we decided to deprecate this approach and instead replace it with a custom section per field.

Because we are designing this with the express goal of being compatible with the OCI Annotations Spec, the value encodings must be compatible. We do however have liberty over the names of the custom sections, and we can define the mappings between Wasm and OCI in the Wasm OCI specification.

I'm opening this issue to explain what the current plan for this is before opening up any PRs that implement this. My current plan for this is as follows:

1. Remove support for `RegistryMetadata` from wasm modules
2. Replace `RegistryMetadata` with dedicated custom sections for components
3. Add support for any remaining custom sections
4. Document these sections in `wasm-tool-conventions` 
5. Document these sections in the Wasm OCI Spec

I hope this all makes sense. Happy to answer questions if anyone has any. Thanks!

## Tasks

The following sections would need to be encoded to fully match the OCI Annotations Spec:

- [x] author
- [x] description custom section (supported by [GitHub packages][gh])
- [x] licenses custom section (supported by [GitHub packages][gh])
- [x] source custom section (supported by [GitHub packages][gh])
- [x] homepage custom section
- [x] version custom section
- [x] revision custom section
- [x] ~~created custom section~~

[gh]: https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#labelling-container-images

## References

- https://github.com/WebAssembly/WASI/issues/608
- https://github.com/WebAssembly/tool-conventions/issues/230

yoshuawuyts:
I'm actually having some reservations about including `created` in a custom section. I for one like reproducible builds, and I believe that might run directly counter to it.

yoshuawuyts:
[...]

PR #1925, opened 2024-11-29
[wasm-metadata] parse OCI author custom section

yoshuawuyts:
This PR adds support for parsing the `author` custom section, implementing parts of #1922. Thanks!

## Draft Notes

I'm filing this PR early and as a draft because I'm not 100% sure I'm following best practices here. It's my first time using `wasmparser` and `wasm-encoder`, and I'm for example not sure whether `parse_wasm` should take a `BinaryReader` or `CustomSectionReader`.

Also: I was looking for something like a `Parse` trait, but I don't think `wasmparser` has one? I quite liked the encoder side of things.

Finally: in this PR I'm basically just wrapping `CustomSection` in a newtype - however that means storing the internal data as a byte array rather than a string. How do we feel about that?

I'm asking these questions since I'm planning to repeat this pattern a number of times for the other custom sections, and it seems worth making sure it's early on. Thanks!

### DIFF
--- a/crates/wasm-metadata/src/lib.rs
+++ b/crates/wasm-metadata/src/lib.rs
@@ -5,6 +5,7 @@
 pub use add_metadata::AddMetadata;
 pub use metadata::Metadata;
 pub use names::{ComponentNames, ModuleNames};
+pub use oci_annotations::Author;
 pub use producers::{Producers, ProducersField};
 pub use registry::{CustomLicense, Link, LinkType, RegistryMetadata};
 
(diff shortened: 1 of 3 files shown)


alexcrichton (commented):
Thanks for this! What I might recommend is to take the `&[u8]` entire wasm in memory as opposed to just a custom section. That way you can bake in how to handle a missing section, duplicated sections, the section name, etc. Internally you can use a `BinaryReader` to parse the binary data if necessary, but yeah the `Parse`-trait side of things  is [sort of there with `FromReader`](https://docs.rs/wasmparser/latest/wasmparser/struct.BinaryReader.html#method.read) but it's no the same as the encoding side of things.

alexcrichton (approved):
Looks great to me, thanks for this!

PR #1930, opened 2024-12-02
 Reimplement fast-path validation of MVP types 

alexcrichton:
This commit fixes an issue from https://github.com/bytecodealliance/wasm-tools/pull/1906 which is preventing the upgrade of
wasm-tools in Wasmtime. That commit implemented a fast path by skipping
a function and reimplementing parts of it internally, but that then
caused Wasmtime to panic when other data structures weren't filled out.
The intention was that the user-facing interface doesn't change
depending on features, so this is an attempt at fixing the mistake in
that commit.

The fix here is to remove the fast path added and restructure it
differently. Instead now the fast and normal paths have much less
divergence which should prevent this issue from re-surfacing. This is
15% slower than `main` but it doesn't have the same bug as `main` so for
now that may be the best that can be done.

### DIFF
--- a/crates/wasmparser/benches/benchmark.rs
+++ b/crates/wasmparser/benches/benchmark.rs
@@ -273,6 +273,9 @@ fn define_benchmarks(c: &mut Criterion) {
     fn validator() -> Validator {
         Validator::new_with_features(WasmFeatures::all())
     }
+    fn old_validator() -> Validator {
+        Validator::new_with_features(WasmFeatures::WASM2)
+    }
 
     let test_inputs = once_cell::unsync::Lazy::new(collect_benchmark_inputs);
 
@@ -330,6 +333,14 @@ fn define_benchmarks(c: &mut Criterion) {
                 validator().validate_all(&wasm).unwrap();
             })
         });
+        if old_validator().validate_all(&wasm).is_ok() {
+            c.bench_function(&format!("validate-old/{name}"), |b| {
+                Lazy::force(&wasm);
+                b.iter(|| {
+                    old_validator().validate_all(&wasm).unwrap();
+                })
+            });
+        }
         c.bench_function(&format!("parse/{name}"), |b| {
             Lazy::force(&wasm);
             b.iter(|| {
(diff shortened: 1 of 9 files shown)


alexcrichton:
cc @Robbepop

Robbepop:
@alexcrichton looks reasonable.

Can you provide me a link to the Wasmtime issue? It is nicer that you managed to unify both implementations. The 15% slowdown is a bummer. Do you have an idea how to make it faster again? According to my benchmarks `wasmparser v0.221` is still ~5-15% slower than `wasmparser v0.100` for Wasmi but I deemed that an acceptable trade-off. However, I would rather want to make the newer `wasmparser` faster. :) I am hopeful we find more ways to do so.

Robbepop (approved):
LGTM

alexcrichton:
No issue on the wasmtime side, just a [failed build](https://github.com/bytecodealliance/wasmtime/actions/runs/12126250834/job/33807995246). The failure is in [this loop](https://github.com/bytecodealliance/wasmtime/blob/213459835229ef436c9f5eebf884e7beaae64b0f/crates/environ/src/compile/module_environ.rs#L249-L290) which is iterating by rec groups and specifically [this one was the failure](https://github.com/bytecodealliance/wasmtime/blob/213459835229ef436c9f5eebf884e7beaae64b0f/crates/environ/src/compile/module_environ.rs#L270) where the fast path wasn't maintaining the rec group lists.

My guess is we could probably skip maintenance of the rec group lists entirely in the fast path, but only if the other paths that read these lists understand that they may be optional and/or not populated. That might require plumbing `WasmFeatures` more. I've long wanted to just store `WasmFeatures` on `self` in more places as it's currently plumbed in about a million places which is a bummer (and it's super small at 4-bytes so basically inconsequential to duplicate)

Robbepop:
> My guess is we could probably skip maintenance of the rec group lists entirely in the fast path, but only if the other paths that read these lists understand that they may be optional and/or not populated. That might require plumbing WasmFeatures more. I've long wanted to just store WasmFeatures on self in more places as it's currently plumbed in about a million places which is a bummer (and it's super small at 4-bytes so basically inconsequential to duplicate)

Sounds great and worthwhile!

PR #1938, opened 2024-12-06
Add `no_std` support to `wasm-encoder`

ark0f:


alexcrichton:
Thanks for the PR! For posterity can you describe as well how you're planning on using this? (e.g. the platform/use case you have in mind). Also mind updating [this comment](https://github.com/bytecodealliance/wasm-tools/blob/96c1aa1afadffa816b6dd0ffe03c6181a1e9baf4/Cargo.toml#L67-L68) as well to mention the new build script?

ark0f:
> For posterity can you describe as well how you're planning on using this? (e.g. the platform/use case you have in mind).

Sure. Great part of [Gear Protocol](https://github.com/gear-tech/gear) business logic is compiled into WASM to be deterministically executed on blockchain and thus we have `no_std` environment. We want to instrument user uploaded WASM programs with `wasmparser` and `wasm-encoder` libraries

ark0f:
> Also mind updating [this comment](https://github.com/bytecodealliance/wasm-tools/blob/96c1aa1afadffa816b6dd0ffe03c6181a1e9baf4/Cargo.toml#L67-L68) as well to mention the new build script?

Done

alexcrichton (approved):

PR #1946, opened 2024-12-10
[wasm-metadata] pretty-print cli output

yoshuawuyts:
Closes #1923. While this doesn't quite deliver on the vision I sketched in the issue, I believe it's a meaningful step up from the status quo. I've also filed https://github.com/Nukesor/comfy-table/issues/160, which if accepted will hopefully allow us to more closely match `duf`'s output.

I'm not quite sure what the process for introducing new dependencies here is. To pretty-print tables, we're going to need *some* printing library - and `comfy-table` seemed like a reasonable choice. I looked at the deps, and nothing particularly stood out to me as a potential problem?

Thanks!

## Screenshot
This is what the output currently looks like. I might do another pass to add colors, etc - but this seemed like a good starting point.

Screenshot 2024-12-10 at 03 08 38

PR #1948, opened 2024-12-10
Fix test bless

yoshuawuyts:
While working on https://github.com/bytecodealliance/wasm-tools/pull/1946 I noticed that the test case bless flag was changing a lot of files (174). That's because we currently include a mix of files where some have trailing newlines and others do not. This fixes the test case bless flag to always include trailing newlines - and then blesses the test suite to ensure all files have a trailing newline.

This should make it a lot easier to bless subsequent runs. Thanks!


tests/cli/merge-with-similar-versions.wit.invalid-semver.stdout
```diff
@@ -0,0 +1 @@
+
```
alexcrichton, line 1: Mind adding a special case that skips the output if the output is empty? (avoids adding lots of empty files)

alexcrichton (approved):

alexcrichton:
I think this ended up happening in https://github.com/bytecodealliance/wasm-tools/pull/1954, so I'm going to close this

yoshuawuyts:
oops, sorry I forgot I had this open. Ty for closing this!

PR #1951, opened 2024-12-14
Cap the limit of empty rec groups in `wasm-smith`

alexcrichton:
Looks like OSS-Fuzz has found an input where it creates lots and lots of empty rec groups triggering quadratic behavior so cap the number of empty rec groups to avoid triggering this case.

### DIFF
--- a/crates/wasm-smith/src/core.rs
+++ b/crates/wasm-smith/src/core.rs
@@ -51,6 +51,7 @@ pub struct Module {
     /// All types locally defined in this module (available in the type index
     /// space).
     types: Vec<SubType>,
+    empty_rec_groups: u32,
 
     /// Non-overlapping ranges within `types` that belong to the same rec
     /// group. All of `types` is covered by these ranges. When GC is not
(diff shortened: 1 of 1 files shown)


fitzgen:
> triggering quadratic behavior

Having a limit on the max number of rec groups that `wasm-smith` generates (empty or not) makes sense, generally. However, accidental quadratic behavior seems like something we should fix in its own right. Can you expand a bit on how we the quadratic behavior emerges?

crates/wasm-smith/src/core.rs
```diff
@@ -568,7 +570,12 @@ impl Module {
             if !keep_going {
                 break;
             }
-            self.arbitrary_rec_group(u, AllowEmptyRecGroup::Yes)?;
+            let allow_empty = if self.empty_rec_groups > 10 {
+                AllowEmptyRecGroup::Yes
+            } else {
+                AllowEmptyRecGroup::No
+            };
```
fitzgen, line 577: Can we either reuse the types limit here or add a rec groups limit to the config? Hardcoding is not ideal and 10 is also super small.

alexcrichton:
I'll just go ahead and close this in favor of https://github.com/bytecodealliance/wasm-tools/pull/1952 to address the quadratic behavior, I don't really want to add a config knob for this.

PR #1955, opened 2024-12-17
Update hashbrown deps due to security

quanterion:
In Microsoft we recently got a security notification due to usage of hashbrown 0.15.0
The borsh serialization of the HashMap did not follow the borsh specification. It potentially produced non-canonical encodings dependent on insertion order. It also did not perform canonicty checks on decoding.

This can result in consensus splits and cause equivalent objects to be considered distinct.

This was patched in 0.15.1.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -728,7 +728,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40ecd4077b5ae9fd2e9e169b102c6c330d0605168eb0e8bf79952b256dbefffd"
 dependencies = [
  "fallible-iterator",
- "indexmap 2.6.0",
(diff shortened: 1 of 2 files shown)


alexcrichton (approved):
Thanks!

quanterion:
@alexcrichton is there a place where I can look for approximate release schedule?

sunfishcode:
@quanterion I've now published a new wasm-tools including this PR.

alexcrichton:
@quanterion note that a release isn't necessary for this PR (although I realize it's already been done) as that's not how cargo dependencies work. The dependency directives here are a minimum bound, not a range, so the fixed versions could already be used in downstream crates.

PR #1956, opened 2024-12-17
[wit-component] Extend the WIT printer for use in syntax highlighting

tomasol:
Introduce Output trait as a visitor for collecting syntax tokens.
This trait enables syntax highlighting and supports advanced use cases, such as filtering tokens by package, interface, or function.

### DIFF
--- a/crates/wit-component/src/printing.rs
+++ b/crates/wit-component/src/printing.rs
@@ -86,10 +86,11 @@ impl WitPrinter {
         }
 
         if is_main {
-            self.print_semicolon();
-            self.output.push_str("\n\n");
+            self.output.semicolon();
+            self.output.push_str("\n");
         } else {
             self.output.push_str(" {\n");
+            //self.output.indent_start();
         }
 
         for (name, id) in pkg.interfaces.iter() {
@@ -121,10 +122,6 @@ impl WitPrinter {
             writeln!(&mut self.output, "}}")?;
         }
         Ok(())
-    }
-
-    fn print_semicolon(&mut self) {
-        self.output.push_str(";");
     }
 
     fn new_item(&mut self) {
@@ -166,8 +163,7 @@ impl WitPrinter {
             self.print_name(name);
             self.output.push_str(": ");
             self.print_function(resolve, func)?;
-            self.print_semicolon();
-            self.output.push_str("\n");
+            self.output.semicolon();
         }
 
         self.any_items = prev_items;
@@ -251,8 +247,7 @@ impl WitPrinter {
                 }
             }
             write!(&mut self.output, "}}")?;
-            self.print_semicolon();
-            self.output.push_str("\n");
+            self.output.semicolon();
         }
 
         for id in types_to_declare {
@@ -278,8 +273,7 @@ impl WitPrinter {
         self.output.push_str(" ");
         self.print_name(ty.name.as_ref().expect("resources must be named"));
         if funcs.is_empty() {
-            self.print_semicolon();
-            self.output.push_str("\n");
+            self.output.semicolon();
             return Ok(());
         }
         self.output.push_str(" {\n");
@@ -302,8 +296,7 @@ impl WitPrinter {
                 FunctionKind::Freestanding => unreachable!(),
             }
             self.print_function(resolve, func)?;
-            self.print_semicolon();
-            self.output.push_str("\n");
+            self.output.semicolon();
         }
         self.output.push_str("}\n");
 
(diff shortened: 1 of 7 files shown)


crates/wit-component/src/printing.rs
```diff
@@ -1,15 +1,56 @@
...
+use std::ops::Deref;
 use wit_parser::*;
 
 // NB: keep in sync with `crates/wit-parser/src/ast/lex.rs`
 const PRINT_F32_F64_DEFAULT: bool = true;
 
 /// A utility for printing WebAssembly interface definitions to a string.
 pub struct WitPrinter {
-    output: Output,
+    ext: WitPrinterExt<OutputToString>,
```
alexcrichton, line 13: FWIW it's ok to have breaking changes to this crate, so it might make more sense to repurpose the existing `WitPrinter` as `WitPrinter<O>` now?

crates/wit-component/src/printing.rs
```diff
@@ -33,98 +75,122 @@ impl Default for WitPrinter {
...
+            self.output.indent_end();
         }
         Ok(())
     }
 
-    fn print_semicolon(&mut self) {
-        self.output.push_str(";");
+    /// Print the specified package without its content.
+    /// Does not print the semicolon nor starts the indentation.
+    pub fn print_package_outer(&mut self, pkg: &Package) -> Result<()> {
```
alexcrichton, line 156: How come some methods were made `pub` in this PR? It's totally ok to do so, but I just wanted to confirm this wasn't by accident and it was intentional

crates/wit-component/src/printing.rs
```diff
@@ -990,66 +1115,223 @@ fn is_keyword(name: &str) -> bool {
...
+    fn newline(&mut self) {
+        self.output.push('\n');
+        self.needs_indent = true;
+    }
+
+    fn keyword(&mut self, src: &str) {
+        self.indent_and_print(src);
+    }
+
+    fn r#type(&mut self, src: &str, _kind: TypeKind) {
```
alexcrichton, line 1259: Could this be named perhaps `ty` or `type_`? (personally I prefer to avoid `r#foo` where possible)

crates/wit-component/src/printing.rs
```diff
@@ -990,66 +1115,223 @@ fn is_keyword(name: &str) -> bool {
...
+    fn case(&mut self, src: &str);
+    /// Generic argument section starts. In WIT this represents the `<` character.
+    fn generic_args_start(&mut self);
+    /// Generic argument section ends. In WIT this represents the '>' character.
+    fn generic_args_end(&mut self);
+    /// Called when a single documentation line is added.
+    /// The `doc` parameter starts with `///` omitted, and can be an empty string.
+    fn doc(&mut self, doc: &str);
+    /// A semicolon is added.
+    fn semicolon(&mut self);
```
alexcrichton, line 1139: Do you think it might make sense to forward these methods to `str` by default? That way the implementation for `OutputString` could be a bit simpler perhaps

alexcrichton (commented):
Thanks for this! Are you planning on integrating this externally into an IDE or similar perhaps? 

As an optional addition to this PR, it'd be pretty neat if this could get integrated into `wasm-tools component wit` in the same way that `wasm-tools print` will use terminal colors by default. No need to do that here though if you're not interested in doing so.

crates/wit-component/src/printing.rs
tomasol: Sure.

crates/wit-component/src/printing.rs
tomasol: Sure, I've renamed it to `ty`.

tomasol:
> Thanks for this! Are you planning on integrating this externally into an IDE or similar perhaps?

Thanks for the quick review! I am building an HTML highlighter as part of a larger project - not just for displaying colors, but also for deeper integration, such as showing icons next to certain functions and filtering out parts of a WIT document.

This is also why some higher-level functions are marked as pub in the PR.

Here is how it looks now:
image

crates/wit-component/src/printing.rs
tomasol: It was intentional, answering here: https://github.com/bytecodealliance/wasm-tools/pull/1956#issuecomment-2551305004

alexcrichton (approved):
Thanks for this, and that looks pretty cool! One small thing here but otherwise should be good to go

alexcrichton (approved):

PR #1958, opened 2024-12-18
fix overly-aggressive pruning of types based on features

dicej:
Previously, `ast::Resolve` would sometimes assign a `Stability` to types according to the first function it found that type in as a parameter or result. That's a problem if such a function is marked `@unstable`, since it will effectively poison that type and prevent it from being used by any other function -- even if the type itself is ungated.

This fixes the problem by removing the `&Stability` parameter from `resolve_params` and `resolve_results` and instead passing `&Stability::Unknown` to `resolve_type` from those functions.  Additionally, I've added a new `find_stability` function which recursively inspects a `&TypeDefKind`, looking for a non-unknown stability.  If it finds one, it will use that; otherwise, it will default to `Stability::Unknown`.

For example, if `option<T>` appears as a parameter type in a function, we'll start by assuming that type has unknown stability, then look to see if `T` has a stability (directly or transitively) and use that if so.  Things get interesting in the case of e.g. `tuple<T, U, V>`, where `T`, `U`, and `V` each have their own stability.  Currently we just use the first known stability we find, but perhaps there's a better approach to consider?

### DIFF
--- a/crates/wit-parser/src/ast/resolve.rs
+++ b/crates/wit-parser/src/ast/resolve.rs
@@ -1045,8 +1045,8 @@ impl<'a> Resolver<'a> {
     ) -> Result<Function> {
         let docs = self.docs(docs);
         let stability = self.stability(attrs)?;
-        let params = self.resolve_params(&func.params, &kind, func.span, &stability)?;
-        let results = self.resolve_results(&func.results, &kind, func.span, &stability)?;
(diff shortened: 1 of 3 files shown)


pchickey (approved):
For the tuple and related cases you'd need to redesign the definition of Stability to have a way to represent a union of multiple features, but I think its fine to fix this without opening up that can of worms.

PR #1959, opened 2024-12-18
add stream/future/error-context to wit-smith

dicej:
I had to tweak fuzz/src/wit64.rs because otherwise the old version of `wit-component` it was using choked on these new types.  Now we just return early when that happens.

Otherwise, there was just one other failure in the 15+ minutes I've been running the fuzzer: I had neglected to support the new types in `WitPrinter::declare_type` thinking it wasn't necessary, but I was wrong.

### DIFF
--- a/crates/wit-component/src/printing.rs
+++ b/crates/wit-component/src/printing.rs
@@ -754,9 +754,13 @@ impl<O: Output> WitPrinter<O> {
                         }
                         None => bail!("unnamed type in document"),
                     },
-                    TypeDefKind::Future(_) => panic!("no need to declare futures"),
-                    TypeDefKind::Stream(_) => panic!("no need to declare streams"),
-                    TypeDefKind::ErrorContext => panic!("no need to declare error-contexts"),
+                    TypeDefKind::Future(inner) => {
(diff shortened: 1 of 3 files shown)


crates/wit-component/src/printing.rs
```diff
@@ -893,6 +897,54 @@ impl WitPrinter {
         Ok(())
     }
 
+    fn declare_stream(&mut self, resolve: &Resolve, name: Option<&str>, ty: &Type) -> Result<()> {
```
alexcrichton, line 900: Could this take a non-optional name and an error be generated above if `ty.name` is none?

alexcrichton (approved):

crates/wit-component/src/printing.rs
dicej: I was following the precedent set by `declare_list`, `declare_option`, etc. above, which all take a `Option<&str>`.  Should all of them be changed to non-optional?  Or if not, what's different about `stream`?

alexcrichton: Ah no I didn't realize that's how everything worked. It feels weird to me to have a method that when it doesn't do anything it's silently a bug, but if it's how everything else is there's no need to change just this method.

PR #1960, opened 2024-12-19
add async support to `wit_component::dummy_module`

dicej:
This allows us to round-trip fuzz test using the async ABI(s) as well as the sync one.  I've also added corresponding `--async-callback` and `--async-stackful` options to the `component embed --dummy` subcommand for generating dummy modules which use the new ABIs.

Note that this currently only generates ultra-minimal, non-functional modules. A real module would import the `task.return` intrinsic with the appropriate signature for each async export, and would presumably use other new intrinsics such as `subtask.drop`, `task.backpressure`, etc. -- not to mention the various `stream.*`, `future.*`, and `error-context.*` intrinsics.  For more thorough fuzz testing, we'll want to generate imports for all known intrinsics (although we probably wouldn't do that for `component embed --dummy` modules, since it would be more confusing than helpful).

### DIFF
--- a/crates/wit-component/src/dummy.rs
+++ b/crates/wit-component/src/dummy.rs
@@ -1,18 +1,18 @@
-use wit_parser::abi::{AbiVariant, WasmType};
+use wit_parser::abi::WasmType;
 use wit_parser::{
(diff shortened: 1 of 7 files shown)


alexcrichton (approved):
Thanks!

PR #1961, opened 2024-12-20
Fix `std` feature in `wasm-encoder`

ark0f:


alexcrichton:
Thanks for the PR! Could you describe a bit more what's going on here? I'm not sure what the fix is and I'm also not familiar enough with the `leb128fmt` crate to know how best it solves such a problem off the top of my head.

ark0f:
> Could you describe a bit more what's going on here?

I found out that wasm-encoder does not switch `std` feature of its `wasmparser` dependency after #1938. I also disabled `std` feature in the workspace for `wasmparser` (and `wasm-encoder` for consistency) so it can manage this feature properly.

> I'm also not familiar enough with the `leb128fmt` crate

`leb128` doesn't support `no-std` when `leb128fmt` does

alexcrichton:
Ok sounds good, thanks! Can you update CI configuration [around here](https://github.com/bytecodealliance/wasm-tools/blob/6e9164c5db03892c9dc603ba1f783e84d0eacdd7/.github/workflows/main.yml#L239-L241) to check that `wasm-encoder` builds for no_std? Looks like we forgot that from before which is why this wasn't caught earlier.

I'll take some time tomorrow to review the `leb128fmt` crate as well as we want to review new dependencies before including them.

alexcrichton (approved):
Thanks again for your work here!

PR #1963, opened 2024-12-23
[wasmprinter] support custom indent text

### LINKED ISSUE #1962: `wasmprinter`: make indent character configurable
I find the indentation is not very noticeable when reading nested blocks.

I wish it could be increased to 8 spaces.

Do I just need to modify this place and add the configuration item:

https://github.com/bytecodealliance/wasm-tools/blob/57b1acee689e49f43338e4efb15b65e9f922831d/crates/wasmprinter/src/lib.rs#L1416

oovm:
Close #1962

### DIFF
--- a/crates/wasmprinter/tests/all.rs
+++ b/crates/wasmprinter/tests/all.rs
@@ -13,7 +13,6 @@ fn no_oom() {
     assert!(wat.len() < 500_000_000);
 }
 
-
 #[test]
 fn offsets_and_lines_smoke_test() {
     const MODULE: &str = r#"
--- a/crates/wasmprinter/tests/all.rs
+++ b/crates/wasmprinter/tests/all.rs
@@ -13,19 +13,19 @@ fn no_oom() {
     assert!(wat.len() < 500_000_000);
 }
 
-const MODULE_1: &str = r#"
-    (;@0     ;) (module
-    (;@b     ;)   (type (;0;) (func (param i32) (result i32)))
-    (;@1f    ;)   (func (;0;) (type 0) (param i32) (result i32)
-    (;@20    ;)     local.get 0
-                  )
-    (;@17    ;)   (export "f" (func 0))
-                )
-"#;
 
 #[test]
 fn offsets_and_lines_smoke_test() {
-    let bytes = wat::parse_str(MODULE_1).unwrap();
+    const MODULE: &str = r#"
+        (;@0     ;) (module
+        (;@b     ;)   (type (;0;) (func (param i32) (result i32)))
+        (;@1f    ;)   (func (;0;) (type 0) (param i32) (result i32)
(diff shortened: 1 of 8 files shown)


crates/wasmprinter/tests/all.rs
```diff
@@ -46,3 +47,29 @@ fn offsets_and_lines_smoke_test() {
 
     assert_eq!(actual, expected);
 }
+
+#[test]
+fn custom_indent_text() {
```
alexcrichton, line 52: Nowadays I try to avoid adding tests here since they're difficult to update, so mind moving this test to `tests/cli/*.wat` whch uses the `--indent-text` argument?

src/bin/wasm-tools/print.rs
```diff
@@ -30,6 +30,10 @@ pub struct Opts {
     /// Print instructions in the folded format.
     #[clap(short, long)]
     fold_instructions: bool,
+
+    /// The string to use when indenting.
+    #[clap(long)]
+    indent_text: Option<String>,
```
alexcrichton, line 36: Do you think it'd be worthwhile to also have `--indent 4` as a shortcut for `--indent-text '    '`?

oovm: It's a good idea, but maybe someone would like '\t'?

oovm:
It seems that '\t' is not supported, so I agree to use `--indent`

```yaml
---- "tests/cli\\print-custom-indent-text.wat" ----
failed test "tests/cli\\print-custom-indent-text.wat"

Caused by:
    0: failed RUN directive `` on line 1
    1: failed to check stdout expectation (auto-update with BLESS=1)
    2: failed test: result is not as expected:Diff < left / right > :
        (module
       <        (type (;0;) (func (param i32) (result i32)))
       <        (func (;0;) (type 0) (param i32) (result i32)
       <                local.get 0
       <        )
       <        (export "f" (func 0))
       >'/t'(type (;0;) (func (param i32) (result i32)))
       >'/t'(export "f" (func 0))
       >'/t'(func (;0;) (type 0) (param i32) (result i32)
       >'/t''/t'local.get 0
       >'/t')
        )

failures:
    "tests/cli\\print-custom-indent-text.wat"
```


test result: FAILED. 125 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 10.92s

oovm:
It seems that run_test has some problems, and does not support strings and escapes, whether it is `\t`, `'\t'` or `"   "`

alexcrichton (approved):
Thanks!

alexcrichton:
Oh sorry, yes it's ok to not test `\t` or `"    "` directly, the parser for these tests is pretty primitive. You could also test `--indent-text xxxx` and that'll assert "xxxx" starts each line

PR #1965, opened 2024-12-30
smith: Fix types recorded in `ref.func`

alexcrichton:
This commit fixes a minor issue in `wasm-smith` when generating the `ref.func` instruction when the GC proposal is disabled but shared-everything-threads is enabled. In this situation all `ref.func` instructions were recorded as `RefType::FUNCREF` but this wasn't correct for shared function types. The logic here is applicable both with and without the GC proposal is enabled so the precise type is always pushed now.

### DIFF
--- a/crates/wasm-smith/src/core.rs
+++ b/crates/wasm-smith/src/core.rs
@@ -900,7 +900,7 @@ impl Module {
     }
 
     fn arbitrary_super_type_of_val_type(
-        &mut self,
+        &self,
         u: &mut Unstructured,
         ty: ValType,
     ) -> Result<ValType> {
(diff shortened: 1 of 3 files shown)


fitzgen:
> when the GC proposal is disabled but shared-everything-threads is enabled

But shared-everything extends the GC proposal, so I don't think we should even allow this set up. It's like disabling bulk memory but enabling reference types.

crates/wasm-smith/src/core.rs
```diff
@@ -938,8 +938,23 @@ impl Module {
...
+        // If GC is disabled then favor the "top" type of the input type.
         if !self.config.gc_enabled {
-            return Ok(ty);
+            return Ok(match ty {
+                HeapType::Concrete(i) => {
+                    let CompositeType { inner, shared } = &self.ty(i).composite_type;
+                    let ty = match inner {
+                        CompositeInnerType::Array(_) => AbstractHeapType::Array,
+                        CompositeInnerType::Struct(_) => AbstractHeapType::Struct,
+                        CompositeInnerType::Func(_) => AbstractHeapType::Func,
```
fitzgen, line 949: The only case that this would be valid for is `(ref null? func)`. If GC is disabled, then struct and array references will still be invalid.

Which again points me towards this trying to do something that doesn't make sense...

alexcrichton:
Seems reasonable to [apply that fix too yeah](https://github.com/bytecodealliance/wasm-tools/pull/1973). Personally the two proposals are only loosely coupled in my head because we'll probably want pretty separate gates/paths for both, but it doesn't really matter in the interim.

> It's like disabling bulk memory but enabling reference types.

Technically we do allow this, but we don't allow disabling reference types and enabling gc. Or at least not in `wasm-smith`, I think we allow that in `wasmparser`.

> The only case that this would be valid for is `(ref null? func)`. If GC is disabled, then struct and array references will still be invalid.

True, but if GC is disabled, then the struct/array stuff won't show up. I gambled that an insertion of `=> unreachable!()` would have fewer reviewer comments and lost heh

Issue #1967, opened 2025-01-07
[wasmparser] validate traps with "indexmap: index out of bounds" when validating composed components with GC subtypes


tanishiking:
### tools
- `wasm-tools` version: 1.222.0 (892d4b6ef)
- `wac` version: Custom build from [6c22d80fc](https://github.com/bytecodealliance/wac/commit/6c22d80fc5fe6331b28c6d1482e8f9b60a8785f3) to support GC types
- `cargo component` version: 1.17.0

### description
It seems like `wasm-tools validate` fails with a panic when validating a component that includes GC types with subtypes, after composing it with another component using `wac`. (Note that, there's no problem with validating a WasmGC module with subtypes).

```
thread 'main' panicked at crates/wasmparser/src/validator/types.rs:998:70:
indexmap: index out of bounds
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::option::expect_failed
   3: wasmparser::validator::core::canonical::internrecgroup::check_subtype
   4: wasmparser::validator::core::module::add_types
   5: wasmparser::validator::validator::type_section
   6: wasmparser::validator::validator::payload
   7: wasm_tools::validate::opts::run
   8: wasm_tools::wasmtools::run
   9: wasm_tools::main
note: some details are omitted, run with `rust_backtrace=full` for a verbose backtrace.
```

### reproduction
Full example: https://github.com/tanishiking/kitchensink/tree/main/wasmgc-component-validation-failed

(1) Create a module (`socket.wat`) that uses GC types with subtypes:

```wat
(module
  (type $test (func))
  (rec
    (type $Base (sub (struct (field $foo i32))))
    (type $Derived (sub $Base (struct (field $foo i32))))
  )
  (import "tanishiking:wasmgc-component-test/tester@0.0.1" "test" (func $test (type $test)))
  (export "memory" (memory $mem))
  (memory $mem 1)
)
```

```wit
/// wit/world.wit
package tanishiking:wasmgc-component-test@0.0.1;

world plug {
    export tester;
}

world socket {
    import tester;
}

interface tester {
    test: func();
}
```

(2) Componetize it:

```sh
wasm-tools parse socket.wat -o socket.wasm
wasm-tools component embed wit -o socket.wasm -w socket socket.wasm
wasm-tools component new -o socket.wasm socket.wasm
```

(3) Build a plugin component (using cargo component)

```rust
// src/lib.rs
#[allow(warnings)]
[...]

PR #1968, opened 2025-01-07
wasmparser: Fix subtyping depth indexing

tanishiking:
fix: https://github.com/bytecodealliance/wasm-tools/issues/1967

This patch addresses an issue in subtyping depth checking where `set_subtyping_depth` indexes the depth by `CoreTypeId`, but `get_subtyping_depth` accesses it using `CoreTypeId#index` (`u32`). This leads to "IndexMap: index out of bounds" error when validating WasmGC components with subtypes.

### DIFF
--- a/crates/wasmparser/src/validator/types.rs
+++ b/crates/wasmparser/src/validator/types.rs
@@ -995,7 +995,7 @@ impl TypeList {
         let depth = self
             .core_type_to_depth
             .as_ref()
-            .expect("cannot get subtype depth from a committed list")[id.index()];
+            .expect("cannot get subtype depth from a committed list")[&id];
         debug_assert!(usize::from(depth) <= crate::limits::MAX_WASM_SUBTYPING_DEPTH);
         depth
     }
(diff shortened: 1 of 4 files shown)


alexcrichton:
Thanks for the report, the investigation, and a PR! I believe this is the right fix as well. Playing around with this locally this is a small test case to trigger the bug from before and this PR fixes it. If you'd like to include that here I'd be happy to flag this for merge:

```diff
diff --git a/tests/local/component-model/gc.wast b/tests/local/component-model/gc.wast
index 1b409903f..d7a47456e 100644
--- a/tests/local/component-model/gc.wast
+++ b/tests/local/component-model/gc.wast
@@ -124,3 +124,12 @@
     ))
   )
   "sub type cannot have a final super type")
+
+(component
+  (type $t (resource (rep i32)))
+  (core func (canon resource.drop $t))
+  (core module
+    (type $t1 (sub (func)))
+    (type (sub $t1 (func)))
+  )
+)
```

tanishiking:
Thank you so much for taking a look & minimizing the reproduction @alexcrichton !
I added the test case you mentioned :)

alexcrichton (approved):

PR #1973, opened 2025-01-13
smith: Disable shared-everything-threads without GC

alexcrichton:
Alternative fix from #1965

### DIFF
--- a/crates/wasm-smith/src/config.rs
+++ b/crates/wasm-smith/src/config.rs
@@ -796,6 +796,13 @@ impl Config {
             self.shared_everything_threads_enabled = false;
         }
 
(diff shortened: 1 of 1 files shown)


crates/wasm-smith/src/config.rs
```diff
@@ -796,6 +796,13 @@ impl Config {
             self.shared_everything_threads_enabled = false;
         }
 
+        // shared-everything-threads depends on GC, so if gc is disabled then
+        // also disable shared-everything-threads.
+        // capped at 1 and disable gc as well.
```
fitzgen, line 801: bad copy paste or something?

fitzgen (approved):

PR #1976, opened 2025-01-14
disallow post-return functions for async lifts

dicej:
I can't find where in the spec this is disallowed, but Luke agreed that it _should_ be disallowed, so we'll update the spec to make that explicit.  Indeed, we hope to deprecate post-return functions for both sync and async lifts in favor of `task.return`.


alexcrichton (approved):

lukewagner:
You're right, at the moment, `post-return` is simply ignored when `async` is set, but it should be rejected.  I'll make a PR to the spec.

PR #1986, opened 2025-01-27
Fix `cargo doc` warnings

samestep:
This PR fixes a few warnings that currently appear when running `cargo doc`.


alexcrichton (approved):
Thanks! 

If you're up for it I think it'd be reasonable to add this as a CI check, something along the lines of `-Dwarnings` or something like that

PR #1987, opened 2025-01-27
Catch `cargo doc` warnings in CI

samestep:
Following up on https://github.com/bytecodealliance/wasm-tools/pull/1986#pullrequestreview-2576325105. That PR fixed all warnings from `cargo doc` but not from `cargo doc --all`, so this PR fixes those too.


pchickey:
Thank you for working on this - we let this go for far too long and I appreciate your efforts correcting it.

alexcrichton (approved):

PR #1988, opened 2025-01-28
Sort dependencies in `Cargo.toml` files

samestep:
Currently most of the `[dependencies]` lists in `Cargo.toml` files are close to sorted but not quite, making it a bit confusing to figure out where to insert a new dependency. This PR sorts them all.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,37 +70,37 @@ version = "0.224.0"
 rust-version = "1.76.0"
 
 [workspace.dependencies]
+ahash = { version = "0.8.11", default-features = false }
 anyhow = "1.0.58"
 arbitrary = "1.1.0"
+bitflags = "2.5.0"
 clap = { version = "4.0.0", features = ["derive"] }
 clap_complete = "4.4.7"
 comfy-table = { version = "7.1.3", default-features = false }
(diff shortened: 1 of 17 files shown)


pchickey (approved):
I wonder if we should leave a comment at the top of these lists indicating they are sorted and to please respect the sort order on changes? On the other hand that seems noisy. I wish this was something `cargo fmt` did automatically.

samestep:
@pchickey I'm not sure; I was going to say it might not be too hard to add a Python script using [`tomllib`](https://docs.python.org/3/library/tomllib.html) to check in CI that all the dependencies are sorted, but then I remembered that the following cases wouldn't be completely trivial:

- [different groups of dependencies separated by comments](https://github.com/bytecodealliance/wasm-tools/blob/866672f4fd04fbd7fc0387c64547a09f02a17852/Cargo.toml#L120-L177)
- [individual dependencies specified in separate TOML tables](https://github.com/bytecodealliance/wasm-tools/blob/866672f4fd04fbd7fc0387c64547a09f02a17852/fuzz/Cargo.toml#L32-L38)

Still possible, though.

pchickey:
Yeah, I don't think we should invent nonstandard tooling here. Best effort is good enough.

PR #1989, opened 2025-01-28
update `task.return` to match spec

dicej:
As of https://github.com/WebAssembly/component-model/pull/431, `task.return` takes component-level result types rather than a core function type representing the flattening of those types.  This updates `wasm-tools` to match.


alexcrichton:
This reminds me that the support for multiple component model return values (plus names) is all deprecated/gated at this point and I need to get around to removing it. Mind codifying that here with a `task_return(Option<ComponentValType>)` constructor only? (for both parsing/encoding/etc)

dicej:
> This reminds me that the support for multiple component model return values (plus names) is all deprecated/gated at this point and I need to get around to removing it. Mind codifying that here with a `task_return(Option<ComponentValType>)` constructor only? (for both parsing/encoding/etc)

Done.

crates/wasmparser/src/validator/component.rs
```diff
@@ -1114,20 +1114,30 @@ impl ComponentState {
...
+                    ))
+                })
+                .collect::<Result<_>>()?,
+            results: Box::new([]),
+        }
+        .lower(types, Abi::LiftSync);
 
-        self.core_funcs.push(id);
+        self.core_funcs
+            .push(types.intern_func_type(FuncType::new(info.params.iter(), []), offset));
```
alexcrichton, line 1140: Mind asserting here that `info.results` is empty?

alexcrichton (approved):

PR #1992, opened 2025-02-02
Fixed length list support

cpetig:
Implement fixed length list support as defined in https://github.com/WebAssembly/component-model/pull/384

 - [X] wit-parser
 - [X] wasm-encoder
 - [X] wasmparser
 - [X] wasmprinter
 - [X] wat
 - [X] wit-encoder
 - [X] wit-smith


cpetig:
Fixed size lists became a separate type because they have a different ABI representation and map to different Rust/C types.

alexcrichton (commented):
Thanks for this! In addition to the comments below, mind adding some more tests? For example:

* WIth a feature gate there should be a test that using this type is invalid without the feature gate (e.g. `tests/local/missing-features`)
* Tests for the `*.wat` syntax in `tests/local/component-model/*`. This also tests round-trip parsing and such
* Tests for `list<ty, 10000000000000000>` (something out of the 64-bit range but still looks integer-like)

Also with the new support in `wit-smith` (thanks!) mind running the fuzzer for a few minutes locally to ensure it doesn't turn up anything? (`FUZZER=roundtrip_wit cargo +nightly fuzz run -s none run -j30`)

alexcrichton:
Also, as a possible bikeshed, which you can take or leave, in Rust at least there's the term "slice" which is `&[T]` which is sort of like a WIT "list", and in Rust the term for `[T; N]` is "array" which could possibly be used here instead of "fixed size list". It's easy to confuse "list" and "array" though and harder to confuse "fixed sized list" and "list". The downside of "fixed size list" is that it's a bit of a mouthful to say and type

crates/wasmparser/src/validator/component_types.rs
```diff
@@ -1134,6 +1136,9 @@ impl ComponentDefinedType {
                 lowered_types,
             ),
             Self::List(_) => lowered_types.push(ValType::I32) && lowered_types.push(ValType::I32),
+            Self::FixedSizeList(ty, length) => {
+                (0..*length).all(|_n| ty.push_wasm_types(types, lowered_types))
```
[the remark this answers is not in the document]
cpetig: I hesitated about limits because I also have support for efficient storage of AI tensors or image data in mind, but of course with these you want the outermost list a normal list to avoid expanding all elements on the stack.

On the other hand if you nest fixed width lists of a maximum size of 1000 you can still deflate bomb `types` here, perhaps a global fuel like argument/result limit makes sense.

crates/wasmparser/src/validator/component.rs
```diff
@@ -3277,6 +3277,12 @@ impl ComponentState {
             crate::ComponentDefinedType::List(ty) => Ok(ComponentDefinedType::List(
                 self.create_component_val_type(ty, offset)?,
             )),
+            crate::ComponentDefinedType::FixedSizeList(ty, elements) => {
+                Ok(ComponentDefinedType::FixedSizeList(
+                    self.create_component_val_type(ty, offset)?,
+                    elements,
+                ))
+            }
```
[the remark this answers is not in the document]
cpetig: > From a high level we're going to want to gate this feature by default and have it off-by-default. It'll take some time to percolate this feature throughout tools/etc and until that's all working I don't think we'll want to have everything accessible by default.
> 
> Can you add something to `features.rs` and then add a check of the feature gate here?

Done, but this will still need all code using wit-parser to handle the new type in match expressions, right?

crates/wit-component/src/encoding.rs
```diff
@@ -320,6 +320,7 @@ impl TypeContents {
                 }
                 TypeDefKind::Enum(_) => Self::empty(),
                 TypeDefKind::List(t) => Self::for_type(resolve, t) | Self::LIST,
+                TypeDefKind::FixedSizeList(t, _elements) => Self::for_type(resolve, t) | Self::LIST,
```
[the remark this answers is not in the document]
cpetig: Now, that I checked what LIST is about, it is obviously wrong (LIST indicates requiring memory and realloc). :facepalm:

crates/wit-encoder/src/ty.rs
```diff
@@ -25,7 +25,7 @@ pub enum Type {
     Borrow(Ident),
     Option(Box<Type>),
     Result(Box<Result_>),
-    List(Box<Type>),
+    List(Box<Type>, Option<usize>),
```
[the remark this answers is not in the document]
cpetig: I though about this as well, but I feel that on the level of representing WIT Syntax representing both as List makes sense, while on the level of ABI and API they are very different.

I considered this a case of WIT Syntax representation.

crates/wit-parser/src/ast.rs
```diff
@@ -1347,13 +1348,26 @@ impl<'a> Type<'a> {
...
 
             // list<T>
+            // list<T, N>
             Some((span, Token::List)) => {
                 tokens.expect(Token::LessThan)?;
                 let ty = Type::parse(tokens)?;
+                let size = if tokens.eat(Token::Comma)? {
+                    let number = tokens.next()?;
+                    if let Some((span, Token::Integer)) = number {
+                        let size: usize = tokens.get_span(span).parse()?;
```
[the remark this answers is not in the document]
cpetig: This was mostly out of habit of representing array sizes in Rust as usize, but when it comes to this feature it makes sense to use the architecture independent u32 instead (everywhere) - as the wasm-parser code will never directly connect it to an array in its code.

cpetig:
@alexcrichton do you prefer me to resolve the ones I implemented or do you want to check first?

I will implement wat roundtrip tests later today (I didn't find them although I did one manually) and think about limiting on-stack elements.

cpetig:
> Also with the new support in `wit-smith` (thanks!) mind running the fuzzer for a few minutes locally to ensure it doesn't turn up anything? (`FUZZER=roundtrip_wit cargo +nightly fuzz run -s none run -j30`)

It found something ;-) 
`Fixed size lists require the component model fixed size list feature (at offset 0x26)`
so I need to figure out where to activate this feature inside the fuzzer.

cpetig:
> I need to figure out where to activate this feature inside the fuzzer.

The default Validator used for the component encoder has this feature disabled (as it is correctly disabled by default), I don't see an easy way to only activate it for the WIT fuzzing, is there a secret environment variable?

crates/wasmparser/src/validator/component.rs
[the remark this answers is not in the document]
alexcrichton: Right yeah, tooling and such like wit-parser` and `wit-component` don't deal much with features and instead generally just support everything. It's primarily the validator where we want to gate features as that's the bedrock of Wasmtime for example and we'll want to take time to implement this in Wasmtime and evaluate it before it's all insta-stable.

cpetig:
> Thanks for this!

Thank you for your in-depth review!

>     * Tests for `list<ty, 10000000000000000>` (something out of the 64-bit range but still looks integer-like)

I tried but wasn't able to figure out the right way to set up this test: https://github.com/bytecodealliance/wasm-tools/pull/1992/files#diff-1532b9f29c0a59c8d44ce87b39476ec5242d1b2a2329e0b72bf7b2f23c2ae485R24-R38

alexcrichton:
> do you prefer me to resolve the ones I implemented or do you want to check first?

Feel free to do so yourself, I'll give the code a once-over anyway once it's ready too.

> The default Validator used for the component encoder has this feature disabled (as it is correctly disabled by default), I don't see an easy way to only activate it for the WIT fuzzing, is there a secret environment variable?

For this I'd recommend in the various locations validation is performed for tests/etc to just enable the new feature unconditionally. Adding for example `Validator::new_with_features(WasmFeatures::all())` I think is reasonable since the validation in the tooling is mostly test assertions and we want to exercise it all during fuzzing anyway.

> I tried but wasn't able to figure out the right way to set up this test

Ah yeah it's a lexical error right in `*.wat` if the integer literal is out-of-bounds so I think it's reasonable to skip the `*.wat` test, but I think `*.wit` may want to be tested? It might also be reasonable to just leave the `*.wat` test around to ensure that some error message comes out, and it'd use `assert_malformed` I think or whatever directive supports `module quote ...`

crates/wasmparser/src/validator/component_types.rs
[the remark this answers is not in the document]
alexcrichton: That makes sense to me yeah. My basic worry is loops like this (there's a few throughout this PR) which run a risk of taking time proportional to `N` which is not a property validation/processing otherwise has at this time. I think a larger limit like 128k or something may be ok (probably coupled with fuel) but I suspect we probably shouldn't allow `u32::MAX` regardless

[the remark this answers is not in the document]
cpetig: I went for an 8k limit for now

tests/local/component-model/component-model-fixed-size-list/fixed-size-list.wast
```diff
@@ -0,0 +1,38 @@
+(component
+  (core module $m
+    (memory (export "memory") 1)
+    (func (export "ret-list") (result i32) unreachable)
+  )
+  (core instance $i (instantiate $m))
+
+  (func (export "ret-list") (result (list u32 4))
```
alexcrichton, line 8: I'm actually a little confused, how does this test pass? The fixed-size-list is appropriately gated in wasmparser (or so I thought) but I would expect a change to `tests/roundtrip.rs` to enable the fixed-size-list feature here

crates/wasmparser/src/validator/component_types.rs
```diff
@@ -1171,6 +1173,9 @@ impl ComponentDefinedType {
                 lowered_types,
             ),
             Self::List(_) => lowered_types.push(ValType::I32) && lowered_types.push(ValType::I32),
+            Self::FixedSizeList(ty, length) => {
+                (0..(*length).min(8192)).all(|_n| ty.push_wasm_types(types, lowered_types))
```
alexcrichton, line 1177: I'm a bit wary  of this in that it's sort of just a constant sitting here. For example if this stays as-is it should be documented as to why `min` is used and why 8k is chosen. Otherwise though what I was envisioning was an update to the validator to gate the structure of types to require that they fit under the 8k limit.

tests/local/component-model/component-model-fixed-size-list/fixed-size-list.wast
alexcrichton, line 1: While you're here it might also be good to add some tests for failure cases such as [this](https://github.com/bytecodealliance/wasm-tools/blob/ae180192f2cd781bee0ebb8b12d5b4d6af2994c9/tests/local/component-model/instantiate.wast#L913-L983) where the type-check is tested to ensure that the lengths of two fixed-size-lists are the same

tests/local/component-model/component-model-fixed-size-list/fixed-size-list.wast
alexcrichton: I'm still sort of surprised here though because this test is indeed passing on CI, and it looks like it is indeed testing fixed-size-lists, so I'm not sure why it's passing...

alexcrichton: Ok nevermind I see that `WasmFeatures::all()` is used in tests, so disregard this.

alexcrichton:
I've added a merge commit here which resolves the conflicts with https://github.com/bytecodealliance/wasm-tools/pull/2096, but feel free to discard that if you'd like.

cpetig:
Oh, dear, I got a first hand impression on why I should get this merged soon … adding fixed size lists to wit-bindgen is moderate effort, but to run any runtime test you need to add support for the extension to wasm-component-ld (shipped with the Rust compiler), wac, and wasmtime.

And this is the improved situation after Alex simplified the runtime test to no longer require full support on the host (wasmtime) side.

crates/wasmparser/src/validator/component_types.rs
alexcrichton: Ping on this constant, this is still sitting here. Can you add documentation here as to why this was chosen? (ideally moved to `limits.rs` as well)

crates/wit-parser/src/abi.rs
```diff
@@ -288,6 +288,12 @@ impl Resolve {
                     result.push(WasmType::Length);
                 }
 
+                TypeDefKind::FixedSizeList(ty, size) => {
+                    for _ in 0..*size {
+                        self.push_flat(ty, result);
+                    }
+                }
```
alexcrichton, line 295: I'm re-reviewing this PR in light of https://github.com/bytecodealliance/wasmtime/pull/10619#discussion_r2052556773 as I forgot about recursive lists, and the consequence of debug mode here. Although even just with large numbers here this can blow up, for example:

```wit
package a:b;

world foo {
  import x: func(x: list<list<u32, 2147483648>, 100>);
}
```

as input to:

```
$ cargo run component embed --dummy foo.wit | cargo run component new -t
```
 
is effectively infinite looping locally due to this loop.

Basically I think that another way to put this concern is that processing, at a compiler/tooling level, a type in WIT should take O(...) space relative to how much space was taken to define the type. That's why records for example iterate over fields (each field is explicitly defined) but why I don't think for lists that loops like this can be used because it makes processing O(n) when the definition takes O(1) space.

This is especially a bit trickier since I think tooling should generally work reasonably even without the validator since the validator is generally optional until you actually hand a module to the runtime. Otherwise passing around components between tooling ideally wouldn't result in large processing time of modules.

crates/wasmparser/src/validator/component_types.rs
alexcrichton: I also made a comment down below, but I rethinking this again and I don't think we'll be able to get by reusing existing helpers for this, I think that there needs to be a dedicated helper/path for pushing a fixed-size-list into a flat list of types which doesn't take O(n) time, even in debug mode, to help prevent accidental exponential blowups.

crates/wit-parser/src/abi.rs
cpetig: Thank you for highlighting the critical shortcoming in this current implementation, I perfectly understand the severity of the issue (also why it doesn't apply to tuples and why lists are different as they never get flattened to the stack).

I will give this a try in the coming days. 🤔 Likely the part which flattens a `list<list<u32, 1000>,1000>` to lowered types should exit early.

cpetig: > Thinking about this, I'm not sure an early-return will be enough. In the case that a type has zero size the loop would still iterate N times (possibly exponentially for nested lists). Zero-sized types are pretty rare now that tuples/structs require at least one field, but `list<T, 0>` I think is a new flavor of ZST?

Update: This ZST loophole was closed by https://github.com/WebAssembly/component-model/pull/504, so I will need to change the validator to reject zero size lists.

Do you think that with ZST off the table early return is still not enough?

alexcrichton: That's true yeah, I think it's ok to have early-return now.

crates/wasmparser/src/validator/component_types.rs
cpetig: I took a closer look and found that it should exit early due to restrictions of the lowered_types container, but I added an additional if to make it more visible. `if *length as usize > lowered_types.max { return false; }`

crates/wit-parser/src/abi.rs
cpetig: I fell back to using `MAX_FLAT_PARAMS` which to me looks reasonable to constrain the explosion (nested lists aren't limited, but same applies to nested tuples which could exit more early as well)

cpetig:
> > Thanks for this!
> 
> Thank you for your in-depth review!
> 
> > ```
> > * Tests for `list<ty, 10000000000000000>` (something out of the 64-bit range but still looks integer-like)
> > ```
> 
> I tried but wasn't able to figure out the right way to set up this test: https://github.com/bytecodealliance/wasm-tools/pull/1992/files#diff-1532b9f29c0a59c8d44ce87b39476ec5242d1b2a2329e0b72bf7b2f23c2ae485R24-R38

fixed now

tests/local/component-model/component-model-fixed-size-list/fixed-size-list.wast
cpetig: Added, it's now part of `tests/cli/component-model/fixed-size-list.wast`

crates/wit-encoder/src/ty.rs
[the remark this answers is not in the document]
cpetig: I got rid of all `Option<u32>` in list data structures, it feels more clean even though the wit/wat parsing is now more complex.

crates/wasmparser/src/validator/component_types.rs
```diff
@@ -1174,6 +1176,12 @@ impl ComponentDefinedType {
                 lowered_types,
             ),
             Self::List(_) => lowered_types.push(ValType::I32) && lowered_types.push(ValType::I32),
+            Self::FixedSizeList(ty, length) => {
+                if *length as usize > lowered_types.max {
+                    return false;
+                }
```
alexcrichton, line 1182: Upon further reflection of this I'm starting to have second thoughts about this approach. This input for example:

```wit
package a:b;

world x {
  type t = list<t2, 16>;
  type t2 = list<t3, 16>;
  type t3 = list<t4, 16>;
  type t4 = list<t5, 16>;
  type t5 = list<t6, 16>;
  type t6 = list<t7, 16>;
  type t7 = list<t8, 16>;
  type t8 = list<u32, 16>;
  import x: func(t: t);
}
```

when locally run as:

```
$ cargo run component embed --dummy foo.wit -t
```

takes ~30 seconds to produce the result. I think the reason is that this case isn't hit and then the recursion gets to go hog wild for a bit.

Perhaps an alternative fix would be to iterate from 0 to `length` but then check each iteration of the loop if the max has been reached and bail out if so? That way each level would check the break-out and once that happens it'd quickly exit.

Also if you're up for it I think that these would be good test cases to add to the test suite to ensure processing large or nested lists doesn't accidentally take forever

cpetig: Good catch, ... but the same problem shows with 
```wit
package a:b;

world x {
  type t = tuple<t2,t2,t2,t2,t2,t2,t2,t2,t2,t2,t2,t2,t2,t2,t2,t2,>;
  type t2 = tuple<t3,t3,t3,t3,t3,t3,t3,t3,t3,t3,t3,t3,t3,t3,t3,t3,>;
  type t3 = tuple<t4,t4,t4,t4,t4,t4,t4,t4,t4,t4,t4,t4,t4,t4,t4,t4,>;
  type t4 = tuple<t5,t5,t5,t5,t5,t5,t5,t5,t5,t5,t5,t5,t5,t5,t5,t5,>;
  type t5 = tuple<t6,t6,t6,t6,t6,t6,t6,t6,t6,t6,t6,t6,t6,t6,t6,t6,>;
  type t6 = tuple<t7,t7,t7,t7,t7,t7,t7,t7,t7,t7,t7,t7,t7,t7,t7,t7,>;
  type t7 = tuple<t8,t8,t8,t8,t8,t8,t8,t8,t8,t8,t8,t8,t8,t8,t8,t8,>;
  type t8 = tuple<u32,u32,u32,u32,u32,u32,u32,u32,u32,u32,u32,u32,u32,u32,u32,u32,>;
  import x: func(t: t);
}
```

I will take a look into whether I can shrink the time needed to process this.

alexcrichton: Oh that's a really good point! Ok in that case I'm going to go ahead and flag this for merge and follow-ups can handle fixes for cases like this.

PR #2021, opened 2025-02-05
Updates to wasm-tool's release process

alexcrichton:
* Enable releases to happen on `release-*` branches, not just `main`
* PRs to bump versions from `release-*` branches now have the correct base branch.
* Logic for publication was moved from `.github/workflows/publish.yml` to `.github/actions/publish-release` to be able to share logic with `main.yml`. This is because releases from `main` always have the merge queue first, hence the `publish.yml` workflow, but releases from a `release-*` branch don't have a merge queue meaning it has to run at the end of normal CI.
* Some various conditions were tweaked for various conditions to assist with external testing of the release workflow, e.g. dropinng requirements for repos to be in the `bytecodealliance` organization.
* The `main.yml` CI now runs for all pushes to `release-*` branches, enabling artifacts being built for publication.

### DIFF
--- a/.github/actions/publish-release/action.yml
+++ b/.github/actions/publish-release/action.yml
@@ -1,0 +1,84 @@
+name: 'Maybe perform a release'
+description: 'Steps to perform a conditional release of this repository'
+
+inputs:
+  cargo_token:
+    description: 'token used to publish crates'
+    required: false
+
+runs:
+  using: composite
+  steps:
+    # If this commit log has an indicator saying that a tag should be made. If
+    # so create one and push it.
+    - name: Test if tag is needed
+      run: |
+        git log ${{ github.event.before }}...${{ github.event.after }} | tee main.log
+        version=$(./ci/print-current-version.sh)
+        echo "version: $version"
+        echo "version=$version" >> $GITHUB_OUTPUT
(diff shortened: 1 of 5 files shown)


alexcrichton:
This is being done to enable the patch releases on various release branches in PRs right now. In case anyone's curious:

* Rust nightly changed some code which exposed a preexisting bug in wasm-tools. That caused tests to fail and was fixed in https://github.com/bytecodealliance/wasm-tools/pull/1999. The exact bug was that `derive(PartialOrd)` disagreed with `impl Ord` which manifested in new behavior on nightly.
* Wasmtime no longer builds on nightly because of this bug. This is because outputs of `wit-parser` are used in compile-time-assertions in Wasmtime to ensure that size/align agree with Wasmtime's own computation.
* This means that historical releases of Wasmtime using historical releases of `wit-parser` will, in the future, cease building with rustc. I figured this is bad enough to warrant patch releases.
* To handle this I manually created `release-*` branches and have sent a bunch of PRs.
* This PR is necessary to actually get the releases to work. 

My plan is to land this, backport it to all the release branches, and then actually push through the releases. At the end of all this I plan on yanking all the broken versions as well.

alexcrichton:
I'll also probably do a proof-of-concept point release from the main branch here to after this merges to make sure that it still works as well.

dicej (approved):

PR #2042, opened 2025-02-07
threads: add `thread.spawn_indirect`

abrown:
This propagates the upstream spec changes to the component model in [#447] here. See commit messages for more details.

[#447]: https://github.com/WebAssembly/component-model/pull/447


crates/wasmparser/src/readers/component/canonicals.rs
```diff
@@ -269,9 +282,12 @@ impl<'a> FromReader<'a> for CanonicalFunction {
             0x04 => CanonicalFunction::ResourceRep {
                 resource: reader.read()?,
             },
-            0x05 => CanonicalFunction::ThreadSpawn {
+            0x05 => CanonicalFunction::ThreadSpawnRef {
                 func_ty_index: reader.read()?,
             },
+            0x24 => CanonicalFunction::ThreadSpawnIndirect {
+                table_index: reader.read()?,
+            },
```
alexcrichton, line 290: This makes sense logically to put here but for easily scanning this list mind moving it down to the bottom so it's more clear where the "holes" are in decoding?

crates/wast/src/component/resolve.rs
```diff
@@ -383,9 +383,16 @@ impl<'a> Resolver<'a> {
...
-            CanonicalFuncKind::ThreadSpawn(info) => {
+            CanonicalFuncKind::ThreadSpawnRef(info) => {
                 self.resolve_ns(&mut info.ty, Ns::CoreType)?;
             }
+            CanonicalFuncKind::ThreadSpawnIndirect(info) => {
+                self.core_item_ref(&mut info.table)?;
+                // Eventually this should resolve the specific type associated
+                // with this canonical function, e.g.,
+                // `self.resolve_type_use(&mut info.ty)?;` (TODO: spawn indirect
+                // types).
```
alexcrichton, line 394: Adding a comment to flag this. This looks though like it's a TODO at the spec-level, right? (as opposed to just exclusively the implementation here)

alexcrichton (commented):
Looks good, thanks! Mind adding a test as well for using a table of the wrong type? (e.g. a non-shared funcref or a shared-externref)

crates/wasmparser/src/readers/component/canonicals.rs
abrown: This is now resolved by moving to the `0x40+` space.

abrown:
@alexcrichton, I rebased this on top of `main` so a lot has changed. This also incorporates the conclusions we reached in https://github.com/WebAssembly/component-model/pull/447 so there's new encoding-related implementation related to that.

crates/wasmparser/src/validator/component.rs
```diff
@@ -1934,11 +1938,76 @@ impl ComponentState {
...
+        let _ = self.validate_spawn_type(func_ty_index, types, offset)?;
+
+        // Check this much like `call_indirect` (see
+        // `OperatorValidatorTemp::check_call_indirect_ty`), but loosen the
+        // table type restrictions to just a `funcref`. See the component model
+        // for more details:
+        // https://github.com/WebAssembly/component-model/blob/6e08e283/design/mvp/CanonicalABI.md#-canon-threadspawn_indirect.
+        let table = self.table_at(table_index, offset)?;
+        // TODO: check that the table is shared once components allow shared
+        // tables.
```
alexcrichton, line 1982: Is this TODO still applicable?

crates/wast/src/component/func.rs
```diff
@@ -471,23 +474,46 @@ impl<'a> Parse<'a> for CanonResourceRep<'a> {
...
+}
+
+impl<'a> Parse<'a> for CanonThreadSpawnIndirect<'a> {
+    fn parse(parser: Parser<'a>) -> Result<Self> {
+        parser.parse::<kw::thread_spawn_indirect>()?;
+        let ty = parser.parse()?;
+        let table = parser.parens(|parser| {
+            let span = parser.parse::<kw::table>()?.0;
+            parse_trailing_item_ref(kw::table(span), parser)
+        })?;
```
alexcrichton, line 512: Can this be replaced with `let table = parser.parens(|p| p.parse())?;`?

Or let me rephrase after investigating this, this should be replaced with that and I'll update the other places you copied this from since those should all be updated as well.

alexcrichton (approved):

crates/wasmparser/src/validator/component.rs
alexcrichton: Oh that was due to https://github.com/bytecodealliance/wasm-tools/pull/1970. What I added there was a very coarse version of "you can't lower into a 64-bit memory".

What really needs to happen is that every location that uses a memory (or table) needs to perform a subtype check on what the actual table/memory received is. For example all canonical options referring to a `memory` have to refer to a non-shared 32-bit linear memory. 

Basically #1970 was a coarse over-approximation that didn't break anyone "on stable". To remove the logic from #1970 we'll need to audit usage of `memory_at` and `table_at` to ensure everything is doing a subtype check as appropriate.

PR #419, opened 2021-12-09
A bunch more random `wasm-mutate` clean ups

fitzgen:

PR #2043, opened 2025-02-07
threads: rename `thread.hw_concurrency` to `thread.available_parallelism`

abrown:
This propagates @sunfishcode's renaming of the built-in upstream (see [#419]) to these crates.

[#419]: https://github.com/WebAssembly/component-model/pull/419


sunfishcode (approved):
Looks good!

PR #2045, opened 2025-02-09
Solve the future/errror binary spec discrepancy, fixes #2044

### LINKED ISSUE #2044: Discrepancy in binary representation of future and error-context
Probably because https://github.com/WebAssembly/component-model/pull/384 was merged a few weeks before https://github.com/WebAssembly/component-model/pull/405 , the component model [binary specification](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Binary.md#type-definitions) defines futures as 0x65 and error-contexts as 0x64.

Wasm-tools on the other hand uses [future 0x67](https://github.com/bytecodealliance/wasm-tools/blob/main/crates/wasmparser/src/readers/component/types.rs#L553) and error-context 0x65. I guess the implementation predates the spec.

Both agree on streams as 0x66.

I can create a PR to make wasm-tools follow the spec, but I would like to clarify that this is indeed the right path.

PS: I found this problem while implementing fixed size lists in https://github.com/bytecodealliance/wasm-tools/pull/1992 ...

cpetig:


cpetig:
Fixes https://github.com/bytecodealliance/wasm-tools/issues/2044 (doesn't seem to link this MR from just the title)

alexcrichton (approved):
Thanks for this!

alexcrichton:
cc @dicej

PR #2048, opened 2025-02-11
wit-parser/sizealign: Introduce format() variant with optional brackets around the sum for codegen

cpetig:
wit-bindgen generated Rust code warns because of unnecessary brackets generated by the format() function


cpetig:
Part of https://github.com/bytecodealliance/wit-bindgen/pull/1098

alexcrichton (approved):

PR #2049, opened 2025-02-12
Use typed indices via a new `wasm-types` crate

samestep:
Following up on some conversation in #1985, this draft PR is a start on what it might look like to use typed indices throughout the whole codebase. I was originally planning to _finish_ this before coming back to #1985, but it eventually became clear to me that this is considerably bigger than I expected.

In any case, I'm opening this PR to gauge whether there's any interest in going further down this path, or if I should just call this an interesting experiment and abandon it. There were some unintuitive cases, so maybe there is some potential value in these from a codebase documentation/consistency perspective? For example:

https://github.com/bytecodealliance/wasm-tools/blob/beac7de36a6747a9061b8fa497a130aa4f7da185/crates/wasm-encoder/src/reencode/component.rs#L963-L966

This code made me mark the `func_ty_index` field of the `CanonicalFunction::ThreadSpawn` variant as `TypeIdx` instead of `ComponentTypeIdx`, unlike all the fields in all the `Resource` variants. Is this correct? I don't know much about the component model, so maybe threads just happen to use core type indices while other similar-looking things use component type indices. Maybe it's clear from the spec (?), but it's not clear from the docstring:

https://github.com/bytecodealliance/wasm-tools/blob/beac7de36a6747a9061b8fa497a130aa4f7da185/crates/wasmparser/src/readers/component/canonicals.rs#L70-L74

Another random question I had while working on this: why do both [`wasm_encoder::NameSection::tag`](https://docs.rs/wasm-encoder/0.225.0/wasm_encoder/struct.NameSection.html#method.tag) and [`wasm_encoder::NameSection::tags`](https://docs.rs/wasm-encoder/0.225.0/wasm_encoder/struct.NameSection.html#method.tags) exist? Don't they do the same thing?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -970,6 +970,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "index_vec"
+version = "0.1.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "44faf5bb8861a9c72e20d3fb0fdbd59233e43056e2b80475ab0aacdc2e781355"
+
+[[package]]
 name = "indexmap"
 version = "1.9.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1904,8 +1910,10 @@ name = "wasm-encoder"
 version = "0.225.0"
 dependencies = [
  "anyhow",
+ "index_vec",
  "leb128fmt",
  "tempfile",
+ "wasm-types",
  "wasmparser 0.225.0",
  "wasmprinter 0.225.0",
 ]
@@ -1932,6 +1940,7 @@ version = "0.225.0"
 dependencies = [
  "anyhow",
  "clap",
+ "index_vec",
  "indexmap 2.7.1",
  "serde",
  "serde_derive",
@@ -1950,10 +1959,12 @@ dependencies = [
  "clap",
  "egg",
  "env_logger",
+ "index_vec",
  "log",
  "rand",
  "thiserror",
  "wasm-encoder 0.225.0",
+ "wasm-types",
  "wasmparser 0.225.0",
  "wasmprinter 0.225.0",
  "wat",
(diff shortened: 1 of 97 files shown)


alexcrichton:
Thanks again for your work on this! I wanted to give a heads up that I'm at least a bit swamped right now so it may take a bit before I get back to this, but it's in my inbox so I don't forget about it.

gkgoat1:
> Following up on some conversation in #1985, this draft PR is a start on what it might look like to use typed indices throughout the whole codebase. I was originally planning to _finish_ this before coming back to #1985, but it eventually became clear to me that this is considerably bigger than I expected.
> 
> In any case, I'm opening this PR to gauge whether there's any interest in going further down this path, or if I should just call this an interesting experiment and abandon it. There were some unintuitive cases, so maybe there is some potential value in these from a codebase documentation/consistency perspective? For example:
> 
> https://github.com/bytecodealliance/wasm-tools/blob/beac7de36a6747a9061b8fa497a130aa4f7da185/crates/wasm-encoder/src/reencode/component.rs#L963-L966
> 

This could also extend outside the codebase; it would be useful for interoperability with `walrus` and `waffle` (both the original and my fork). A similar thing could also be done with `Operator`. +1, though I'm not in BA

> This code made me mark the `func_ty_index` field of the `CanonicalFunction::ThreadSpawn` variant as `TypeIdx` instead of `ComponentTypeIdx`, unlike all the fields in all the `Resource` variants. Is this correct? I don't know much about the component model, so maybe threads just happen to use core type indices while other similar-looking things use component type indices. Maybe it's clear from the spec (?), but it's not clear from the docstring:
> 
> https://github.com/bytecodealliance/wasm-tools/blob/beac7de36a6747a9061b8fa497a130aa4f7da185/crates/wasmparser/src/readers/component/canonicals.rs#L70-L74
> 
> Another random question I had while working on this: why do both [`wasm_encoder::NameSection::tag`](https://docs.rs/wasm-encoder/0.225.0/wasm_encoder/struct.NameSection.html#method.tag) and [`wasm_encoder::NameSection::tags`](https://docs.rs/wasm-encoder/0.225.0/wasm_encoder/struct.NameSection.html#method.tags) exist? Don't they do the same thing?

gkgoat1:
+1 again, I am developing a new WASM framework for storing and processing modules in-memory and this would help, again.

PR #2052, opened 2025-02-13
[wasm-metadata] rename `Author` to `Authors`

yoshuawuyts:
While filing https://github.com/bytecodealliance/wasm-pkg-tools/pull/150 I realized the OCI annotation is called `org.opencontainers.image.authors`. While in `wasm-metadata` we called it `Author`. The lack of pluralization was unintentional from my part, and this PR fixes that. Thanks!

cc/ @bytecodealliance/sig-packaging


yoshuawuyts:
oops, hold on - submodule issues. Let me fix that.

lann (approved):
I find it annoying that it is specified as plural but no guidance is given as to how to actually encode multiple authors. :roll_eyes:

Not our fault though so lgtm.

alexcrichton (approved):

PR #2059, opened 2025-02-20
Make `ConstExpr` implementation a bit more concise

samestep:
Tiny followup on #1985, since I couldn't figure out how to do this at the time I was writing that PR.


alexcrichton (approved):
Thanks!

PR #2061, opened 2025-02-20
Gate future/stream/error-context types on async feature

alexcrichton:
This commit updates the validation of components to ensure that future/stream/error-context types are all gated behind the component-model-async feature rather than being unconditionally allowed as they were previously.

### DIFF
--- a/crates/wast/src/component/types.rs
+++ b/crates/wast/src/component/types.rs
@@ -317,6 +317,7 @@ impl Peek for PrimitiveValType {
                 | Some(("float64", _))
                 | Some(("char", _))
                 | Some(("string", _))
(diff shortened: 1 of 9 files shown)


crates/wast/src/component/types.rs
```diff
@@ -317,6 +317,7 @@ impl Peek for PrimitiveValType {
                 | Some(("float64", _))
                 | Some(("char", _))
                 | Some(("string", _))
+                | Some(("error-context", _))
```
alexcrichton, line 320: I'll note that this was an accidental omission from https://github.com/bytecodealliance/wasm-tools/pull/2060

dicej (approved):

PR #2065, opened 2025-02-22
 Synchronize some component model intrinsics with the spec 

alexcrichton:
This commit goes through a few component model intrinsics and
synchronizes them with the current specification, mainly centered around
async intrinsics. This doesn't synchronize all of them but is at least a
start. Changes here include:

* Support for `resource.drop async` intrinsics.
* Renaming `task.backpressure` to `backpressure.set`.
* Adding canonical ABI options to `task.return`.
* A number of refactorings to simplify some constructs, reduce
  duplication, etc.

The validation of `task.return` has been updated to take
https://github.com/WebAssembly/component-model/issues/453 into account with the new set of options
that are allowed on it. Additionally `wit-component` was updated to set
options for `task.return` in the same manner as the `canon lift`'d
function on the other end.


alexcrichton:
I'll note that this is based on https://github.com/bytecodealliance/wasm-tools/pull/2064, so only the last commit is exclusive to the PR (at this time)

crates/wit-component/src/encoding.rs
```diff
@@ -2006,6 +2045,19 @@ enum ShimKind<'a> {
     /// A shim used for the `task.poll` built-in function, which must refer to
     /// the core module instance's memory to which results will be written.
     TaskPoll { async_: bool },
+    /// TODO
```
dicej, line 2048: Were you planning to fill in these `TODO` comments now or later?

crates/wit-component/src/encoding.rs
```diff
@@ -2418,6 +2503,32 @@ impl<'a> Shims<'a> {
     }
 }
 
+fn task_return_options_and_type(
+    resolve: &Resolve,
+    ty: Option<Type>,
+    async_: bool,
```
dicej, line 2509: When would this be `true`?

dicej (approved):

crates/wit-component/src/encoding.rs
alexcrichton: Ah this is a holdover from prior to the clarification of https://github.com/WebAssembly/component-model/issues/453, I'll remove it

PR #2073, opened 2025-02-25
 Implement the WIT async keyword 

alexcrichton:
This commit is an implementation of https://github.com/WebAssembly/component-model/pull/442 in
this repository, namely:

* The `async` identifier is now a keyword in WIT.
* Functions in WIT can be annotated `async`.
* New kebab-names are now recognized to round-trip this attribute
  through the wasm binary format:
  * `[async]foo` => `foo: async func()`
  * `[async method]foo.bar` => `resource foo { bar: async func() }`
  * `[async static]foo.bar` => `resource foo { bar: static async func() }`
* Various new tests were updated to ensure these names are gated behind
  the component-model-async feature and additionally test some
  validation of these names.
* The `wit-component` convention of "name mangling" was updated to use
  `[async-lower]` instead of `[async]` for lowered functions (to not
  conflict with the kebab-name otherwise. Additionally in exports
  `[async-lift]` is used instead of `[async]` as well as
  `[async-lift-stackful]` instead of `[async-stackful]`.
* Support was added to `wit-smith` to generate `async` functions.


alexcrichton:
Note that the first few commits come from https://github.com/bytecodealliance/wasm-tools/pull/2072, https://github.com/bytecodealliance/wasm-tools/pull/2070, and https://github.com/bytecodealliance/wasm-tools/pull/2069, so only the last commit here is the one in question for this PR

dicej (approved):
Thanks for doing this!

Issue #1995, opened 2025-02-04
Mismatch in stability attributes error in wit-parser when using `Unstable` attribute in wit


jsturtevant:
When a new interface is marked as `@unstable (feature = somefeaturegate)` and it uses a stable type from another package via `use package:interface/type.{name}` the resulting package can't be imported into other packages that use it. The resulting error is:

```
wasm-tools component wit --features tls ./wit/
error: mismatch in stability attributes
```

### example
I've demonstrated this by adding a test: https://github.com/jsturtevant/wasm-tools/commit/4495325817c25de8d2048846973b21f368ebde30

As well as trying to integrate it into another world: https://github.com/WebAssembly/wasi-cli/compare/main...jsturtevant:wasi-cli:wasi-tls-example?expand=1  (note that this was just a test to see if I was getting in the a set up similiar to other imports)

I was getting this original from wit-bindgen macro: https://github.com/bytecodealliance/wit-bindgen/tree/main?tab=readme-ov-file#supported-guest-languages

### workaround

for now I've patched wit-parser like https://github.com/jsturtevant/wasm-tools/commit/22cfe2ed1c2289bc55cfeb149f053725018efeb1 but it seems like there is probably a better way to approach?

alexcrichton:
Thanks for the report! I'm traveling for a bit so may take me a bit to dig in to this but I plan on doing so in the not-too-distant future

alexcrichton:
Digging into this, can you confirm that the reduced test case you have reproduces the original error you were seeing? Locally if I revert your fix on your branch and run the test suite I see:

```
failures:

---- "tests/ui/multi-package-features" ----
test "tests/ui/multi-package-features" failed

Caused by:
    0: failed to resolve directory while parsing WIT for path [tests/ui/multi-package-features]
    1: type `unstable-resource` not defined in interface
            --> tests/ui/multi-package-features/deps/unstable/root.wit:12:33
             |
          12 |     use wasi:dep2/stable@0.2.3.{unstable-resource};
             |                                 ^----------------


failures:
    "tests/ui/multi-package-features"
```

which looks different from "mismatch in stability attributes" that you were seeing with `wasi-tls`.

I'm definitely seeing some funny behavior that I wouldn't expect with `use` and feature attributes too. The test case you have can be reduced further to:

```wit
package wasmtime:test;

interface types {
  @unstable(feature = notsure)
  use wasi:dep2/stable@0.2.3.{unstable-resource};
}

[...]

PR #2076, opened 2025-02-26
fix mismatch in stability attributes error in wit-parser

jsturtevant:
fixes: #1995

When a new interface is marked as `@unstable (feature = somefeaturegate)` and it uses a stable type from another package via `use package:interface/type.{name}` the resulting package should be able to be imported into other packages that use it. The use case for this is adding a unstable interface such as `wasi-tls` to `wasi-cli`.  Once we get to `resolve_include`'s processing we've already verified that everything checks out from a feature/version perspective for those features. i.e. we can't be trying to consolidate an interface type with a feature that hasn't already been resolved.

### DIFF
--- a/crates/wit-parser/tests/ui/gated-include-use-with-stable.wit.json
+++ b/crates/wit-parser/tests/ui/gated-include-use-with-stable.wit.json
@@ -1,0 +1,397 @@
+{
+  "worlds": [
+    {
+      "name": "imports",
+      "imports": {
+        "interface-0": {
+          "interface": {
+            "id": 0,
+            "stability": {
+              "unstable": {
+                "feature": "active"
+              }
+            }
+          }
+        }
+      },
+      "exports": {},
+      "package": 0,
+      "stability": {
+        "unstable": {
+          "feature": "active"
+        }
+      }
+    },
+    {
+      "name": "imports",
+      "imports": {
+        "interface-1": {
+          "interface": {
+            "id": 1
+          }
+        }
+      },
+      "exports": {},
+      "package": 1
+    },
+    {
+      "name": "imports",
+      "imports": {
+        "interface-2": {
+          "interface": {
+            "id": 2,
+            "stability": {
+              "stable": {
+                "since": "0.2.0"
+              }
+            }
+          }
+        }
+      },
+      "exports": {},
+      "package": 2,
+      "stability": {
+        "stable": {
+          "since": "0.2.0"
+        }
+      }
+    },
+    {
+      "name": "imports",
+      "imports": {
+        "interface-2": {
+          "interface": {
+            "id": 2,
+            "stability": {
+              "stable": {
+                "since": "0.2.0"
+              }
+            }
+          }
+        },
+        "interface-1": {
+          "interface": {
+            "id": 1
+          }
+        },
+        "interface-0": {
+          "interface": {
+            "id": 0,
+            "stability": {
+              "unstable": {
+                "feature": "active"
+              }
+            }
+          }
+        }
+      },
+      "exports": {},
+      "package": 3,
+      "stability": {
+        "stable": {
+          "since": "0.2.0"
+        }
+      }
+    },
+    {
+      "name": "imports",
+      "imports": {
+        "interface-2": {
+          "interface": {
+            "id": 2,
+            "stability": {
+              "unstable": {
+                "feature": "active"
+              }
+            }
+          }
+        },
+        "stable-resource": {
+          "type": 3
+        },
+        "interface-1": {
+          "interface": {
+            "id": 1,
+            "stability": {
+              "unstable": {
+                "feature": "active"
+              }
+            }
+          }
+        },
+        "unversioned-resource": {
+          "type": 4
+        },
+        "interface-0": {
+          "interface": {
+            "id": 0,
+            "stability": {
+              "unstable": {
+                "feature": "active"
+              }
+            }
+          }
+        },
+        "unstable-resource": {
+          "type": 5
+        }
+      },
+      "exports": {},
+      "package": 4,
+      "stability": {
+        "unstable": {
+          "feature": "active"
+        }
+      }
+    },
+    {
+      "name": "test",
+      "imports": {
+        "interface-2": {
+          "interface": {
+            "id": 2,
+            "stability": {
+              "stable": {
+                "since": "0.2.0"
+              }
+            }
+          }
+        },
+        "stable-resource": {
+          "type": 3
+        },
+        "interface-1": {
+          "interface": {
+            "id": 1,
+            "stability": {
+              "unstable": {
+                "feature": "active"
+              }
+            }
+          }
+        },
+        "unversioned-resource": {
+          "type": 4
+        },
+        "interface-0": {
+          "interface": {
+            "id": 0,
+            "stability": {
+              "unstable": {
+                "feature": "active"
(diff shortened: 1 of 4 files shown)


alexcrichton (approved):

alexcrichton:
Hm ok the CI failure is kind of worrisome since I think this looks like we're getting nondeterministic output? I forget what affects the order things are fed in to the resolve (maybe filesystem ordering?) but it may mean that we still want some sort of "unifying" here but with a defined ordering rather than that `update_stability` method

jsturtevant:
I've been looking into it and found that it is because I am not preserving the versioning that was previously in `update_stability`, specifically:

```
// Otherwise if `into` is unknown then inherit the stability listed in
    // `from`.
    if into.is_unknown() {
        *into = from.clone();
        return Ok(());
    }
```

but I found that there might be a bug prior to this due to ordering in https://github.com/jsturtevant/wasm-tools/tree/fix-mismatch-attributes-on-include/crates/wit-component/tests/merge/success/merge.

For instance the final merge preserves the `@since` for the `into` but not the `from`:

```
world shared-world-with-versions-and-include {
  import shared-version-on-from;
  @since(version = 1.0.0)
  import shared-version-on-into;
}
```

And it does the opposite with the `since` when I merge `into` onto `from` (so ordering matters here):

```
world shared-world-with-versions-and-include {
  @since(version = 1.0.0)
  import shared-version-on-from;
  import shared-version-on-into;
}
```

Where I would expect to see:

```
world shared-world-with-versions-and-include {
  @since(version = 1.0.0)
  import shared-version-on-from;
  @since(version = 1.0.0)
  import shared-version-on-into;
}
```

My thought is replace the `update_stability` with something that copies those appropriately, but might need to sort through some changes with `usntable` attribute.

alexcrichton:
> My thought is replace the update_stability with something that copies those appropriately, but might need to sort through some changes with unstable attribute which isn't covered by that set of tests currently.

That sounds reasonable to me, sort of a "define an `Ord` for stability and preserve the max" strategy of sorts

jsturtevant:
> For instance the final merge preserves the `@since` for the `into` but not the `from`:

This ended up being a separate issue.  I'd be happy to split this out into a separate change and do a little refactoring to cleanup the change which I sorta shove into the world merge section. 

> That sounds reasonable to me, sort of a "define an Ord for stability and preserve the max" strategy of sorts

I added `derive ord` to the enum, the order ends up being the order in which it is in the enum so `stable` is less than `unstable` which I find a little awkward.  I am not sure if it's OK that way or if we should reorder the enum (would that be a breaking change?)

I'll clean up some of the comments tomorrow too

alexcrichton:
> This ended up being a separate issue. I'd be happy to split this out into a separate change and do a little refactoring to cleanup the change which I sorta shove into the world merge section.

Ah ok good catch! I'm happy to leave that in here, too, but if you'd be up for refactoring to deduplicate a bit I think that'd be good 👍 

>  I am not sure if it's OK that way or if we should reorder the enum (would that be a breaking change?)

Definitely ok to reorder an enum, so how about a swap of `Unknown` and `Stable` so that way `>` should work? Also, mind leaving a comment on the enum that the `derive(Ord)` ordering is significant?

ydnar:
@alexcrichton found an issue with this patch: when WIT is recreated from the resolved JSON, it has versioned feature gates on the unversioned `wasmtime:test` package:

```wit
package wasi:foo@0.2.3;

@since(version = 0.2.0)
world imports {
	@since(version = 0.2.0)
	import wasi:dep2/stable@0.2.3;
	import wasi:dep-unversioned/unversioned;
	@unstable(feature = active)
	import wasi:dep-unstable/unstable;
}

package wasmtime:test {
	world test {
		@since(version = 0.2.0)
		import wasi:dep2/stable@0.2.3;
		@unstable(feature = active)
		use wasi:dep2/stable@0.2.3.{stable-resource};
		@unstable(feature = active)
		import wasi:dep-unversioned/unversioned;
		@unstable(feature = active)
		use wasi:dep-unversioned/unversioned.{unversioned-resource};
		@unstable(feature = active)
		import wasi:dep-unstable/unstable;
		@unstable(feature = active)
		use wasi:dep-unstable/unstable.{unstable-resource};
	}

	world test-ordered {
		@since(version = 0.2.0)
		import wasi:dep2/stable@0.2.3;
		@unstable(feature = active)
		import wasi:dep-unversioned/unversioned;
		@unstable(feature = active)
		import wasi:dep-unstable/unstable;
		@unstable(feature = active)
		use wasi:dep2/stable@0.2.3.{stable-resource};
		@unstable(feature = active)
		use wasi:dep-unversioned/unversioned.{unversioned-resource};
		@unstable(feature = active)
		use wasi:dep-unstable/unstable.{unstable-resource};
	}
}

package wasi:unstable@0.2.3 {
	@unstable(feature = active)
	world imports {
		@unstable(feature = active)
		import wasi:dep2/stable@0.2.3;
		@unstable(feature = active)
		use wasi:dep2/stable@0.2.3.{stable-resource};
		@unstable(feature = active)
		import wasi:dep-unversioned/unversioned;
		@unstable(feature = active)
		use wasi:dep-unversioned/unversioned.{unversioned-resource};
		@unstable(feature = active)
		import wasi:dep-unstable/unstable;
		@unstable(feature = active)
		use wasi:dep-unstable/unstable.{unstable-resource};
	}
}

package wasi:dep-unstable {
	@unstable(feature = active)
	interface unstable {
		@unstable(feature = active)
		resource unstable-resource;
	}

	@unstable(feature = active)
	world imports {
		@unstable(feature = active)
		import unstable;
	}
}

package wasi:dep-unversioned {
	interface unversioned {
		resource unversioned-resource;
	}

	world imports {
		import unversioned;
	}
}

package wasi:dep2@0.2.3 {
	@since(version = 0.2.1)
	interface stable {
		resource stable-resource;
	}

	@since(version = 0.2.0)
	world imports {
		@since(version = 0.2.0)
		import stable;
	}
}
```

crates/wit-parser/tests/ui/gated-include-use-with-stable.wit.json
```diff
@@ -0,0 +1,397 @@
...
+        }
+      }
+    },
+    {
+      "name": "test",
+      "imports": {
+        "interface-2": {
+          "interface": {
+            "id": 2,
+            "stability": {
```
ydnar, line 153: This should be feature gate, not version gate. The package that world `test` belongs to (`wasmtime:test`) is unversioned, so this stability is impossible.

ydnar: During resolution, when a world includes another world from a different package, should version gates be resolved then, and stripped out entirely?

alexcrichton: I think that everything new coming from `include` should probably forcibly get whatever feature gate is on the `include` itself, but otherwise I do agree that whatever the source is should likely be discarded entirely as features/versions aren't guaranteed to make sense in the destination world

jsturtevant: @ydnar calls out in https://github.com/bytecodealliance/go-modules/issues/325  the test suite there does `WIT → JSON → WIT`.  This seems like a good addition that would have caught this. I will open a new issue and see if can get this resolved.

PR #2077, opened 2025-02-27
Fix the docstring for `wasmprinter::Config::print`

samestep:
The method signature was changed in #1593 and #1596 but the docstring never got updated.


alexcrichton (approved):
Thanks!

PR #2084, opened 2025-03-07
Enhance dummy async module generation

alexcrichton:
This commit enhances the `--async-callback` and `--async-stackful` arguments to the `wasm-tools component embed` subcommand when paired with `--dummy-names legacy`. Specifically the dummy module generated now additionally includes imports for various component model intrinsics such as waitable-set management, future management, `task.return`, etc.  The goal of this commit is to stress `wit-component`'s processing of these intrinsics during normal fuzzing.

After the dummy module support was updated the `roundtrip_wit` fuzzer was run locally for awhile and a number of bugs popped out. This commit additionally includes various fixes and test cases. All tests "pass" before this commit because intrinsics weren't generated in dummy mode, but tests all fail without the `wit-component` fixes but with the dummy module support.


alexcrichton:
My plan is to make a point release after this is merged.

dicej (approved):

PR #2090, opened 2025-03-10
Reduce dependencies of `wasm-metadata`

alexcrichton:
This crate is depended on by `wit-component` which shows up in a lot of places, but most of the dependencies of `wasm-metadata` are not needed for `wit-component`. This commit gates most of the OCI-related information behind a compile-time `oci` feature which is enabled by default. The workspace dependency now disables default features and only the CLI itself enables the default features of `wasm-metadata`.


yoshuawuyts (approved):
Seems reasonable to me; ship it!

PR #2092, opened 2025-03-11
`wasm-smith`: the largest `i31ref` value is interesting

fitzgen:
This is the largest value that can be stored in an `i31ref` without any truncation. That is interesting and deserves to be in the interesting set.


fitzgen:
Nevermind, the negative powers of two just above this covers this case.

Issue #2093, opened 2025-03-11
Add a `wasm-tools wast` subcommand


fitzgen:
A little helper subcommand to test the parsing/validation parts of `.wast` tests, without needing to copy the file into `tests/local` and run `cargo test --test roundtrip`.

alexcrichton:
Done in #2096

PR #2096, opened 2025-03-12
Add a `wast` subcommand to the CLI

alexcrichton:
I was duly inspired by https://github.com/bytecodealliance/wasm-tools/issues/2093 to not only add such a subcommand but also restructure the test suite. I've long felt that the current test suite is a bit of a pain especially around per-test configuration since right now it's all configured in `tests/roundtrip.rs` and it's also not obvious how much configuration applies to each test. With the concept of a `wast` subcomand I realized that we could migrate all tests to `tests/cli/*` which easily afforts per-test configuration, yay!

This commit thusly migrates `tests/roundtrip.rs` to `wasm-tools wast` and then moves all tests into the `tests/cli/*` directory. Tests aren't really cleaned up at this time but it should be much easier in the future to clean things up as needed.


alexcrichton:
I still want to put some finishing touches on this in terms of documentation, but in terms of review it should otherwise be good to go

alexcrichton:
documented now 👍

src/bin/wasm-tools/wast.rs
```diff
@@ -0,0 +1,584 @@
...
+        let (test, contents) = self.read_wast()?;
+        let contents = contents.as_str();
+        macro_rules! adjust {
+            ($e:expr) => {{
+                let mut e = wast::Error::from($e);
+                e.set_path(test);
+                e.set_text(contents);
+                e
+            }};
+        }
```
fitzgen, line 97: This seems like it could just be a closure, and doesn't actually need to be a macro?

fitzgen: Oh I guess because it is generic over the type of `$e` it can't be a closure. blech.

README.md
```diff
@@ -155,6 +155,7 @@ that can be use programmatically as well:
 | `wasm-tools addr2line` |  |  | Translate wasm offsets to filename/line numbers with DWARF |
 | `wasm-tools completion` |  |  | Generate shell completion scripts for `wasm-tools` |
 | `wasm-tools json-from-wast` |  |  | Convert a `*.wast` file into JSON commands |
+| `wasm-tools -wast` |  |  | Validate the structure of a `*.wast` file |
```
fitzgen, line 158: `-wast` should be `wast`

fitzgen (approved):
Nice!! This should also fix the issue of `BLESS=1` and running a particular test, right?

alexcrichton:
> This should also fix the issue of BLESS=1 and running a particular test, right?

You mean where you run one test, set `BLESS=1`, and all of `tests/snapshots` is deleted? If so, yes!

PR #2100, opened 2025-03-14
Implement `context.{get,set}` intrinsics

alexcrichton:
This fills out the last missing set of intrinsics from the component-model async proposal. These two intrinsics are pretty simple and use preexisting idioms everywhere for integration into tooling.


dicej (approved):
I'm always impressed with how fast you can whip up PRs like this.  Thanks!

PR #2102, opened 2025-03-15
Add public interface for adding continuation types to `CoreTypeEncoder`

pufferfish101007:
The machinery for encoding continuation types (from the stack switching proposal) already exists, but there was no public method to add them. This resolves that - it's such a minor change that I didn't think it was worth opening an issue for it.


alexcrichton (approved):
Thanks!

PR #2104, opened 2025-03-18
Improve support for memory64/shared-everything-threads in components

alexcrichton:
Previously the validator in wasm-tools was updated to conservatively reject all aliases of exports of core memories/tables that were 64-bit or shared. This is a coarse approximation of the actual rules though which today disallow using 64-bit memories in the canonical ABI. Instead this commit updates with a few minor changes to improve the situation here, namely:

* Aliasing core exports is now allowed for any table and memory type.
* Usage of memories in canonical ABI positions now verifies that the memory is of the expected type, using the standard sub-typing.
* Usage of tables in `canon thread.spawn_indirect` now performs a type-check via a similar method to ensure the table is of the right type.

This adjusts the behavior of a few tests here and there as well to account for some slight differences in behavior with these proposals.

### DIFF
--- a/crates/wasmparser/src/validator/component.rs
+++ b/crates/wasmparser/src/validator/component.rs
@@ -1833,7 +1833,7 @@ impl ComponentState {
             )
         }
 
-        self.memory_at(memory, offset)?;
+        self.cabi_memory_at(memory, offset)?;
 
         self.core_funcs
             .push(types.intern_func_type(FuncType::new([ValType::I32; 2], [ValType::I32]), offset));
@@ -1861,7 +1861,7 @@ impl ComponentState {
             )
         }
 
-        self.memory_at(memory, offset)?;
+        self.cabi_memory_at(memory, offset)?;
 
         self.core_funcs
             .push(types.intern_func_type(FuncType::new([ValType::I32; 2], [ValType::I32]), offset));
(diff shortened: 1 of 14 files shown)


crates/wasmparser/src/validator/component.rs
```diff
@@ -1978,14 +1978,24 @@ impl ComponentState {
...
-        // TODO: check that the table is shared once components allow shared
-        // tables.
-        let expected_ty = RefType::FUNCREF
-            .shared()
-            .expect("a funcref can always be shared");
-        if table.element_type != expected_ty {
-            bail!(offset, "expected a table of shared `funcref`");
-        }
+
+        SubtypeCx::table_type(
```
abrown, line 1982: I learn something new every day...

tests/cli/component-model/shared-everything-threads/not-accepted.wast
```diff
@@ -1,20 +1,34 @@
...
+  (core module $A
+    (table (export "m") shared 1 2 (ref null (shared func)))
   )
-  "shared linear memories are not compatible with components yet")
+  (core instance $A (instantiate $A))
+  (alias core export $A "m" (core table $m))
+
+  (core module $B (import "" "" (table shared 1 2 (ref null (shared func)))))
+  (core instance (instantiate $B (with "" (instance (export "" (table $m))))))
+)
```
abrown, line 22: You want to keep these tests in `not-accepted.wast`? Or rename this file? Not a big deal if we leave things here, I suppose, just noting the file name.

abrown (approved):
LGTM!

tests/cli/component-model/shared-everything-threads/not-accepted.wast
alexcrichton: I tend to pay far less attention to filenames than I should unfortunately... I think I'll go ahead and merge this PR, but I think it'd be best to clean up some organization from the old `tests/local/*` test suite which is no longer necessary in the future

PR #2105, opened 2025-03-20
Update the spec test suite

alexcrichton:
No changes necessary in spec files, just some tests have shifted around.


pchickey:
Don't understand this one - where did all the memory64 prints go? Did you forget to git add after they moved?

alexcrichton:
Ah I think this is because of https://github.com/WebAssembly/testsuite/pull/116 where memory64 no longer has a dedicated test suite since it's all merged into wasm-3.0 now.

pchickey (approved):

PR #2109, opened 2025-03-21
wast.rs: add explicit --assert for test-folded/no-test-folded

keithw:
This PR gives the `wast` subcommand a command-line `--assert` to configure whether to test the folded-instructions printer. Currently this is set more implicitly (`!self.features.features().legacy_exceptions()`).


alexcrichton (approved):
Thanks!

PR #2114, opened 2025-03-26
Fix/improve validation of components with tags

alexcrichton:
This commit fixes an issue where the `Tag` kind of core item was allowed to "leak through" validation of components without the exceptions proposal active. Additionally creating a core instance with a tag was consulting the wrong index space and needed fixing too. The end result is that the `exceptions` proposal is now required for tags to appear within components.

This so far has been glossed over because any valid component would require a core module with a tag and those are properly gated, but it's possible to construct an invalid component with a tag "kind" within it which doesn't actually have any tags. By forgetting to gate on the exception-handling proposal it was leaking this "kind" to callers which weren't necessarily prepared for it.

Finally this also fixes a panic in the text format for components where named tags still had an `unimplemented!()` for when they were being registered.

### DIFF
--- a/tests/cli/component-model/gated-tags.wast
+++ b/tests/cli/component-model/gated-tags.wast
@@ -1,0 +1,18 @@
+;; RUN: wast --features=-exceptions %
+
+(assert_invalid
+  (component
+    (core module $m (func (export "")))
+    (core instance $i (instantiate $m))
+    (alias core export $i "" (core tag $t))
+  )
+  "exceptions proposal not enabled")
+
+(assert_invalid
+  (component
+    (core module $m (func (export "")))
+    (core instance $i (instantiate $m))
+    (core instance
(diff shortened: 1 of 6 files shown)


alexcrichton:
Thanks to @0xalpharush for identifying this in Wasmtime which led to this bugfix here

tests/cli/component-model/gated-tags.wast
```diff
@@ -0,0 +1,18 @@
+;; RUN: wast --features=-exceptions %
```
alexcrichton, line 1: All of our existing `*.wast` tests also do `--assert default` with `--snapshot ...`, but personally I find it kind of a pain to set that all up and manage it, so my thinking is that it's reasonable to have one-off tests like this which just run the directives themselves.

fitzgen (approved):

PR #2115, opened 2025-03-27
Refactor how liveness of imports is calculated

alexcrichton:
In `wit-component` there's a pass when creating a component to prune types/functions/etc from imported interfaces if they're not actually needed by the final component. This logic has not stood well against the test of time and the loop nowadays looks like: for each import if it's found within the imports then mark it live. This logic had some special handling of resource intrinsics but did not take into account any async/stream/future-related intrinsics.

To fix this issue I've replaced the outdated loop with "for all module imports add them, as necessary, to the set of live types". This starts with the source of truth, module imports, and works backwards to WIT types as opposed to the other way around, starting with WIT types and testing to see if it's included in the intrinsics list. That fixes the various cases with async intrinsics and should ensure that all future intrinsics are also accounted for.

Closes bytecodealliance/wasip3-prototyping#82


fuzz/src/roundtrip_wit.rs
```diff
@@ -47,6 +49,19 @@ pub fn run(u: &mut Unstructured<'_>) -> Result<()> {
...
+                u,
+                removed_funcs: 0,
+            };
+            if reencode
+                .parse_core_module(&mut dst, Default::default(), &dummy)
+                .is_ok()
+            {
+                dummy = dst.finish();
+            }
+        }
```
alexcrichton, line 64: I'll note that this update to the fuzzer did not actually turn up this bug after fuzzing for ~1 hour on 30 cores. My best guess is that the shape of this bug is pretty specific so it's relatively rare to come up, but nonetheless I'm hoping that this stresses the optional import logic of wit-component.

dicej (approved):

PR #2116, opened 2025-03-27
Update some async intrinsic signatures

alexcrichton:
* `future.new` - takes a pointer parameter instead of returning a result.
* `future.close-{readable,writable}` - drop the `error-context` argument second field.

### DIFF
--- a/crates/wit-component/src/dummy.rs
+++ b/crates/wit-component/src/dummy.rs
@@ -192,13 +192,13 @@ fn push_imported_future_and_stream_intrinsics(
             TypeDefKind::Future(_) => {
                 wat.push_str(&format!(
                     r#"
-(import {module:?} "[future-new-{i}]{name}" (func (result i32)))
+(import {module:?} "[future-new-{i}]{name}" (func (param i32)))
 (import {module:?} "[future-read-{i}]{name}" (func (param i32 i32) (result i32)))
 (import {module:?} "[future-write-{i}]{name}" (func (param i32 i32) (result i32)))
 (import {module:?} "[future-cancel-read-{i}]{name}" (func (param i32) (result i32)))
 (import {module:?} "[future-cancel-write-{i}]{name}" (func (param i32) (result i32)))
-(import {module:?} "[future-close-readable-{i}]{name}" (func (param i32 i32)))
-(import {module:?} "[future-close-writable-{i}]{name}" (func (param i32 i32)))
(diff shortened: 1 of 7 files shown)


alexcrichton:
This is intended to encompass upstream issues such as https://github.com/WebAssembly/component-model/issues/474 and https://github.com/WebAssembly/component-model/issues/475. Not complete as more changes may be necessary. I hope to add intrinsics for task cancellation soon to. Nevertheless I'm using this in my wit-bindgen refactor.

cc @dicej

crates/wit-component/src/dummy.rs
```diff
@@ -192,13 +192,13 @@ fn push_imported_future_and_stream_intrinsics(
...
-(import {module:?} "[future-new-{i}]{name}" (func (result i32)))
+(import {module:?} "[future-new-{i}]{name}" (func (result i64)))
 (import {module:?} "[future-read-{i}]{name}" (func (param i32 i32) (result i32)))
 (import {module:?} "[future-write-{i}]{name}" (func (param i32 i32) (result i32)))
 (import {module:?} "[future-cancel-read-{i}]{name}" (func (param i32) (result i32)))
 (import {module:?} "[future-cancel-write-{i}]{name}" (func (param i32) (result i32)))
-(import {module:?} "[future-close-readable-{i}]{name}" (func (param i32 i32)))
-(import {module:?} "[future-close-writable-{i}]{name}" (func (param i32 i32)))
+(import {module:?} "[future-close-readable-{i}]{name}" (func (param i32)))
+(import {module:?} "[future-close-writable-{i}]{name}" (func (param i32)))
```
alexcrichton, line 201: This is a basic summary of the changes here, notably:

* `{future,stream}.new` intrinsics return a two 32-bit handles as a result packed into a 64-bit integer. It kind of sucks we don't have multi-value and I also didn't want to go the route of adding canonopts to these intrinsics just to write values into memory. It's a bit wonky since bit packing doesn't happen anywhere else in the canonical ABI but it happens in other async intrinsics so maybe ok?
* For `{future,stream}.close-{readable,writable}` the only change is to drop the error-context parameter.

alexcrichton (commented):
Ok I think this is good to go @dicej, just some minor changes here and I'll follow-up later with task cancellation intrinsics in a future PR.

dicej (approved):

PR #2125, opened 2025-04-04
chore: preserve original `resolve` when decoding WIT package twice

0xsenty:
The second call to `wit_component::decode` reused the `resolve` identifier from the first decode. This silently overwrites the original `resolve`, which can be misleading — especially since both `resolve` and `resolve2` are referenced later in the function.

I’ve updated the second match arm to bind to `resolve2` instead, so both decoded values are preserved independently as originally intended.

This should clarify the logic and avoid unintentional data loss.


alexcrichton:
Thanks! Is this perhaps the wrong change though? This current PR looks like it's a small refactoring but no actual behavior changes.

alexcrichton:
I'm going to close this due to lack of response

PR #2128, opened 2025-04-08
Improve text-format parsing of offset/align and update spec testsuite

keithw:
The memory64/wasm-3.0 features update the type of memop `align=`_N_ and `offset=`_N_ to u64 in the text format. The `align.wast` test now includes an example with the largest possible values and also uses underscores in them (https://github.com/WebAssembly/spec/pull/1888). This updates the `wast` parsing logic to support this (by using the lexer), and updates the spec testsuite.


alexcrichton (approved):
Thanks!

PR #2130, opened 2025-04-09
[test] Reclassify some tests between "invalid" and "malformed"

keithw:
This PR is preparation for an invalid/malformed separation in the wast subcommand and for wasm-tools to be able to roundtrip any non-malformed module or component (the overall patchset is at https://github.com/keithw/wasm-tools/tree/malformed-invalid-separation).

It reclassifies some of the local tests between "malformed" (a violation of the [binary-format](https://webassembly.github.io/spec/core/binary/index.html) or [text-format](https://webassembly.github.io/spec/core/text/index.html) syntaxes) vs. ["invalid"](https://webassembly.github.io/spec/core/valid/index.html). I think these tests now match the spec more closely, although I haven't checked them all in the reference interpreter because the error messages aren't exactly the same and the reference interpreter doesn't make it easy to ignore them.

This doesn't have any effect yet because `wasm-tools wast` currently handles `assert_malformed` and `assert_invalid` the same way, but it will make a difference once they are separated.

### DIFF
--- a/tests/cli/code-after-end.wast
+++ b/tests/cli/code-after-end.wast
@@ -1,36 +1,36 @@
 ;; RUN: wast --assert default --snapshot tests/snapshots %
 
-(assert_invalid
-  (module
-    (func end))
+(assert_malformed
+  (module quote
+    "(func end)")
   "operators remaining after end of function")
 
-(assert_invalid
-  (module
-    (func end block))
+(assert_malformed
+  (module quote
+    "(func end block)")
   "operators remaining after end of function")
 
-(assert_invalid
-  (module
-    (func end i32.add))
+(assert_malformed
+  (module quote
+    "(func end i32.add)")
   "operators remaining after end of function")
 
-(assert_invalid
-  (module
-    (func end unreachable))
+(assert_malformed
+  (module quote
+    "(func end unreachable)")
   "operators remaining after end of function")
 
-(assert_invalid
-  (module
-    (func end br 0))
+(assert_malformed
+  (module quote
+    "(func end br 0)")
   "operators remaining after end of function")
 
-(assert_invalid
-  (module
-    (func end return))
+(assert_malformed
+  (module quote
+    "(func end return)")
   "operators remaining after end of function")
 
-(assert_invalid
-  (module
-    (func end return_call 0))
+(assert_malformed
+  (module quote
+    "(func end return_call 0)")
   "operators remaining after end of function")
--- a/tests/cli/code-after-end.wast
+++ b/tests/cli/code-after-end.wast
@@ -5,9 +5,21 @@
     "(func end)")
   "operators remaining after end of function")
 
+;; binary version of previous test
+(assert_malformed
+  (module binary
+    "\00asm\01\00\00\00\01\04\01\60\00\00\03\02\01\00\0a\05\01\03\00\0b\0b")
+  "operators remaining after end of function")
+
 (assert_malformed
   (module quote
     "(func end block)")
+  "operators remaining after end of function")
+
+;; binary version of previous test
+(assert_malformed
+  (module binary
+    "\00asm\01\00\00\00\01\04\01\60\00\00\03\02\01\00\0a\07\01\05\00\0b\02\40\0b")
   "operators remaining after end of function")
 
 (assert_malformed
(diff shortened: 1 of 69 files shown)


tests/cli/missing-features/shared-everything-threads-globals.wast
```diff
@@ -1,14 +1,14 @@
 ;; RUN: wast --assert default --snapshot tests/snapshots % -f mvp
 
-(assert_invalid
-  (module
-    (global (shared i32) (i32.const 0))
+(assert_malformed
+  (module quote
+    "(global (shared i32) (i32.const 0))"
```
alexcrichton, line 5: How come this, and some tests below, switched to `module quote`?

Or possible answering my own question I was looking at some of the future patches and I saw more handling of features in `BinaryReader`, so is that what's happening here? Personally I've always tried to minimize the interaction of features with parsing and put everything in validation instead. I've felt historically that while, technically yes, this module produces an invalid binary with respect to the current spec it's just way easier to develop a proposal and evaluate it if the validator is the only piece of tooling that needs configuration (as opposed to every layer of tooling needing feature configuration)

In that sense I'd prefer to keep this as-is with `assert_invalid` and the error only coming up with the validator, but I'm also want to get your take on this as well

tests/cli/invalid/func.wast
```diff
@@ -63,27 +63,27 @@
...
+    "(func end br 100000)")
   "operators remaining after end of function")
 
-(assert_invalid
-  (module
-    (func end br_table 0))
+(assert_malformed
+  (module quote
+    "(func end br_table 0)")
   "operators remaining after end of function")
```
alexcrichton, line 89: Similar comment on these tests as to the ones above, would you be up for adding a `module binary` dual for these malformed tests?

tests/cli/missing-features/shared-everything-threads-globals.wast
keithw: Re: `assert_invalid` vs. `assert_malformed`, I think the issue is that this error is currently detected in the binary reader (https://github.com/bytecodealliance/wasm-tools/blob/main/crates/wasmparser/src/readers/core/globals.rs#L42-L54). This seems like the right behavior because of the upstream tests make this malformed (https://github.com/WebAssembly/spec/blob/main/test/core/global.wast#L414).

If it were `accept_invalid`, it should be parseable and printable and then parseable again to produce the same binary; `accept_malformed` means we're not expecting to be able to parse or print it which reflects the status quo in this case.

Re: `module` vs. `module quote`, at least for wasm-tools this doesn't *need* to be a `module quote`, because the error is only detected by the binary parser (not the text parser). The method I'm proposing for the `assert_malformed` tests is basically not to change the text parsers at all, and instead run each test by first (1) using the `wast` crate to convert WAT to bytes, and then (2) running wasmparser on those bytes to see if they parse as binary Wasm. That way all the safety checks are done in the binary reader/parser and it doesn't matter where the module originally came from (text or binary). But in general I think it's good practice for an `assert_malformed` test to use `module quote` or `module binary`, because if the text parser is what catches the malformedness (e.g. for something like mismatched parentheses or quotation marks), then unless the module is quoted, it would doom the whole .wast file. If we want these tests to pass in the reference interpreter (once the shared-everything-threads proposal has a reference interpreter) or another engine, I think they would need `module quote` or `module binary`.

tests/cli/code-after-end.wast
```diff
@@ -1,36 +1,36 @@
...
+  (module quote
+    "(func end return)")
   "operators remaining after end of function")
 
-(assert_invalid
-  (module
-    (func end return_call 0))
+(assert_malformed
+  (module quote
+    "(func end return_call 0)")
```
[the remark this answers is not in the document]
keithw: I'm happy to do it, but fwiw I'm basically not anything to the text parser and the text parser won't catch this by itself -- it's only the binary parser that will keep track of the blocks and make sure the `end`s are well-matched. But I don't mind extra safety -- will add. :-)

(Per above: the method I'm proposing for the assert_malformed tests is basically not to change the text parsers at all, and instead run each test by first (1) using the wast crate to convert WAT to bytes, and then (2) running wasmparser on those bytes to see if they parse as binary Wasm. That way all the safety checks are done in the binary reader/parser and it doesn't matter where the module originally came from (text or binary).)

tests/cli/invalid/func.wast
keithw: Done.

tests/cli/code-after-end.wast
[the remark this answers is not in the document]
keithw: Done.

alexcrichton (approved):
Thanks! I'll continue some discussion of malformed/invalid over on https://github.com/bytecodealliance/wasm-tools/pull/2123, and in the meantime I'll flag this for merge

PR #2131, opened 2025-04-09
wast.rs: Validate, roundtrip and snapshot "unlinkable" modules

keithw:
This PR adds test coverage for the `assert_unlinkable` cases in the spec testsuite.

These are valid modules, so this PR makes the `wast` subcommand treat them the same as a `module` declaration (i.e. validate, roundtrip, and snapshot). There are some tests in the testsuite that use `assert_unlinkable` to test extreme (valid) modules without causing the test runner to actually instantiate a gigantic module (e.g. https://github.com/WebAssembly/custom-page-sizes/blob/main/test/core/custom-page-sizes/memory_max.wast), so it would be good to have wasm-tools run these test cases through the validator instead of ignoring them.

Treating them like a `module` seems the most consistent, but the snapshots are a lot of new files in the repo and I'm not sure they're totally necessary. We could also just validate and roundtrip without the snapshot.

This is part of the "malformed-invalid separation + roundtrip invalid" patchset (https://github.com/keithw/wasm-tools/tree/malformed-invalid-separation).


alexcrichton (approved):
Thanks!

I'm also not a huge fan of the quantity of snapshot files but that's sort of where we're at for now so this isn't unreasonable in that sense.

PR #2133, opened 2025-04-10
wast: ignore multiple start sections (already detected in wasmparser)

keithw:
The wast crate has a `validate` function (on `Wat`s, `Module`s and `Component`s), but:

- it's not getting called totally consistently ([Wat `parse`](https://github.com/bytecodealliance/wasm-tools/blob/main/crates/wast/src/wat.rs#L62) calls it, but [Module `parse_without_module_keyword`](https://github.com/bytecodealliance/wasm-tools/blob/main/crates/wast/src/core/module.rs#L124) and [Component `parse_without_component_keyword`](https://github.com/bytecodealliance/wasm-tools/blob/main/crates/wast/src/component/component.rs#L125) don't call it, and those are used instead for some of the wast directives -- this led to some confusing debugging!)
- it doesn't do a ton (it checks for multiple start sections, which `wasmparser` is also capable of detecting)
- I think it would be safer to put all the eggs in wasmparser's basket to reduce the risk that there's an error condition that only gets caught when a module/component is expressed in the text format

This PR eliminates the `validate` function and defers detection of the error to the binary format.

### DIFF
--- a/crates/wast/src/component/component.rs
+++ b/crates/wast/src/component/component.rs
@@ -90,21 +90,6 @@ impl<'a> Component<'a> {
         crate::core::EncodeOptions::default().encode_component(self)
     }
 
-    pub(crate) fn validate(&self, parser: Parser<'_>) -> Result<()> {
-        let mut starts = 0;
-        if let ComponentKind::Text(fields) = &self.kind {
-            for item in fields.iter() {
-                if let ComponentField::Start(_) = item {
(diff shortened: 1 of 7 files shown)


alexcrichton (approved):
Huh I'm not actually sure where this came from... I know I had a fair number of misunderstandings about the text format when I first wrote all this and maybe that's what happened here... In any case thanks!

PR #2134, opened 2025-04-10
wasmparser: detect "malformed" cases in parser alone (without validator)

keithw:
This PR moves the checks required to detect "malformed" modules/components out of the validator and into the reader/parser. The biggest (textual) change is moving all of the operators reading logic into the OperatorsReader, which also now maintains a stack of blocks required to check that `end` opcodes are matched with the beginning of a block.


crates/wasm-shrink/tests/tests.rs
```diff
@@ -44,8 +44,10 @@ fn shrink_to_empty_allowed() -> Result<()> {
 #[test]
 fn smoke_test() -> Result<()> {
     let info = WasmShrink::default().attempts(100).run(wasm(), |wasm| {
-        let wat = wasmprinter::print_bytes(&wasm)?;
-        Ok(wat.contains("local.get"))
+        match wasmprinter::print_bytes(&wasm) {
+            Ok(wat) => Ok(wat.contains("local.get")),
+            Err(_) => Ok(false),
+        }
     })?;
```
alexcrichton, line 51: This change technically shouldn't be required, `wasm-mutate`, when given a valid module, should always produce a valid module. What you've run into here though was a preexisting bug which I've fixed at https://github.com/bytecodealliance/wasm-tools/pull/2135, so I'll try to remember to ensure this gets reverted back depending on the order things land

crates/wasmparser/src/readers/core/memories.rs
```diff
@@ -32,23 +32,40 @@ impl<'a> FromReader<'a> for MemoryType {
...
+        }
+        if shared && !reader.threads() {
+            bail!(pos, "threads must be enabled for shared memories");
+        }
+        if has_page_size && !reader.custom_page_sizes() {
+            bail!(
+                pos,
+                "the custom page sizes proposal must be enabled to customize a memory's page size"
+            );
+        }
```
alexcrichton, line 46: Could these feature checks be moved back to the validator? Locally it looks like `tests/cli/missing-features/custom-page-sizes.wast` fails but that can be switched to `assert_invalid` from `assert_malformed`.  

Or otherwise, more generally, I'd prefer to keep feature flag checking entirely in the validator where possible. If a local test is the one forcing a feature flag check then it's definitely ok to change local tests to `assert_invalid` instead of `assert_malformed`.

crates/wasmparser/src/readers/core/memories.rs
```diff
@@ -32,23 +32,40 @@ impl<'a> FromReader<'a> for MemoryType {
...
                 Some(reader.read_var_u64()?)
             } else {
                 Some(reader.read_var_u32()?.into())
             },
             page_size_log2: if has_page_size {
-                Some(reader.read_var_u32()?)
+                let val = reader.read_var_u32()?;
+                if val >= 64 {
+                    bail!(pos, "invalid custom page size");
+                }
```
alexcrichton, line 67: At least with the [current spec](https://github.com/WebAssembly/custom-page-sizes/blob/main/proposals/custom-page-sizes/Overview.md) I think that it's a valid binary to specify an arbitrary u32 here, but only validation-invalid if it's >= 64. Does this cause problems with the preexisting upstream spec tests though?

crates/wasmparser/src/readers/core/tables.rs
```diff
@@ -74,17 +74,28 @@ impl<'a> FromReader<'a> for TableType {
...
+
+        if shared && !reader.shared_everything_threads() {
+            bail!(
+                pos,
+                "shared tables require the shared-everything-threads proposal"
+            );
+        }
+        if table64 && !reader.memory64() {
+            bail!(pos, "memory64 must be enabled for 64-bit tables");
+        }
```
alexcrichton, line 86: Similar to above with memories, can these be moved back to the validator and the local tests changed to `assert_invalid`?

crates/wasmparser/src/readers/core/tags.rs
```diff
@@ -24,6 +24,12 @@ impl<'a> FromReader<'a> for TagType {
         if attribute != 0 {
             bail!(reader.original_position() - 1, "invalid tag attributes");
         }
+        if !reader.exceptions() {
+            bail!(
+                reader.original_position(),
+                "exceptions proposal not enabled"
+            );
+        }
```
alexcrichton, line 32: Mind moving this back to the validator? (and tests we have to `assert_invalid`)

crates/wasmparser/src/readers/core/operators.rs
```diff
@@ -251,22 +358,42 @@ impl<'a> OperatorsReader<'a> {
...
+        if !self.eof() {
+            bail!(
+                self.original_position(),
+                "unexpected data at the end of operators"
+            );
+        }
+        match (data_index_allowed, self.data_index_occurred) {
+            (false, Some(pos)) => bail!(pos, "data count section required"),
+            _ => Ok(()),
+        }
```
alexcrichton, line 376: Question on this: why was this moved out of the validator? From a binary-syntax point of view my read is that `data.drop` is allowed with any index, and it's only validation that requires that `data.drop` is preceded by a data count section. If I remove this check here the spec tests all look like they pass as well, so is the change here unnecessary?

alexcrichton: And to clarify, the reason I ask is that threading this around is pretty cumbersome it looks like so I'd prefer to remove it if possible and leave it purely to validation to figure this out

crates/wasmparser/src/validator/core.rs
```diff
@@ -394,7 +306,7 @@ impl ModuleState {
             // (aka a fuzz bug) if we somehow forget to emit an error somewhere
             // else.
             fn insert_ref_func(&mut self, index: u32) {
-                if self.order == Order::Data {
+                if self.resources.module.as_mut().is_none() {
```
alexcrichton, line 309: I'm a bit wary of coupling the order of traversal to `as_mut`, mind maintaining a boolean flag of sorts in the validator to avoid needing to lean on `as_mut`?

crates/wasmprinter/src/lib.rs
```diff
@@ -1385,32 +1396,41 @@ impl Printer<'_, '_> {
...
-            self.newline(body.original_position())?;
+            self.newline(reader.original_position())?;
         }
 
         Ok(())
     }
 
     fn print_operators<'a, O: OpPrinter>(
         body: &mut BinaryReader<'a>,
+        data_index_allowed: bool,
```
alexcrichton, line 1407: re my question above about this as well - this is an example of where it's a bit cumbersome to manage this and I'd prefer to remove this if possible, but if it's required by tests I mostly just want to understand in more detail what's requiring it.

alexcrichton (commented):
Thank you again for putting this all together, I'm sure it took a lot of back-and-forth with the test suite!

I've got some questions on particulars below, but I'm also realizing now at this point after going through this that the changes to `src/bin/wasm-tools/wast.rs` aren't here so my comments below about "tests still pass" may be less relevant. I assume though that the changes to `src/bin/wasm-tools/wast.rs` are relatively small, do you think they'd be reasonable to fold into this PR as well?

Otherwise, at a high-level, I do very much prefer this as a solution than #2123. I think it's worth pushing on this instead and trying to find a balance between validation/parsing and checks and such.

crates/wasm-shrink/tests/tests.rs
alexcrichton: Ah ok with #2135 now merged if you rebase this PR then this change should no longer be necessary

tests/cli/missing-features/missing-exceptions.wast
```diff
@@ -1,12 +1,5 @@
 ;; RUN: wast --assert default --snapshot tests/snapshots % -f wasm1
 
-(assert_malformed
-  (module binary
-    "\00asm" "\01\00\00\00"
-    "\0d\01\00"
-    )
-  "exceptions proposal not enabled")
```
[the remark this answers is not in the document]
keithw: Yes (with the other changes discussed below) -- pushed.

crates/wasm-shrink/tests/tests.rs
keithw: Done!

crates/wasmparser/src/readers/core/memories.rs
keithw: Got it. I was able to move two out of the three -- the `shared` check is currently required by the spec `binary.wast` test but can get moved to the validator once the wasm-3.0 branch is merged to main in the spec repo.

crates/wasmparser/src/readers/core/memories.rs
keithw: Yeah, this is required by an upstream test: https://github.com/WebAssembly/custom-page-sizes/blob/edb9c5b5/test/core/custom-page-sizes/custom-page-sizes-invalid.wast#L90 .

I think it needs to be malformed because otherwise there would be no way to represent bigger values in the text format (the text format has the $2^n$ page size and the binary format just stores $n$).

crates/wasmparser/src/readers/core/tags.rs
keithw: Done.

crates/wasmparser/src/readers/core/operators.rs
keithw: I found this to be definitely the most cumbersome part of the binary format spec. :-( The tests that require this are at https://github.com/WebAssembly/spec/blob/05949f5/test/core/binary.wast#L492-L533 (which is probably clearer now that I pushed a wast.rs that runs the `assert_malformed` tests through only the parser -- you'll see the failure if you run it now). The textual part of the binary format spec for this is https://webassembly.github.io/spec/core/binary/modules.html#binary-module :
> if $m^{?} \ne \epsilon \lor \texttt{dataidx}(\textit{code}^n) = \emptyset $

I assume the reason it works this way is that in the land of the spec, the validator runs on the abstract syntax (not the text or binary formats), and "the data count section disagreed with the data section about the number of data segments" or "the data count section was missing" aren't expressible in the abstract syntax.

I tried a few different ways to handle this (e.g. trying to keep state in the BinaryReader about whether a data index was currently allowed or not, which meant being careful to always reset that state when entering a new sub-parser which I wasn't confident I could get safely, etc.) but this seemed to be the least-bad way in the end. :-/

crates/wasmparser/src/validator/core.rs
keithw: Ai, I've now gone back and forth on this a few times and after some consideration I honestly think this is the right way to go. I think this is really the appropriate coupling in the end -- we really **do** want to insert the function index as referenceable (no matter what segment we're in), and the only reason we're not doing it is that the module is no longer mutable.

Of course we do have side knowledge that a `ref.func` as part of a data segment offset expression is not going to be valid (so it doesn't really have to be inserted), but that's for a pretty narrow reason -- because `drop` isn't allowed in an extended-const expression, there can't be a `ref.func` in an i32-typed `offset` expression. Which is why this is just an assertion that never gets hit because it's a validation error before that. But I think the real thing that matters here is the module mutability, not the `Order::Data`.

If you don't buy it, happy to make the change but wanted to give it a shot. :-)

crates/wasmparser/src/readers/core/operators.rs
alexcrichton: Oof ok I see what you mean, I was looking in the wrong spot for the binary validation...

For this issue specifically, what do you think about diverging from the spec on this? My impression is that the spec is viewed by most authors/engine implementors as a guideline but engines aren't expected to match 1:1 with all the nuances of the spec in cases like this (e.g. precisely where an error shows up, precisely an error message, etc). In that sense I'm hesitant to take design principles from one engine, the spec interpreter, and force that to guide design decisions of other engines (e.g. wasm-tools/wasmtime/etc). More-or-less, I don't think it's worth the baggage necessary to implement this one rule.

That being said I also at the same time would not want to champion a change in the spec. I fear pushback along the lines of "well just don't do that in your engine and reject it in the validator", or resistance along the lines of that. Given that we nonetheless have to handle this test somehow with the change to `assert_malformed`. WDYT of special-casing this error message and, in the case of this exact match, assert that the binary parsing is valid while the binary validation fails with the error message? That removes the need for all the infrastructure plumbing around whether a data index is allowed and keeps the "hack" somewhat scoped to just `wast.rs` file.

crates/wasmprinter/src/lib.rs
keithw: Understood -- discussed at https://github.com/bytecodealliance/wasm-tools/pull/2134#discussion_r2038328446

keithw:
> I've got some questions on particulars below, but I'm also realizing now at this point after going through this that the changes to `src/bin/wasm-tools/wast.rs` aren't here so my comments below about "tests still pass" may be less relevant. I assume though that the changes to `src/bin/wasm-tools/wast.rs` are relatively small, do you think they'd be reasonable to fold into this PR as well?

Yes, done! I just added the parts that run the `assert_malformed` tests through the parser alone -- will hold back the roundtripping of invalid modules for a later PR (this will require some more changes to wasmprinter/wasm-encoder).

crates/wasmparser/src/readers/core/operators.rs
keithw: Fine with me -- just done in https://github.com/bytecodealliance/wasm-tools/pull/2134/commits/5be6fcd693fd4e14dc372b8320757664c0eb4070

keithw: I guess, now that I've done this, we probably could use the same trick to move the "shared" feature check out of the reader too if you want.

alexcrichton (approved):
I've pushed up a few commits to this myself, notably taking you up on your suggestion to use permissive assertions for the threads/shared-everything-threads related flags as that I think looks pretty good in the end. I did a few other misc changes here and there as well.

One final comment about the default input to the CLI tooling, as to why the parse is needed, but otherwise looks good to me 👍

crates/wasmparser/src/readers/core/operators.rs
rossberg: FWIW, there is no need for implementations to distinguish "malformed" and "invalid" at all — and most engines fuse decoding and validation anyway. For all practical purposes, they have the exact same effect anyway. (Well, with one caveat, which is lazy function validation, but I don't think any current engine does that anymore.)

Technically, the two cases differentiate illegal programs that are still inter-convertible between binary and text format (but are semantically invalid) from those that do not even have a representation in the other (so are syntactically malformed). So, @keithw is correct about why datacount is handled the way it is. Not that I like that much either, but datacount is a terrible hack either way.

alexcrichton:
I'm looking to do a release in the next day or so, so I'm going to go ahead and merge this. Thanks again for your work here @keithw!

keithw:
Okay! I think we may have a challenge with these extra commits trying to roundtrip everything that is assert_invalid -- the problem is that I think some of these error messages (like "integer too large") really *are* caught by the wasmparser, so turning them into AssertInvalid then fails if we try to assert that everything that's AssertInvalid can be parsed and roundtripped through wasmprinter. But I will deal with that in the next PR...

PR #2137, opened 2025-04-10
add `subtask.cancel` intrinsic

dicej:
This has not yet been added to the spec, but we've got a general idea of how it will look and behave: https://github.com/WebAssembly/component-model/issues/495

Note that, since there's not spec for it yet, I've picked the next unused binary discriminant in the sequence: 0x24.  That should be easy enough to change if necessary once the spec is updated.


alexcrichton:
Mind adding some tests for this as well? For example wit-component tests that the import gets hooked up in addition to `*.wast` tests that it validates and has the right signature and such

alexcrichton (approved):

PR #2138, opened 2025-04-10
Require `[..]l.l` is not present with `l`

alexcrichton:
This updates validation of component model names to account for WebAssembly/component-model#494. This technically affects all existing users as well but it's hoped that it won't actually hit any cases of this in the wild.


jsturtevant:
LGTM, I am not super familiar with this code but changes look sane to me

pchickey (approved):

PR #2143, opened 2025-04-15
Prevent over-deep type hierarchies in wasm-smith

alexcrichton:
Currently wasm-smith can produce invalid GC modules because there's no limit placed on subtyping depth. This commit keeps track of subtyping depth and appropriately manages the `can_subtype` list as a result.

### DIFF
--- a/crates/wasm-smith/src/core.rs
+++ b/crates/wasm-smith/src/core.rs
@@ -592,7 +592,24 @@ impl Module {
         list.push(index);
 
         if !ty.is_final {
-            self.can_subtype.push(index);
+            // Calculate the recursive depth of this type, and if it's beneath a
+            // threshold then allow future types to subtype this one. Otherwise
+            // this can no longer be subtyped so despite this not being final
+            // don't add it to the `can_subtype` list.
+            //
+            // Note that this limit is intentinally a bit less than the
+            // wasm-defined maximum of 63.
+            const MAX_SUBTYPING_DEPTH: u32 = 60;
+
+            let mut depth = 1;
+            let mut cur = ty.supertype;
+            while let Some(parent) = cur {
+                depth += 1;
+                cur = self.types[parent as usize].supertype;
+            }
+            if depth < MAX_SUBTYPING_DEPTH {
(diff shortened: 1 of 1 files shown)


alexcrichton:
I don't know why it took fuzzing in Wasmtime to find this.

alexcrichton:
Well I'm flummoxed. This is fixing a preexisting bug which by all means should have been discovered by fuzzing of just this repo, fuzzing in Wasmtime shouldn't be necessary. Ever since #1793, landed 6 months ago, this should have been discoverable by the validate-valid-module fuzzer.

I double-checked all configuration options Wasmtime is using/setting are at least possible to configure from this fuzzer. Given that I guess it just requires a relatively specific shape of fuzz input and that didn't crop up until later...

fitzgen (commented):
Rather than calculating this on-demand, and potentially repeating the traversal up the parent chain which could lead to quadratic behavior or worse, I think we want to have a side table and compute the subtyping depth for each type as we generate each type.

More concretely, I'm thinking we want to have a side table similar to [`super_to_sub_types`](https://github.com/bytecodealliance/wasm-tools/blob/aa2e152b6fab38092720187924c29036e77fde39/crates/wasm-smith/src/core.rs#L60-L61) that maps from a type to its subtyping depth, and we eagerly insert an entry in this table when generating a type. This way, we only ever have to add one to the super type's depth, rather than calculate the whole depth and traverse the whole hierarchy every time.

fitzgen:
> I double-checked all configuration options Wasmtime is using/setting are at least possible to configure from this fuzzer. Given that I guess it just requires a relatively specific shape of fuzz input and that didn't crop up until later...

Huh. I don't have a good answer for this mystery either, other than we just got "lucky" in Wasmtime, or maybe it was somehow rewarding the fuzzer with additional coverage for deeper subtyping in a way that just the parser and validator was not.

alexcrichton:
Given the maximum recursion depth here is fixed I figured there was no need for quadratic bits, but sure it's not too bad to do that too.

fitzgen (approved):

PR #2146, opened 2025-04-15
For `assert_invalid`, assert modules all parse

alexcrichton:
Sort of the inverse of `assert_malformed`, but assert that invalid modules all parse successfully and just fail validation.

### DIFF
--- a/src/bin/wasm-tools/wast.rs
+++ b/src/bin/wasm-tools/wast.rs
@@ -198,6 +198,30 @@ impl Opts {
                 mut module,
                 message,
             } => {
+                // For `assert_malformed` it means that either converting this
+                // test case to binary should fail, or that parsing the binary
+                // should fail. Currently there's no distinction as to which
+                // step should fail.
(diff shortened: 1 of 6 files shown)


src/bin/wasm-tools/wast.rs
```diff
@@ -198,6 +198,30 @@ impl Opts {
...
+                        return Ok(());
+                    }
+                };
+                let mut parser = wasmparser::Parser::new(0);
+                parser.set_features(self.features.features());
+                if let Err(e) = parse_binary_wasm(parser, &bytes) {
+                    self.assert_error_matches(test, &format!("{e:?}"), message)?;
+
+                    // make sure validator also rejects module (not necessarily
+                    // with same error)
```
fitzgen, line 218:
```suggestion
                    // Make sure validator also rejects the module (not necessarily
                    // with same error).
```

fitzgen (approved):

PR #2153, opened 2025-04-22
Fix typos

omahs:
Fix typos


abrown (approved):

pchickey:
I believe those wit fixes are already upstream and the right fix is to update the tests to a later release. Either way, theyre causing a CI failure, so they need to be backed out from this PR.

PR #2154, opened 2025-04-22
Add a few doc comments around validation, and fix some typos.

### LINKED ISSUE #2063: Tolerate duplicate imports when creating components
Multiple imports of the same name are in-spec for core-wasm modules but [not allowed for components](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#import-and-export-definitions). Currently, attempting to componentize a module with non-uniquely named imports yields…

```
% wasm-tools component new main-with-dupe-imports.wasm -o component-with-dupes.wasm --adapt wasi_snapshot_preview1.wasm
error: failed to encode a component from module

Caused by:
    1: module was not valid
    2: module has duplicate import for `wasi_snapshot_preview1::fd_write`
```

We should pick a unique import of each name and ignore the rest when componentizing. Since inter-component linking is done solely by name, multiple same-named imports have no way of being distinguished in order to be satisfied individually—even if they had unique types, for example—and so nothing is lost by ignoring them.

erikrose:
Breaking these free of an old attempt to fix #2063.

### DIFF
--- a/crates/wasmparser/src/validator.rs
+++ b/crates/wasmparser/src/validator.rs
@@ -346,7 +346,7 @@ pub enum ValidPayload<'a> {
     /// This result indicates that the specified parser should be used instead
     /// of the currently-used parser until this returned one ends.
     Parser(Parser),
-    /// A function was found to be validate.
+    /// A function was found to be validated.
     Func(FuncToValidate<ValidatorResources>, FunctionBody<'a>),
     /// The end payload was validated and the types known to the validator
     /// are provided.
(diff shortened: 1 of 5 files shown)


crates/wasmparser/src/validator.rs
```diff
@@ -346,7 +346,7 @@ pub enum ValidPayload<'a> {
     /// This result indicates that the specified parser should be used instead
     /// of the currently-used parser until this returned one ends.
     Parser(Parser),
-    /// A function was found to be validate.
+    /// A function was found to be valid.
```
alexcrichton, line 349: Oh "validate" here should have been "validated", the function isn't valid yet in this variant and it still needs to be validated.

alexcrichton (approved):

crates/wasmparser/src/validator.rs
erikrose: Fixed and push -f'd!

PR #2159, opened 2025-04-25
Drop stability for external package includes

### LINKED ISSUE #2113: `Includes` in Resolved json are copying additional information in the resolved world.
when WIT is recreated from the resolved JSON, it has versioned feature gates on the unversioned `wasmtime:test` package.

Given:
```
package wasmtime:test;

world test{
    @unstable(feature = active)
    include wasi:unstable/imports@0.2.3; 
    include wasi:foo/imports@0.2.3;
} 

world test-ordered{
    include wasi:foo/imports@0.2.3;
    @unstable(feature = active)
    include wasi:unstable/imports@0.2.3; 
} 

package wasi:unstable@0.2.3 {
    @unstable(feature = active)
    world imports {
        @unstable(feature = active)
        use wasi:dep2/stable@0.2.3.{stable-resource};
        @unstable(feature = active)
        use wasi:dep-unversioned/unversioned.{unversioned-resource};
        @unstable(feature = active)
        use wasi:dep-unstable/unstable.{unstable-resource};
    }
}

package wasi:foo@0.2.3 {
    @since(version = 0.2.0)
    world imports {
        @since(version = 0.2.0)
        include wasi:dep2/imports@0.2.3;
        include wasi:dep-unversioned/imports;
        include wasi:dep-unstable/imports;
    }
}

package wasi:dep2@0.2.3 {
    @since(version = 0.2.0)
    world imports {
        @since(version = 0.2.0)
        import stable;
    }
    @since(version = 0.2.1)
    interface stable {
        resource stable-resource {
        }
    }
}

package wasi:dep-unversioned{
    world imports {
        import unversioned;
    }
    interface unversioned {
        resource unversioned-resource {

        }
    }
}

package wasi:dep-unstable{
    @unstable(feature = active)
    world imports {
        @unstable(feature = active)
        import unstable;
    }
    @unstable(feature = active)
    interface unstable {
        @unstable(feature = active)
        resource unstable-resource {

        }
    }
}

```

the world test should not have the `@since`
```
@since(version = 0.2.0)
import wasi:dep2/stable@0.2.3;``` when regenerated:
```

the full resolved wit from the json:

```wit
package wasmtime:test;

world test {
	@since(version = 0.2.0)
	import wasi:dep2/stable@0.2.3;
	@unstable(feature = active)
	use wasi:dep2/stable@0.2.3.{stable-resource};
	@unstable(feature = active)
	import wasi:dep-unversioned/unversioned;
	@unstable(feature = active)
	use wasi:dep-unversioned/unversioned.{unversioned-resource};
	@unstable(feature = active)
	import wasi:dep-unstable/unstable;
	@unstable(feature = active)
	use wasi:dep-unstable/unstable.{unstable-resource};
}

world test-ordered {
	@since(version = 0.2.0)
	import wasi:dep2/stable@0.2.3;
	@unstable(feature = active)
	import wasi:dep-unversioned/unversioned;
	@unstable(feature = active)
	import wasi:dep-unstable/unstable;
	@unstable(feature = active)
	use wasi:dep2/stable@0.2.3.{stable-resource};
	@unstable(feature = active)
	use wasi:dep-unversioned/unversioned.{unversioned-resource};
	@unstable(feature = active)
	use wasi:dep-unstable/unstable.{unstable-resource};
}

package wasi:foo@0.2.3 {
	@since(version = 0.2.0)
	world imports {
		@since(version = 0.2.0)
		import wasi:dep2/stable@0.2.3;
		import wasi:dep-unversioned/unversioned;
		@unstable(feature = active)
		import wasi:dep-unstable/unstable;
	}
}

package wasi:unstable@0.2.3 {
	@unstable(feature = active)
	world imports {
		@unstable(feature = active)
		import wasi:dep2/stable@0.2.3;
		@unstable(feature = active)
		use wasi:dep2/stable@0.2.3.{stable-resource};
		@unstable(feature = active)
		import wasi:dep-unversioned/unversioned;
		@unstable(feature = active)
		use wasi:dep-unversioned/unversioned.{unversioned-resource};
		@unstable(feature = active)
		import wasi:dep-unstable/unstable;
		@unstable(feature = active)
		use wasi:dep-unstable/unstable.{unstable-resource};
	}
}

package wasi:dep-unstable {
	@unstable(feature = active)
	interface unstable {
		@unstable(feature = active)
		resource unstable-resource;
	}

	@unstable(feature = active)
	world imports {
		@unstable(feature = active)
		import unstable;
	}
}

package wasi:dep-unversioned {
	interface unversioned {
		resource unversioned-resource;
	}

	world imports {
		import unversioned;
	}
}

package wasi:dep2@0.2.3 {
	@since(version = 0.2.1)
	interface stable {
		resource stable-resource;
	}

	@since(version = 0.2.0)
	world imports {
		@since(version = 0.2.0)
		import stable;
	}
}
```

_Originally posted by @ydnar in https://github.com/bytecodealliance/wasm-tools/issues/2076#issuecomment-2745897580_
 

A round trip test would help catch this.

jsturtevant:
fixes #2113

### DIFF
--- a/crates/wit-component/tests/merge/success/merge/foo.wit
+++ b/crates/wit-component/tests/merge/success/merge/foo.wit
@@ -65,9 +65,13 @@ world shared-world-with-versions {
   export shared-version-on-into;
 }
 world shared-world-with-versions-and-include {
+  @since(version = 1.0.0)
   import shared-version-on-from;
+  @since(version = 1.0.0)
   import shared-version-on-into;
 
+  @since(version = 1.0.0)
   export shared-version-on-from;
+  @since(version = 1.0.0)
   export shared-version-on-into;
 }
--- a/crates/wit-component/tests/merge/success/merge/foo.wit
+++ b/crates/wit-component/tests/merge/success/merge/foo.wit
@@ -65,13 +65,9 @@ world shared-world-with-versions {
   export shared-version-on-into;
 }
 world shared-world-with-versions-and-include {
-  @since(version = 1.0.0)
   import shared-version-on-from;
-  @since(version = 1.0.0)
   import shared-version-on-into;
 
-  @since(version = 1.0.0)
   export shared-version-on-from;
-  @since(version = 1.0.0)
   export shared-version-on-into;
 }
--- a/crates/wit-parser/src/lib.rs
+++ b/crates/wit-parser/src/lib.rs
@@ -1243,6 +1243,10 @@ impl Stability {
     pub fn is_unknown(&self) -> bool {
         matches!(self, Stability::Unknown)
     }
+
+    pub fn is_stable(&self) -> bool {
+        matches!(self, Stability::Stable { .. })
+    }
 }
 
 impl Default for Stability {
--- a/crates/wit-parser/src/resolve.rs
+++ b/crates/wit-parser/src/resolve.rs
@@ -7,6 +7,7 @@ use std::path::{Path, PathBuf};
 
 use anyhow::{anyhow, bail, Context, Result};
 use id_arena::{Arena, Id};
+use indexmap::map::Entry;
 use indexmap::{IndexMap, IndexSet};
 use semver::Version;
 #[cfg(feature = "serde")]
@@ -3404,13 +3405,13 @@ impl Remap {
         assert_eq!(world.includes.len(), spans.includes.len());
         let includes = mem::take(&mut world.includes);
         let include_names = mem::take(&mut world.include_names);
-        for (((stability, include_world), span), names) in includes
+        for (((include_stability, include_world), span), names) in includes
             .into_iter()
             .zip(&spans.includes)
             .zip(&include_names)
         {
             if !resolve
-                .include_stability(&stability, pkg_id, Some(*span))
+                .include_stability(&include_stability, pkg_id, Some(*span))
                 .with_context(|| {
                     format!(
                         "failed to process feature gate for included world [{}] in package [{}]",
@@ -3421,7 +3422,14 @@ impl Remap {
             {
                 continue;
             }
-            self.resolve_include(world, include_world, names, *span, resolve)?;
+            self.resolve_include(
+                world,
+                include_world,
+                names,
+                *span,
+                resolve,
+                &include_stability,
+            )?;
         }
 
         Ok(())
@@ -3444,6 +3452,7 @@ impl Remap {
         names: &[IncludeName],
         span: Span,
         resolve: &Resolve,
+        include_stability: &Stability,
     ) -> Result<()> {
         let include_world_id = self.map_world(include_world, Some(span))?;
         let include_world = &resolve.worlds[include_world_id];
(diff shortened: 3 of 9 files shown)


alexcrichton:
I'm trying to think through the implications of this in terms of bindgen, and I'm wondering if we perhaps want a different merging algorithm related to `unstable`. My worry here is that we could have a chain of events like:

* `wasi:tls/imports` has `@unstable import wasi:tls/types;`
* `wasi:cli/imports` has `@unstable include wasi:tls/imports;`
* `wasi:cli/run` has `@stable include wasi:cli/imports`

I think that the instability will get propagated to `wasi:cli/imports`, but I think that the instability won't get propagated to `wasi:cli/run`? Wouldn't the `@stable` tag there override the `@unstable` of the world item for `wasi:tls/types`?

Basically I'm worried about accidentally flagging something originally marked `@unstable` as `@stable` as it transitions between worlds and `include`. One possible solution might be to reorder the `Stable` and `Unstable` variants so unstable is "greater", but that may cause havoc elsewhere

jsturtevant:
I think the individual items would get upgraded.  Do I understand correctly that `include` and `use` also do different things in the types system.  (ie. `use` is a reference and include would redefine it (unless it then calls a use))

I found there are all sorts of edge cases here based on what `use`/`included` first. And gets tricky when there are nested `includes` too.  This took the simple approach of overriding everything of with the value on the include.

I was re-reading the spec and the implications of `using`/`including` to different worlds means that we break one of the ["rules" of features gates](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#rules-for-feature-gate-usage): `Either @since or @unstable should be used, but not both (exclusive or).`  When we add setting `feature-gates` on the `include` or `use` statement I think we further add ambiguity and complexity to solving for the correct version: i.e. Does that apply to all sub items?

Maybe we need to detail this out a bit more in the design/spec?  I can think of a couple possibilities:
- either `includes` can't have `since`/`unstable`  and the underlying stability info comes from the items
- `since`/`unstable`  can only determine if the `include`/`use` are in scope of current package but the stability of the underlying items that are being included are the ones that get applied.  
- must match the stabilities of individual items across all imports/includes (this wouldn't work in practice i think)

alexcrichton:
You're right yeah that `include` and `use` are different, and you're also right that the spec is pretty amibugous about what to do here. One of the main differences is that `include` attempts to inline one world into another, moving items around, where `use` just refers to something somewhere else, and nothing is moved. In that sense stability makes a lot more sense with `use` and less so with `include`.

Even if we disallow stability on `include` though I think we have the same issues. For example `include wasi:cli/imports` will transitively refer to `import wasi:tls/types` which is unstable, and somehow that should be resolved. If it's unstable it's ok to copy over the instability, but if it's stable it's not valid to copy over the stable information since the version of the current package could be completely different (and IIRC was one of the impetuses for this change)

How about a solution along the lines of:

* intra-package `include` (e.g. one world in a package include-ing a world in the same package) requires no stability on the `include` and the annotation of each item is just copied over from the source, failing on conflict if something was already in the `world`. This is IIRC mostly the original behavior of `include`
* inter-package `include` would also require no stability on `include` and would discard `@stable` annotations while preserving `@unstable` annotations, and would probably return an error on clash with an explicit import.

That way the stability of `include` is never considered and unstable things at least get propagated as unstable while stable things just sort of lose their annotation which shouldn't be the end of the world

alexcrichton (approved):
Oh wow that was a pretty small implementation! Was there other tests you wanted to add or other examples? Otherwise this looks good to go to me

jsturtevant:
I saw you added some support to `wit-smith` for includes.  Does it make sense to add a simple round trip to the tests in wit-parser as an initial test of the samples or do you think the fuzzing tests running later is enough to catch issues?

alexcrichton:
I did add support yeah but `wit-smith` still has no support for stability annotations, so while `include` is now there it still doesn't use `@stable` or `@unstable` at all, so fuzzing won't catch this.

Fuzzing does though do round-tripping so once it generates all the things here it should catch this, in theory.

alexcrichton:
I'm going to go ahead and land this, and if necessary can follow-up with extra tests/fuzzing.

Issue #2110, opened 2025-03-21
`wasm-smith` does not support `--available-imports` with GC types


mobsceneZ:
Hi developers, I try to leverage **wasm-smith** to generate random Wasm modules with the specified imports and exports. However, when passing the following Wasm module to the `--available-imports` option, **wasm-smith (commit 43556041e)** panicked: 
> *thread 'main' panicked at crates/wasm-smith/src/core.rs:1309:37:
could not parse type section: BinaryReaderError { inner: BinaryReaderErrorInner { message: "gc proposal not supported", kind: Custom, offset: 11, needed_hint: None } }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace*

```wat
(module
  (rec
    (type (sub final (array (mut i32))))
    (type (sub final (func (param i32) (result i32))))
  )
  (import "a" "f1" (func (type 1)))
  (import "a" "m1" (memory (;0;) 43153 57698))
)
```

The provided module itself is **valid**, thus I manually checked out the panicked location. As it turns out, the `into_iter_err_on_gc_types()` public  function of the **wasmparser** crate assumes each recursive group contains exactly one function type, which IMO makes no sense for **wasm-smith** since *GC* proposal is now standardized:
https://github.com/bytecodealliance/wasm-tools/blob/43556041e072465c4a658da3d5b3ee7393211cc9/crates/wasmparser/src/readers/core/types.rs#L1952-L1973

I think maybe we should update the related code logic for **wasm-smith** crate to better support the *GC* proposal? If you think this is appropriate but don't have time to implement these changes yet, I can submit a PR.

mobsceneZ:
Besides, it looks like the support for generating exports for `tag`, `memory`, and `table` is still missing, maybe **wasm-smith** can directly copy the corresponding elements from the user-provided module 🤔?

alexcrichton:
Thanks! This is a "correct panic" in the sense that it's correctly indicating that this isn't implemented yet, but otherwise is also very much indeed a bug and worth fixing. I believe what's necessary here is that the original module's types will have to be reflected into the `wasm_smith::Module` being constructed, so sort of converting wasmparser to wasm-smith/wasm-encoder. That shouldn't be too too hard but will likely be a bit nontrivial.

If you're up for sending a PR that'd be most welcome! And supporting other types like tag/memory/table makes sense to me too

mobsceneZ:
[...]

PR #2160, opened 2025-04-28
wasm-smith: add support for `--exports` and `--available-imports` option

mobsceneZ:
This PR fixes issue https://github.com/bytecodealliance/wasm-tools/issues/2110 (sorry for the lateness):

1. It allows wasm-smith to generate the corresponding tag, table, and memory exports as specified by the `--exports` option. Note that I have also modified the function signatures of **arbitrary_const_expr** and **globals_for_const_expr** and added a new parameter called *allow_defined_globals*. This is because the **required_exports** is called after **arbitrary_globals** in the **build** function, which results in wasm-smith potentially *picking up defined globals for table initialization expressions* and that is not allowed in the current Wasm standard;
2. It allows Wasm modules with GC types to appear in the `--available-imports` option. To achieve this, the current solution is to copy all the recursive groups in the provided module into the module being constructed. And the necessary trait implementation is also added (i.e., `From` trait);

Please feel free to share any suggestions/concerns you might have :).

### DIFF
--- a/crates/wasm-smith/src/core.rs
+++ b/crates/wasm-smith/src/core.rs
@@ -1656,9 +1656,7 @@ impl Module {
                     return Ok(false);
                 }
                 let ty = arbitrary_table_type(u, self.config(), Some(self))?;
-                let init = self.arbitrary_table_init(u, ty.element_type)?;
-                self.defined_tables.push(init);
-                self.tables.push(ty);
+                self.add_arbitrary_table_of_type(ty, u)?;
                 Ok(true)
             },
         )
@@ -1683,7 +1681,9 @@ impl Module {
         if ty.nullable && u.arbitrary()? {
             return Ok(None);
         }
-        let expr = self.arbitrary_const_expr(ValType::Ref(ty), u)?;
+        // Only imported globals are allowed in the constant initialization
+        // expressions for tables.
+        let expr = self.arbitrary_const_expr(ValType::Ref(ty), u, false)?;
         Ok(Some(expr))
     }
 
@@ -1696,8 +1696,8 @@ impl Module {
                 if !self.can_add_local_or_import_memory() {
                     return Ok(false);
                 }
-                self.num_defined_memories += 1;
-                self.memories.push(arbitrary_memtype(u, self.config())?);
+                let ty = arbitrary_memtype(u, self.config())?;
+                self.add_arbitrary_memory_of_type(ty)?;
                 Ok(true)
             },
         )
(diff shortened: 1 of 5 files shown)


crates/wasm-smith/src/core.rs
```diff
@@ -3111,3 +3168,68 @@ impl FromStr for InstructionKind {
...
+
+        CompositeType {
+            inner: inner_type,
+            shared: value.shared,
+        }
+    }
+}
+
+#[cfg(feature = "wasmparser")]
+impl From<wasmparser::SubType> for SubType {
```
alexcrichton, line 3220: Could these impls be modeled as `TryFrom` to avoid the internal `.unwrap()`s?

crates/wasm-smith/src/core.rs
```diff
@@ -1980,13 +1981,65 @@ impl Module {
...
+                                func_type: new_type,
+                            });
+                            self.num_defined_tags += 1;
+                            tag_index
+                        }
+                        _ => panic!(
+                            "Unable to handle type {:?} from `exports` Wasm",
+                            subtype.composite_type
+                        ),
+                    }
```
alexcrichton, line 2042: This looks like a large copy/paste of what functions are doing above, so could this be refactored to not duplicate the logic?

crates/wasm-smith/src/core.rs
```diff
@@ -1683,7 +1660,9 @@ impl Module {
         if ty.nullable && u.arbitrary()? {
             return Ok(None);
         }
-        let expr = self.arbitrary_const_expr(ValType::Ref(ty), u)?;
+        // Only imported globals are allowed in the constant initialization
+        // expressions for tables.
+        let expr = self.arbitrary_const_expr(ValType::Ref(ty), u, false)?;
```
alexcrichton, line 1665: I suspect this `bool` argument to `arbitrary_const_expr` was added due to tests failing at some point? I'm curious why though because `arbitrary_tables`, which calls this function, is called before globals are generated so I'm not sure how this would otherwise refer to a defined global instead of an imported global.

Could you expand on what led you to adding this `bool` argument?

alexcrichton: My hunch is that this was added due to `required_exports` being called after all items, and in that case I wonder if it might make sense to split apart the `required_exports` function to being called at the start of section generation? For example when generating tables the first thing that would be done is that any required tables, because of exports, would be generated. After that though more arbitrary tables might be generated.

alexcrichton (commented):
Thanks for your work on this!

crates/wasm-smith/src/core.rs
mobsceneZ: > My hunch is that this was added due to `required_exports` being called after all items, and in that case I wonder if it might make sense to split apart the `required_exports` function to being called at the start of section generation? For example when generating tables the first thing that would be done is that any required tables, because of exports, would be generated. After that though more arbitrary tables might be generated.

Yes, that's exactly the reason why I'm adding the `bool` argument, otherwise the `required_exports` function may pick up defined globals for table initialization expressions. Spliting apart the `required_exports` function to being called in each section generation function might be feasible, but intuitively, this would require:

1. Either duplicating the module parsing logic within each section generation function, which definitely introduces unnecessary time cost;
2. Or parsing the module in advance in the `build` function and storing relevant information in the `Module` structure for subsequent use, still, this demands additional bookkeeping cost;

Thus, I personally believe that introducing a boolean argument is the simplest solution, as the resulting code changes and side effects are both clear and straightforward to manage. Of course, if you insist, I think the spliting approach could be worth a try.

crates/wasm-smith/src/core.rs
mobsceneZ: > Could these impls be modeled as `TryFrom` to avoid the internal `.unwrap()`s?

The `TryFrom` trait requires defining an associated type `Error`, do you think it would be acceptable to reuse `wasm_encoder::reencode::Error` in wasm-smith for this purpose?

alexcrichton: Since these are mostly internal I think it's fine to use something like `()` for the error, my hope would be to basically push the `.unwrap()` to the location where all the other `.unwrap()`s are happening to avoid future code relying on `From::from` here without realizing there's unwraps internally

crates/wasm-smith/src/core.rs
alexcrichton: Ok that seems reasonable yeah, let's leave it as-is. I'll look into possibly refactoring this as a follow-up since I'm interested but I agree that amount of work is larger-than-expected so doesn't need to happen here

mobsceneZ:
Hey @alexcrichton, I have made the necessary code changes in https://github.com/bytecodealliance/wasm-tools/pull/2160/commits/a43b9828aeb6be2e1b6b3242619d94307f2fdcf2, and you can check it out to see if any further changes are required :).

alexcrichton (approved):

PR #2161, opened 2025-04-28
Add support for `include`  to `wit-smith`

alexcrichton:
This commit updates the fuzz-generation of WIT documents to use `include` in worlds. Getting this implemented was surprisingly tricky and required some careful balancing of various constructs in generation in addition to support for round-tripping.

My hope is that this makes it easier to next add support for stability in fuzzing to enable the fuzzers to find https://github.com/bytecodealliance/wasm-tools/issues/2113


pchickey (approved):
Looks like some more reordered test output needs to be blessed, otherwise good?

PR #2163, opened 2025-04-28
wasmparser: Refactor canonical option validation and func type lowering

fitzgen:
Rather than repeatedly searching the list of canonical options, create a `struct CanonicalOptions` that summarizes their meaning when taken altogether. This `struct` is returned from `check_options`, where some initial, light validation occurs. Then, additional methods like `options.require_memory()?` can be called on the resulting `struct` depending on the use site (e.g. lifting vs. lowering vs. `task.return` vs. etc...).

This also makes it so that `check_options` does not require a `LoweringInfo` as a parameter. That presupposes that we already know the core function signature we will be lowering into. That will no longer be true with the combination of Wasm GC and the component model, where there will be a canonical option to specify the desired core function type, and this controls whether GC is used or not, among other things.

Finally, the `ComponentFuncType::lower` method is now invoked after canonical options are checked and takes a `CanonicalOptions` argument. Rather than setting `info.requires_memory = true` when it determines that a memory is required (for example) which was then validated in `check_options` later on, it has become fallible and calls `options.require_memory()?` itself to validate that a memory is declared. Again, this fallibility will be required for Wasm GC and component model integration in the future.

### DIFF
--- a/crates/wasmparser/src/validator/component.rs
+++ b/crates/wasmparser/src/validator/component.rs
@@ -2376,8 +2376,8 @@ impl ComponentState {
             (true, _, true) => {
                 bail!(offset, "cannot specify post-return function in async")
             }
-            (false, _, _) => Concurrency::Sync,
-            (true, callback, _) => Concurrency::Async { callback },
+            (false, None, _) => Concurrency::Sync,
+            (true, callback, false) => Concurrency::Async { callback },
         };
 
         Ok(CanonicalOptions {


crates/wasmparser/src/validator/component.rs
```diff
@@ -2266,62 +2369,23 @@ impl ComponentState {
...
-                if !self.features.cm_async_stackful() {
-                    bail!(
-                        offset,
-                        "`async` without `callback` requires the async stackful feature"
-                    )
-                }
+            (true, _, true) => {
+                bail!(offset, "cannot specify post-return function in async")
             }
+            (false, _, _) => Concurrency::Sync,
```
alexcrichton, line 2379: Can this bind `false, None, _` to guarantee the callback isn't provided at-a-glance?

crates/wasmparser/src/validator/component.rs
```diff
@@ -2266,62 +2369,23 @@ impl ComponentState {
...
+            (true, callback, _) => Concurrency::Async { callback },
```
alexcrichton, line 2380: Like above, can this bind `true, callback, false` to at-a-glance guarantee post-return isn't here?

crates/wasmparser/src/validator/component.rs
```diff
@@ -2170,10 +2306,11 @@ impl ComponentState {
...
                         None => {
-                            let ty = types[self.core_function_at(*idx, offset)?].unwrap_func();
-                            if ty.params()
+                            let ty_id = self.core_function_at(*idx, offset)?;
+                            let func_ty = types[ty_id].unwrap_func();
+                            if func_ty.params()
                                 != [ValType::I32, ValType::I32, ValType::I32, ValType::I32]
-                                || ty.results() != [ValType::I32]
+                                || func_ty.results() != [ValType::I32]
                             {
```
alexcrichton, line 2314: Do you think it might make sense to defer this type-check until `CanonicalOptions::check_{lower,lift}`? That's how post-return and callback options work for somewhat unrelated reasons, but in the future with memory64 integration it might be required for realloc as well

alexcrichton (approved):
Nice :+1:

crates/wasmparser/src/validator/component.rs
fitzgen: I was trying to validate as much up front as possible, so that it is harder to forget to add things in the implementation of subsequent methods and so that it can be shared as much as possible. That said, I don't feel super strongly about it. My attitude is probably best summarized as

> maybe? but it isn't clear to me. let's wait and cross that bridge when we get to it

fitzgen:
Going to put this in the merge queue, if you feel more strongly about when/where we validate the `realloc` function's type, feel free to remove it from the queue and drop another comment.

PR #2166, opened 2025-04-28
wasmparser(CM+GC): Allow declaring a core function type when lowering

fitzgen:
When lowering a component function into a core function, allow the declaration of which core function type the lowered function (or lowered CM intrinsic) should have. Right now, this is effectively just double checking that the declared type matches the actual lowered type. After future PRs, this will allow declaring a Wasm GC-using function type, and lowering will adjust accordingly. This also lets the Wasm declare which of the "equivalent" function types it wants to lower to in the case that there are multiple rec groups containing the "same" function type, or in the face of function type subtyping. Again, subsequent PRs will start adding support for rec groups and subtyping and such to CM core types.

### DIFF
--- a/crates/wasm-encoder/src/component/canonicals.rs
+++ b/crates/wasm-encoder/src/component/canonicals.rs
@@ -27,6 +27,8 @@ pub enum CanonicalOption {
     /// The function to use if the async lifting of a function should receive task/stream/future progress events
     /// using a callback.
     Callback(u32),
+    /// The core function type to lower a component function into.
+    CoreType(u32),
 }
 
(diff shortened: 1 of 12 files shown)


alexcrichton (approved):

alexcrichton:
Ah the test failures are something I was slightly worried about where this is peeking for just `core` and not `core type` which is ambiguous with `(core func ...)` and I wasn't sure if we used that anywhere, but looks like we do in `task.return`

fitzgen:
Yeah I am renaming the canonical option to `core-type` to work around this

PR #2167, opened 2025-04-29
Improve time complexity of handling large WIT types

### LINKED ISSUE #2164: Nested tuples and fixed size lists use a lot of memory
Originally posted by @alexcrichton and @cpetig in https://github.com/bytecodealliance/wasm-tools/pull/1992#discussion_r2064834026_

```wit
package a:b;

world x {
  type t = tuple<t2,t2,t2,t2,t2,t2,t2,t2,t2,t2,t2,t2,t2,t2,t2,t2,>;
  type t2 = tuple<t3,t3,t3,t3,t3,t3,t3,t3,t3,t3,t3,t3,t3,t3,t3,t3,>;
  type t3 = tuple<t4,t4,t4,t4,t4,t4,t4,t4,t4,t4,t4,t4,t4,t4,t4,t4,>;
  type t4 = tuple<t5,t5,t5,t5,t5,t5,t5,t5,t5,t5,t5,t5,t5,t5,t5,t5,>;
  type t5 = tuple<t6,t6,t6,t6,t6,t6,t6,t6,t6,t6,t6,t6,t6,t6,t6,t6,>;
  type t6 = tuple<t7,t7,t7,t7,t7,t7,t7,t7,t7,t7,t7,t7,t7,t7,t7,t7,>;
  type t7 = tuple<t8,t8,t8,t8,t8,t8,t8,t8,t8,t8,t8,t8,t8,t8,t8,t8,>;
  type t8 = tuple<u32,u32,u32,u32,u32,u32,u32,u32,u32,u32,u32,u32,u32,u32,u32,u32,>;
  import x: func(t: t);
}
```

when locally run as:

`$ cargo run component embed --dummy foo.wit -t`

takes ~30 seconds to produce the result. I think the reason is that this case isn't hit and then the recursion gets to go hog wild for a bit.

alexcrichton:
Update a few places in tooling which iterate over the structure of a WIT type to contain early-returns or similar to ensure that processing these large types doesn't take an exponential amount of time.

Closes #2164

### DIFF
--- a/tests/cli/wit-deep-variant.wit
+++ b/tests/cli/wit-deep-variant.wit
@@ -1,0 +1,47 @@
+// FAIL[parse]: component wit % -t
+
+// TODO: this should use "FAIL" and should actually be a test, but this
+// currently takes too long to run and the long runtime is in and of itself a
+// bug.
+// fail[abi]: component embed --dummy % | component new
+
+package a:b;
+
+world variants {
+  variant t1 {
+    f1(t2), f2(t2), f3(t2), f4(t2), f5(t2), f6(t2), f7(t2), f8(t2),
+    f9(t2), f10(t2), f11(t2), f12(t2), f13(t2), f14(t2), f15(t2), f16(t2),
+  }
+  variant t2 {
+    f1(t3), f2(t3), f3(t3), f4(t3), f5(t3), f6(t3), f7(t3), f8(t3),
+    f9(t3), f10(t3), f11(t3), f12(t3), f13(t3), f14(t3), f15(t3), f16(t3),
+  }
+  variant t3 {
+    f1(t4), f2(t4), f3(t4), f4(t4), f5(t4), f6(t4), f7(t4), f8(t4),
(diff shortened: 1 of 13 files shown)


alexcrichton:
cc @cpetig

tests/cli/wit-deep-variant.wit
```diff
@@ -0,0 +1,47 @@
+// FAIL[parse]: component wit % -t
+
+// TODO: this should use "FAIL" and should actually be a test, but this
+// currently takes too long to run and the long runtime is in and of itself a
+// bug.
+// fail[abi]: component embed --dummy % | component new
```
alexcrichton, line 6: This is an interesting test where the test below actually fits in the flat representation, unlike all of the above tests which don't fit flat. This test takes quite a long time to produce the flat representation.

This isn't a problem for validation since this type is "too large and complex" and fails validation before we get to the flat representation, but is still an issue for tooling which I'm not sure how best to handle.

pchickey (approved):

PR #2170, opened 2025-04-29
wasmparser(CM+GC): Assert that `SubtypeCx`s are always using the same typing context

fitzgen:
A `SubtypeCx`, used for subtype-checking in the component model, is constructed with two `TypesRef`s. These `TypesRef`s must be associated with the same validator, so that we know that core types are canonicalized the same across the two of them and we can (for example) compare `CoreTypeId`s for equality.

I had originally hoped to move the `SubtypeCx` to wrapping a single `TypesRef`, but this cannot work because (unlike core types) new component model types can be pushed into each different `TypesRef` for different components.

### DIFF
--- a/crates/wasmparser/src/validator/component_types.rs
+++ b/crates/wasmparser/src/validator/component_types.rs
@@ -730,7 +730,13 @@ pub enum ComponentEntityType {
 
 impl ComponentEntityType {
     /// Determines if component entity type `a` is a subtype of `b`.
-    pub fn is_subtype_of(a: &Self, at: TypesRef, b: &Self, bt: TypesRef) -> bool {
+    ///
+    /// # Panics
+    ///
+    /// Panics if the two given `TypesRef`s are not associated with the same
+    /// `Validator`.
+    pub fn is_subtype_of(a: &Self, at: TypesRef<'_>, b: &Self, bt: TypesRef<'_>) -> bool {
+        assert_eq!(at.id(), bt.id());
         SubtypeCx::new(at.list, bt.list)
             .component_entity_type(a, b, 0)
             .is_ok()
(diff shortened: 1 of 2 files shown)


crates/wasmparser/src/validator/component_types.rs
```diff
@@ -2496,7 +2505,13 @@ macro_rules! limits_match {
...
     /// Create a new instance with the specified type lists
+    ///
+    /// # Panics
+    ///
+    /// Panics if the two given `TypesRef`s are not associated with the same
+    /// `Validator`.
     pub fn new_with_refs(a: TypesRef<'a>, b: TypesRef<'a>) -> SubtypeCx<'a> {
+        assert_eq!(a.id(), b.id());
         Self::new(a.list, b.list)
     }
```
alexcrichton, line 2516: Could this go further and take a single `TypesRef` as a constructor argument?

alexcrichton (approved):

crates/wasmparser/src/validator/component_types.rs
fitzgen: I tried that, but it doesn't work because while the `TypesRef`s are always associated with the same validator, they can still be distinct and have different sets of component types. See my last paragraph in the OP:

> I had originally hoped to move the `SubtypeCx` to wrapping a single `TypesRef`, but this cannot work because (unlike core types) new component model types can be pushed into each different `TypesRef` for different components.

alexcrichton: Wait hang on, while `TypesRef` can be different aren't the`TypeList` values the same? Otherwise how are core type canonicalizations shared?

fitzgen: The underlying `TypeList`s are not necessarily the same, but one will always be the extension of another, so any types they both share are canonicalized the same.

alexcrichton: Ah ok right, those are snapshots of a list at a point in time, ok makes sense!

PR #2172, opened 2025-04-30
wasmparser(CM+GC): Validate components' core function linking via type canonicalization

fitzgen:
Instead of structurally checking parameters and return types, rely on type canonicalization to dedupe types so that we can simply compare ids. This gets instantiation with the "same" function type defined in a different rec group to properly fail to link.

This does not yet handle function subtyping. That will happen in follow up commits.


crates/wasmparser/src/readers/core/types.rs
```diff
@@ -798,6 +859,19 @@ pub struct StructType {
...
+    pub(crate) fn desc(&self) -> String {
+        let mut s = String::from("(struct");
+        for f in &self.fields {
+            s.push(' ');
+            s.push_str(&f.desc());
+        }
+        s.push_str(")");
+        s
+    }
+}
```
alexcrichton, line 873: Could this be done as `fmt::Display` or similar to avoid the intermediate string allocations?

crates/wasmparser/src/validator/component_types.rs
```diff
@@ -3094,16 +3090,16 @@ impl<'a> SubtypeCx<'a> {
         }
     }
 
-    fn core_func_type(&self, a: &FuncType, b: &FuncType, offset: usize) -> Result<()> {
+    fn core_func_type(&self, a: CoreTypeId, b: CoreTypeId, offset: usize) -> Result<()> {
         if a == b {
```
alexcrichton, line 3094: Could this include a debug assert that both types are in the same original shared tail of the type list?

crates/wasmparser/src/readers/core/types.rs
fitzgen: I was going to move all the `desc` (including ones already defined before this PR) to `Display` impls in a follow up. Would you prefer that happen in this PR?

crates/wasmparser/src/validator/component_types.rs
fitzgen: Specifically in the case that they are equal? Because if they are not equal, that is not guaranteed to be true.

alexcrichton (approved):

PR #2176, opened 2025-05-02
wasmparser(CM+GC): Use matching when checking actual lowered signatures

fitzgen:
When checking that a component function's lowered parameters and results are of the correct types compared to the declared `core-type` canonical option, use spec matching rather than strict equality. This shouldn't change anything today, since reference types will never currently appear in params and results, but opens the door to their proper subtyping in future commits that allow reference types in these signatures.


fitzgen:
Actually, this stuff is going to need to be folded into `lower` itself, so we don't actually want this.

PR #2177, opened 2025-05-02
wasmparser(CM+GC): validate `core-type` canonical option during lowering

fitzgen:
Rather than checking that the lowered type matches the declared `core-type` canonical option post facto, check it during the process of lowering. This sets us up (in future commits) for peeking at the expected lowering and deciding whether to lower to the regular linear memory version of the canonical ABI or the GC version that I am prototyping.


fitzgen:
Hm beyond just the test failures in `wasm-compose` that I haven't diagnosed yet, this is going to run into trouble with the way that we first try to flatten and then back up and pass via linear memory pointer.

Will comment when this is actually ready for review.

fitzgen:
Okay I worked out the kinks and this should be ready for review now!

crates/wasmparser/src/validator/component_types.rs
```diff
@@ -34,56 +35,174 @@ const MAX_FLAT_FUNC_RESULTS: usize = 1;
...
+type LowerResult<T = ()> = Result<T, LowerError>;
+
+enum Variance {
+    Covariant,
+    Contravariant,
+}
+
+/// A list of expected types and how to validate them.
+struct ExpectedLoweredTypes<'a> {
+    types: &'a TypeList,
```
alexcrichton, line 72: This stores a `TypeList` but elsewhere when we're passing around a `LoweredTypes` we also pass around a `TypeList`. Could this change to not storing a `TypeList` and taking a parameter, or instead storing this within `LoweredTypes` to uniformly either use-a-field or use-a-parameter?

crates/wasmparser/src/validator/component_types.rs
```diff
@@ -928,20 +1064,106 @@ impl ComponentFuncType {
...
+        sig: &mut LoweredSignature<'_>,
+    ) -> Result<()> {
+        match self.lower_params_flat(types, options, abi, offset, sig) {
+            Ok(()) => Ok(()),
+            Err(LowerError::OutOfCapacity) => self.lower_params_indirect(options, abi, offset, sig),
+            Err(LowerError::TypeMismatch(e)) => self
+                .lower_params_indirect(options, abi, offset, sig)
+                .with_context(|| e.to_string()),
+            Err(LowerError::Other(e)) => Err(e),
+        }
```
alexcrichton, line 1156: Could you add some comments to this `match`? I think it'd be good to explicitly comment why the `OutOfCapacity` error is caught here for example.

Additionally I don't understand how `TypeMismatch` is handled here. That's taking what I thought would be a fatal error, trying to catch it, and then adding context if the indirect option fails.

crates/wasmparser/src/validator/component_types.rs
```diff
@@ -34,56 +35,174 @@ const MAX_FLAT_FUNC_RESULTS: usize = 1;
...
+        self.expected.get(self.index).copied()
+    }
+
+    fn match_ty(&mut self, actual: ValType) -> LowerResult {
+        let expected = match self.peek() {
+            Some(e) => e,
+            // Arity mismatch, but ignore it for now. The arity check at the
+            // end of `ComponentFuncType::lower` will catch this and can
+            // give a more descriptive error than we can here.
+            None => return Ok(()),
```
alexcrichton, line 122: Do you think it might make sense to have a `self.arity_mismatch` which is debug-assert'd to be false right before we leave without an error higher up the stacsk?

crates/wasmparser/src/validator/component_types.rs
```diff
@@ -34,56 +35,174 @@ const MAX_FLAT_FUNC_RESULTS: usize = 1;
...
-        } else {
-            None
-        }
-    }
-
+    /// Like `try_push` but asserts that this list has capacity for the type.
     #[track_caller]
-    fn assert_push(&mut self, ty: ValType) {
-        assert!(self.try_push(ty));
+    fn assert_push(&mut self, ty: ValType) -> Result<()> {
```
alexcrichton, line 180: I think we should probably rename this to something else since it's returning a `Result`. 

How about renaming `try_push` to `push`, removing `assert_push`, and using test coverage to test the `assert_*` cases? Everything already returns a `Result` anyway so using `?` liberally seems fine here.

crates/wasmparser/src/validator/component_types.rs
```diff
@@ -1183,34 +1455,40 @@ impl ComponentDefinedType {
...
-
+        lowered_types.try_push(ValType::I32)?;
+
+        // First, simultaneously check that the `lowered_types` has capacity for
+        // the variants' types and join the `i`th type from each variant
+        // together to get the actual types we will push onto
+        // `lowered_types`. This phase ignores the expected types, if any, since
+        // we don't know what we will actually push onto `lowered_types` until
+        // after all the joining has taken place.
+        let mut joined_tys = SmallVec::<[_; MAX_FLAT_FUNC_PARAMS]>::new();
```
alexcrichton, line 1468: To avoid adding `SmallVec` as a dependency, could this use `clone_without_expected_types` as the storage here, similar-ish to before?

alexcrichton (commented):
FWIW I'm having a bit of a difficult time seeing how this is leading towards GC stuff, so I could be off-the-mark on some of these comments.

alexcrichton:
Also, high-level, what are you expecting the canonical ABI to be for GC types? I would imagine that it wants to be handled almost entirely differently, not just with types, from the linear memory lowering. For example passing a `record` should probably be a GC struct meaning there's basically no need to recurse into a type. I'm also not sure whether you want to have a max param or result limit?

It'd be helpful to me at least I think to have an idea of where this is all ending up. Much of the structure of these methods you're modifying is very highly-tailored to linear memory and I'm not sure any of it is going to get reused for GC. This refactoring is (in isolation) making this code less readable and harder to understand, and while that's fine if it's paving the way for GC in the future I'm not sure it'd be worth it if GC isn't going to end up using any of this anyway.

fitzgen:
> I'm also not sure whether you want to have a max param or result limit?

We have to because core wasm has such a limit

crates/wasmparser/src/validator/component_types.rs
```diff
@@ -34,56 +35,174 @@ const MAX_FLAT_FUNC_RESULTS: usize = 1;
...
+    Contravariant,
+}
+
+/// A list of expected types and how to validate them.
+struct ExpectedLoweredTypes<'a> {
+    types: &'a TypeList,
+    offset: usize,
+
+    /// The expected types themselves.
+    expected: smallvec::SmallVec<[ValType; MAX_LOWERED_TYPES]>,
```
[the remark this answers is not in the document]
fitzgen: We could probably get away with `&'a [ValType]` here, might take a little finessing.

crates/wasmparser/src/readers/core/types/matches.rs
```diff
@@ -319,3 +353,9 @@ impl Matches for WithRecGroup<RefType> {
         )
     }
 }
+
+impl Matches for RefType {
+    fn matches(types: &TypeList, a: Self, b: Self) -> bool {
+        types.reftype_is_subtype_impl(a, None, b, None)
```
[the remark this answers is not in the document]
fitzgen: Good call.

crates/wasmparser/src/validator/component_types.rs
fitzgen: The thing is that we don't always have an `ExpectedLoweredTypes`, so we could add a `TypeList` to `LoweredTypes` or something to remove the threading, but that is additional plumbing work.

Alternatively, we could remove it from this type and pass it at every call site. But that seems kind of like unnecessary busy work as well, and also won't remove the lifetime parameter if we make `ExpectedTypes` hold a `&'a [ValType]` instead of a `SmallVec`.

crates/wasmparser/src/validator/component_types.rs
fitzgen: We initially try to lower flat, but that can fail if we have too many args to pass flatly, in which case we then try to pass indirect. The core types of the flat vs indirect versions are going to be different, so you can fail to flatten due to a type mismatch (which would otherwise become a out-of-capacity error) because the declared type is expecting to be indirect.

crates/wasmparser/src/validator/component_types.rs
fitzgen: I'm not sure I am following this suggestion

alexcrichton:
I'm still a bit confused in that what I'm picturing GC validation looks like is completely different from what core wasm does. GC validation I would imagine does a zip of component-level params and wasm-level params and then does a "diff" to make sure the two match. Perhaps also with "if > N everything is a single `struct` in GC". That's completely different from core wasm where we're flattening component types into core wasm types and building up a function signature as we go, falling back to deferring everything to being a single in-memory pointer.

While these two code paths could share "iterate over the type structure" that's not really all that hard so I'd imagine that it isn't really necessary to share. Otherwise the two are so logically different I'm not sure they want to be inter-woven into the same validation functions?

Also, how are you thinking of handling variants? Would they unconditionally be something like a 32-bit discriminant followed by an `anyref`?

alexcrichton:
Er sorry for repeat comments, but you're thinking that there's no explicit opt-in via a canonical option to a GC lowering? If so that seems a bit surprisingly subtle to me and instead I was expecting that you'd explicitly specify a `gc` canonical option. Validation of `gc` would then require that `core-type` was specified, and we could even reject `core-type` if `gc` isn't specified. Either that or rename `core-type` to `gc` so it's just always got the type specified there.

crates/wasmparser/src/validator/component_types.rs
fitzgen: `try_push` and `assert_push` return two different kinds of results tho, which actually works out pretty nice right now. `try_push` returns `LowerResult`s and notably that means it can return `OutOfCapacity` errors, while `assert_push` returns a `Result<(), BinaryReaderError>` and can never fail with out-of-capacity errors. This is useful for stratifying the try-and-lower-into-a-max-size-arg-list use cases versus the infallible indirect-through-memory use cases.

fitzgen: Don't love the names either tho.

fitzgen:
> you're thinking that there's no explicit opt-in via a canonical option to a GC lowering

I've been thinking that the explicit opt in is providing a `(core-type $some_gc_type)`

fitzgen:
> we could even reject `core-type` if `gc` isn't specified

You might want a linear-memory-style function in a particular rec group or with a particular subtyping relationship (specified via `core-type`) even if you aren't using GC structs in the canonical ABI.

PR #2178, opened 2025-05-05
Fix `var_u{32,64}` parsing in `memories` and `tables`

Robbepop:
I tried to update Wasmi to `wasmparser v0.229` and encountered the following bug in the existing Wasm spec testsuite:

```
thread 'wasm_binary_leb128' panicked at crates/wast/tests/mod.rs:11:9:
failed directive on testsuite/binary-leb128:217:1: module succeeded to compile and validate but should have failed with: integer representation too long
```

After some debugging I found that [this PR](https://github.com/bytecodealliance/wasm-tools/pull/2134) recently changed the logic to parse `var_u32` or `var_u64` depending on `memory64` flags [for `memories.rs`](https://github.com/keithw/wasm-tools/blob/68ec57dfbbbae243d58b662371945ad1072002ea/crates/wasmparser/src/readers/core/memories.rs#L38) and [for `tables.rs`](https://github.com/keithw/wasm-tools/blob/68ec57dfbbbae243d58b662371945ad1072002ea/crates/wasmparser/src/readers/core/tables.rs#L81).

With this PR I fixed the bugs I found on Wasmi, however, the Wasm spec testsuite of `wasm-tools` is unhappy now. This is probably due to Wasmi and `wasm-tools` using different versions of the Wasm spec testsuite. Was there a Wasm spec divergence in what Wasm files are supposed to be accepted in the recent past?

### DIFF
--- a/crates/wasmparser/src/readers/core/memories.rs
+++ b/crates/wasmparser/src/readers/core/memories.rs
@@ -35,14 +35,14 @@ impl<'a> FromReader<'a> for MemoryType {
         Ok(MemoryType {
             memory64,
             shared,
-            initial: if reader.memory64() {
+            initial: if memory64 {
                 reader.read_var_u64()?
             } else {
                 reader.read_var_u32()?.into()
             },
             maximum: if !has_max {
                 None
-            } else if reader.memory64() {
+            } else if memory64 {
                 Some(reader.read_var_u64()?)
             } else {
                 Some(reader.read_var_u32()?.into())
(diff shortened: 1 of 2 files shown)


Robbepop:
Never mind, I just noticed that Wasmi never uses [`Parser::set_features`](https://docs.rs/wasmparser/0.228.0/wasmparser/struct.Parser.html#method.set_features) and so far that has never been an issue. But probably with [this PR](https://github.com/bytecodealliance/wasm-tools/pull/2134) the issue just got uncovered.

Closed.

PR #2179, opened 2025-05-05
reencoder: allow implementers to fail to reencode indices and such

nagisa:
I have found myself wanting to return an error from `Reencode::type_index`. This is admittedly a pretty invasive change though and I punted on component stuff.

Thoughts?

### DIFF
--- a/crates/wasm-encoder/src/reencode.rs
+++ b/crates/wasm-encoder/src/reencode.rs
@@ -24,36 +24,36 @@ use alloc::vec::Vec;
 pub trait Reencode {
     type Error;
 
-    fn data_index(&mut self, data: u32) -> u32 {
-        utils::data_index(self, data)
+    fn data_index(&mut self, data: u32) -> Result<u32, Error<Self::Error>> {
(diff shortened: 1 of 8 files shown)


alexcrichton:
I do agree `Error<U>` is pretty clunky, but one nice part is that you can stick in `std::convert::Infallible` and not have to worry much about it. With extra constructors that would require extra boilerplate on each impl I think?

Perhaps as a further refinement of your idea, a `ReencodeError` trait instead of methods directly on `Reencode`?

nagisa:
I updated this to use Error<E> universally.

alexcrichton (approved):

alexcrichton:
Ah that's a merge conflict with https://github.com/bytecodealliance/wasm-tools/pull/2145 which just landed prior, I can help resolve too as well if you'd like.

nagisa:
If you don't find an opportunity to, I'll get to it tonight or tomorrow myself.

nagisa:
This is ready to go again now.

PR #2182, opened 2025-05-07
Add the GC canonical option

fitzgen:
Disallow using the `core-type` option without `gc`.


crates/wasmparser/src/validator/component.rs
```diff
@@ -2475,6 +2485,15 @@ impl ComponentState {
                         }
                     };
                 }
+                CanonicalOption::Gc => {
```
alexcrichton, line 2488: Can this also check for `self.features.cm_gc()` being enabled?

alexcrichton (approved):

crates/wasmparser/src/validator/component.rs
fitzgen: Ah yes indeed, thanks

PR #2183, opened 2025-05-08
CM+GC: Define and validate GC lowerings for most component model types

fitzgen:
This commit defines the expected lowerings for most component model types (primitives, records, strings, tuples, etc..) and implements their validation.

This commit also defines futures and streams as `externref`s in the GC ABI, but it does not plumb that through to all of the async intrinsics yet.

Finally, this commit does *not* yet define the expected lowerings for variants and variant-like types (options and results). This is because it isn't 100% clear how to lower sum types to GC types yet, especially with the way that we are doing shallow-matching to allow for the "same" type from any rec group to be used.


crates/wasmparser/src/readers/component/types.rs
```diff
@@ -226,6 +232,165 @@ impl PrimitiveValType {
         // component functions.
         a == b
     }
+
+    #[cfg(all(feature = "component-model", feature = "validate"))]
```
alexcrichton, line 236: To avoid the #[cfg] could this move inside of `mod component_types`? (or somewhere around there)

crates/wasmparser/src/readers/component/types.rs
```diff
@@ -226,6 +232,165 @@ impl PrimitiveValType {
...
+                        _ => type_mismatch_err(),
+                    },
+                    _ => type_mismatch_err(),
+                }
+            }
+
+            (PrimitiveValType::ErrorContext, _) => bail!(
+                offset,
+                "GC support for error-context is not yet implemented",
+            ),
```
alexcrichton, line 391: If it's easy, this'll probably be externref like futures/streams

crates/wasmparser/src/validator.rs
```diff
@@ -50,6 +50,8 @@ fn test_validate() {
 #[cfg(feature = "component-model")]
 mod component;
 #[cfg(feature = "component-model")]
+pub(crate) use component::{CanonicalOptions, StringEncoding};
```
alexcrichton, line 53: Would moving more validation within `component_types` avoid the need for this? Or is this needed for other reasons?

crates/wasmparser/src/validator/component.rs
```diff
@@ -2503,6 +2511,14 @@ impl ComponentState {
             }
         }
 
+        let string_encoding = match encoding {
```
alexcrichton, line 2514: Instead of `encoding: Option<CanonicalOption>`, could `encoding: Option<StringEncoding>` be the local variable?

crates/wasmparser/src/validator/component_types.rs
```diff
@@ -918,6 +932,87 @@ pub(crate) enum Abi {
...
+            _ => None,
+        }
+    }
+
+    pub(crate) fn as_ref_type(self) -> Option<RefType> {
+        match self {
+            ArgOrField::Arg(ValType::Ref(r))
+            | ArgOrField::Field(StorageType::Val(ValType::Ref(r))) => Some(r),
+            _ => None,
+        }
```
alexcrichton, line 986: `self.as_val_type()?.as_reference_type()` perhaps?

crates/wasmparser/src/validator/component_types.rs
```diff
@@ -1033,6 +1173,46 @@ pub struct RecordType {
...
+                Ok(())
+            }
+            other => bail!(
+                offset,
+                "expected to lower component `record` type to core `(ref null? (struct ...))`, \
+                 but found `{other}`",
+            ),
+        }
+    }
+}
```
alexcrichton, line 1215: One thing I've done a lot with canonical ABI things is that there's one helper for packed things and one helper for variant things. For example records/tuples [go here](https://github.com/bytecodealliance/wasmtime/blob/c213302828268411e90a3ea454d42ae254c793e2/crates/environ/src/component/types.rs#L671-L687) while variant/option/result all go [here](https://github.com/bytecodealliance/wasmtime/blob/c213302828268411e90a3ea454d42ae254c793e2/crates/environ/src/component/types.rs#L757-L794). That'll help cut down on duplication between records/tuples here and, in the future, cut down on duplication between variants/option/result.

alexcrichton (commented):
Nice suite of tests 👍 

Two questions:

* What are you thinking of doing with the `Abi` field here? It looks like it's not used, although I could be missing it. I was naively expecting one direction to require a precise `struct` where the other direction would accept up to an `anyref` or something like that but I didn't see that.
* Do you have an idea for how variants will be supported? I figure that's something we should at least have a plan for before landing the other bits since it's such a foundational data type in the component model

fitzgen:
> What are you thinking of doing with the `Abi` field here? It looks like it's not used, although I could be missing it. I was naively expecting one direction to require a precise `struct` where the other direction would accept up to an `anyref` or something like that but I didn't see that.

Yeah I wanted to leave the option open for doing just that, but I didn't want to bite it off quite yet.

> Do you have an idea for how variants will be supported? I figure that's something we should at least have a plan for before landing the other bits since it's such a foundational data type in the component model

My thinking is that we will match against an empty `struct` type (in whatever rec group and with whatever subtyping of other empty `struct` types), check that it is not final, and then when lowering a variant at runtime, we lower it to a subtype of that empty `struct` type with the the variant's lowered payload values defined as fields that is defined in a singleton rec group. It doesn't give quite as much flexibility as product types get, but I can't think of anything better.

The main alternative I've come up with is adding new canonical options for saying something like "the `i`th variant of this type should be lowered to core type `$foo`", but that seems super wordy and annoying to use...

fitzgen:
Rebased and review comments addressed. This should be good for a final review.

PR #2186, opened 2025-05-11
Add the "lime1" feature set for validation

alexcrichton:
This commit adds `WasmFeatures::LIME1` which is a set of features added recently to LLVM to encompass a just-a-bit-more-than-MVP baseline which includes useful features for producers without placing too much of an implementation burden on engines. This required splitting the reference-types and bulk-memory features into sub-portions to handle the special cases of the lime1 proposal.

A new test is added here which asserts that all the right features are turned on while all the wrong features are indeed all turned off. This commit additionally includes a bug fix where passive data segments were not properly gated by the bulk-memory feature prior to this commit.


abrown (approved):
Makes sense to me!

sunfishcode (approved):
Looks good!

PR #2194, opened 2025-05-17
Add stability attributes to the worlds in wit-smith

jsturtevant:
Follow up to https://github.com/bytecodealliance/wasm-tools/pull/2159#issuecomment-2842851791 to include stability attributes for fuzzing.


alexcrichton (approved):
Thanks for this!

PR #2198, opened 2025-05-21
Update the ABI of async imports/exports

alexcrichton:
This commit updates wasm-tools with the changes from WebAssembly/component-model#520 which updates the ABI of async imports and exports. The changes here are:

* Flat function parameters are now allowed.
* If there are no function parameters then the parameters are entirely omitted.
* The result out-ptr is now omitted if there are no function results.


alexcrichton:
This is going to require a lot of adjustments in downstream tools, so I'm going to work my way through those first before landing this.

PR #2199, opened 2025-05-21
Revise MSRV policy for this repository

alexcrichton:
In writing the current policy for this repository's minimum supported version of Rust (MSRV) is "stable minus two" which matches Wasmtime. This does not reflect the current reality, however, where the current MSRV is 1.76 where the current Rust is 1.87. This additionally does not reflect what I think is the best decision for this repository. While "stable minus two" matches Wasmtime this repository is expected to be possibly used in more contexts than Wasmtime, for example any wasm-processing tools which can be lower-level than anything needing a wasm runtime.

Given all this the change here is to update the comments on MSRV policy for this repository to "stable minus 10, not proactively updated". That is a much larger window of Rust but the code in this repository is much less likely than Wasmtime, for example, to benefit from newer features of Rust. The hope is that this won't be too onerous to maintain while still allowing to eventually use newer features as necessary.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -86,8 +86,22 @@ allow_attributes_without_reason = 'warn'
 edition = '2021'
 license = "Apache-2.0 WITH LLVM-exception OR Apache-2.0 OR MIT"
 version = "0.231.0"
-# Current policy for wasm-tools is the same as Wasmtime which is that this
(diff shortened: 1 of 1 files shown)


sunfishcode (approved):
This makes sense to me.

PR #2201, opened 2025-05-22
Turn on new WIT metadata format by default

alexcrichton:
This commit is a follow-up to #1508 where a new JSON format was added for storing docs/stability and accounting for world imports/exports allowing overlap. This support should have permeated far enough that it's reasonable to turn this on-by-default now.

This additionally fixes fuzz bugs cropping up after adding support for stability annotations in the generated WIT, the other motivation for this.

### DIFF
--- a/crates/wit-parser/src/metadata.rs
+++ b/crates/wit-parser/src/metadata.rs
@@ -43,7 +43,7 @@ const PACKAGE_DOCS_SECTION_VERSION: u8 = 1;
 /// tools we'll still try to emit the v0 format by default, if the input is
 /// compatible. This will be turned off in the future once enough published
 /// versions support the v1 format.
-const TRY_TO_EMIT_V0_BY_DEFAULT: bool = true;
+const TRY_TO_EMIT_V0_BY_DEFAULT: bool = false;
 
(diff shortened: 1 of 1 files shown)


crates/wit-component/tests/interfaces/console.wat
```diff
@@ -11,7 +11,7 @@
     )
   )
   (export (;1;) "console" (type 0))
-  (@custom "package-docs" "\00{}")
+  (@custom "package-docs" "\01{}")
```
jsturtevant, line 14: Why does this get bumped to 01?  I don't really see reasoning in this PR but I assume it is due to the face this feature is no turned on?

alexcrichton: Ah that's [around here](https://github.com/bytecodealliance/wasm-tools/blob/84d3a4ce8328925e5691877cbf9084ddcab26caa/crates/wit-parser/src/metadata.rs#L121-L133) where by changing the static this is changing the default emission as well

abrown (approved):
I'm not sure I'm qualified to review this PR but @jsturtevant seems to think it is OK, so +1!

Issue #2180, opened 2025-05-06
`wasmparser`: 15-30% performance regressions from v228 -> v229


Robbepop:
I just finished updating Wasmi to the new `wasm-tools` v229 coming from v228.

Benchmarking before and after revealed the following performance regressions:
![Image](https://github.com/user-attachments/assets/a441592f-6a15-4acc-abb6-ecb03236788d)

- Roughly 15% performance regressions across the board.
- Up to 30% performance regressions when Wasmi translation was lazy and thus `wasmparser` parsing and validation made up a larger chunk of the work. So that's probably the real regression in `wasmparser`.

alexcrichton:
Would you be able to help profile and track down where this is coming from? Perhaps via bisection?

Robbepop:
> Would you be able to help profile and track down where this is coming from? Perhaps via bisection?

Yes I can do that.

keithw:
I can replace this in-repo with `cargo bench -p wasmparser -- parse/tests` ; let me try to produce a flamegraph and take a look asap (but please don't let me stop anybody else).

keithw:
I did some measurements on parsing (without validating) a [single huge module](https://github.com/turbolent/w2c2/blob/main/examples/clang/clang.wasm) on an unloaded AMD Ryzen 7 PRO 4750U, comparing two release builds built with rustc 1.86.0 (05f9846f8 2025-03-31).

For the "before", I tested the commit just before that PR (90c156ff15a81cb0360fb949e6504a4799d3b5cc), backporting the "full parse" logic from https://github.com/bytecodealliance/wasm-tools/blob/main/src/lib.rs#L299 (with the exception of the [ops.finish()?;](https://github.com/bytecodealliance/wasm-tools/blob/main/src/lib.rs#L327) line since the "before" parser doesn't have a finish function to check anything at the end of a function).

For the "after", I used that PR (0354dde10184d7ff813282a2837d9a4ed8e766e6).

In both cases I tested `wasm-tools parse` without output (i.e. removing https://github.com/bytecodealliance/wasm-tools/blob/main/src/bin/wasm-tools/parse.rs#L25). I ran the command under `perf stat` a few hundred times and recorded the cycle count. I ran it enough that the standard error of the mean is <1M cycles for both measurements.

I was surprised to find that in this "single huge module" test, I didn't see a regression:

| Commit | mean cycle count | diff |
| ------------ | ------------------------- | ------ |
| Before | 823,830,564 | |
[...]

PR #2202, opened 2025-05-22
Relax the block check in `OperatorsReader<'_>`

alexcrichton:
This commit relaxes a check added in #2134 which maintains a stack of frame kinds in the operators reader, in addition to the validator. The goal of #2134 was to ensure that spec-wise-syntactically-invalid-modules are caught in the parser without the need of the validator, but investigation in #2180 has shown that this is a source of at least some of a performance regression. The change here is to relax the check to still be able to pass spec tests while making such infrastructure cheaper.

The reader now maintains just a small `depth: u32` counter instead of a stack of kinds. This means that the reader can still catch invalid modules such as instructions-after-`end`, but the validator is required to handle situations such as `else` outside of an `if` block.

This required some adjustments to tests as well as some workarounds for the upstream spec tests that assert legacy exception-handling instructions are malformed, not invalid.

### DIFF
--- a/ci/generate-spec-tests.rs
+++ b/ci/generate-spec-tests.rs
@@ -44,9 +44,19 @@ fn copy_test(src: &Path, dst: &Path) {
     contents.push_str(";;      --assert default \\\n");
 
     // Allow certain assert_malformed tests to be interpreted as assert_invalid
-    if src.ends_with("binary.wast") || src.ends_with("global.wast") || src.ends_with("select.wast")
+    if src.ends_with("binary.wast")
+        || src.ends_with("global.wast")
+        || src.ends_with("select.wast")
+        || src.ends_with("try_table.wast")
     {
         contents.push_str(";;      --assert permissive \\\n");
+    }
+
+    // For this test it has legacy instructions which, for roundabout reasons,
+    // are attempted to be printed in the folded format but that doesn't work.
+    // Exclude this test for now.
+    if src.ends_with("try_table.wast") {
+        contents.push_str(";;      --assert no-test-folded \\\n");
     }
 
     contents.push_str(";;      --snapshot tests/snapshots \\\n");
--- a/crates/wasmparser/src/readers/core/operators.rs
+++ b/crates/wasmparser/src/readers/core/operators.rs
@@ -336,7 +336,13 @@ crate::for_each_operator!(define_operator);
 #[derive(Clone)]
 pub struct OperatorsReader<'a> {
     reader: BinaryReader<'a>,
-    blocks: Vec<FrameKind>,
+    /// Block depth this reader is currently operating at.
+    ///
+    /// Increments for instructions like `block` and decrements for instructions
+    /// like `end`. Used to rule out syntactically invalid modules (as defined
+    /// by the wasm spec currently) in a cheap but not 100% rigorous fashion.
+    /// More checks are still performed in the validator about block depth.
+    depth: u32,
     data_index_occurred: Option<usize>,
 }
 
@@ -345,7 +351,7 @@ impl<'a> OperatorsReader<'a> {
     pub fn new(reader: BinaryReader<'a>) -> OperatorsReader<'a> {
         OperatorsReader {
             reader,
-            blocks: vec![FrameKind::Block],
+            depth: 1,
             data_index_occurred: None,
         }
     }
@@ -377,7 +383,7 @@ impl<'a> OperatorsReader<'a> {
     }
 
     fn ensure_stack_empty(&self) -> Result<()> {
-        if !self.blocks.is_empty() {
+        if self.depth != 0 {
             bail!(
                 self.original_position(),
                 "control frames remain at end of function body or expression"
@@ -408,33 +414,6 @@ impl<'a> OperatorsReader<'a> {
     pub fn read_with_offset(&mut self) -> Result<(Operator<'a>, usize)> {
         let pos = self.reader.original_position();
         Ok((self.read()?, pos))
-    }
-
-    fn enter(&mut self, k: FrameKind) {
-        self.blocks.push(k)
-    }
-
-    fn expect_block(&mut self, k: FrameKind, found: &str) -> Result<()> {
-        match self.blocks.last() {
-            None => bail!(
-                self.original_position(),
-                "empty stack found where {:?} expected",
-                k
-            ),
-            Some(x) if *x == k => Ok(()),
-            Some(_) => bail!(
-                self.original_position(),
-                "`{}` found outside `{:?}` block",
-                found,
-                k
-            ),
-        }
-    }
-
-    fn end(&mut self) -> Result<()> {
-        assert!(!self.blocks.is_empty());
-        self.blocks.pop();
-        Ok(())
     }
 
     /// Visit the next available operator with the specified [`VisitOperator`] instance.
@@ -486,7 +465,7 @@ impl<'a> OperatorsReader<'a> {
     where
         T: VisitOperator<'a>,
     {
-        if self.blocks.is_empty() {
+        if self.depth == 0 {
             bail!(
                 self.original_position(),
                 "operators remaining after end of function body or expression"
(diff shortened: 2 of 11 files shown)


alexcrichton:
cc @keithw I'm curious for your thoughts on this, I'm not wed to this approach and would be happy to close without merging, but if you don't have an objection reducing the cost of the parsing I think would be good.

cc @Robbepop in relation to https://github.com/bytecodealliance/wasm-tools/issues/2180

Robbepop:
@alexcrichton thank you for trying to come up with a solution for this! :)

keithw:
Hmm... Honestly, I really tried to replicate a slowdown in parsing individual large modules, and to date my measurements (https://github.com/bytecodealliance/wasm-tools/issues/2180#issuecomment-2875193522) are that parsing **alone** doesn't seem to have gotten slower, we're talking about parsing a single huge module, e.g. https://github.com/turbolent/w2c2/blob/main/examples/clang/clang.wasm, vs. something like the testsuite which is about parsing thousands of tiny modules.

So my preference would be to find and be able to replicate the problem first because I think that will help inform how worth it is is to complicate the story and weaken the spec adherence. My view is that the spec defines a notion of well-formedness in chapter 5, and validity in chapter 3, and it would be nice if the parser conformed to chapter 5 by itself (and the validator conformed to chapter 3) rather than just saying that the parser+validator is going to jointly conform to chapter 5+3. This is particularly relevant to me because we're building a text editor that enforces the syntactic well-formedness on entry (and shows validation errors but doesn't prevent them), so the distinction matters to us for that reason. :-)

keithw:
To be a little more concrete, as far as I can tell so far, the slowdowns seem to be in "parsing and validating thousands of tiny modules" but not in "parsing one gigantic module." But I'm not sure if we're all on the same page about this? That's certainly what I was trying to figure out in #2180. :-)

If we are on the same page about where the slowdown is, then I guess my thinking would be that it's worth it to reduce the startup overhead of the Parser (maybe in being able to share its heap allocations across runs like the Validator can do?), but I'd be less eager to use startup overhead as a reason to reclassify a module like `(func else)` as well-formed if chapter 5 says otherwise.

Robbepop:
> Hmm... Honestly, I've really tried to replicate a slowdown in parsing individual large modules

@alexcrichton Did you also use `cargo bench -p wasmparser` instead of using the `wasm-tools` CLI tool as @alexcrichton [has commented earlier](https://github.com/bytecodealliance/wasm-tools/issues/2180#issuecomment-2894881384)?

I ran the benchmarks on an `Apple M2 Pro`, you ran them on a `AMD Ryzen 7 PRO 4750U`. What did you use @alexcrichton? Both alex and me saw could reproduce the regressions via `cargo bench -p wasmparser`.

> [..] parsing alone doesn't seem to have gotten slower, when we're talking about parsing a single huge module [..]
> [..] To be a little more concrete, as far as I can tell so far, the slowdowns seem to be in "parsing and validating thousands of tiny modules" but not in "parsing one gigantic module." [..]

I also wondered about this. It was a hypothesis so far since I could not proof if it really behaves like this. If true then indeed we should see nice speed-ups by introducing `ParserAllocations` similar to `ValidatorAllocations`. Thought, the parser's state looks fairly small in contrast to the validator's state.

> My view is that the spec defines a notion of binary well-formedness in chapter 5, and validity in chapter 3, and it would be nice if the parser enforced chapter 5 by itself, and the validator enforced chapter 3 [..]

I fully agree with you. It would certainly be better and we should ideally find a set of changes that allow for this while keeping good performance.

> If we are on the same page about where the slowdown is, then I guess my thinking would be that it's worth it to reduce the startup overhead of the Parser [..]

@alexcrichton has shown in a prior post which parts of the parser causes which slowdowns: https://github.com/bytecodealliance/wasm-tools/issues/2180#issuecomment-2894881384

I am not saying that the mentioned fixes should be implemented but they provide a clue to what is actually causing the regressions and at least in this comments it doesn't look like the regressions are stemming entirely from the startup of the parser.

Robbepop:
> In wamsparser using a variety of modules the regression is in the 5% range, while wasmi shows it's in the 30% range. I don't think any of us know at this time where such a difference comes from.

Indeed, that is very confusing to me as well. I tried to have minimal refactorings for the `wasmparser` update for Wasmi but maybe some things went sideways; OR (as seen prior) some old invariants are no longer upheld and since that once worked efficiently are not anymore.

One area where Wasmi uses `wasmparser` very differently from `wasmparser`'s own benchmarks is its lazy translation. And it is exactly where we see the most regressions in Wasmi.

I will re-benchmark Wasmi and take a deeper dive to see what is going on as soon as time and capacity permits. I will inform you about my findings (if any) at https://github.com/bytecodealliance/wasm-tools/issues/2180.

keithw:
Hi @alexcrichton,

I don't want to derail the practical questions of dealing with the performance issues in the code, but I feel like I owe you a response on these more policy-oriented questions -- I'm sorry I missed this the first time.

One high-level response would be to ask if you think it would be a good idea to start a #wasm-tools and #git-wasm-tools topic on the Bytecode Alliance Zulip. That might be a useful place to have longer-horizon discussions vs. GitHub PR/issue threads, especially among stakeholders from different projects.

> Ok that's all a bunch of background which leads me to my articulation: I'm not sure where to draw the line of where wasmparser can/should be different from the spec. Is the current state of "just a little different" ok? Is this PR's state of "ok a bit more significantly different" ok? Personally I don't really feel equipped to answer these questions myself as it feels like we're stuck between a rock and a hard place with the way the spec is developed and tested.

I think if we're looking for a semi-principled position, let me shop a proposal for an intermediate place to "draw the line," which happens to match exactly where wasm-tools is right now.

Namely: I think it's fine if wasmparser (without validation) over-accepts some malformed binary modules, _as long as these modules still have a plausible representation in the abstract syntax_ defined by Chapter 2. That means these binaries can be translated to well-formed text and back to binary and back to the same text as before. In other words, the exceptions are only relevant to the binary format.

So the current affordances (e.g. missing DataCount section, malformed mutability flags, integer too large) are all fine under that rubric, because those modules can still be represented in the abstract syntax, and they don't cause wasmprinter to output malformed text.

By contrast, if wasmparser allowed `(func else)`, that would be on the other side of this particular line. That module can't be represented in the abstract syntax of chapter 2 (and it's malformed even in the text format). So I think that why I was personally fine with the existing affordances but reluctant about relaxing this kind of check -- to me this is where I would draw the line.

(For context, my group is writing a structured editor to teach freshman CS where the editor enforces that the input buffer is always syntactically well-formed WAT, so when I picture myself in front of a classroom of 300 angry freshman trying to defend these distinctions to people who have never programmed before, I'm hoping for a certain moral clarity. I realize this isn't a typical use-case... I think you're totally right that a browser doesn't have to worry about these distinctions, but, the browser engines aren't trying to be general Wasm toolkits; they don't have text-mode parsers and I don't think they worry about roundtripping, or performing manipulations on well-formed Wasm modules that might take them through an invalid state.)

> Do you think that it would be worthwhile to bring these concerns to the CG perhaps? For example these are two possible (and radical) states to shift the development of the spec hypothetically:
> 
>     Remove the distinction between assert_{malformed,invalid} and instead have one single predicate of "a module is valid or not". Precisely how it's valid/invalid is irrelevant for the tests, engines, etc. This would amount to dropping the assert_malformed assertion and possible updating the wording of the specification in various places (maybe? unsure.)
>     Massage the specification to taking into account performance/maintainability concerns on behalf of engines in exactly how the AST of a module is constructed. For example I'd like block-like instructions to look more like normal instructions for this instead of having stricter syntactic rules. I'm not sure if this is even possible though.

I think a concern these would have to overcome is that most of the spec (including the validation and execution rules) are specified in terms of the abstract syntax of chapter 2. So if we wanted to allow stuff that isn't currently representable in the abstract syntax, that syntax would have to be widened to allow, and then the validation rules would have to be adapted to account for it.

E.g. looking at 2.4.8 (https://webassembly.github.io/spec/core/syntax/instructions.html#control-instructions), the abstract syntax for `instr` currently includes `if blocktype instr* else instr* end`. The "consequent" and "alternative" instruction sequences are components of the production, and the validation and execution rules later just refer to them as intrinsically present. Changing this to have `else` be a plain instruction that can syntactically appear anywhere would be I think a substantial change.

alexcrichton:
> One high-level response would be to ask if you think it would be a good idea to start a #wasm-tools and #git-wasm-tools topic on the Bytecode Alliance Zulip.

I think this would be and [excellent idea](https://bytecodealliance.zulipchat.com/#narrow/channel/511605-wasm-tools) (working on making that web-public)

> let me shop a proposal for an intermediate place to "draw the line," which happens to match exactly where wasm-tools is right now.

Ok that all sounds pretty reasonable to me. I agree with you that changing `else` or control flow would be a pretty big spec change and is likely not on the table at this time. I'm still slightly uncomfortable about data count bits but I can see it going either way myself really.

---

Overall I'm personally warming up to https://github.com/bytecodealliance/wasm-tools/pull/2228 so my current plan is to close this when that PR merges, but I'll leave this open in the meantime for now.

(thanks again for all the discussion on this!)

alexcrichton:
Closing in favor of https://github.com/bytecodealliance/wasm-tools/pull/2228

PR #2203, opened 2025-05-22
wasmparser OperatorsReader: remove unused `data_index_occurred`

keithw:
The OperatorsReader tracks whether a data index is used in the code section, but this is vestigial from an earlier version of #2134 that used this to enforce a well-formedness requirement. This eliminates the (basically unused) field.


alexcrichton (approved):
Good find!

PR #2204, opened 2025-05-23
wasmparser: avoid skipping locals to get OperatorsReader for a function body

keithw:
This PR lets users turn a finished LocalsReader into an OperatorsReader, avoiding a need to ask the FunctionBody to skip through the locals again (as suggested at https://github.com/bytecodealliance/wasm-tools/issues/2180#issuecomment-2894881384 -- @alexcrichton if you prefer a different/better way to do this please feel free to ignore this one!).

This seems to yield roughly a 3-4.5% improvement in the `parse/tests`, `parse/bz2`, `parse/spidermonkey`, etc. benchmarks.

The validator was already doing this (https://github.com/bytecodealliance/wasm-tools/blob/main/crates/wasmparser/src/validator/func.rs#L82-L87) and is unchanged by this PR, so I didn't expect this to affect validation time -- although I am seeing a ~1.5-2% slowdown in the validate tests for reasons unknown (hopefully just random variation from how this is being compiled).

### DIFF
--- a/crates/wasmparser/benches/benchmark.rs
+++ b/crates/wasmparser/benches/benchmark.rs
@@ -157,10 +157,11 @@ fn read_all_wasm(wasm: &[u8]) -> Result<()> {
                 }
             }
             CodeSectionEntry(body) => {
-                for item in body.get_locals_reader()? {
+                let mut locals = body.get_locals_reader()?.into_iter();
+                for item in locals.by_ref() {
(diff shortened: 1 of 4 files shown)


alexcrichton (approved):

alexcrichton:
Nah yeah this all looks reasonable to me, thanks!

PR #2215, opened 2025-06-02
Fix internal assertions renaming types using `with`

alexcrichton:
This commit fixes a fuzz bug that cropped up trying to rename a type using `with`. The destination type that was created didn't have its new name registered.


abrown:
I'm probably not the best WIT syntax reviewer. Who is?

alexcrichton:
Perhaps @fitzgen would you be up for taking a look?

Issue #2211, opened 2025-05-28
AddMetadata should be able to clear existing fields


vigoo:
### Feature request
I would like to modify `wasm_metadata`'s `AddMetadata` struct to be able to express _clearing_ some of the existing metadata fields.

### Why
The example use case I'm having is that previously I was storing custom _name_ and _version_ in WASM components as part of our component build flow, and by a recent change in `cargo-component`, it is now storing the crate name and version (from Cargo.toml) in the same metadata fields. I want our build flow to still store our name and version but unfortunately I am also rely on not having version in some cases (actually I'm storing the wit root package name and version - see https://github.com/WebAssembly/component-model/issues/488). So I need to be able to clear the version metadata put there by cargo-component, if there was no version part of the root wit. 

### Proposal
I made this change in a fork by introducing something like
```rust
pub enum AddMetadataField<T: Debug + Clone> {
    Keep,
    Clear,
    Set(T),
}
``` 

and using this in place of the `AddMetadata` struct's `Option` fields, happy to open a PR if needed.

alexcrichton:
This seems reasonable to me to add, thanks! And yes I'm happy to review a PR too. Bonus points if this can be integrated into the CLI too, but that's not required.

PR #2216, opened 2025-06-03
Ability to clear metadata fields

vigoo:
Resolves https://github.com/bytecodealliance/wasm-tools/issues/2211


crates/wasm-metadata/src/add_metadata.rs
```diff
@@ -27,48 +28,79 @@ pub struct AddMetadata {
...
     /// Version of the packaged software
-    #[cfg_attr(feature = "clap", clap(long, value_name = "NAME"))]
+    #[cfg_attr(feature = "clap", clap(long, value_name = "NAME", default_value_t, value_parser = parse_add_metadata_field::<crate::Version>))]
     #[cfg(feature = "oci")]
-    pub version: Option<crate::Version>,
+    pub version: AddMetadataField<crate::Version>,
 }
 
+#[cfg(feature = "clap")]
+pub const KEEP: &str = "::keep";
```
vigoo: This (and the `Display` instance below) is quite ugly, but that's the best I could come up with if we want to keep `AddMetadata` both the API and the clap interface because of https://github.com/clap-rs/clap/issues/4558

Alternatively we could have a separate struct for `clap` with option fields that can default to `None` without going through the value parser. Let me know if you prefer that.

pchickey: I agree the behavior here was hard to parse for me, and I don't understand how it would be rendered in the help text either. I don't love the `::keep` sigil on the cli. Without knowing clap's facilities for this very well: is there some way to get `--<name>=value` to have the current set behavior, and expose an additional `--clear-<name>` as an additional flag?

pchickey (commented):
Thanks for working on this, this is useful functionality to add but I want to understand some alternatives for the cli interface.

vigoo:
> is there some way to get --<name>=value to have the current set behavior, and expose an additional --clear-<name> as an additional flag?

I would prefer that command line interface too! (Currently the rendered help is the worst part as it shows these `::keep` (or whatever we choose) values which should not ever be actually used. Otherwise `--name value` sets the attribute and `--name ""` clears it; definitely not something I'm happy with)

Unfortunately (?) I think it is only possible if we give up that the `AddMetadata` struct is both used in the Rust API and for `clap`. Then we can define whatever fields and attributes are needed for clap to make the desired user experience.

Are you OK with that? 
Personally I think it's not a big loss compared to the much better user interface it could provide

pchickey:
Yes, please write whatever datastructures Clap requires to provide the ideal cli interface, and then translate the output of clap's parsing to the AddMetadata enums used by the Rust API. You can sequester all of the clap-specific intermediate representations into either a private mod, or a `#[cfg(feature = "clap")] #[doc(hidden)] pub mod __clap {...}` as needed.

pchickey (approved):
Looks great. Could you please add a quick cli test to cover the clap portion of this as well? for example https://github.com/bytecodealliance/wasm-tools/blob/main/tests/cli/add-metadata-merge-sections.wat

vigoo:
Added one

pchickey:
Thanks very much!

PR #2217, opened 2025-06-03
wasmparser(CM+GC): Add support for lowering variants, options, and results to GC types

fitzgen:
This completes prototype support for GC in the component model by filling out the lowering of sum types (variants, options, and results) to GC types.


alexcrichton:
Thanks for writing this up and the proposal upstream as well, cleared up all possible questions I'd have about variants 👍

fitzgen:
(Whoops I thought your approval was before the second commit, nvm)

PR #2222, opened 2025-06-06
update `future.{write,read}` ABIs

dicej:
This implements https://github.com/WebAssembly/component-model/pull/524 insofar as it affects validation, componentization, etc.

- `future.write` now accepts its payload value as up to 4 flat parameters before spilling to linear memory.

- `future.read` takes no payload pointer when it has no payload type

- `{stream,future}.close-{readable,writable}` have been renamed to `{stream,future}.drop-{readable,writable}`


dicej:
CI failures appear to be due to me using the latest Rust release; will downgrade to the MSRV locally and fix the issues.

crates/wasmparser/src/validator/component.rs
```diff
@@ -1776,19 +1783,50 @@ impl ComponentState {
...
+            types,
+            &CanonicalOptions {
+                string_encoding: StringEncoding::Utf8,
+                memory: options.memory,
+                realloc: None,
+                post_return: None,
+                concurrency: Concurrency::Async { callback: None },
+                core_type: None,
+                gc: false,
+            },
```
alexcrichton: Instead of creating a new `CanonicalOptions` could this `clone()` and manipulate the `options` above? I mostly want to keep things in sync in terms of gc/string encoding/etc for future compatibliity

alexcrichton (approved):
Looks good 👍 

In the `*.wast` tests I think I only saw renamings, so could you add tests for futures for both reads/writes to ensure properties such as `future.read` has no pointer with no payload and `future.write` uses flat lowerings or a pointer if the flat lowering is too big? Additionally I think it'd be good to have a test that `future.write` doesn't require memory if the lowering fits

crates/wasmparser/src/validator/component.rs
dicej: I was thinking the same thing, but then check with Luke, who confirmed that we use `MAX_FLAT_ASYNC_PARAMS` for `future.write` _regardless_ of whether the `future.write` is called sync or async.  I guess I could still clone, but then force the `concurrence` to async.

alexcrichton: Oh right yeah I think this would still overwrite some options, but I want to be sure to preserve options that aren't otherwise overwritten for the purposes of the internal API calls here (e.g. the canonical ABI for GC strings depends on the string encoding which isn't preserved here, nor is GC)

dicej:
> In the `*.wast` tests I think I only saw renamings, so could you add tests for futures for both reads/writes to ensure properties such as `future.read` has no pointer with no payload and `future.write` uses flat lowerings or a pointer if the flat lowering is too big? Additionally I think it'd be good to have a test that `future.write` doesn't require memory if the lowering fits

FWIW, I've tested those cases end-to-end in wasip3-prototyping, plus the fuzz tester is hitting those cases also.  I agree we should have some handwritten WAST tests as well; I'll add those.

alexcrichton:
Yeah all the code here looks correct so I don't expect any failures, mostly just handwritten tests for the well-known edge cases are good to have to catch any possible issues in future refactorings earlier

crates/wasmparser/src/validator/component.rs
dicej: The other wrinkle is that `ComponentFuncType::lower` may return a `LoweredFuncType::Existing(_)` rather than `LoweredFuncType::New(_)` when `gc` is `true`, and I couldn't figure out how to extract the core parameter list from the former.  Do you have any tips?

alexcrichton: Nah that's fine to leave as a panic for now, it's more the silent/subtle bugs of switching options by accident I'm worried about. Overall this is a really minor concern so if you'd prefer it's ok to leave as-is.

PR #2224, opened 2025-06-06
Revert "update `future.{write,read}` ABIs (#2222)"

dicej:
This reverts commit 95f36d27a71c91cd210dd50bd5a26001875c61a4.

Alex, Luke, and I decided to scale back these changes.  I'll open a new PR next week which reflects that discussion.


alexcrichton:
Would it be easier to leave as-is and only update with the back-out of the future.write ABI changes? (e.g. keeping the renames basically)

dicej:
> Would it be easier to leave as-is and only update with the back-out of the future.write ABI changes? (e.g. keeping the renames basically)

Yes, it would be easier overall, except I won't have time to do that until Monday, and I was concerned that other PRs might be merged before that, making it harder to partially revert changes.  If you're not concerned about that, we can close this.

alexcrichton:
Famous last words: I'm not worried

(aka I'll take responsibility if things come up)

Issue #2235, opened 2025-06-12
`wasmparser` reports incorrect arity for `end` instruction for `if` with no `else`?


fitzgen:
Given the following Wasm module, `wasmparser` reports the `if`'s `end` as popping 2 values and pushing 2 values, but I think it should be popping and pushing 1 value, no?

```wat
                (module
                    (func (export "f") (param i64 i32) (result i64)
                        (local.get 0)
                        (local.get 1)
                        ;; If with no else. Same number of params and results.
                        (if (param i64) (result i64)
                            (then
                                (drop)
                                (i64.const -1)
                            )
                        )
                    )
                )
```

cc @keithw

fitzgen:
Here is a WIP regression test:

```rust
    #[test]
    fn arity_if_no_else_same_params_and_results() {
        let wasm = wat::parse_str(
            r#"
                (module
                    (func (export "f") (param i64 i32) (result i64)
                        (local.get 0)
                        (local.get 1)
                        ;; If with no else. Same number of params and results.
                        (if (param i64) (result i64)
                            (then
                                (drop)
                                (i64.const -1)
                            )
                        )
                    )
                )
            "#,
        )
        .unwrap();
        assert!(Validator::new().validate_all(&wasm).is_ok());

        let expected_arity = vec![vec![
            // local.get 0
            (0, 1),
            // local.get 1
            (0, 1),
            // if
            (2, 1),
            // drop
            (1, 0),
            // i64.const -1
            (0, 1),
            // end
            (1, 1),
            // implicit end
            (1, 1),
        ]];

        let parser = Parser::new(0);
        let mut validator = Validator::new();

        let mut actual_arity = vec![];

        for payload in parser.parse_all(&wasm) {
            let payload = payload.unwrap();
            match payload {
                crate::Payload::CodeSectionEntry(body) => {
                    let mut arity = vec![];
                    let mut func_validator = validator
                        .code_section_entry(&body)
                        .unwrap()
                        .into_validator(FuncValidatorAllocations::default());
                    let ops = body.get_operators_reader().unwrap();
                    for op in ops.into_iter() {
                        let op = op.unwrap();
                        dbg!(&arity);
                        arity.push(
                            op.operator_arity(&func_validator)
                                .expect("valid operators should have arity"),
                        );
                        func_validator.op(usize::MAX, &op).expect("should be valid");
                    }
                    actual_arity.push(arity);
                }
                p => {
                    validator.payload(&p).unwrap();
[...]

PR #2236, opened 2025-06-12
Simplify arity implementation

alexcrichton:
Originally the annotations in the `for_each_operator!` macro were intended to be used to implement the validator in addition to arity, but during the PR this part was removed and the annotations were only used for arity. As such the macro/annotations were a bit more complicated than they might otherwise need to be. Coupled with #2235 where some changes may be needed to arity calculations the macro was a bit difficult to understand. As a result this commit simplifies the arity annotations to either:

* `arity N -> N`
* `arity custom`

All "custom" instructions are now written as Rust functions in the `arity.rs` module. They delegate to each other internally as appropriate in a similar manner to how the macro calculated before.

Overall this passes tests but I can clarify that I only have a tenuous understanding at best of the previous macro. I did my best to faithfully translate it from what it used to be but it's possible I got something wrong along the way.


keithw:
On changing the macro, no real objection from me if you prefer this style. I did think there was some elegance in being able to define the arity of each instruction in terms of a small number of composable primitives, e.g.:

```rust
Block { blockty: $crate::BlockType } => visit_block (arity block -> ~block)`
Loop { blockty: $crate::BlockType } => visit_loop (arity block -> ~block)
If { blockty: $crate::BlockType } => visit_if (arity 1 block -> ~block)
```

vs.

```rust
Block { blockty: $crate::BlockType } => visit_block (arity custom)
Loop { blockty: $crate::BlockType } => visit_loop (arity custom)
If { blockty: $crate::BlockType } => visit_if (arity custom)
```

and having a separate custom function for each one of these. But I think this is a matter of taste, and I'm new to Rust, so fine with me if you want to change it to a less macro-heavy approach.

What I would prudentially maybe avoid is changing the FuncValidator code (that tests the correctness of the arity computation) at the same time as changing the arity computation itself. Although I realize it's a pretty small change here so the risk is low.

dicej (approved):

fitzgen (approved):
I find this version easier to follow, thanks!

alexcrichton:
@keithw for me and historically my experience in Rust has been that heavy usage of macros can lead to difficult-to-understand code so I try to avoid it where possible, and then if a big macro is necessary I try to keep it as simple as possible. In that sense I do agree with your sentiments but the way I would frame it is that the building blocks are the `ModuleArity` trait and instead of a custom DSL (e.g. `block -> ~block`) the definitions are now Rust functions. It's definitely wordier than before but understanding/modifying arity doesn't require first learning the DSL now which I think will be good for maintainability over time.

Also agreed on not changing `FuncValidator`! I improved the error message but otherwise have no intention on changing it more deeply.

PR #2237, opened 2025-06-12
wasmparser: Fix arity for `if`'s `end` with no `else` and `br_table`

fitzgen:
This also makes the push/pop tracking ignore "temporary" pushes and pops that only happen as an internal detail of an instruction's validation and are not exposed outside the instruction's encapsulation.

Fixes https://github.com/bytecodealliance/wasm-tools/issues/2235

This will need a rebase now that #2236 merged.

### DIFF
--- a/crates/wasmparser/src/validator/operators.rs
+++ b/crates/wasmparser/src/validator/operators.rs
@@ -56,6 +56,10 @@ pub(crate) struct OperatorValidator {
     /// Whether validation is happening in a shared context.
     shared: bool,
 
+    /// A trace of all operand push/pop operations performed while validating an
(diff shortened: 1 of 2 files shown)


fitzgen:
Rebased.

crates/wasmparser/src/validator/operators.rs
```diff
@@ -57,7 +57,7 @@ pub(crate) struct OperatorValidator {
     shared: bool,
 
     #[cfg(debug_assertions)]
-    pub(crate) pop_push_count: (u32, u32),
+    pub(crate) pop_push_log: Vec<bool>,
```
alexcrichton, line 64: Mind leaving a comment for what the `bool` values are here?

PR #2251, opened 2025-06-28
wasm-tools print: add option to print operand stack between instructions

### LINKED ISSUE #1381: Add option to show operand stack's types in disassembly
When debugging some fiddly wasm bits I often annotate the WAT disassembly with the types on the stack in between each instruction:

```wat
;; []
i32.const 32
;; [i32]
i32.const 10
;; [i32 i32]
i32.add
;; [i32]
```

Would be nice if we added a flag to `wasm-tools print` to do this automatically.

I guess `wasmprinter` can handle invalid Wasm at the moment, so we would have to maybe make this best effort or require valid input for when this is enabled or something.

keithw:
Closes #1381 (mentioned in https://github.com/bytecodealliance/wasm-tools/issues/2235#issuecomment-2968469805)

Example output:

image


alexcrichton:
This is awesome, thank you so much for implementing this! Overall this looks great and the points I'd comment on are mostly Rust-related so I went ahead and sketched them out and PR'd as https://github.com/keithw/wasm-tools/pull/2. Notably I personally prefer to cut down on `#[cfg]` wherever possible, so that uses a strategy of having two modules, one which is just a skeleton, for handling the enabled-vs-not code with respect to runtime dependencies. I also added a few minor tests but nothing major.

keithw:
That's a nice technique! Thanks, merged your PR.

alexcrichton (approved):

fitzgen:
This is fantastic!! Thanks so much @keithw!!

Issue #2250, opened 2025-06-25
Wasm-smith's `--exports` option does not support exporting globals with GC types


mobsceneZ:
When providing the following module to the `--exports` option, `wasm-smith` panicked with the error messsage "***called Result::unwrap() on an Err value: CanonicalizedHeapTypeReference***":

```wat
(module
  (rec (type $t0 (sub final (array i32))))
  (global $g1 (mut (ref null $t0)) ref.null $t0)
  (export "global$1" (global $g1))
)
```

A quick look at the source code reveals that the problem is with the **`entity_type_from_export()`** function:
https://github.com/bytecodealliance/wasm-tools/blob/bc0a78578a764b3f43489e210066bc7ea2449ac5/crates/wasm-smith/src/core.rs#L1971-L1975

It returns a **`wasmparser::GlobalType`** that contains canonicalized **`CoreTypeId`**. Then, the **`TryFrom`** trait implementation is called to convert the original **`wasmparser::GlobalType`** into **`wasm_encoder::GlobalType`**, but the conversion demands *type space indices* rather than *core type ids*, which eventually results in a panic.

I guess maybe we could remove the validation code inside **`_required_exports`** function? The **`_arbitrary_imports_from_available`** function doesn't seem to perform any validity checks too, so probably it's ok to do so 🤔.

alexcrichton:
Yes handling of `--exports` for wasm GC will need some work. The type section of the input wasm module effectively needs to be copied over into the generated module (as opposed to just function types from today). I think it goes a bit beyond just validating a few things and structurally needs to be implemented. That being said it would be great to have!

mobsceneZ:
> Yes handling of `--exports` for wasm GC will need some work. The type section of the input wasm module effectively needs to be copied over into the generated module (as opposed to just function types from today). I think it goes a bit beyond just validating a few things and structurally needs to be implemented. That being said it would be great to have!

Thanks for the reply! I just want to check if I understand your point correctly, are you suggesting that we copy the type section from the input wasm module and discard the validation part, similar to what **`_arbitrary_imports_from_available`** function does? That seems to be the simplest solution I can think of that doesn't require significant modifications to the source code 🤔.

fitzgen:
That would make sense except that this would then mean that using both available imports and `--exports` with GC types would be incompatible with each other, as they would both try to copy over their respective type sections and only one or the other would have its types' indices preserved in the new types index space.

[...]

PR #2255, opened 2025-07-14
wasm-smith: add --module-shape option.

mobsceneZ:
This PR fixes https://github.com/bytecodealliance/wasm-tools/issues/2250 and the key changes are summarized as follows:

- Add a new option called **`module_shape`** into wasm-smith which automatically disables **`available_imports`** and **`exports`** when enabled;
- For simplicity, we handle the imports and exports declared in the shape module together in a single function: ***_imports_exports_from_module_shape***. And this function presents two notable differences when compared with function ***_arbitrary_imports_from_available*** and ***_required_exports***:
  1. To adhere more closely to the meaning of "module shape", it copies ***all*** declared imports into the target module, rather than randomly selecting a subset. As a result, defined limits such as *max_funcs* may be exceeded;
  2. Relying on **`wasmparser::Validator`** for type information, as the ***_required_exports*** does, would introduce the inconvenient step of converting **CoreTypeId** back to **ModuleIndex**. To avoid this, this function directly copies the entire type section from the shape module to the target module, which also aligns with the implementation of ***_arbitrary_imports_from_available***;

Please feel free to share any suggestions/concerns.

### DIFF
--- a/crates/wasm-smith/src/core.rs
+++ b/crates/wasm-smith/src/core.rs
@@ -441,15 +441,22 @@ impl Module {
     fn build(&mut self, u: &mut Unstructured) -> Result<()> {
         self.valtypes = configured_valtypes(&self.config);
 
+        let mut generate_arbitrary_imports = true;
+        let mut generate_arbitrary_exports = true;
+        if self.imports_exports_from_module_shape(u)? {
+            generate_arbitrary_imports = false;
+            generate_arbitrary_exports = false;
+        }
         // We attempt to figure out our available imports *before* creating the types section here,
         // because the types for the imports are already well-known (specified by the user) and we
         // must have those populated for all function/etc. imports, no matter what.
         //
         // This can affect the available capacity for types and such.
         if self.arbitrary_imports_from_available(u)? {
-            self.arbitrary_types(u)?;
-        } else {
-            self.arbitrary_types(u)?;
+            generate_arbitrary_imports = false;
+        }
+        self.arbitrary_types(u)?;
+        if generate_arbitrary_imports {
             self.arbitrary_imports(u)?;
         }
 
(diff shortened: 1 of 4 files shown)


crates/wasm-smith/src/core.rs
```diff
@@ -1182,6 +1192,289 @@ impl Module {
...
+
+        #[cfg(feature = "wasmparser")]
+        {
+            self._imports_exports_from_module_shape(u, &example_module)?;
+            Ok(true)
+        }
+        #[cfg(not(feature = "wasmparser"))]
+        {
+            let _ = (example_module, u);
+            panic!("support for `module_shape` was disabled at compile time");
```
dicej, line 1210: There are a lot of `panic!` and `expect` expressions in this function and in `_imports_exports_from_module_shape` despite the fact that they both return `Result<_>` types.  What's the reason for panicking rather than returning `Err` in those cases?

mobsceneZ: > There are a lot of `panic!` and `expect` expressions in this function and in `_imports_exports_from_module_shape` despite the fact that they both return `Result<_>` types. What's the reason for panicking rather than returning `Err` in those cases?

Well, the reason why I use `panic!()` and `expect()` is to align with the current implementation of wasm-smith. As you can see in the source code of ***_required_exports*** or ***_arbitrary_imports_from_available***, both of them heavily use `panic!()` to report errors, e.g.:
https://github.com/bytecodealliance/wasm-tools/blob/cef5a0cf7573312e8feded47c5a55f1a339e6966/crates/wasm-smith/src/core.rs#L1363-L1376

dicej: Good point; I hadn't noticed that.

dicej (approved):

Issue #2257, opened 2025-07-21
Add linux musl to release packages


thomas9911:
Currently we use an Alpine docker image with wasm-tools installed. Currently we have to manually build an package this because the target linux gnu doesn't work in Alpine. It would be nice that wasm-tools also builds and releases the binary with musl instead of gnu.

alexcrichton:
Would you be up for helping to add this release binary? It should mostly be some edits to CI configuration here and there

thomas9911:
created a pull request, but I think there are some changes that need to made to support aarch64-musl

alexcrichton:
Added in https://github.com/bytecodealliance/wasm-tools/pull/2259 (thanks!)

PR #2259, opened 2025-07-22
fix: added musl builds x86_64 and aarch64

thomas9911:
https://github.com/bytecodealliance/wasm-tools/issues/2257

The build are copied from https://github.com/bytecodealliance/wasmtime/blob/v29.0.1/ci/build-build-matrix.js

I see it uses version 28.0 but I don't see aarch64-musl in that tag. However in 29.0 I dont see the 'bytecodealliance/wasmtime/.github/actions/binary-compatible-builds@release-28.0.0' workflow anymore 🤔

### DIFF
--- a/ci/docker/aarch64-musl/Dockerfile
+++ b/ci/docker/aarch64-musl/Dockerfile
@@ -1,0 +1,5 @@
+FROM ghcr.io/cross-rs/aarch64-unknown-linux-musl
+
+RUN apt-get update -y && apt-get install -y ninja-build
+RUN git config --global --add safe.directory '*'
+ENV EXTRA_RUSTFLAGS=-Ctarget-feature=-crt-static
--- a/ci/docker/aarch64-musl/Dockerfile
+++ b/ci/docker/aarch64-musl/Dockerfile
@@ -2,4 +2,4 @@ FROM ghcr.io/cross-rs/aarch64-unknown-linux-musl
 
 RUN apt-get update -y && apt-get install -y ninja-build
 RUN git config --global --add safe.directory '*'
-ENV RUSTFLAGS=-Ctarget-feature=-crt-static
+ENV EXTRA_RUSTFLAGS=-Ctarget-feature=-crt-static
--- a/ci/docker/aarch64-musl/Dockerfile
+++ b/ci/docker/aarch64-musl/Dockerfile
@@ -2,4 +2,4 @@ FROM ghcr.io/cross-rs/aarch64-unknown-linux-musl
 
 RUN apt-get update -y && apt-get install -y ninja-build
 RUN git config --global --add safe.directory '*'
-ENV EXTRA_RUSTFLAGS=-Ctarget-feature=-crt-static
+ENV RUSTFLAGS=-Ctarget-feature=-crt-static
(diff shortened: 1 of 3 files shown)


thomas9911:
ah I see it uses wasm-tools (I though it would use wasmtime) to get the builder docker images

alexcrichton:
Ah yeah the Wasmtime action in-use here basically just builds some Docker containers in `ci/docker/*`, so copying over the files from Wasmtime should be sufficient

ci/docker/aarch64-musl/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM ghcr.io/cross-rs/aarch64-unknown-linux-musl
+
+RUN apt-get update -y && apt-get install -y ninja-build
+RUN git config --global --add safe.directory '*'
+ENV EXTRA_RUSTFLAGS=-Ctarget-feature=-crt-static
```
alexcrichton, line 5: This is a bit different from Wasmtime in that `EXTRA_RUSTFLAGS` isn't recognized in this repository, could this be updated to `RUSTFLAGS` and/or could `EXTRA_RUSTFLAGS` get handled? (as-is static binaries are produced when it's intended that dynamic binaries are produced instead)

ci/docker/x86_64-musl/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM ubuntu:16.04
+
+RUN apt-get update -y && apt-get install -y gcc gcc-aarch64-linux-gnu ca-certificates git
+RUN git config --global --add safe.directory '*'
+
+ENV PATH=$PATH:/rust/bin
+ENV CARGO_BUILD_TARGET=aarch64-unknown-linux-gnu
+ENV CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc
```
alexcrichton: I think there may have been a mixup maybe? These variables look like they should be in the above dockerfile, and down here I think the `-crt-static` is needed.

thomas9911: I did the classic not saving the file :') will commit the correct one

ci/docker/aarch64-musl/Dockerfile
thomas9911: updated both to use RUSTFLAGS instead

alexcrichton:
Bah static binaries are still popping out. I believe the reason for that is that the wrapper script here will by default [inherit rustflags from the calling process](https://github.com/bytecodealliance/wasmtime/blob/8bc30af15d22adaa7f7dfdf719934aecc58bf23c/.github/actions/binary-compatible-builds/main.js#L36-L43) (which for CI is `-D warnings`) although below that I found [handling of `EXTRA_RUSTFLAGS`](https://github.com/bytecodealliance/wasmtime/blob/8bc30af15d22adaa7f7dfdf719934aecc58bf23c/.github/actions/binary-compatible-builds/main.js#L50), so actually I think `EXTRA_*` will work and the reason it didn't prior was that the architectures were swapped perhaps?

ci/docker/aarch64-musl/Dockerfile
thomas9911: reverted back to EXTRA_RUSTFLAGS

PR #2268, opened 2025-07-30
feat(wasm-encoder): expose Ieee32 and Ieee64 constructors

michaelficarra:
This allows consumers who already have their float bit pattern as a `u32`/`u64` to not have to go through a cast to `f32`/`f64` just for the `Ieee32` or `Ieee64` constructors to cast it back. If being able to directly read/write the stored bit pattern is undesirable, I could impl `new` instead. It doesn't look like #2157 added any tests for this struct, so I didn't add any either.


alexcrichton (commented):
Thanks! Instead of making fields public could constructors be added? That'll provide ideally more flexibility in the future while also being a bit more idiomatic.

Issue #2274, opened 2025-08-01
unresolved symbol when using wasi-sdk-27


brian14708:
I've been testing out [wasi-sdk](https://github.com/WebAssembly/wasi-sdk) 26/27, when trying to use the dynamic linking features this error occurs:

```
Error: unresolved symbol(s):
        libc++abi.so needs __cxa_thread_atexit_impl (function [] -> [])
```

here the sample code

```rust
fn linking_libcppabi() -> Result<()> {
    // libc.so and libc++abi.so from ./wasi-sdk-27.0-x86_64-linux/share/wasi-sysroot/lib/wasm32-wasip1/
    wit_component::Linker::default()
        .validate(true)
        .library("libc.so", include_bytes!("../libc.so"), false)?
        .library("libc++abi.so", include_bytes!("../libc++abi.so"), false)?
        // .stub_missing_functions(true)
        .encode()?;

    Ok(())
}
```

even when `stub_missing_functions`, the program fails with

```
thread 'linking_libcpp' panicked at crates/wit-component/src/linking.rs:1167:35:
called `Result::unwrap()` on an `Err` value: BinaryReaderError { inner: BinaryReaderErrorInner { message: "duplicate export name `__cxa_thread_atexit_impl` already defined", kind: Custom, offset: 82, needed_hint: None } }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

it works on when using wasi-sdk 25, from a preliminary analysis the new `libc++abi.so` contains the following

```wat
(module $libc++abi.so
  (@dylink.0
    (mem-info (memory 16496 4) (table 262 0))
    (needed "libc.so" "libc.so")
    (import-info "env" "__cxa_thread_atexit_impl" binding-weak undefined)
  )
  ...
```

alexcrichton:
This might best be a wasi-sdk issue for the difference between 25/26/27? I'm not sure if this is something that'd be fixable on the wasm-tools side of things myself

brian14708:
I think I've found the problem, the type of `__cxa_thread_atexit_impl` is actually `[i32, i32, i32] -> i32`

https://github.com/bytecodealliance/wasm-tools/blob/c2fc211e3259292ea91afdabacd93b1e1c7470fb/crates/wit-component/src/linking.rs#L915-L944

inserts the `missing` as `[] -> []`


my attempt at minimizing the bug and fix is at https://github.com/bytecodealliance/wasm-tools/pull/2275

PR #2275, opened 2025-08-02
handle GOT entries for weak env imports in symbol resolution

brian14708:
see https://github.com/bytecodealliance/wasm-tools/issues/2274

for reference the wat is generate using

```c
// clang -fvisibility=default --target=wasm32-wasip1 -O2 -shared -fPIC -nostartfiles test.c

__attribute__((weak))
extern int foobar();

int foo() { return foobar ? foobar() : 0; }
```

### DIFF
--- a/crates/wit-component/src/linking.rs
+++ b/crates/wit-component/src/linking.rs
@@ -48,6 +48,7 @@ const PAGE_SIZE_BYTES: u32 = 65536;
 // This matches the default stack size LLVM produces:
 pub const DEFAULT_STACK_SIZE_BYTES: u32 = 16 * PAGE_SIZE_BYTES;
 const HEAP_ALIGNMENT_BYTES: u32 = 16;
+const STUB_LIBRARY_NAME: &str = "wit-component:stubs";
 
 enum Address<'a> {
     Function(u32),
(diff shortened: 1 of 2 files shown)


dicej:
> Thanks for this! @dicej might be able to review this best as he originally wrote much of this, but naively I'd expect some sort of handling of "weak" symbols specifically here but it instead seems to be doing it base on the presence or non-presence of a definition?

The handling of weak symbols happens later, [here](https://github.com/bytecodealliance/wasm-tools/blob/a2d966c7ba79497953375006a2075a6ed52e9cfc/crates/wit-component/src/linking.rs#L1374-L1376).

This patch works because we will already have added the symbol to the `missing` list [here](https://github.com/bytecodealliance/wasm-tools/blob/a2d966c7ba79497953375006a2075a6ed52e9cfc/crates/wit-component/src/linking.rs#L883-L894); adding it a second time would be redundant, plus the redundant entry would be missing the `BINDING_WEAK` flag.

dicej (approved):
Thanks!

dicej:
@brian14708 Have you tried actually running the code in your example?  I'm afraid it might not work, since [this code](https://github.com/bytecodealliance/wasm-tools/blob/a2d966c7ba79497953375006a2075a6ed52e9cfc/crates/wit-component/src/linking.rs#L1374-L1376) will generate a module of trapping stubs for any missing exports (including weakly-imported ones) and thereby generate a non-zero `GOT.func` value for the function.

dicej:
I just updated the test to actually export the `foo` function from the component and confirmed that it fails:

```rust
const GOT_IMPORT: &str = r#"
(module
  (@dylink.0
    (mem-info)
    (needed "libc.so")
    (import-info "env" "foobar" binding-weak undefined)
  )
  (type (;0;) (func (result i32)))
  (import "env" "foobar" (func (;0;) (type 0)))
  (import "GOT.func" "foobar" (global (;0;) (mut i32)))
  (export "foo" (func 1))
  (func (;1;) (type 0) (result i32)
    global.get 0
    i32.eqz
    if ;; label = @1
      i32.const 0
      return
    end
    call 0
  )
)
"#;

const GOT_IMPORT_WIT: &str = r#"
package test:test;

world bar {
    export foo: func() -> u32;
}
"#;

#[test]
fn linking_got_weak() -> Result<()> {
    let component = [
        ("libfoo.so", GOT_IMPORT, Some(GOT_IMPORT_WIT)),
        ("libc.so", LIBC, None),
    ]
    .into_iter()
    .try_fold(
        wit_component::Linker::default().validate(true),
        |linker, (name, wat, wit)| {
            linker.library(
                name,
                &encode(wat, wit).with_context(|| name.to_owned())?,
                false,
            )
        },
    )?
    .encode()?;
    std::fs::write("/tmp/foo.wasm", &component)?;
    Ok(())
}
```

```
 $ cargo test -p wit-component
...
 $ wasmtime run --invoke 'foo()' /tmp/foo.wasm
Error: failed to run main module `/tmp/foo.wasm`

Caused by:
    0: error while executing at wasm backtrace:
           0:    0x2c0 - wit-component:stubs!<wasm function 0>
           1:    0x38b - libfoo.so!<wasm function 1>
    1: wasm trap: wasm `unreachable` instruction executed
```

I think we'll need to modify [this code](https://github.com/bytecodealliance/wasm-tools/blob/a2d966c7ba79497953375006a2075a6ed52e9cfc/crates/wit-component/src/linking.rs#L403-L412) to use a value of zero for unsatisfied weak imports (i.e. those satisfied by the stub module).

brian14708:
@dicej Thanks for the feedback. I've integrated the test and added the "resolve-to-null" semantics in `make_env_module`, but I'm not sure if it the best way to implement this.

PR #2281, opened 2025-08-13
Use trusted publishing for release automation in this repository

alexcrichton:
This commit updates the publish scripts and CI configuration to use the "Trusted Publishing" workflow on crates.io. This enables removal of the long-lived token assigned to this repository in favor of short-lived ephemeral tokens scoped to just the `publish.yml` workflow.

Here CI configuration is updated to avoid using the token in the environment. Instead it uses `rust-lang/crates-io-auth-action@v1` to acquire a token during publishing. Documentation is added for how to add a crate to this repository as it now requires that the crate is published on crates.io before it's added to this repository. That ensures that the trusted publishing workflow is in place before a PR is merged. Finally the `publish.rs` script is updated to avoid owner management on version bumps but also check on PRs that the owners are configured correctly.

### DIFF
--- a/.github/actions/publish-release/action.yml
+++ b/.github/actions/publish-release/action.yml
@@ -1,11 +1,6 @@
 name: 'Maybe perform a release'
 description: 'Steps to perform a conditional release of this repository'
 
-inputs:
-  cargo_token:
-    description: 'token used to publish crates'
-    required: false
-
 runs:
   using: composite
   steps:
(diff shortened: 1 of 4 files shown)


alexcrichton:
I expect CI to fail temporarily as the [wasm-encoder crate](https://crates.io/crates/wasm-encoder) doesn't have permissions set up. That'll be resolved shortly, however.

After this PR merges I plan on publishing a point release to ensure that this all works.

pchickey (approved):

PR #2283, opened 2025-08-15
Add support for the WASM_DYLINK_RUNTIME_PATH subsection of the dylink.0 custom section

Arshia001:
Hi!

This PR adds support for the WASM_DYLINK_RUNTIME_PATH subsection. `wasm-ld` 21 already supports generating this subsection via the `-rpath` arg, so I feel it's important for `wasm-tools` to support the subsection as well.

I don't know of any tools that generate the subsection in WAT format yet, but judging by the information in https://github.com/WebAssembly/tool-conventions/blob/main/DynamicLinking.md#text-format-of-dylink0, the correct format should be `(runtime-path "a" "b")`; please correct me if I'm wrong.

I believe this is a breaking change to `wasmparser`, in the sense that adding a new case to `Dylink0Subsection` breaks existing code that does exhaustive matching on it; however, I don't know if there's a better way to handle this.


alexcrichton (approved):
Thanks! And no worries about the breaking change, it's expected that this crate has breaking changes over time as wasm evolves.

PR #2286, opened 2025-08-21
WIT: Add support for fallible constructors

badeend:
This implements: https://github.com/WebAssembly/component-model/pull/550

---

**Edit:** What is the expected workflow regarding adding new test cases to `tests/cli/component-model/resources.wast`? If I understand the CI failure correctly, I was not supposed to touch the `tests/snapshots/cli/component-model/resources.wast/*.print` files.


badeend:
Yup, that did it

PR #2288, opened 2025-08-25
Support lists of WITs in `Resolve::select_world`

sunfishcode:
This extends `Resolve::select_world` to be usable for lists of WITs. One such example is wit-bindgen's `generate` macro's `path:` parameter, which can take a list of WIT paths. It currently uses a [custom `select_world` implementation], but with this patch in wit-parser, it would be able to use `Resolve::select_world` instead.

I'm also considering extending the wit-bindgen CLI to accept multiple WIT paths; with this patch, it could use `Resolve::select_world` as well.

[custom `select_world` implementation]: https://github.com/bytecodealliance/wit-bindgen/blob/0e2201b8629b4144f45867161de7bc1fe26133bb/crates/guest-rust/macro/src/lib.rs#L191

Specifically, this PR makes the package argument to `Resolve::select_world` an `Option`, and adds code to handle the case where it's `None`. A `None` means the caller has a list of WITs and as such doesn't have a single main package.

This also rewrites the documentation for `select_world` to clean up already-obsolete references to the `packages` array argument, and to hopefully make it more clear how `select_world` makes its choice.

### DIFF
--- a/crates/wit-component/src/encoding.rs
+++ b/crates/wit-component/src/encoding.rs
@@ -2971,7 +2971,7 @@ world test {
 "#,
             )
             .unwrap();
-        let world = resolve.select_world(Some(pkg), None).unwrap();
+        let world = resolve.select_world(&[pkg], None).unwrap();
 
         let mut module = dummy_module(&resolve, world, ManglingAndAbi::Standard32);
 
--- a/crates/wit-component/src/encoding.rs
+++ b/crates/wit-component/src/encoding.rs
@@ -2971,7 +2971,7 @@ world test {
 "#,
             )
             .unwrap();
-        let world = resolve.select_world(pkg, None).unwrap();
+        let world = resolve.select_world(Some(pkg), None).unwrap();
 
         let mut module = dummy_module(&resolve, world, ManglingAndAbi::Standard32);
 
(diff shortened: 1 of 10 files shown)


alexcrichton:
I know that this function has a long and storied history of flip-flopping between various signatures and semantics, and I also don't remember everything as well. WIth this PR though it has a subtly different semantics than the `generate!` macro where a world name isn't specified where with `generate!` it works if only one of the "main packages" has a world where with this PR it only works if there's one world in the entire `Resolve`. I can't quite place my finger on why but that feels like a crucial difference to me where, for example, WASI APIs might frequently be deps of custom WITs which always have worlds but maybe the custom WITs wouldn't have such a world?

Basically: do you think the "main package" argument should be a "main packages" list instead of `Option<PackageId>`? I forget where that lands on "yes we tried this a year ago and moved away from it" spectrum but it would solve what I'm thinking here because the "no world specified, multiple packages, check for just one world" part of this selection would only consult the list of packages provided, no others.

sunfishcode:
Going by the git logs, the switch to using one main package was in #1700, which introduced the idea of requiring a single root package, which I think still makes sense within a given WIT directory.

I think what's new now is that we're also supporting lists of independent WIT directories, and at this level, there may not be a single root package connecting them all. With that understanding, I think it makes sense to switch back to a list of main packages, so I've now updated the patch to do this.

alexcrichton (approved):
Thanks for digging that up!

alexcrichton:
Oh I should also mention:

> I'm also considering extending the wit-bindgen CLI to accept multiple WIT paths

Sounds reasonable to me!

PR #2295, opened 2025-09-02
Implementation of the custom descriptors proposal

yurydelendik:
Draft implementation of the proposal that can be found at https://github.com/WebAssembly/custom-descriptors/blob/main/proposals/custom-descriptors/Overview.md.

Keeping it as draft:
- [x] `PackedIndex` or `RefType` mask extension for exact type is okay
- [ ] Ensure validation rules properly implemented
- [x] Has some support for wasm-smith
- [ ] Exact / heap types for imports/`ref.func`

There are no tests or valid examples in the proposal yet: providing my own.

### DIFF
--- a/ci/generate-spec-tests.rs
+++ b/ci/generate-spec-tests.rs
@@ -40,7 +40,12 @@ fn copy_test(src: &Path, dst: &Path) {
         return;
     }
 
-    let mut contents = format!(";; RUN: wast \\\n");
+    let directive = match dst.file_name().and_then(|s| s.to_str()) {
+        Some("exact-func-import.wast") => "FAIL",
+        Some(_) | None => "RUN",
+    };
+
+    let mut contents = format!(";; {directive}: wast \\\n");
     contents.push_str(";;      --assert default \\\n");
 
     // Allow certain assert_malformed tests to be interpreted as assert_invalid
--- a/tests/cli/spec/proposals/custom-descriptors/exact-func-import.wast
+++ b/tests/cli/spec/proposals/custom-descriptors/exact-func-import.wast
@@ -1,4 +1,4 @@
-;; RUN: wast \
+;; FAIL: wast \
 ;;      --assert default \
 ;;      --snapshot tests/snapshots \
 ;;      --ignore-error-messages \
--- a/tests/cli/spec/proposals/custom-descriptors/exact-func-import.wast.stderr
+++ b/tests/cli/spec/proposals/custom-descriptors/exact-func-import.wast.stderr
@@ -1,0 +1,5 @@
+error: expected `)`
+     --> tests/testsuite/proposals/custom-descriptors/exact-func-import.wast:3:26
+      |
+    3 |   (import "" "" (func $1 (exact (type 0))))
+      |                          ^


crates/wasmparser/src/features.rs
```diff
@@ -295,6 +295,9 @@ define_wasm_features! {
         ///
         /// This is a subcomponent of the "lime1" feature.
         pub bulk_memory_opt: BULK_MEMORY_OPT(1 << 34) = true;
+
+        // Custom descriptors proposal.
+        pub custom_descriptors: CUSTOM_DESCRIPTORS(1 << 35) = true;
```
alexcrichton: This'll want to default to `false` for now

crates/wasm-smith/src/core.rs
```diff
@@ -568,6 +570,8 @@ impl Module {
                     return false;
                 }
             },
+
+            _ => todo!(),
```
alexcrichton: Mind filling this out to avoid the `_`? For example `(HT::Exact(_), _) | (_, HT::Exact(_)) => todo!()` or something like that

crates/wasmparser/src/validator/core/canonical.rs
```diff
@@ -118,6 +118,14 @@ pub(crate) trait InternRecGroup {
             }
         }
 
+        if is_new {
+            for i in start..end {
+                let i = u32::try_from(i).unwrap();
+                let id = CoreTypeId::from_index(i);
+                self.check_descriptors(rec_group_id, id, types, offset)?;
```
alexcrichton, line 118: Could this get folded into the `if is_new` block above? Or does this semantically need to happen after the above loop?

crates/wasmparser/src/validator/core/canonical.rs
```diff
@@ -162,6 +170,111 @@ pub(crate) trait InternRecGroup {
...
+        let describes_idx = if let Some(i) = ty.describes_idx {
+            Some(map_cannonical(i)?)
+        } else {
+            None
+        };
+
+        if let Some(supertype_index) = types[id].supertype_idx {
+            debug_assert!(supertype_index.is_canonical());
+            let sup_id = map_cannonical(supertype_index)?;
+            if descriptor_idx.is_some() {
```
alexcrichton: Where possible I'd recommend changing this for example to `if let Some(descriptor_idx) = descriptor_idx` since that avoids `descriptor_idx.unwrap()` below within this scope. That's not possible everywhere below without Rust's relatively new stabilized feature of "let chains" but this clause for example should be easy to change

crates/wasmparser/src/validator.rs
```diff
@@ -262,7 +262,7 @@ impl WasmFeatures {
             return Err("reference types support is not enabled");
         }
         match r.heap_type() {
-            HeapType::Concrete(_) => {
+            HeapType::Concrete(_) | HeapType::Exact(_) => {
```
alexcrichton: Here this'll want to check `self.custom_descriptors()` for the `Exact` use case to gate the usage of `Exact` to only when `custom_descriptors()` is enabled.

crates/wasmparser/src/readers/core/types.rs
```diff
@@ -1820,6 +1867,11 @@ impl<'a> FromReader<'a> for HeapType {
                     let ty = reader.read()?;
                     Ok(HeapType::Abstract { shared: true, ty })
                 }
+                0x62 => {
```
alexcrichton, line 1870: Mind updating the table [here](https://github.com/bytecodealliance/wasm-tools/blob/07068fa17507e38310bcbefa83a001423c7548d9/crates/wasmparser/src/readers/core/types.rs#L1725-L1775) as well?

crates/wasmparser/src/readers/core/types.rs
```diff
@@ -1491,6 +1536,8 @@ pub enum HeapType {
     ///
     /// Introduced in the function-references proposal.
     Concrete(UnpackedIndex),
+    /// An exact defined type.
```
alexcrichton: Mind expanding this to indicate that it's part of the custom-descriptors proposal?

Ideally this'd have more comprehensive documentation as to what it actually is (I've no idea what it is myself) but for now if you'd prefer to just delegate to the proposal itself that's fine.

crates/wasmparser/src/readers/core/types.rs
```diff
@@ -83,13 +83,13 @@ fn can_fit_max_wasm_types_in_packed_index() {
...
+    const KIND_MASK: u32 = 0b11 << 19;
+    const INDEX_MASK: u32 = (1 << 19) - 1;
 
-    const MODULE_KIND: u32 = 0b00 << 20;
-    const REC_GROUP_KIND: u32 = 0b01 << 20;
+    const MODULE_KIND: u32 = 0b00 << 19;
+    const REC_GROUP_KIND: u32 = 0b01 << 19;
     #[cfg(feature = "validate")]
-    const ID_KIND: u32 = 0b10 << 20;
+    const ID_KIND: u32 = 0b10 << 19;
```
alexcrichton: For these changes mind updating the documentation on `PackedIndex` above? Also I think that `0b11` as a "kind" is open to represent "exact" so the index bit-space need not be shrunk?

alexcrichton (commented):
I only skimmed the validation/bit-packing-`RefType` bits and I'll cc @fitzgen for those bits too as I'm sure he'll be interested in them when they shape up. Otherwise I left a few comments here and there. Overall thanks for this though!

crates/wasm-encoder/src/core/instructions.rs
```diff
@@ -4609,4 +4609,58 @@ impl<'a> InstructionSink<'a> {
...
     }
+
+    /// Encode [`Instruction::RefGetDesc`].
+    pub fn ref_get_desc(&mut self, type_index: u32) -> &mut Self {
+        self.sink.push(0xFB);
+        34u32.encode(self.sink);
+        type_index.encode(self.sink);
+        self
+    }
+    /// Encode [`Instruction::RefCastDescNonNull`].
```
fitzgen, line 4621: Super nitty pick: all other methods have a newline between each other, we should match that with these new ones.

crates/wasmparser/src/readers/core/types.rs
```diff
@@ -47,10 +47,10 @@ pub(crate) use self::matches::{Matches, WithRecGroup};
 //
 // This is a bit-packed `u32` with the following layout:
 //
-//     [ unused:u10 kind:u2 index:u20 ]
+//     [ unused:u11 kind:u2 index:u19 ]
 //
-// It must fit in 22 bits to keep `RefType` in 24 bits and `ValType` in 32 bits,
-// so the top ten bits are unused.
+// It must fit in 21 bits to keep `RefType` in 24 bits and `ValType` in 32 bits,
+// so the top 11 bits are unused.
```
fitzgen: 19 bits isn't enough to represent the `1_000_000` types implementation limit. That is, we would be taking on a new, more-restrictive-than-everyone-else implementation limit on the number of types with this change. (I'd expect that the `can_fit_max_wasm_types_in_packed_index` test would start failing).

Do we need an `exact` bit for all different kinds of packed indices? If not, then Alex's suggestion about using a new `0b11` kind prefix for exact references should be workable. If we need both exact type indices and exact `CoreTypeId`s, then I don't think it will work.

But also, this is making me think that the `exact` bit should be on `RefType` and not inside `PackedIndex`? `PackedIndex` doesn't really directly reflect a spec entity. If exactness is a property of the reftype, then the bit should be inside `RefType` and if it is a property of the heaptype, then the bit should be inside `HeapType`. So maybe after that adjustment we don't have these bitpacking issues anymore?

fitzgen (commented):
Thanks!

crates/wasmparser/src/readers/core/types.rs
yurydelendik: I changed `RefType` to store the data in u32 instead of [u8;3]

fitzgen:
> I'll whip this up

I was hoping I would finish this up before the weekend but it looks like I won't. Since the CG meeting in Munich is next week, I probably won't finish this until the week afterwards.

fitzgen:
Draft PR here: https://github.com/bytecodealliance/wasm-tools/pull/2369

Alex has ideas about landing this PR before that, though.

alexcrichton:
Ok I've done a bit of benchmarking now. With this PR as-is:

<details>

<summary>bench results</summary>

```
validate/tests          time:   [8.5038 ms 8.5072 ms 8.5128 ms]
                        change: [+3.1259% +3.2126% +3.3074%] (p = 0.00 < 0.05)
                        Performance has regressed.

validate/lots-of-unique-fn-types
                        time:   [1.6182 ms 1.6190 ms 1.6200 ms]
                        change: [+4.2532% +4.3423% +4.4290%] (p = 0.00 < 0.05)
                        Performance has regressed.

validate-old/lots-of-unique-fn-types
                        time:   [764.95 µs 765.34 µs 765.94 µs]
                        change: [+5.6912% +5.7338% +5.7780%] (p = 0.00 < 0.05)
                        Performance has regressed.

validate/intgemm-simd   time:   [725.08 µs 725.26 µs 725.46 µs]
                        change: [-2.5999% -2.5459% -2.5027%] (p = 0.00 < 0.05)
                        Performance has improved.

validate-old/intgemm-simd
                        time:   [713.58 µs 713.74 µs 713.90 µs]
                        change: [-2.7302% -2.6875% -2.6424%] (p = 0.00 < 0.05)
                        Performance has improved.

validate/bz2            time:   [313.12 µs 313.41 µs 313.75 µs]
                        change: [-7.5212% -7.4616% -7.3949%] (p = 0.00 < 0.05)
                        Performance has improved.

validate-old/bz2        time:   [310.04 µs 310.17 µs 310.32 µs]
                        change: [-8.2737% -8.1943% -8.0962%] (p = 0.00 < 0.05)
                        Performance has improved.

validate/lots-of-empty-fn-types
                        time:   [611.75 µs 612.03 µs 612.33 µs]
                        change: [+6.5110% +6.6114% +6.7036%] (p = 0.00 < 0.05)
                        Performance has regressed.

validate-old/lots-of-empty-fn-types
                        time:   [687.19 µs 687.32 µs 687.48 µs]
                        change: [+6.3189% +6.3881% +6.4912%] (p = 0.00 < 0.05)
                        Performance has regressed.

validate/spidermonkey   time:   [18.976 ms 18.981 ms 18.987 ms]
                        change: [-3.8797% -3.8300% -3.7809%] (p = 0.00 < 0.05)
                        Performance has improved.

validate-old/spidermonkey
                        time:   [19.005 ms 19.008 ms 19.012 ms]
                        change: [-3.8858% -3.8447% -3.8097%] (p = 0.00 < 0.05)
                        Performance has improved.

validate/pulldown-cmark time:   [793.03 µs 793.35 µs 793.79 µs]
                        change: [-2.5407% -2.4893% -2.4298%] (p = 0.00 < 0.05)
                        Performance has improved.

validate-old/pulldown-cmark
                        time:   [787.91 µs 788.14 µs 788.41 µs]
                        change: [-3.1519% -3.1091% -3.0608%] (p = 0.00 < 0.05)
                        Performance has improved.
```

</details>

So not obviously bad. It's actually a speedup on all 'real world' modules but is an equivalent regression on synthetic stressor modules.

Next I took https://github.com/bytecodealliance/wasm-tools/pull/2369 and massaged it into the validator as well. The commit is https://github.com/alexcrichton/wasm-tools/commit/c2d95e49235e4ddf6bb11c9c6f8875caf87f1153 and is pretty messy as my goal was just to get to the point of benchmarking. Try as I might it's a net regression across the board:

<details>

<summary>bench numbers</summary>

```
validate/tests          time:   [8.6207 ms 8.6267 ms 8.6365 ms]
                        change: [+4.5500% +4.6617% +4.8015%] (p = 0.00 < 0.05)
                        Performance has regressed.

validate/lots-of-unique-fn-types
                        time:   [1.6303 ms 1.6312 ms 1.6324 ms]
                        change: [+5.0902% +5.2510% +5.4381%] (p = 0.00 < 0.05)
                        Performance has regressed.

validate-old/lots-of-unique-fn-types
                        time:   [756.01 µs 756.24 µs 756.51 µs]
                        change: [+4.5071% +4.5594% +4.6147%] (p = 0.00 < 0.05)
                        Performance has regressed.

validate/intgemm-simd   time:   [793.10 µs 793.51 µs 794.08 µs]
                        change: [+6.5632% +6.6305% +6.6962%] (p = 0.00 < 0.05)
                        Performance has regressed.

validate-old/intgemm-simd
                        time:   [781.38 µs 781.66 µs 781.94 µs]
                        change: [+6.5166% +6.5586% +6.6019%] (p = 0.00 < 0.05)
                        Performance has regressed.

validate/bz2            time:   [347.23 µs 347.40 µs 347.69 µs]
                        change: [+2.5307% +2.5825% +2.6475%] (p = 0.00 < 0.05)
                        Performance has regressed.

validate-old/bz2        time:   [344.52 µs 344.78 µs 345.28 µs]
                        change: [+1.9069% +2.0015% +2.1150%] (p = 0.00 < 0.05)
                        Performance has regressed.

validate/lots-of-empty-fn-types
                        time:   [610.01 µs 610.26 µs 610.52 µs]
                        change: [+6.2306% +6.3265% +6.4184%] (p = 0.00 < 0.05)
                        Performance has regressed.

validate-old/lots-of-empty-fn-types
                        time:   [678.92 µs 679.28 µs 679.86 µs]
                        change: [+5.0427% +5.1188% +5.2261%] (p = 0.00 < 0.05)
                        Performance has regressed.

validate/spidermonkey   time:   [20.722 ms 20.729 ms 20.737 ms]
                        change: [+4.9696% +5.0265% +5.0852%] (p = 0.00 < 0.05)
                        Performance has regressed.

validate-old/spidermonkey
                        time:   [20.741 ms 20.745 ms 20.749 ms]
                        change: [+4.8940% +4.9394% +4.9792%] (p = 0.00 < 0.05)
                        Performance has regressed.

validate/pulldown-cmark time:   [857.67 µs 857.86 µs 858.06 µs]
                        change: [+5.3636% +5.4135% +5.4581%] (p = 0.00 < 0.05)
                        Performance has regressed.


validate-old/pulldown-cmark
                        time:   [855.23 µs 855.41 µs 855.60 µs]
                        change: [+5.0936% +5.1401% +5.1899%] (p = 0.00 < 0.05)
                        Performance has regressed.

```

</details>

I've tried a number of ways to further optimize this in my commit, but to no avail. In retrospect this makes sense though:

* `ValType` is extremely pervasive and this PR inflates it from 4 to 8 bytes. While we're changing the validator to store `ValTypePacked` nowhere else is changing so everywhere else is getting a larger memory footprint. I think this is why the synthetic modules are regressing.
* Validator performance with `ValTypePacked` is pretty bad because, looking at the generated code, there's a lot of conversion between `ValType` and `ValTypePacked`. The theory of `ValTypePacked` is that its conversion gets inlined away and is zero-cost, but in practice that's just not happening. It's certainly happening in some places, and I can see it, but it's not happening in all cases which is basically what's required here.

So my conclusion is that we should land this PR as-is. Update the test to accept an 8-byte `MaybeType` instead of a 4-byte `MaybeType` and just accept that `ValType` is now 8 bytes instead of 4 bytes. I don't think it's worth replacing `ValType` with `ValTypePacked` (e.g. have it be an opaque 32-bit integer) because that loses `match` exhaustiveness. I also don't think it's worth having two representations because it's objectively slower and also a larger maintenance burden. In the end if we want this to be faster I think we just need to wait for future Rust language features such as pattern types and/or the ability to define a type which has a broader niche than just `NonZero` for example.

fitzgen:
> * `ValType` is extremely pervasive and this PR inflates it from 4 to 8 bytes. While we're changing the validator to store `ValTypePacked` nowhere else is changing so everywhere else is getting a larger memory footprint. I think this is why the synthetic modules are regressing.
> 
> * Validator performance with `ValTypePacked` is pretty bad because, looking at the generated code, there's a lot of conversion between `ValType` and `ValTypePacked`. The theory of `ValTypePacked` is that its conversion gets inlined away and is zero-cost, but in practice that's just not happening. It's certainly happening in some places, and I can see it, but it's not happening in all cases which is basically what's required here.

Yeah I was hoping that the conversions would pretty much get inlined away and it is unfortunate that you aren't seeing that.

I think I agree with your conclusion that we can land this PR as-is (modulo CI going green) but I think it is worth thinking a tiny bit more about `ValTypePacked` before giving up there. Specifically, where are the conversions not getting inlined away? I would imagine it is at the places where valtypes enter/exit the validator code itself, not within the validator. Assuming that is correct, are there perhaps a couple choice places outside the validator that we could change to use the packed variant and avoid most of the conversion costs? On the other hand, if the conversions aren't boiling away in the middle of validator's hot loop, then this whole approach is pretty dead in the water.

fitzgen:
> So not obviously bad. It's actually a speedup on all 'real world' modules but is an equivalent regression on synthetic stressor modules.

FWIW, I think we should care about the real world modules ***way*** more than the synthetic ones.

alexcrichton:
> On the other hand, if the conversions aren't boiling away in the middle of validator's hot loop, then this whole approach is pretty dead in the water.

More-or-less this is what I saw. For example, this is one of the hottest methods in the validator:

```rust
    fn visit_local_set(&mut self, local_index: u32) -> Self::Output {
        let ty = self.local(local_index)?;
        self.pop_operand(Some(ty))?;
        self.local_inits.set_init(local_index);
        Ok(())
    }
```

Here `self.local` is reading from some storage of `ValTypePacked`. Later `pop_operand` reads a `MabyeTypePacked` representation. Here the `ty` isn't a "constant" like it is in other places (e.g. `pop_operand(Some(ValType::I32))`) but it's genuinely a runtime variable. LLVM was unable to boil away all the conversions so `self.local`, which returns `ValType`, would decode and then it's get re-encoded in `pop_operand` to `MaybeTypePacked`.

We could fix this by changing `self.local` to return `ValTypePacked` I think, but that starts bringing up the question of just wholesale replacing `ValType` with `ValTypePacked` everywhere in the validator. Once we do that there's the question of everythign else and .... then I reached the conclusion we shouldn't do this

fitzgen:
> We could fix this by changing `self.local` to return `ValTypePacked` I think, but that starts bringing up the question of just wholesale replacing `ValType` with `ValTypePacked` everywhere in the validator. Once we do that there's the question of everythign else and .... then I reached the conclusion we shouldn't do this

Yeah, makes sense to me.

Really wish Rust didn't make us choose between packing and ergonomics!

alexcrichton:
@yurydelendik alright if you're ok with it the ball's now back in your court. I pushed up two commits here (feel free to rebase/squash them away as you see fit) as well as a merge with `main`. Notably in https://github.com/bytecodealliance/wasm-tools/pull/2295/commits/162dff044290c0669c45066341f8f2d895d0b6e3 I raised the assertion limit for `MaybeType` to unblock this PR. In the next commit I updated the spec test suite submodule which has a lot more custom-descriptors tests. Would you be ok adjusting text/validation/etc as necesary to get the spec test suite passing?

yurydelendik:
> @yurydelendik alright if you're ok with it the ball's now back in your court. I pushed up two commits here (feel free to rebase/squash them away as you see fit) as well as a merge with `main`. Notably in [162dff0](https://github.com/bytecodealliance/wasm-tools/commit/162dff044290c0669c45066341f8f2d895d0b6e3) I raised the assertion limit for `MaybeType` to unblock this PR. In the next commit I updated the spec test suite submodule which has a lot more custom-descriptors tests. Would you be ok adjusting text/validation/etc as necesary to get the spec test suite passing?

Thanks. I rebased and adjusted most of syntax. Changed logic around ref.get_desc and ref.cast_desc too.

Not implement is parsing of some 'exact' case, e.g. in imports -- that requires some deep diving into data structures. Will be looking at that next

alexcrichton:
Happy to help out if you get stuck! I just pushed up a commit which is a re-`BLESS=1` of the cli testsuite which cuts things down to just a few failures

alexcrichton:
Also, if you'd like, these tests are pretty minor so I think it would be reasonable to land what you have here and address the missing functionality in a follow-up too. Whichever you'd prefer.

yurydelendik:
> Also, if you'd like, these tests are pretty minor so I think it would be reasonable to land what you have here and address the missing functionality in a follow-up too. Whichever you'd prefer.

That sounds good. We can land what is here so far, at any moment. I'll continue working on exact types in imports.

alexcrichton:
Ok sounds good, I'll do one final pass on review and frob CI to get the test to be expected to fail and then land this

PR #2298, opened 2025-09-04
Implement co-op threading builtins

TartanLlama:
Implements https://github.com/WebAssembly/component-model/pull/557

- Renames `async` variants of `waitable-set.wait`, `waitable-set.poll`, and `yield` to `cancellable`
- Renames `yield` to `thread.yield`
- Adds `[cancellable]` legacy name mangling prefix
- Adds support to `wasmparser`, `wasm-encoder`, and `wit-component` for the new builtins:
  - `thread.index`
  - `thread.new_indirect`
  - `thread.switch-to`
  - `thread.suspend`
  - `thread.resume-later`
  - `thread.yield-to`
- Replaces the `cm_async_stackful` and `cm_async_builtins` features with `cm_threading` 
- Makes `context.get` and `context.set` callable with a cell ref of `1` in addition to `0`
- Updates all tests for breaking changes
- Adds tests for new functionality

The `thread.new_indirect` builtin required adding support for tracking exports of `funcref` tables so that an alias can be made for the table that stores the thread start function, which is then referenced by `(canon thread.new_indirect ..)`

Marking as a draft for now while I play around with `wasmtime` support.


crates/wasmparser/src/validator/component.rs
```diff
@@ -1456,8 +1476,8 @@ impl ComponentState {
                 "`context.get` requires the component model async feature"
             )
         }
-        if i > 0 {
-            bail!(offset, "`context.get` immediate must be zero: {i}")
+        if i > 1 {
```
alexcrichton: Could this and below check or `cm_threading` for 1? e.g. cm-async for 0, cm-threading + cm-async for 1

crates/wasmparser/src/features.rs
```diff
@@ -254,47 +254,41 @@ define_wasm_features! {
...
+        /// Corresponds to the 🧵 character in
         /// <https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md>.
-        pub cm_async_stackful: CM_ASYNC_STACKFUL(1 << 28) = false;
-        /// Gates some intrinsics being marked with `async` in the component
-        /// model async proposal.
-        ///
-        /// Corresponds to the 🚝 character in
-        /// <https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md>.
-        pub cm_async_builtins: CM_ASYNC_BUILTINS(1 << 29) = false;
+        pub cm_threading: CM_THREADING(1 << 28) = false;
```
alexcrichton: I'm not personally sold just yet on merging these features all together. That being said I realize you're just implementing [the spec PR here](https://github.com/WebAssembly/component-model/pull/557) so I'm also going to leave a comment over there (but wanted to leave this here as well)

tests/cli/missing-features/component-model/threading.wast
```diff
@@ -0,0 +1,101 @@
+;; RUN: wast % --assert default --snapshot tests/snapshots \
+;;        -f=cm-async,-cm-threading
```
alexcrichton, line 2: I think it'll be worth splitting this test in two: one for `-cm-threading` which asserts that all the various constructs are indeed an error if they're present. Then a separate test as well for `cm-threading` which asserts that the constructs are all valid. The valid test would also test things like: the table must be of the right type, each intrinsic produces a core wasm function of an expected signature, validation of intrinsic arguments like the function type to spawn, etc.

crates/wit-component/src/validation.rs
```diff
@@ -558,6 +591,11 @@ impl ImportMap {
         } else {
             (false, name)
         };
+        let (cancellable, name) = if let Some(name) = names.cancellable_name(name) {
+            (true, name)
+        } else {
+            (false, name)
+        };
```
alexcrichton: In retrospect I realize this is how `[async]` is modeled but it's pretty brittle. I think it might be best to instead refactor this such that the `names` implementation does the `[async]` or `[cancellable]` stripping for each individual intrinsic name rather than doing it up-front and having to remember to call `validate_not_cancellable()` everywhere. That being said this is preexisting, so feel free to defer this and I can open an issue for it for later.

crates/wit-component/src/validation.rs
```diff
@@ -569,111 +607,200 @@ impl ImportMap {
...
                 return Ok(Import::ContextSet(i));
             }
+            if Some(name) == names.thread_index() {
+                validate_not_async()?;
+                validate_not_cancellable()?;
+                let expected = FuncType::new([], [ValType::I32]);
+                validate_func_sig(name, &expected, ty)?;
+                return Ok(Import::ThreadIndex);
+            }
+            if let Some((func_ty_idx, table_idx)) = names.thread_new_indirect(name) {
```
alexcrichton: Toolchain-wise it's basically impossible to talk about indexes at a source-language-level (which is where these imports would show up). For example wasi-libc would have no way of correctly importing this intrinsic because it would have no idea what index the function type is assigned. It can generally rely on there being only one table, but that's somewhat brittle as well. All that being said I think it's still appropriate for the component model intrinsic to stay as-is, so the question then remains still of how to bridge what a source language can generate and what the component model wants.

What I might recommend instead here is:

* To handle `func_ty_idx` given that there's only a single signature valid initially I'd say skip this entirely. Basically don't try to name a function at all but instead fold the "this is the type" into the name somehow. For example `[thread-new-indirect-v0]` where there's no function type at all but "v0" means the initial version of the proposal. In the future we could do something like `[thread-new-indirect-v1-(func (param i32 i32))]` if we really wanted, but for now we can leave that for future design.
* To handle `table_idx` I think it's probably best to handle this the same way as linear memory. Basically the table is so ubiquitous at the LLVM layer it's not really worthwhile trying to specifically call it out here. So effectively update wit-component to find a particular table of a particular export name. Looks like LLD currently names the exported table `__indirect_function_table` with no option of changing that. Currently we probe for a memory export named `memory` (also LLD's default naming) and the "standard" mangling uses a different name. We can customize all this in the future, but just looking for `__indirect_function_table` for now I believe is sufficient.

crates/wit-component/src/validation.rs
```diff
@@ -1164,6 +1299,12 @@ impl ExportMap {
                 }
                 return Ok(None);
             }
+            ExternalKind::Table => {
+                if types.table_at(export.index).element_type.is_func_ref() {
+                    return Ok(Some(Export::FuncTable(export.index)));
+                }
```
alexcrichton: I alluded above to this a bit too but here I think we'll want to check the export name as well (e.g. `names.export_table()` to be added here)

crates/wit-component/src/validation.rs
```diff
@@ -1335,6 +1476,15 @@ impl ExportMap {
         self.find(|m| matches!(m, Export::Memory))
     }
 
+    /// Returns the funcref table that matches the given index,
+    /// if exported, for this module.
+    pub fn func_table(&self, index: u32) -> Option<&str> {
```
alexcrichton: Given my changes above this'd drop the `index` argument and end up looking similar to `memory` in implementation

crates/wit-component/src/validation.rs
```diff
@@ -1772,8 +1951,8 @@ impl NameMangling for Legacy {
     fn waitable_join(&self) -> Option<&str> {
         Some("[waitable-join]")
     }
-    fn yield_(&self) -> Option<&str> {
-        Some("[yield]")
+    fn thread_yield(&self) -> Option<&str> {
+        Some("[thread-yield]")
     }
```
alexcrichton: One thing I mentioned earlier today was we need to be careful about this change because we can get ourselves into a theoretical state where we can't update either wit-bindgen or wasmtime. Thinking about this specific change though I think we should be ok if we update wit-bindgen first. In updating wit-bindgen's usage of wasm-tools we'll update the name emitted in core wasm, but the binary encoding in the component will stay the same so it'll still work in Wasmtime today. After wit-bindgen is updated we can update Wasmtime's version of wasm-tools which will also require updating wit-bindgen at the same time since wasmtime's tests use wit-bindgen and the yield intrinsic.

Basically I wanted to write down "I don't think we need to handle the old name" which is nice, yay! Just requires updating wit-bindgen once this is published, then updating wasmtime.

crates/wit-component/src/validation.rs
TartanLlama: Makes sense! I've had a go at implementing this scheme

crates/wit-component/src/validation.rs
TartanLlama: Updated

crates/wit-component/src/validation.rs
TartanLlama: Also updated

crates/wasmparser/src/validator/component.rs
TartanLlama: Yep, I've factored this out

tests/cli/missing-features/component-model/threading.wast
TartanLlama: I think that's all covered by my changes here: https://github.com/bytecodealliance/wasm-tools/pull/2298/files#diff-906df3769b98fd0435b815a8d3adda4ce5ab21d4a947df40ee600ad7dcc37fdd

alexcrichton: Oh right, thanks!

It's ok to leave that as-is in this PR, but if you're up for it the file structure here was already a bit messy and needs cleanup. We had a number of tests originally added for component-model-async and then retroactively we carved out part of the specification under new feature flags. At that time we didn't actually split any tests so the tests are still all lumped together. For new features though it's typically best to have all the tests in a new file as opposed to intermingling with old tests.

Although as I type this out I realize that this hinges on the merging-or-not of the old features into the threading feature. I'm assuming in my head that we won't want to do that, but if we did move forward with that what I'm describing is probably already matching the current organization.

Basically it's fine to disregard this for now, I can always try to come back to it later.

crates/wit-component/src/validation.rs
TartanLlama: I've had a go at refactoring this

alexcrichton: Looks great! Thanks for taking this on!

alexcrichton:
For the CI failure to resolve that you might need to blow away `tests/snapshots` and then recreate it with `BLESS=1` to delete the now-stale files

crates/wasmparser/src/features.rs
TartanLlama: I've updated the PR to have the old features + a new threading one

alexcrichton:
@TartanLlama WDYT about merging this? I was thinking I'd try to tackle https://github.com/WebAssembly/component-model/pull/560 today and I don't want to cause unnecessary merge conflicts for this

TartanLlama:
> @TartanLlama WDYT about merging this?

Fine by me!

PR #2302, opened 2025-09-12
wasmparser: reject function bodies larger than implementation limit.

cfallin:
The implementation limits for JS embedders at [1] describe various upper bounds for entities in Wasm modules, including the size of a single function body in bytes. wasmparser is not strictly required to abide by these implementation limits because they are conventions for the JS embedding of Wasm rather than part of the core Wasm standard. However, it does seem to enforce other limits, such as the number of types or locals; this PR updates it to enforce function body size as well.

This came up in bytecodealliance/wasmtime#11682, where a very large function (larger than the implementation limit) led to out-of-bounds SSA value numbers in Cranelift when they exceeded the range allowed by our data-structure bitpacking. Rather than doing major surgery to plumb the exact failure through all of Cranelift (including its public API being `Result`-ified on every builder interface) it seems better to have a single limit on the size of incoming functions. It turns out that the Wasm implementation limits were designed for just this purpose, so let's use them in our tooling as well.

[1]: https://webassembly.github.io/spec/js-api/#limits

### DIFF
--- a/crates/wasmparser/src/limits.rs
+++ b/crates/wasmparser/src/limits.rs
@@ -17,6 +17,8 @@
 
 // The following limits are imposed by wasmparser on WebAssembly modules.
 // The limits are agreed upon with other engines for consistency.
+//
+// See https://webassembly.github.io/spec/js-api/#limits for details.
 pub const MAX_WASM_TYPES: usize = 1_000_000;
 pub const MAX_WASM_SUPERTYPES: usize = 1;
 pub const MAX_WASM_FUNCTIONS: usize = 1_000_000;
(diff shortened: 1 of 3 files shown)


crates/wasmparser/src/limits.rs
```diff
@@ -26,7 +28,7 @@ pub const MAX_WASM_GLOBALS: usize = 1_000_000;
 pub const MAX_WASM_ELEMENT_SEGMENTS: usize = 100_000;
 pub const MAX_WASM_DATA_SEGMENTS: usize = 100_000;
 pub const MAX_WASM_STRING_SIZE: usize = 100_000;
-pub const MAX_WASM_FUNCTION_SIZE: usize = 128 * 1024;
+pub const MAX_WASM_FUNCTION_SIZE: usize = 7_654_321;
```
cfallin, line 31: I'll note that I'm not entirely sure what was going on with the previous limit of 128K here -- it wasn't actually used to enforce function-size limits (despite the name of the constant), only the number of branch-table targets. Given the name it seems reasonable to use it for what it says on the tin.

PR #2305, opened 2025-09-15
fix command 'cargo test --no-default-features'

alexanderkjall:


alexcrichton:
Thanks for the PR! Could you expand on the motivation for this though? There are more than a few instances of this throughout the repository and we wouldn't have any guarantee that this continues working without CI changes too. I've not run across a case for this in the past so I've never tried to keep this working myself, so I'm curious to learn a bit more about your motivations for having this work.

alexanderkjall:
Sorry for not providing any context!

The test suite in Debian typically run all tests for each feature, and with all features enabled, and no features enabled.

This might be a bit extreme, but since Debian run on a lot of architectures it's sadly common that we need to patch things to work on those, and having as good test coverage as possible helps :)

The reason I sent this patch was that I thought it would be good to try to contribute those changes back, in order to try to keep the distance between the package in Debian and upstream as small as possible, the PR is based on: https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/wasmparser/debian/patches/gate-tests.patch

alexcrichton:
Is it possible to request that that style of feature testing is not done for this crate? The cost of this is mostly boilerplate annotations and extra CI cycles spent for not a whole lot of benefit. Other crates I help maintain, such as Wasmtime, are in a similar situation where we explicitly don't test all features combinations and we instead rely on coding conventions and CI "this builds" checks to try to keep combinations of features working.

alexanderkjall:
no problem, I was just trying to offload the work of maintaining this patch upstream :P I fully understand why it's not wanted.

PR #2307, opened 2025-09-16
Add debug names to components by default

alexcrichton:
This commit updates `wit-component` to add a debug `component-name` section to generated components by default. I find navigating the output otherwise quite hard to do and find myself juggling lots of numbers. This can be disabled with `--debug-names=false`.


pchickey (approved):
Looks good - this will be very helpful, I've needed this for years. Spot checked all the expected test output changes, but this is pretty straightforward.

PR #2311, opened 2025-09-19
Support `--features` CLI arguments for `wasm-tools component new`

### LINKED ISSUE #2310: Support `--features` CLI arguments for `wasm-tools component new`
Currently, it is not possible to make use of WebAssembly proposal/features that aren't enabled by default in components. This is because unlike `wasm-tools validate` which allows a `--features` CLI arguments to extend (or reduces) the set of Web Assembly proposals/features supported. 
This prevents users from making components out of valid Wasm modules, for example ones that use the [`custom-page-sizes`](github.com/WebAssembly/custom-page-sizes/) proposal which is otherwise supported by `wasmtime` and `wasm-tools`. As far as I can tell, nothing in theory prevents this proposal from being used in components so it's a shame that this is currently impossible when it could easily be supported.

anlavandier:
Closes #2310 

~~This PR adds a `--features` argument to `wasm-tools component new` using the model of `wasm-tools validate`. This is done by parsing the `--features` arguments and propagating the features down to `ValidatedModule::new` which is the function responsible for initializing the `Validator` used when parsing the provided Wasm Module~~

This PR enables all of the features supported by `wasm-tools` in the `Validator` used when parsing the provided Wasm Module in `wasm-tools component new`

This PR is trying to be as targeted as possible but I would argue that  `--features` should be a argument global to `wasm-tools` and forwarded to every validator used when parsing `.wasm` modules.

### DIFF
--- a/crates/wit-component/src/validation.rs
+++ b/crates/wit-component/src/validation.rs
@@ -8,7 +8,7 @@ use wasm_encoder::ExportKind;
 use wasmparser::names::{ComponentName, ComponentNameKind};
 use wasmparser::{
     Encoding, ExternalKind, FuncType, Parser, Payload, TypeRef, ValType, ValidPayload, Validator,
-    types::TypesRef,
+    WasmFeatures, types::TypesRef,
 };
 use wit_parser::{
     Function, InterfaceId, PackageName, Resolve, Type, TypeDefKind, TypeId, World, WorldId,
@@ -59,7 +59,7 @@ impl ValidatedModule {
         import_map: Option<&ModuleImportMap>,
         info: Option<&LibraryInfo>,
     ) -> Result<ValidatedModule> {
-        let mut validator = Validator::new();
+        let mut validator = Validator::new_with_features(WasmFeatures::all());
         let mut ret = ValidatedModule::default();
 
         for payload in Parser::new(0).parse_all(bytes) {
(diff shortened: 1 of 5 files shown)


alexcrichton:
Thanks for the PR! In situations like this thought I've often found it easiest to avoid passing features around and instead just turn on all the features. That way it's a bit easier since there's much less plumbing and it's not really less-wrong since `wasm-tools component new` isn't really a load-bearing validation function anyway. What do you think about that approach instead?

anlavandier:
I'm completely fine with this. I'll change the PR title, comment and code to reflect this. Once this is merged, will this directly affect newer releases of `wasm-component-ld` or will a similar PR need to be raised in that repo as well ?

alexcrichton:
To update `wasm-component-ld` the process is (a) merge the PR here, (b) make a crates.io release from this repo, (c) update the `wasm-component-ld` source, (d) make a crates.io release of `wasm-component-ld`, and (e) update rust-lang/rust and wasi-sdk to the new versions. I'm able to help out with these parts too and it's fine to make a release for this.

Would you be up for adding a test here as well? For example something in `crates/wit-component/tests/components` I think would work. Something small is fine, but it'd be good to verify that everything works with a proposal like custom-page-sizes.

anlavandier:
Done in the latest commit.

PR #2312, opened 2025-09-19
Print offset of final `end` in `wasm-tools print -p`

alexcrichton:
I was looking at a validation error recently where the offset printed for an error wasn't present in `wasm-tools print -p`, and that was because the error was on the final `end` instruction of a function which isn't printed in the text format. This commit updates the output of `-p` to include a comment with an offset for the `end` instruction.

### DIFF
--- a/crates/wasmprinter/src/lib.rs
+++ b/crates/wasmprinter/src/lib.rs
@@ -1470,7 +1470,7 @@ impl Printer<'_, '_> {
         let fold_instructions = self.config.fold_instructions;
         let mut operator_state = OperatorState::new(self, OperatorSeparator::Newline);
 
-        if fold_instructions {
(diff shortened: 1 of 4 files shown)


keithw:
Would it be possible to print this offset with the closing parenthesis (which currently lacks an offset) rather than with a new `(; end ;)` line? Semantically, the 0x0b opcode represents the end of the expression rather than an instruction in the function body. Happy to send a followup PR if it would be welcomed.

alexcrichton:
@keithw I tried to do that at first yeah as I agree it'd be nice to not print the extra line, but I ended up giving up because it felt a bit weird threading things around and figured this was "good enough". All that to say, please feel free to send a follow-up! You'll probably have more success than I

PR #2315, opened 2025-09-20
wasmprinter: print end offsets with closing parenthesis

keithw:
Followup of #2312. This is a purely cosmetic change that puts the offsets for function-end and module-end on the same line as the closing parenthesis.


alexcrichton (approved):
Thanks for this!

PR #2317, opened 2025-09-21
Initial skeleton of async support

alexcrichton:
This commit starts to fill out the support for async in `wit-dylib`. In doing so this has filled out a number of pieces of functionality elsewhere and implemented more thorough testing. As a result of this some issues were discovered which are also fixed in this PR. Notable changes are:

* Naming of the `task-return` intrinsic is now in a helper in `wit-parser` to avoid needing to duplicate that elsewhere.
* A new `roundtrip` test was added to `wit-dylib` which generates arbitrary WITs and arbitrary values for the WIT and ensures that the values are communicated successfully.
* Composition code shared between the `all` and `roundtrip` tests was moved to the `artifacts` crate as a shared dependency of the two.
* In-memory data structures for `wit_func_t` are updated to reflect async support.
* Deferred deallocation of bytes is pushed as a responsibility to the embedder. Storing values on the stack was getting too tricky and felt like a too-clever-by-half implementation. To simplify the generated code the responsibility is now on the `wit_dylib_*` implementor to deallocate bytes in a deferred fashion.
* The `wit-bindgen` crate is a dependency of `wit-dylib-ffi`, the guest bindings of `wit_dylib.h`, for async support.
* The `wasm-tools wit-dylib` subcommand now has configuration, in the same manner as `wit-bindgen` for which functions should be async and which should be sync.
* The `with_temp_stack` helper was removed as that was also getting too tricky to manage and instead functions just always allocate some temporary space.
* Existing bindings were updated for async functions when that ABI is selected, and new bindings were added for various intrinsics related to the lifecycle of async imports/exports.

Perhaps the biggest feature in this commit is the addition of the `roundtrip` test which is hoped to provide a much more thorough implementation of testing this crate. Notably it generates an arbitrary WIT structure *and additionally* generates arbitrary values of this structure. This has weeded out a number of bugs in the bindings generator and necessitated various refactorings here. This is intended to be a crucial part of making this a solid dependency to build on.

### DIFF
--- a/crates/wit-dylib/src/bindgen.rs
+++ b/crates/wit-dylib/src/bindgen.rs
@@ -32,10 +32,12 @@ macro_rules! intrinsics {
 intrinsics! {
     initialize : [ValType::I32] -> [] = "wit_dylib_initialize",
     cabi_realloc : [ValType::I32; 4] -> [ValType::I32] = "cabi_realloc",
-    dealloc_bytes : [ValType::I32; 3] -> [] = "wit_dylib_dealloc_bytes",
+    dealloc_bytes : [ValType::I32; 5] -> [] = "wit_dylib_dealloc_bytes",
 
     export_start : [ValType::I32] -> [ValType::I32] = "wit_dylib_export_start",
     export_call : [ValType::I32; 2] -> [] = "wit_dylib_export_call",
+    export_async_call : [ValType::I32; 2] -> [ValType::I32] = "wit_dylib_export_async_call",
+    export_async_callback : [ValType::I32; 4] -> [ValType::I32] = "wit_dylib_export_async_callback",
     export_finish : [ValType::I32; 2] -> [] = "wit_dylib_export_finish",
     resource_dtor : [ValType::I32; 2] -> [] = "wit_dylib_resource_dtor",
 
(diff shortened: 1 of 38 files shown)


alexcrichton:
Draft for now until https://github.com/bytecodealliance/wit-bindgen/pull/1390 is merged

crates/wit-dylib/src/bindgen.rs
```diff
@@ -104,17 +106,157 @@ pub fn import(
...
+        .chain(func.result.iter())
+}
+
+/// Returns the offset, in bytes, within the async_abi_area allocated for an
+/// async import call, to the return value.
+///
+/// Async parameters, if indirect, and async results, always, are required to be
+/// in a long-lived location outside of the original call itself. This function
+/// returns the offset from the start of this allocation to the return value.
+/// This is plubmed through to the guest in `async_abi_area_{size,align}`.
```
dicej:
```suggestion
/// This is plumbed through to the guest in `async_abi_area_{size,align}`.
```

crates/wit-dylib/src/bindgen.rs
```diff
@@ -125,73 +267,78 @@ pub fn export(
...
-        align: 2,
-        offset: 0,
-    });
-    c.free_temp_local(l_ctx);
+    if abi.is_async() {
+        // Async exports use the lifted params from above, but do not handle the
+        // lowered results. That's the responsibility of the
+        // `task_return`-generated function. Here the `cx` is passed along with
+        // the function index. Notably this has different ownership semantics
+        // around `cx` where the callee gets onwership, unlike sync where
```
dicej:
```suggestion
        // around `cx` where the callee gets ownership, unlike sync where
```

PR #2320, opened 2025-09-30
Expand --help output for `parse` and `print` commands

catamorphism:
Add example usage sections to the --help output for these commands.


catamorphism:
I'm going to be working on improving the `--help` output for wasm-tools subcommands. If this format for example usage looks good, I'll continue with the other commands in the same way.

alexcrichton (approved):
Thanks!

PR #2323, opened 2025-09-30
Start adding tests for -h and --help output

catamorphism:
Add tests for the -h and --help output for the
`parse`, `print`, and `validate` commands.

### DIFF
--- a/tests/cli.rs
+++ b/tests/cli.rs
@@ -207,6 +207,7 @@ fn assert_output(bless: bool, output: &[u8], path: &Path, tempdir: &TempDir) -> 
     let mut output = String::from_utf8_lossy(output)
         .replace(tempdir, "%tmpdir")
         .replace("\\", "/")
+        .replace("wasm-tools.exe", "wasm-tools")
         .lines()
         .map(|line| {
             if let Some(start) = line.find("(processed-by \"wit-component\"") {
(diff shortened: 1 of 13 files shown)


alexcrichton:
Poking through clap sources, setting the `COLUMNS=...` env var may work for the failures here, but that'll only work if the `terminal-size` crate otherwise can't find the dimensions of stdout/stderr/stdin/etc. I think though we could arrange for everything to be a pipe (or /dev/null) to subprocesses in tests to force that to fail, and then we'd always set `.env("COLUMNS", "80")` or something like that and it might force the same-width everywhere?

catamorphism:
I think it's already a pipe? https://github.com/bytecodealliance/wasm-tools/blob/main/tests/cli.rs#L155

I added `.env("COLUMNS", "80")` and it works for me locally, although of course it worked for me locally in the first place.

alexcrichton:
Ah the old nemesis of Windows... Looks like the issue is:

```
       <Usage: wasm-tools parse [OPTIONS] [INPUT]
       >Usage: wasm-tools.exe parse [OPTIONS] [INPUT]
```

Maybe a [hammer like this](https://github.com/bytecodealliance/wasm-tools/blob/8937cdc509924752ebe745e06f904fb3d561ed2f/tests/cli.rs#L209) could work where the tests just replace `.exe` with nothing and see how that works?

catamorphism:
Green now :)

tests/cli.rs
```diff
@@ -206,6 +207,7 @@ fn assert_output(bless: bool, output: &[u8], path: &Path, tempdir: &TempDir) ->
     let mut output = String::from_utf8_lossy(output)
         .replace(tempdir, "%tmpdir")
         .replace("\\", "/")
+        .replace("wasm-tools.exe", "wasm-tools")
```
alexcrichton, line 210: Much better than my idea!

PR #2330, opened 2025-10-03
Revise help text for --adapt option to `component new` / `component link`

catamorphism:
The message was also duplicated across the `new` and `link` commands, so this factors it out.


catamorphism:
Note: I'm not positive my new text for `--adapt` is correct, but it seems more correct than the old text, which implied that the filename is optional (as far as I can tell, it's not?) Please let me know if it's wrong and I'll revise.

PR #2331, opened 2025-10-03
Add more traits to iterators in `wit-dylib-ffi`

alexcrichton:
I found these handy when using the crate in a downstream scenario.


cpetig:
Looks good and like a benefit without downsides to me. (I am not an official reviewer, but I approve)

PR #2333, opened 2025-10-06
When renaming resources, resolve their methods as well

### LINKED ISSUE #2316: WIT `include` does not rename resources correctly
This input:

```wit
package a:b;

world a {
  resource r {
    foo: func();
  }
}


world name {
  include a with { r as other }
}
```

yields:

```
$ wasm-tools component wit -t a.wit
error: import name `[method]r.foo` is not valid
function does not match expected resource name `other` (at offset 0x4c)
```

primoly:
Closes #2316


alexcrichton (approved):
Thanks!

PR #2340, opened 2025-10-07
Allow long symbol names

### LINKED ISSUE #2332: String size out of bounds in linking section
I have received some compiled rust project with very long function names. When compiled with `-Clink-args=--emit-relocs`, the names llvm emits for some calls to `core::ptr::drop_in_place` get especially long since they include the full typename. There seems to be no limit enforced upstream by llvm when emitting this. The limit enforced during parsing is thus incorrect and not covered [by the linked implementation defined limits](https://webassembly.github.io/spec/js-api/#limits).

The name in question is obviously too long to repeat here, but makes it "barely" over the limit with a length of 110012 characters (verified by hand that this length is correct, string goes from offset 0x3889a307 to 0x388b50c3 in the file).

WorldSEnder:
Read symbol names in the custom linking section without the usual `limits::MAX_WASM_STRING_SIZE`.

Internal symbols can contain the name of generated closure types and monomorphized generic arguments, leading to very long symbol names. While often not semantically meaningful to an engine, these strings are useful for debugging and not subject to any length restriction when emitted by llvm.

Fixes #2332

### DIFF
--- a/crates/wasmparser/src/readers/core/linking.rs
+++ b/crates/wasmparser/src/readers/core/linking.rs
@@ -299,7 +299,7 @@ impl<'a> FromReader<'a> for SymbolInfo<'a> {
             SYMTAB_FUNCTION | SYMTAB_GLOBAL | SYMTAB_EVENT | SYMTAB_TABLE => {
                 let index = reader.read_var_u32()?;
                 let name = match defined || explicit_name {
-                    true => Some(reader.read_string()?),
+                    true => Some(reader.read_unlimited_string()?),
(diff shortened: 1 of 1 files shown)


WorldSEnder:
There is one more call to `read_string` in the impl for `Comdat<'_>` but I'm not too familiar with this section's use and if it might be worth to adjust it in this PR. Let me know.

alexcrichton:
Yeah that should be reasonable to change, basically anything in custom sections probably shouldn't be using `read_string`

PR #2355, opened 2025-10-15
feat(wit-parser): allow the non-first segments of the kebab name to start with a number.

chenyan2002:
Partially address https://github.com/WebAssembly/component-model/issues/345.

Allow the non-first segments of the kebab name to start with a number.

### DIFF
--- a/tests/cli/component-model/naming.wast
+++ b/tests/cli/component-model/naming.wast
@@ -128,3 +128,7 @@
     (import "[static]a.a" (func))
   )
   "import name `[static]a.a` conflicts with previous name `a`")
+
+(component
+  (type (flags "a-1-c"))
+)
--- a/tests/snapshots/cli/component-model/naming.wast.json
+++ b/tests/snapshots/cli/component-model/naming.wast.json
@@ -130,6 +130,12 @@
       "filename": "naming.18.wasm",
       "module_type": "binary",
       "text": "import name `[static]a.a` conflicts with previous name `a`"
+    },
+    {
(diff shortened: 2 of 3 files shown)


alexcrichton:
Thanks! In working on https://github.com/WebAssembly/component-model/issues/345 one thing I want to consider though is the roadmap for how it's going to be implemented. For example I think it would best to avoid situations where half the system works and half the system doesn't due to the order in which things were done. Do you have an idea of how much of he rest of wasm-tools/wasmtime/etc need to be updated for this change? I would naively expect that wasmparser validation might need changing as well but not much else other than that. I wanted to see if you had plans already made though perhaps.

chenyan2002:
I think that's all we needed. I tried locally with my wit file. Both `wasm-tools` and `wac` (after a bump) are working as expected.

tests/cli/component-model/naming.wast
```diff
@@ -24,9 +24,9 @@
 
 (assert_invalid
   (component
-    (type (flags "a-1-c"))
+    (type (flags "0-a-1-c"))
   )
-  "flag name `a-1-c` is not in kebab case"
+  "flag name `0-a-1-c` is not in kebab case"
 )
```
alexcrichton, line 30: In the interest of having `*.wast` tests for as much as possible, could the preexisting test be re-added as an "assert this is valid"-style component?

chenyan2002: Done

PR #2356, opened 2025-10-17
feat(wit): add map type support

yordis:
Related [Add a 'map<K,V>' type](https://github.com/WebAssembly/component-model/pull/554)


alexcrichton:
Thanks for this! Are you looking to get this merged ASAP? Or is this more of a proof-of-concept to support the development of the PR on the spec itself? If you're looking to get this merged sooner-rather-than-later, do you have plans to work on other parts of the stack such as wit-bindgen or Wasmtime?

yordis:
@alexcrichton no rush at all in my end. I want to learn how to do the whole thing if possible if that is ok with you

yordis:
Note for self: `0x63` why is that typed in so many places instead of having a enum somewhere?

crates/wit-component/src/encoding.rs
```diff
@@ -324,6 +324,9 @@ impl TypeContents {
                 }
                 TypeDefKind::Enum(_) => Self::empty(),
                 TypeDefKind::List(t) => Self::for_type(resolve, t) | Self::LIST,
+                TypeDefKind::Map(k, v) => {
+                    Self::for_type(resolve, k) | Self::for_type(resolve, v) | Self::LIST
```
yordis:
```suggestion
                    Self::for_type(resolve, k) | Self::for_type(resolve, v)
```

Should this be `Self::MAP` or just removed?

crates/wit-parser/src/ast/resolve.rs
```diff
@@ -1373,6 +1405,9 @@ impl<'a> Resolver<'a> {
                 TypeDefKind::List(ty)
                 | TypeDefKind::FixedSizeList(ty, _)
                 | TypeDefKind::Option(ty) => find_in_type(types, *ty),
+                TypeDefKind::Map(k, v) => {
+                    find_in_type(types, *k).or_else(|| find_in_type(types, *v))
```
yordis, line 1398: I am confused on this one

crates/wit-parser/src/ast/resolve.rs
```diff
@@ -1186,6 +1187,37 @@ impl<'a> Resolver<'a> {
                 let ty = self.resolve_type(&list.ty, stability)?;
                 TypeDefKind::List(ty)
             }
+            ast::Type::Map(map) => {
+                let key_ty = self.resolve_type(&map.key, stability)?;
+                let value_ty = self.resolve_type(&map.value, stability)?;
+
+                // Validate key type according to spec: only bool, integers, char, and string allowed
+                let is_valid_key = match key_ty {
```
yordis: should this be done here?

crates/wasmparser/src/validator/component_types.rs
```diff
@@ -1646,7 +1654,9 @@ impl ComponentDefinedType {
 
             ComponentDefinedType::Variant(ty) => ty.lower_gc(types, abi, options, offset, core),
 
-            ComponentDefinedType::List(ty) | ComponentDefinedType::FixedSizeList(ty, _) => {
+            ComponentDefinedType::List(ty)
+            | ComponentDefinedType::Map(ty, _)
```
alexcrichton: For this I think it would be best for `Map` to be routed to some sort of "not implemented" error and you can defer the work to someone fleshing out GC support in the future

crates/wit-component/src/encoding.rs
alexcrichton: If you're up for it renaming `LIST` to something like `NEEDS_MEMORY` or something like that would be appropriate. Otherwise though this looks correct where `LIST` is mostly a proxy for "needs a `memory` canon option" which is always true for `map`

crates/wit-dylib/ffi/src/ffi.rs
```diff
@@ -159,6 +160,15 @@ pub struct wit_list {
...
 #[derive(Debug, Copy, Clone)]
+pub struct wit_map {
+    pub interface: *const ::std::os::raw::c_char,
+    pub name: *const ::std::os::raw::c_char,
+    pub key_ty: wit_type_t,
+    pub value_ty: wit_type_t,
+}
+pub type wit_map_t = wit_map;
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
```
alexcrichton: For this I'd recommend skipping wit-dylib entirely. I ended up skipping fixed-size-lists mostly for example and figure it's fine to come back later and add support as needed.

alexcrichton: Although, if you're specifically interested in getting this working mind splitting it out to a separate follow-up PR?

crates/wit-parser/src/ast/resolve.rs
alexcrichton: Personally I'd say that this should be deferred entirely. WIT has some validation predicates which aren't always captured in WIT itself and are mostly enforced by the component validator itself. Given that wasm validation will be running this check anyway I think it'd be reasonable to just skip this entirely and to lean on wasm validation

crates/wit-parser/src/ast/resolve.rs
alexcrichton: This is fine yeah, and close enough to an approximation.

alexcrichton:
@yordis it looks like the feature gating still isn't implemented?

yordis:
@alexcrichton 🤦🏻 sorry ... I completely forgot focusing on making CI to pass again resolving the comments ... let me work on that

yordis:
> cargo fmt              
> Error writing files: failed to resolve mod `bindings`: /Users/yordisstudio/Developer/github.com/bytecodealliance/wasm-tools/playground/component/src/bindings.rs does not exist


@alexcrichton when I ran `cargo fmt` it seems that there is a missing file here, what is reason for that file to be missing from the code base? Or rather, at the very least be added to a `.gitignore` so I can keep a dummy file to make the formatter to work

alexcrichton:
Sorry I'm not sure what you're referring to? If you run `find . -name '*.rs' | xargs rustfmt --check --edition 2021` like CI does, does that fix things? (no CI is currently being run due to a merge conflict)

yordis:
> Sorry I'm not sure what you're referring to? If you run find . -name '*.rs' | xargs rustfmt --check --edition 2021 like CI does, does that fix things?

Yeah, just wondering why the `cargo fmt` by itself wouldn't work here, no biggy

crates/wit-component/tests/interfaces.rs
```diff
@@ -61,7 +62,7 @@ fn run_test(path: &Path, is_dir: bool) -> Result<()> {
     let wasm = wit_component::encode(&resolve, package)?;
     let wat = wasmprinter::print_bytes(&wasm)?;
     assert_output(&path.with_extension("wat"), &wat)?;
-    wasmparser::Validator::new()
+    wasmparser::Validator::new_with_features(WasmFeatures::all())
```
yordis, line 65: Is this required? I got lost in terms of why my tests didn't work

alexcrichton: Yeah this makes sense to me that it's required, thanks for adding!

crates/wasmparser/src/features.rs
```diff
@@ -286,6 +286,11 @@ define_wasm_features! {
         /// Corresponds to the 🛸 character in
         /// <https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md>.
         pub cm_gc: CM_GC(1 << 33) = false;
+        /// Support for maps in the component model proposal.
+        ///
+        /// Corresponds to the 🗺️ character in
+        /// <https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md>.
+        pub cm_map: CM_MAP(1 << 37) = false;
```
alexcrichton: I think the 37 here conflicts with `compact_imports` below now. I think it's reasonable to leave this adjacent to `cm_gc`, but feel free to renumber the below bits to ensure the biggest number is last

yordis: It scares me that my tests passed 😭 I do not like things no breaking with these type of changes.

Could this use a enum thou?

crates/wasmparser/src/features.rs
```diff
@@ -306,6 +306,12 @@ define_wasm_features! {
 
         // Compact import section proposal.
         pub compact_imports: COMPACT_IMPORTS(1 << 37) = false;
+
+        /// Support for maps in the component model proposal.
+        ///
+        /// Corresponds to the 🗺️ character in
+        /// <https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md>.
+        pub cm_map: CM_MAP(1 << 38) = false;
```
yordis, line 314: I rather keep it based on the flag value than grouping things ... I think that caused the issue to begin with

PR #2362, opened 2025-10-23
various `wit-dylib` API tweaks

dicej:
I've recently ported `componentize-py` to use `wit-dylib` instead of its own Wasm code generator and found it useful to tweak the `wit-dylib` APIs a bit.

The main idea here is to give `componentize-py` (and any similar build tool) access to the metadata generated by `wit-dylib` _before_ the guest runs, as well as match the types in that metadata up with the `wit_parser::TypeId`s from the input world.  That allows the build tool to generate high-level binding code ahead of time and later match it up with the metadata at runtime.

Specifically, this commit does the following:

- Add a `wit_dylib::create_with_metadata` function which returns both the generated module and the `Metadata` object produced for it.
- Add `id: TypeId` fields to all the "WIT type types" so that they can be matched up with the corresponding `wit-parser` types
- Add `index` methods to the `wit-dylib-ffi` wrapper types, allowing the interpreter to determine where the instance appears in its respective array
- Split the metadata `Function` type into separate `ImportFunction` and `ExportFunction` types, each with its own array in the `Wit` structure.  Imported and exported functions are used very differently by interpreters; separating them makes them easier to deal with.

### DIFF
--- a/crates/wit-dylib/ffi/src/types.rs
+++ b/crates/wit-dylib/ffi/src/types.rs
@@ -338,6 +338,10 @@ pub struct ImportFunction {
 impl_extra_traits!(ImportFunction);
 
 impl ImportFunction {
+    pub fn index(&self) -> usize {
+        unsafe { (&raw const *self.ptr).offset_from_unsigned(self.wit.ptr.import_funcs) }
+    }
+
     pub fn interface(&self) -> Option<&'static str> {
         unsafe { opt_str(self.ptr.interface) }
     }
@@ -917,6 +921,10 @@ pub struct List {
 impl_extra_traits!(List);
 
 impl List {
+    pub fn index(&self) -> usize {
+        unsafe { (&raw const *self.ptr).offset_from_unsigned(self.wit.ptr.lists) }
+    }
+
     pub fn interface(&self) -> Option<&'static str> {
         unsafe { opt_str(self.ptr.interface) }
     }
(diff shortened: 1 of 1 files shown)


crates/wit-dylib/ffi/src/types.rs
```diff
@@ -410,9 +427,57 @@ impl Function {
...
+    wit: Wit,
+    ptr: &'static ffi::wit_export_func_t,
+}
+
+impl_extra_traits!(ExportFunction);
+
+impl ExportFunction {
+    pub fn index(&self) -> usize {
+        usize::try_from(unsafe { (&raw const *self.ptr).offset_from(self.wit.ptr.export_funcs) })
+            .unwrap()
```
alexcrichton: For these `index` functions, perhaps [offset_from_unsigned](https://doc.rust-lang.org/stable/std/primitive.pointer.html#method.offset_from_unsigned) to avoid the `try_from`?

crates/wit-dylib/wit_dylib.h
alexcrichton: Mind changing this to `WIT_CURRENT_VERSION` and bump it to 1? (and then update the encoding to emit that instead)

crates/wit-dylib/ffi/src/types.rs
alexcrichton: Also, I tink `ImportFunction` might be missing the new `index` method?

dicej: > Also, I tink `ImportFunction` might be missing the new `index` method?

I don't have any need for an `index` method there, and I'm not sure if it will be useful for anyone.  Maybe we wait and see if anyone asks for it?

dicej: BTW, I didn't add an `index` method to `List`, `FixedSizeList`, `Future`, etc. either since I don't have any use for them (at least not yet).  I guess we could proactively add them for consistency, though.

dicej: I'll go ahead and add them; can't hurt.

PR #2367, opened 2025-11-04
flesh out async support in `wit-dylib`

dicej:
This does two things:

First, it adds a low-level API to `wit-dylib-ffi` for async imports and exports as an alternative to the high-level API based on `wit-bindgen`'s `rt` library and enabled with the `async` feature.  Thus enabling the `async` feature now really just means "use the high-level API", while disabling it means "use the low-level API".  The low-level API can make more sense for e.g. Python, where the coroutine model is different enough from Rust's that it's easier to map it directly to the component model ABI rather than try to plumb it through Rust's async APIs.

Second, it populates the `wit_future` and `wit_stream` types with all the relevant intrinsics, ABI size/align metadata, and generated payload lift and lower functions.

I've tested this all end-to-end with `componentize-py`, but could use some guidance on how to add tests here as well.

### DIFF
--- a/.github/actions/install-rust/action.yml
+++ b/.github/actions/install-rust/action.yml
@@ -70,4 +70,4 @@ runs:
     - name: Setup `wasmtime`
       uses: bytecodealliance/actions/wasmtime/setup@v1
       with:
-        version: "37.0.0"
+        version: "dev"


crates/wit-dylib/ffi/src/lib.rs
```diff
@@ -372,12 +372,19 @@ pub trait Interpreter: 'static {
...
         wit: Wit,
         func: ExportFunction,
         cx: Box<Self::CallCx<'static>>,
     ) -> impl std::future::Future<Output = ()>;
 
+    #[cfg(not(feature = "async"))]
+    fn export_async_start(wit: Wit, func: ExportFunction, cx: Box<Self::CallCx<'static>>) -> u32;
+
+    #[cfg(not(feature = "async"))]
+    fn export_async_callback(event0: u32, event1: u32, event2: u32) -> u32;
```
alexcrichton: Mind splitting the two bits of functionality here behind two features? Maybe `async-raw` and `async-runtime` or something like that. And then perhaps a `compile_error!` if they're both enabled.

crates/wit-dylib/src/bindgen.rs
```diff
@@ -425,6 +425,48 @@ pub fn task_return(
     )
 }
 
+pub fn lift_payload(adapter: &mut Adapter, resolve: &Resolve, ty: Type) -> Function {
+    let mut c = FunctionCompiler::new(adapter, resolve, 2);
+    c.ctx = Some(TempLocal::new(0, ValType::I32));
+    let frame = StackFrame {
```
alexcrichton: Mind leaving a comment here as to why a stack frame is allocated?

dicej: Turns out we only actually needed it for `lower_payload`, so I've removed this one and added a comment to `lower_payload`.

dicej:
The CI failures are due to issues fixed in https://github.com/bytecodealliance/wasmtime/pull/11981.  The tests pass when using Wasmtime `main`.

PR #2368, opened 2025-11-05
WIT annotation examples

mxgordon:
Claude generated some good examples I think. They contain a lot of annotations, so we can always cut down, but I think it has some good ideas to keep in mind while designing.


mxgordon:
Apologies, ignore this

PR #2369, opened 2025-11-07
WIP: create 4-byte `struct ValTypePacked` to offset larger `enum ValType`

fitzgen:


fitzgen:
What still needs to be done is to integrate with `MaybeValType` and the validator

alexcrichton:
Some more discussion [here](https://github.com/bytecodealliance/wasm-tools/pull/2295#issuecomment-3504159080), but for now the conclusion is that it's best to leave `ValType` as-is and avoid `*Packed` types.

PR #2376, opened 2025-11-13
Bake `async` into component model function types

alexcrichton:
This commit is a breaking change to the component-model-async implementation in this repository. This will effectively invalidate all WASIp3 binaries created prior to this commit. Given that this feature is off-by-default everywhere (right? right?) it's expected that while this will have churn it's the appropriate time to make such a change as this.

Concretely the changes here are:

* Function types in the component model now reflect whether they're `async`-or-not. This is no longer WIT-level sugar.
* Kebab names no longer support `[async]`, `[async method]`, or `[async static]`.
* The text format now supports `(func async? (param ...))`.

The binary format previously used 0x40 as the prefix byte for "this is a component function type" and now 0x43 is used as "this is an async function type". The `async_` boolean is plumbed throughout component function type locations and is required to be handled by callers. For now there is no subtyping relationship between async and non-async functions, they must be exactly the same when linking.

The rationale for this change is going to be expanded on more fully in an upcoming PR to the component-model repository itself. The rough tl;dr; is that to fully support JS in/out of components the component model is going to need to require traps in some dynamic situations such as when a synchronous function-type function tries to block. This is required to respect JS's "cannot block the main thread" semantics, for example. This will be more fully explain in the component-model PR and this PR is not intended to serve as justification alone for this change. Instead this PR is intended to change/update Wasmtime as quickly as possible to understand the new binary format so implementation work can proceed as most of the work here is on the runtime side of things, not validation.

One minor caveat for this change is that the `wit-dylib` crate has some tests which exercise async functionality. Those are now all disabled or set as "expect this to fail" because Wasmtime, the runner for these tests, no longer understands the binaries it's ingesting.

### DIFF
--- a/crates/wit-component/src/encoding/types.rs
+++ b/crates/wit-component/src/encoding/types.rs
@@ -10,6 +10,7 @@ use wit_parser::{
 /// Represents a key type for interface function definitions.
 #[derive(Hash, PartialEq, Eq, Clone)]
 pub struct FunctionKey<'a> {
+    async_: bool,
     params: &'a [(String, Type)],
     result: &'a Option<Type>,
 }
@@ -98,6 +99,7 @@ pub trait ValtypeEncoder<'a> {
     /// document.
     fn encode_func_type(&mut self, resolve: &'a Resolve, func: &'a Function) -> Result<u32> {
         let key = FunctionKey {
+            async_: func.kind.is_async(),
             params: &func.params,
             result: &func.result,
         };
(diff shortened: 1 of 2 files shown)


tschneidereit:
@vados-cosmonic as a heads-up, this will be quite important for JCO.

cpetig:
I believe the proper link to the spec change is https://github.com/WebAssembly/component-model/pull/578 (I was expecting a link from there but didn't find it)

PR #2382, opened 2025-11-17
require memory for async lifts as needed

dicej:
Per https://github.com/WebAssembly/component-model/pull/575, both `wasm-tools` and Wasmtime have been too permissive, allowing async lifts to optionally omit a memory canonopt if only the return type (and not the parameter types) would need one.  This fixes validation to require the option in such cases.


alexcrichton:
I'll work on the test failures, but can you add a test along the lines of https://github.com/bytecodealliance/wasmtime/issues/11726 to ensure that wit-component correctly adds the canonical options necessary for the lift too?

dicej:
> I'll work on the test failures, but can you add a test along the lines of [bytecodealliance/wasmtime#11726](https://github.com/bytecodealliance/wasmtime/issues/11726) to ensure that wit-component correctly adds the canonical options necessary for the lift too?

Yeah, I realized there was more to do right after I opened this 🤦

alexcrichton:
Ok pushed up a commit which should resolve the test failure

alexcrichton:
Also, on second thought, shouldn't this also require `realloc` if the result contains a pointer? E.g. if you return a string we want the same restriction in place where the lift requires a realloc even if the params don't need it

dicej:
> Also, on second thought, shouldn't this also require `realloc` if the result contains a pointer? E.g. if you return a string we want the same restriction in place where the lift requires a realloc even if the params don't need it

When would lifting ever require `realloc`?  Any allocations would have already been performed by the guest by the time it calls `task.return`.

alexcrichton (approved):
Oops yes carry on, got my polarities swapped

Issue #2384, opened 2025-11-21
`wasm-smith`: add support for generating instructions from the `threads` proposal


fitzgen:
Right now, `wasm_smith::Config::threads_enabled` only configures whether we will generate shared memories or not. We should also emit atomic instructions and wait/notify instructions.

The instructions we should support: https://webassembly.github.io/threads/core/syntax/instructions.html#syntax-instr-atomic-memory

The place we need to add that support:

https://github.com/bytecodealliance/wasm-tools/blob/44b5ea4280afa4fd41d118c19d3df11e3e2c23aa/crates/wasm-smith/src/core/code_builder.rs#L76-L93

danielvallance:
Hello @fitzgen , I have not worked on this repo before however I would like to get involved and so I have created a pull request for this issue: https://github.com/bytecodealliance/wasm-tools/pull/2386

If you would like me to change anything please let me know, thanks!

abrown:
This was completed in #2386; eventually we should add the shared-everything-threads instructions as well!

PR #2386, opened 2025-11-23
Generate atomic instructions defined in the threads proposal

danielvallance:
This commit adds suppport for emitting atomic instructions when wasm_smith::Config::threads_enabled and
wasm_smith::Config::shared_everything_threads_enabled are true.

A link to the atomic instructions in the threads proposal: https://webassembly.github.io/threads/core/syntax/instructions.html#syntax-instr-atomic-memory

Addresses https://github.com/bytecodealliance/wasm-tools/issues/2384

### DIFF
--- a/crates/wasm-smith/src/core/code_builder.rs
+++ b/crates/wasm-smith/src/core/code_builder.rs
@@ -594,6 +594,74 @@ instructions! {
     (Some(wide_arithmetic_binop128_on_stack), i64_sub128, NumericInt),
     (Some(wide_arithmetic_mul_wide_on_stack), i64_mul_wide_s, NumericInt),
     (Some(wide_arithmetic_mul_wide_on_stack), i64_mul_wide_u, NumericInt),
+    // Atomic instructions
+    (Some(atomic_load_valid), i32_atomic_load, MemoryInt),
+    (Some(atomic_load_valid), i64_atomic_load, MemoryInt),
+    (Some(atomic_load_valid), i32_atomic_load8_u,MemoryInt),
+    (Some(atomic_load_valid), i64_atomic_load8_u, MemoryInt),
+    (Some(atomic_load_valid), i32_atomic_load16_u, MemoryInt),
+    (Some(atomic_load_valid), i64_atomic_load16_u, MemoryInt),
+    (Some(atomic_load_valid), i64_atomic_load32_u, MemoryInt),
+    (Some(i32_atomic_store_valid), i32_atomic_store, MemoryInt),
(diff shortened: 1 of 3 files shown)


dicej:
@abrown Would you have a few minutes to look at this?  I'm not up-to-speed yet on shared-everything threads.

tschneidereit:
FYI, Andrew is on leave right now, so it might be a bit until he'll be able to look at this. (Not providing any timelines here because once he is back, he might have a lot to work through before getting to this for all I know.)

crates/wasm-smith/src/core/code_builder.rs
```diff
@@ -7333,3 +7401,260 @@ fn i64_mul_wide_u(
     instructions.push(Instruction::I64MulWideU);
     Ok(())
 }
+
+#[inline]
+fn atomic_instruction_valid(module: &Module, _builder: &mut CodeBuilder) -> bool {
+    module.config.threads_enabled && module.config.shared_everything_threads_enabled
+}
+
```
alexcrichton, line 7409: For this it's ok to drop `shared_everything_threads_enabled` since these instructions only require the wasm `threads` proposal.

danielvallance: Ahh ok I see, will do, thanks!

danielvallance: Hi @alexcrichton , I have run this for 15 minutes and I did not see any issues apart from some warnings at the start which appear on the main branch also:

WARNING: Failed to find function "__sanitizer_acquire_crash_state". Reason dlsym(RTLD_DEFAULT, __sanitizer_acquire_crash_state): symbol not found.
WARNING: Failed to find function "__sanitizer_print_stack_trace". Reason dlsym(RTLD_DEFAULT, __sanitizer_print_stack_trace): symbol not found.
WARNING: Failed to find function "__sanitizer_set_death_callback". Reason dlsym(RTLD_DEFAULT, __sanitizer_set_death_callback): symbol not found.

Just to sanity check that I was running the tool correctly and that I could recognise an error, I also deliberately added a mistake to it and it fell over after a couple of seconds with a very clear and obvious error.

Would you like me to attach the output of the successful run?

alexcrichton: Nah that sounds good, thanks for double-checking! And yeah those warnings are ok to ignore and otherwise having no "THIS DIED" sort of output is expected.

alexcrichton (approved):

PR #2390, opened 2025-11-25
Support special import names for unit futures/streams

alexcrichton:
This commit enhances `wit-component`'s import detection to support "unit" `future` and `stream` types which don't have a payload in a special way. Previously it was required that these types were mentioned in some WIT somewhere to get connected to a function, but now it's possible to import these intrinsics directly even if the type isn't mentioned in WIT.

The motivation for this change is to support guest language async runtimes which are world-agnostic to still operate on unit futures/streams as necessary without binding them to any particular WIT import or export. For example it's expected that unit streams are a primitive for inter-task communication in the component model.

### DIFF
--- a/crates/wit-component/src/validation.rs
+++ b/crates/wit-component/src/validation.rs
@@ -142,7 +142,7 @@ pub struct PayloadInfo {
     pub name: String,
     /// The resolved type id for the `stream` or `future` type of interest.
     ///
-    /// If `None` this means that it's a "unit" payload or has no associated
+    /// If `Unit{Future,Stream}` this means that it's a "unit" payload or has no associated
     /// type being sent.
     pub ty: PayloadType,
     /// The world key representing the import or export context of `function`.


crates/wit-component/src/validation.rs
```diff
@@ -135,16 +135,16 @@ pub enum ImportInstance {
...
     /// The original, mangled import name used to import this built-in
     /// (currently used only for hashing and debugging).
     pub name: String,
     /// The resolved type id for the `stream` or `future` type of interest.
-    pub ty: TypeId,
-    /// The component-level function import or export where the type appeared as
-    /// a parameter or result type.
-    pub function: String,
+    ///
+    /// If `None` this means that it's a "unit" payload or has no associated
```
dicej:
```suggestion
    /// If `Unit{Future,Stream}` this means that it's a "unit" payload or has no associated
```

PR #2391, opened 2025-11-25
Fix exact import in custom descriptors

yurydelendik:
Allows exact function imports such as:

     (import "" "" (func (exact (type $f))))
     (import "" "" (func (exact (type $f) (param) (result))))

The patch adds FuncExact to external kind and uses heap type reference (instead of index) in the entity function references field. I tried to match changes to [the overview](https://github.com/WebAssembly/custom-descriptors/blob/main/proposals/custom-descriptors/Overview.md) explanations in the Exact Function Imports sections.

I need help with:
- [x] component changes: is it make sense to have the similar change there too?
- [x] wasm-smith or wasm-mutate changes


crates/wasm-encoder/src/core/imports.rs
```diff
@@ -1,16 +1,26 @@
...
+}
+
 /// The type of an entity.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum EntityType {
     /// A function type.
     ///
     /// The value is an index into the types section.
-    Function(u32),
+    Function(FunctionHeapType),
```
alexcrichton: To bikeshed this a bit, this might integrate better with existing `match`es and such with a similar pattern as to wasmparser where a new `FunctionExact` variant is added to the `enum` rather than using `exact: bool` on preexisting variants

crates/wasmparser/src/readers/core/imports.rs
```diff
@@ -18,13 +18,20 @@ use crate::{
...
+}
+
 /// Represents a reference to a type definition in a WebAssembly module.
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
 pub enum TypeRef {
     /// The type is a function.
-    ///
-    /// The value is an index into the type section.
-    Func(u32),
+    Func(FuncHeapType),
```
alexcrichton: Similar to wasm-encoder, I'd bikeshed this to throw out having `FuncExact` as a variant here.

yurydelendik:
>  in Rust it'd be nicer to have Thing::{Func,FuncExact}(u32) as opposed to Thing::Func(ThingType { id: u32, exact: bool }). Did you try that already though and find that it was unworkable? If not, mind trying that out?

Yep, I can try that.

crates/wasm-encoder/src/reencode.rs
```diff
@@ -74,6 +74,7 @@ pub trait Reencode {
             wasmparser::ExternalKind::Memory => self.memory_index(index),
             wasmparser::ExternalKind::Global => self.global_index(index),
             wasmparser::ExternalKind::Tag => self.tag_index(index),
+            wasmparser::ExternalKind::FuncExact => panic!("Unexpected func_exact"),
```
alexcrichton: For this I think it should be fine to do `self.function_index(index)` -- while technically this'll never happen since exports can't have `FuncExact` we can't statically rule it out and most other places handle this by "just do the func thing" so it'd probably best to mirror that here as "just do the func thing"

crates/wasm-smith/src/core.rs
```diff
@@ -1363,6 +1363,8 @@ impl Module {
                     }
                 }
 
+                wasmparser::TypeRef::FuncExact(_) => panic!("Unexpected func_exact import"),
```
alexcrichton, line 1366: These panics are all fine, they can get filled in later if necessary

alexcrichton:
Nice that looks like it worked out well to me at least, with green CI should be good to go 👍

yurydelendik:
> Judging by [WebAssembly/testsuite#146](https://github.com/WebAssembly/testsuite/pull/146) it looks like some instruction names may be getting an update too? (fine to land after this of course)

Yes, this is my next item. The newest update to the proposal will rollback some code around `struct.new` validation and adds `struct.new_desc`

PR #2394, opened 2025-11-30
Fix foreign interface dependencies which are disabled by feature-gating

danielvallance:
A foreign imported interface can be missing from pkg.interfaces either because it does not exist, or because it was disabled by feature-gating.

Previously, the parser failed in both cases, when it should only fail when it does not exist, and should succeed when the references to it are disabled by feature-gating.

By checking the set of unresolved worlds' imports and exports for references to this interface, it is possible to determine if they contain any references to the interface, and whether or not they were disabled by feature gating.

If they are all disabled, then it does not matter that the interface is absent from pkg.interfaces, and the code should not fail. If any references are active however, then this is an error as the code will not be able to find the interface in pkg.interfaces.

If no references exist at all, then no combination of feature enabling/disabling would satisfy the import/use statement which requires the interface, and the code should fail.

This fixes issue 2285 and a unit test was added to cover this case.

### DIFF
--- a/crates/wit-parser/src/ast/resolve.rs
+++ b/crates/wit-parser/src/ast/resolve.rs
@@ -46,8 +46,10 @@ pub struct Resolver<'a> {
     /// Metadata about foreign dependencies which are not defined in this
     /// package. This map is keyed by the name of the package being imported
     /// from. The next level of key is the name of the interface being imported
-    /// from, and the final value is the assigned ID of the interface.
-    foreign_deps: IndexMap<PackageName, IndexMap<&'a str, AstItem>>,
+    /// from, and the final value is a tuple containing the assigned ID of the
+    /// dependency, and a Vector of the Stability attributes associated with each
+    /// of its imports.
+    foreign_deps: IndexMap<PackageName, IndexMap<&'a str, (AstItem, Vec<Stability>)>>,
 
     /// All interfaces that are present within `self.foreign_deps`.
     foreign_interfaces: HashSet<InterfaceId>,
@@ -233,7 +235,9 @@ impl<'a> Resolver<'a> {
                     (
                         name.clone(),
                         deps.iter()
-                            .map(|(name, id)| (name.to_string(), *id))
+                            .map(|(name, (id, stabilities))| {
+                                (name.to_string(), (*id, stabilities.clone()))
+                            })
                             .collect(),
                     )
                 })
@@ -257,18 +261,20 @@ impl<'a> Resolver<'a> {
         let mut foreign_worlds = mem::take(&mut self.foreign_worlds);
         for decl_list in decl_lists {
             decl_list
-                .for_each_path(&mut |_, _attrs, path, _names, world_or_iface| {
+                .for_each_path(&mut |_, attrs, path, _names, world_or_iface| {
                     let (id, name) = match path {
                         ast::UsePath::Package { id, name } => (id, name),
                         _ => return Ok(()),
                     };
 
+                    let stability = self.stability(attrs)?;
+
                     let deps = foreign_deps.entry(id.package_name()).or_insert_with(|| {
                         self.foreign_dep_spans.push(id.span);
                         IndexMap::new()
                     });
-                    let id = *deps.entry(name.name).or_insert_with(|| {
-                        match world_or_iface {
+                    let (id, stabilities) = deps.entry(name.name).or_insert_with(|| {
+                        let id = match world_or_iface {
                             WorldOrInterface::World => {
                                 log::trace!(
                                     "creating a world for foreign dep: {}/{}",
@@ -287,10 +293,13 @@ impl<'a> Resolver<'a> {
                                 );
                                 AstItem::Interface(self.alloc_interface(name.span))
                             }
-                        }
+                        };
+                        (id, Vec::new())
                     });
 
-                    let _ = match id {
+                    stabilities.push(stability);
+
+                    let _ = match *id {
                         AstItem::Interface(id) => foreign_interfaces.insert(id),
                         AstItem::World(id) => foreign_worlds.insert(id),
                     };
@@ -520,7 +529,7 @@ impl<'a> Resolver<'a> {
                                 )
                             })?,
                             ast::UsePath::Package { id, name } => {
-                                self.foreign_deps[&id.package_name()][name.name]
+                                self.foreign_deps[&id.package_name()][name.name].0
                             }
                         };
                         (name.name, item)
@@ -1104,7 +1113,7 @@ impl<'a> Resolver<'a> {
                 }
             }
             ast::UsePath::Package { id, name } => Ok((
-                self.foreign_deps[&id.package_name()][name.name],
+                self.foreign_deps[&id.package_name()][name.name].0,
                 name.name.into(),
                 name.span,
             )),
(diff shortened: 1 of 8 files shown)


danielvallance:
Hi @alexcrichton , yeah my initial approach was to just push None when the interface was not found in pkg.interfaces, however it caused some unit tests to fail as it results in successfully parsing some packages which import nonexistent interfaces. That's what led me to create the is_gated closure as a form of basic validation that the interface actually exists in some form. I could not see another place where it would be more appropriate to check the interface actually exists however I may be overlooking something as I am not super familiar with the code.

Just to confirm, we definitely want to fail in the case where a nonexistent interface is imported but not used, right?

(the unit tests which failed my initial approach were "tests/ui/parse-fail/bad-pkg3", "tests/ui/parse-fail/bad-pkg2" and "tests/ui/parse-fail/use-world")

alexcrichton:
Oof ok yeah that's pretty tricky. Those tests should indeed continue to keep failing and I see how "just push `None`" would result in those tests passing. I unfortunately don't see an easy solution to that...

Would it be possible to update how `include` is processed instead? That's sort of where the culprit lies here as well I think, and that'd be a more focused solution and I think less brittle than this current iteration. (I'm not exactly sure what this route would look like though or if it's viable)

danielvallance:
hi, yeah i can do some research in that area, thanks for the pointer! if i can find a more "principled" way of fixing this then i will update this PR

danielvallance:
i think i am in the right area with this latest change, however i am going to try and add some bookkeepping to keep track of the stability of different imports of the same dependency. i will mark as draft while i do this

danielvallance:
Latest change records the stability of each import of the foreign dependency, and a placeholder is pushed if and only if imports exist, and they are all feature gated

crates/wit-parser/src/ast/resolve.rs
```diff
@@ -47,7 +47,7 @@ pub struct Resolver<'a> {
     /// package. This map is keyed by the name of the package being imported
     /// from. The next level of key is the name of the interface being imported
     /// from, and the final value is the assigned ID of the interface.
-    foreign_deps: IndexMap<PackageName, IndexMap<&'a str, AstItem>>,
+    foreign_deps: IndexMap<PackageName, IndexMap<&'a str, (AstItem, Vec<Stability>)>>,
```
alexcrichton, line 52: Could you expand the above comment to explain what the `Vec<Stability>` here is? (either here or in `wit-parser/src/lib.rs`)

crates/wit-parser/src/resolve.rs
```diff
@@ -3148,6 +3162,7 @@ impl Remap {
         unresolved: &UnresolvedPackage,
         interface_to_package: &HashMap<InterfaceId, (&PackageName, &String, Span)>,
         resolve: &mut Resolve,
+        foreign_interface_stabilities: &HashMap<&Id<Interface>, &Vec<Stability>>,
```
alexcrichton: For this `InterfaceId` is idiomatically used instead of `Id<Interface>`, and `&[T]` is more idiomatic than `&Vec<T>`

crates/wit-parser/src/resolve.rs
```diff
@@ -3157,6 +3172,20 @@ impl Remap {
                 // further interfaces will be non-foreign as well.
                 None => break,
             };
+
+            if let Some(stabilities) = foreign_interface_stabilities.get(&unresolved_iface_id) {
+                let enabled = stabilities.iter().any(|stability| match stability {
+                    Stability::Unknown | Stability::Stable { .. } => true,
+                    Stability::Unstable { feature, .. } => {
+                        resolve.features.contains(feature) || resolve.all_features
+                    }
```
alexcrichton: Can this use `self.include_stability` to avoid duplicating the definition there?

crates/wit-parser/src/resolve.rs
```diff
@@ -3082,17 +3082,21 @@ impl Remap {
...
         let mut world_to_package = HashMap::new();
         let mut interface_to_package = HashMap::new();
+        let mut foreign_interface_stabilities = HashMap::new();
+        let mut foreign_world_stabilities = HashMap::new();
         for (i, (pkg_name, worlds_or_ifaces)) in unresolved.foreign_deps.iter().enumerate() {
-            for (name, item) in worlds_or_ifaces {
+            for (name, (item, stabilities)) in worlds_or_ifaces {
                 match item {
                     AstItem::Interface(unresolved_interface_id) => {
+                        foreign_interface_stabilities.insert(unresolved_interface_id, stabilities);
```
alexcrichton: To avoid the extra map here, could the `stabilities` be inserted into `{interface,world}_to_package`?

crates/wit-parser/src/resolve.rs
danielvallance: have removed this argument since it is included in the {interface, world}_to_package maps now

crates/wit-parser/src/resolve.rs
danielvallance: sure, have changed!

as you can see on the PR the additional logic which include_stability has for the Stable annotations (which is rightfully included now) did cause a build failure which I think was due to invalid WIT files being generated by wit-smith, and I have added a commit which I think fixes this

thanks for the review btw!

alexcrichton (approved):
Thanks!

crates/wit-smith/src/generate.rs
```diff
@@ -518,10 +518,6 @@ impl<'a> InterfaceGenerator<'a> {
     // Generate a feature gate annotation (@since, @unstable, or @deprecated)
     // If version is provided, ensures the annotation is compatible with the version
     fn gen_feature_annotation(&self, u: &mut Unstructured<'_>) -> Result<Option<String>> {
-        if u.arbitrary()? {
-            return Ok(None);
-        }
-
```
alexcrichton, line 524: er this comment didn't go through...

anyway, is this a stray change? Or intentional? (naively this looks related to helping to test this PR but not something that should be included, but wanted to confirm)

danielvallance: hi there, this is intentional!

after i initially addressed your comments in commit (https://github.com/bytecodealliance/wasm-tools/pull/2394/commits/4f9f171f45cc7371f6bcd5cc331b8d0615c19b70), it caused a test failure which seemed to me to be the added call to include_stability picking up on an invalid fuzzed WIT file which had a stable since annotation, but no overall version

This extra commit stops that from happening (i mentioned the rationale in this reply https://github.com/bytecodealliance/wasm-tools/pull/2394#discussion_r2636038430 but maybe i should have put it in a separate thread)

alexcrichton: Oh oops sorry I missed that. Agreed that error looks like invalid WIT is generated, but unconditionally generating stability annotations means that non-annotated things will no longer get fuzzed. Would you be ok updating this to always return `Ok(None)` and file an issue with a `FIXME` for the invalid-generated WIT?

danielvallance: hi there, had a closer look at this and i dont think it was invalid wit being generated, i was just passing the wrong pkgid into include_stability which meant it was looking at the wrong package when checking the consistency between the package's version and any version gating it did

i have changed this now, and also i had to change one of the test output files as invoking include_stability earlier on results in a cosmetic difference between the test outputs

PR #2398, opened 2025-12-08
Implement the compact imports proposal

bvisness:
The [Compact Import Section](https://github.com/WebAssembly/compact-import-section) proposal adds two new binary and text encodings to the import section that can express imports more concisely.


crates/wit-component/src/encoding/dedupe.rs
```diff
@@ -71,10 +71,13 @@ impl ModuleImportMap {
                 // import section in the `module` we're building.
                 ImportSection(i) => {
                     let mut new_import_section = ImportSection::new();
-                    for import in i.clone() {
-                        found_duplicate_imports = ret
-                            .push_import(&mut new_import_section, import?)?
-                            || found_duplicate_imports;
+                    for imports in i.clone() {
+                        let imports = imports?;
+                        for import in imports.iter() {
```
alexcrichton: Another helper impl I might recommend for this is something like:

```rust
impl SectionLimited<'_, Imports<'_>> {
    pub fn into_imports(self) -> impl Iterator<Item = Result<Import<'_>>>;
    pub fn into_imports_with_offsets(self) -> impl Iterator<Item = Result<(usize, Import<'_>)>>;
}
```

(or something like that)

and that way all preexisting users of `for x in section` can use `for x in section.into_imports()` and it should have similar semantics

crates/wasmprinter/src/lib.rs
```diff
@@ -1212,27 +1212,31 @@ impl Printer<'_, '_> {
...
-                    kind: _,
-                    func_type_idx: idx,
-                }) => {
-                    debug_assert!(state.core.tag_to_type.len() == state.core.tags as usize);
-                    state.core.tags += 1;
-                    state.core.tag_to_type.push(Some(idx))
-                }
-                TypeRef::Global(_) => state.core.globals += 1,
+
+            for import in imports.iter() {
```
alexcrichton: Eventually we'll want to update this to print the `(item ...)` form, mind leaving a `// TODO` here?

Although if tests are added in this PR too they'll fail here due to the round-trip-property not being preserved, so I can also help fill this out too if you'd prefer not to.

crates/wasmparser/src/readers/core/imports.rs
```diff
@@ -50,8 +62,85 @@ pub struct Import<'a> {
...
-pub type ImportSectionReader<'a> = SectionLimited<'a, Import<'a>>;
+pub type ImportSectionReader<'a> = SectionLimited<'a, Imports<'a>>;
+
+impl<'a> FromReader<'a> for Imports<'a> {
+    fn from_reader(reader: &mut BinaryReader<'a>) -> Result<Self> {
+        let module = reader.read_string()?;
+        let single_item_name = reader.read_string()?;
+        let discriminator = reader.peek_bytes(1)?[0];
+        match (single_item_name, discriminator) {
+            ("", 0x7F) => {
```
alexcrichton: For here (and below) can you add a check for `reader.compact_imports()` to ensure that the feature is enabled for this to be parsed? That'll require editing `features.rs` to add the feature there as well.

crates/wasm-encoder/src/reencode.rs
```diff
@@ -1435,27 +1435,28 @@ pub mod utils {
...
-            reencoder.parse_import(imports, import?)?;
+        for imports in section {
+            reencoder.parse_imports(import_section, imports?)?;
         }
         Ok(())
     }
 
-    /// Parses the single [`wasmparser::Import`] provided and adds it to the
-    /// `import` section.
-    pub fn parse_import<T: ?Sized + Reencode>(
```
alexcrichton, line 1444: Could this keep the old function the the functionality for `wasmparser::Import` (but keep the new one for `wasmparser::Imports` as well?)

crates/wasmparser/src/readers/core/imports.rs
```diff
@@ -50,8 +62,85 @@ pub struct Import<'a> {
...
+                    }
+                    Ok(())
+                })?;
+                Ok(Imports::Compact1(ImportGroup1 {
+                    module,
+                    items: SectionLimited::new(items)?,
+                }))
+            }
+            ("", 0x7E) => {
+                // Compact encoding 1: one module name / type, many item names
```
alexcrichton: s/encoding 1/encoding 2/

crates/wasmparser/src/readers/core/imports.rs
```diff
@@ -75,3 +173,73 @@ impl<'a> FromReader<'a> for TypeRef {
...
+impl<'a> Iterator for ImportsIter<'a> {
+    type Item = Result<(usize, Import<'a>)>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match self {
+            ImportsIter::Single(it) => it.next(),
+            ImportsIter::Many(it) => it.next(),
+        }
+    }
+}
```
alexcrichton: For a bikeshed here, one possibility would be to remove `Imports::iter` in favor of `SectionLimited::into_imports` entirely. That would means that the iterator could be something like:

```rust
pub struct ImportsIter<'a> {
    cur: None | ImportGroup1<'a> | ImportGroup2<'a>, // extracted out to an enum adjacent to this
    iter: SectionLimited<'a, Imports<'a>>,
}
```

where `fn next` would look like: `loop { /* check self.cur */;  /* check self.iter */ }` or something similar

bvisness: `SectionLimited::into_imports` was a very good suggestion, so thanks for that, but there are two spots where I don't have access to the `SectionLimited` and instead only have access to a specific `Imports` object (and in one case an offset). They are:

- wasmparser/src/validator.rs -> `import_section`, where I get `imports, offset` in a closure
- wasm-encoder/src/reencode.rs -> `parse_imports`, where I get just an Imports (no offset)

These are the two places where it seems I still might want an iterator implementation directly on Imports. But I am not good enough with Rust to understand how to implement what you suggested here.

For now I have all this duplication in here, but if I have an iterator on Imports, it really seems like I ought to be able to use that to drive the `SectionLimited::into_imports*()` methods. Would appreciate guidance on how to do that.

bvisness:
I see there are a gazillion CI failures that I don't get locally when running `cargo test`. I tried running `cargo test --all` and `cargo test --locked --all` (no idea what `--locked` does) but I just get:

```
   Compiling artifacts v0.1.0 (/home/bvisness/Developer/wasm-tools/crates/wit-dylib/test-programs/artifacts)
error: failed to run custom build command for `artifacts v0.1.0 (/home/bvisness/Developer/wasm-tools/crates/wit-dylib/test-programs/artifacts)`

Caused by:
  process didn't exit successfully: `/home/bvisness/Developer/wasm-tools/target/debug/build/artifacts-00f7e71926824b33/build-script-build` (exit status: 101)
  --- stderr

  thread 'main' (224713) panicked at crates/wit-dylib/test-programs/artifacts/build.rs:7:70:
  called `Result::unwrap()` on an `Err` value: NotPresent
```

Would appreciate guidance on how to actually run tests here.

crates/wasmparser/src/readers/core/imports.rs
bvisness: I guess in the reencoder case, I will be handling things differently anyway so that I re-encode to the compact encoding.

alexcrichton: I had to smuggle an offset around, but how about this -- https://github.com/bytecodealliance/wasm-tools/commit/da311734bdf478676fab8b223b9fa62a91d30b24

alexcrichton:
For testing `cargo test --all` should exercise almost all of CI yeah, it's mostly testing other packages in the workspace (`cargo test` only tests the top-level one). It's safe to ignore `--locked` as that's just a CI thing.

For the wit-dylib failure, that's a bad error message about the `WASI_SDK_PATH`  env var being set -- I'll work on improving the error message there at least. In the meantime you can pass `--exclude wit-dylib` to Cargo to ignore that crate for now and it should get all the others building and such

crates/wasmparser/src/readers/core/imports.rs
bvisness: That looks good, thanks for writing that up.

bvisness:
I am making good progress on the text parser, I think, but also, I have found myself in an absolute lifetime disaster here. I would really appreciate any guidance on how to resolve the errors I'm seeing. I am completely stumped. Surely I don't have to annotate the _entire program_ with 'a just to be able to iterate over imports in the AST?

crates/wast/src/core/resolve/deinline_import_export.rs
```diff
@@ -12,21 +12,24 @@ pub fn run(fields: &mut Vec<ModuleField>) {
                 }
                 match f.kind {
                     FuncKind::Import(import, exact) => {
-                        item = ModuleField::Import(Import {
+                        item = ModuleField::Import(Imports {
                             span: f.span,
-                            module: import.module,
-                            field: import.field,
-                            item: ItemSig {
+                            items: ImportItems::Single(Import {
```
alexcrichton: For this I might recommend an `Imports::single(import: Import)` helper constructor, but that's just sort of a stylistic preference

crates/wast/src/core/import.rs
```diff
@@ -16,21 +48,135 @@ pub struct Import<'a> {
...
+    type IntoIter = Box<dyn Iterator<Item = Self::Item> + 'a>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        match &mut self.items {
+            ImportItems::Single(import) => Box::new(std::iter::once(import)),
+            ImportItems::Group1 { module, items } => todo!(),
+            ImportItems::Group2 { module, sig, items } => todo!(),
+        }
+    }
+}
```
alexcrichton: These impls are where the CI issue lies and it can be fixed with this diff:

```diff
diff --git a/crates/wast/src/core/import.rs b/crates/wast/src/core/import.rs
index c98f2dd66..88d8787fc 100644
--- a/crates/wast/src/core/import.rs
+++ b/crates/wast/src/core/import.rs
@@ -151,9 +151,12 @@ impl<'a> Parse<'a> for ImportGroupItem<'a> {
     }
 }

-impl<'a> IntoIterator for &'a Imports<'a> {
-    type Item = &'a Import<'a>;
-    type IntoIter = Box<dyn Iterator<Item = Self::Item> + 'a>;
+impl<'a, 'b> IntoIterator for &'b Imports<'a>
+where
+    'a: 'b,
+{
+    type Item = &'b Import<'a>;
+    type IntoIter = Box<dyn Iterator<Item = Self::Item> + 'b>;

     fn into_iter(self) -> Self::IntoIter {
         match &self.items {
@@ -164,9 +167,12 @@ impl<'a> IntoIterator for &'a Imports<'a> {
     }
 }

-impl<'a> IntoIterator for &'a mut Imports<'a> {
-    type Item = &'a mut Import<'a>;
-    type IntoIter = Box<dyn Iterator<Item = Self::Item> + 'a>;
+impl<'a, 'b> IntoIterator for &'b mut Imports<'a>
+where
+    'a: 'b,
+{
+    type Item = &'b mut Import<'a>;
+    type IntoIter = Box<dyn Iterator<Item = Self::Item> + 'b>;

     fn into_iter(self) -> Self::IntoIter {
         match &mut self.items {
```

references-to-references, especially in `impl`s, are pretty subtle. These impls are effectively requiring an impossible situation which ends up manifesting as a lifetime error at a different use site which can be a bit confusing.

Overall though I think these impls are going to need to change since `&mut Import<'a>` won't exist for `Group{1,2}`. What I might recommend is something like `fn iter_item_sigs_mut(&mut self) -> impl Iterator<Item = &mut ItemSig<'a>>` (or something similar) since I think that's what most of the iterators need in this crate for working on the signatures.

bvisness:
As I mentioned on Zulip, I can't seem to use `wasm-tools parse` on the following file because I get an end-of-file error:

```
(module
  (import "foo" "bar" (func (param i32)))
  (import "foo" "baz" (func (param i32)))
  (import "foo"
    (item "beep" (func (param i64)))
    (item "boop" (func (param i64)))
  )
)
```
```
error: unexpected end-of-file (at offset 0x3f)
```

It seems to me like it's correctly parsing the text, but then blowing up when trying to do some kind of...validation run of the binary parser, except with no features enabled? Or maybe something else is wrong, since the compact encoding should trigger an "unexpected externtype" error.

bvisness:
Ok, I think it's all pretty much working, except that there are WIT linking issues that I don't know how to resolve. Would appreciate any guidance on that. I'll leave this in a draft state for a bit while I add spec tests to the proposal repo and use them to validate both this PR and our SpiderMonkey implementation.

alexcrichton:
Ah those are safe failures to ignore. That's from rustc nightly updates breaking things which I've resolved in https://github.com/bytecodealliance/wasm-tools/pull/2402 by pinning nightly temporarily

alexcrichton:
Code-wise everything looks good to me modulo TODO-docs 👍

bvisness:
No idea what those test failures are...I don't think I changed that file.

If you'd like, I can wait to add spec tests for compact imports before merging this? But I'm not sure what you do for proposal spec tests in this repo.

alexcrichton:
The failures here are related to Rust 1.92 being released, which should be fixed now with a rebase to pick up the changes from https://github.com/bytecodealliance/wasm-tools/pull/2402. 

Otherwise it's ok to defer the rigorous tests to the proposal testsuite, but it would be good to have at least a basic test in this repository for some of the textual forms and basic validation -- for example the binary format should be rejected if the feature is not enabled, and then accepted with the feature enabled.

bvisness:
Well, I rebased, I added some tests, they work locally, and now I'm getting newly inscrutable build errors.

alexcrichton:
```
> #3 ERROR: failed to authorize: failed to fetch oauth token: unexpected status from POST request to https://auth.docker.io/token: 500 Internal Server Error: an unknown error occurred during the request
```

just network errors, I'll retry

Issue #2399, opened 2025-12-09
`wasm_wave`: Make crate `no_std` compatible


wilfreddenton:
In a project I'm working on we are using `wasm_wave` within modules. Ideally we can make these modules compile with `no_std` but the `wasm_wave` dep is currently preventing that. For my use-case I am disabling the default `wit` feature which adds the `wit-parser` dependency so that crate would **not** need to become `no_std`.

Very similar situation to this issues: #3 

I'm happy to do this work and can follow the patterns in this PR: #1493 

But I want to check with the maintainers first that this is an acceptable change and if they have opinions on how it should be done i.e.

Should a non-default `no_std` feature or a default `std` feature that can be disabled be created? Or maybe an `alloc` feature instead?

What Map object should be used instead of `std::collections::HashMap`? I noticed that `indexmap` is a dep but it has the `std` feature enabled.

alexcrichton:
@lann you might have more thoughts on this than I.

From my perspective I'm not familiar enough with the crate to know off-hand whether the same strategy for `wasmparser` would work out for `wasm-wave`. Much of the time `no_std` crates do indeed use the `alloc` crate, and for maps they can be transitioned to `BTreeMap` as well in some situations too.

lann:
I'd have to re-familiarize myself with the code to be sure but I don't recall any reason why this wouldn't work with `alloc` / `BTreeMap`. Looking at the one use of `indexmap` I think it could be removed entirely with a slightly different implementation.

wilfreddenton:
Made a first draft of this at #2401. Since most of the crate doesn't really need `std` I'm just pulling everything from `alloc` and `core` and then anything that depends on `wit-parser` is behind the `wit` flag which, when enabled, forces on the new `std` flag.
Let me know what you think. Thank you!

wilfreddenton:
Thanks all!

PR #2401, opened 2025-12-10
Make `wasm-wave` support `no_std`

wilfreddenton:
- Imports `String`, `Vec`, etc. from `core` and `alloc` instead of `std`.
- Removes `indexmap` dep in favor of `alloc::collections::BTreeMap`
- `std::collections::HashMap` is replaced with `alloc::collections::BTreeMap`
- `ui` test is conditioned on `wit` feature
- added `std` feature and included it in `default` list

Crate tests pass for me locally for both:
```
cargo test --no-default-features
```
and
```
cargo test
```

#2399

### DIFF
--- a/crates/wasm-wave/Cargo.toml
+++ b/crates/wasm-wave/Cargo.toml
@@ -21,7 +21,7 @@ wit = ["dep:wit-parser", "std"]
 
 [dependencies]
 logos = { version = "0.14.2", default-features = false, features = ["export_derive", "forbid_unsafe"] }
-thiserror = { version = "2.0.12", default-features = false }
+thiserror = { workspace = true }
 wit-parser = { workspace = true, optional = true }
 
 [dev-dependencies]
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -287,6 +287,7 @@ jobs:
       - run: cargo check --no-default-features -p wasm-metadata --features serde
       - run: cargo check --no-default-features -p wasm-metadata --features oci
       - run: cargo check --no-default-features -p wasm-wave
+      - run: cargo check --no-default-features -p wasm-wave --target x86_64-unknown-none
       - run: cargo check --no-default-features -p wasm-wave --features std
       - run: cargo check --no-default-features -p wasm-wave --features wit
       - run: cargo check --no-default-features -p wasm-wave --features std,wit
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -139,7 +139,7 @@ spdx = "0.10.1"
 termcolor = "1.2.0"
 url = "2.0.0"
 wasmtime = { version = "34.0.1", default-features = false, features = ['cranelift', 'component-model', 'runtime', 'gc-drc'] }
-thiserror = "2.0.12"
+thiserror = { version = "2.0.12", default-features = false }
 tempfile = "3.2.0"
 wit-bindgen = { git = 'https://github.com/bytecodealliance/wit-bindgen', default-features = false }
 


crates/wasm-wave/src/wasm/val.rs
```diff
@@ -326,4 +324,5 @@ macro_rules! unwrap_val {
         }
     };
 }
+use alloc::{borrow::Cow, boxed::Box};
```
alexcrichton: Can this leave type imports at the top of the file?

crates/wasm-wave/src/wasm/ty.rs
```diff
@@ -136,4 +134,5 @@ macro_rules! maybe_unwrap_type {
         }
     };
 }
+use alloc::{borrow::Cow, boxed::Box};
```
alexcrichton: Similar to below, can type imports stay at the top of the file?

crates/wasm-wave/src/lib.rs
```diff
@@ -4,6 +4,9 @@
 //!
 //! For more information, see the [README](https://github.com/bytecodealliance/wasm-tools/tree/main/crates/wasm-wave#readme).
 #![deny(missing_docs)]
+#![cfg_attr(not(feature = "std"), no_std)]
```
alexcrichton: Personally I find its best for crates wit` no_std` support to unconditionally be `#![no_std]`, and then when `std` is enabled there's `#[cfg(feature = "std")] extern crate std;`

alexcrichton:
Oh, also, mind adding checks to CI that this builds with various combinations of features? We do similar for other crates in `.github/workflows/main.yml`

crates/wasm-wave/Cargo.toml
```diff
@@ -15,13 +15,13 @@ readme = "README.md"
...
+default = ["std", "wit"]
+std = []
+wit = ["dep:wit-parser", "std"]
 
 [dependencies]
-indexmap = { workspace = true, features = ["std"] }
-logos = { version = "0.14.2", features = ["forbid_unsafe"] }
-thiserror = { workspace = true }
+logos = { version = "0.14.2", default-features = false, features = ["export_derive", "forbid_unsafe"] }
+thiserror = { version = "2.0.12", default-features = false }
```
wilfreddenton: To test whether or not the crate is really `no_std`  I compiled using:

```
cargo build --target thumbv7m-none-eabi --no-default-features
````

This informed me that `logos` was not `no_std` by default so I made an adjustment to the features there. It also showed that the `default-features = false` on `thiserror` was being ignored because that doesn't exist in the workspace level dep so I'm setting the version here. I didn't want to make any changes that would affect the workspace but if you think setting `default-features = false` there makes sense I can do that.

With these dep changes the above build command works.

Comment (alexcrichton) [reply crates/wasm-wave/Cargo.toml]: To avoid duplicating the `thiserror` version, could you add `default-features = false` at the workspace level? If other crates need it they can add `features = ['default']` to turn that back on (and seems best to minimize feature deps across the workspace)

We use the `x86_64-unknown-none` target in CI for "this doesn't actually  have `std`" and there's some `wasmparser` configurations that build for that for example. Could you also udpate the CI bits to add `--target x86_64-unknown-none` to the checks for `wasm-wave` where applicable?

**Comment by wilfreddenton** [comment]
Addressing the remaining CI failures:

One of the jobs is failing on:
```
 --> crates/wasm-wave/src/parser.rs:3:12
  |
3 | use core::{error::Error, fmt::Display};
  |            ^^^^^^^^^^^^
  |
  = note: see issue #103765 <https://github.com/rust-lang/rust/issues/103765> for more information
```
`core::error::Error` was stabilized in [`1.81.0`](https://blog.rust-lang.org/2024/09/05/Rust-1.81.0/#core-error-error) but the job is using `1.76.0`. Can this version be updated?

The other job is failing on the `wit-dylib` `tests/all.rs`. After installing the `wasi-sdk` and the `wasmtime` cli I am able to run
```
cargo test --locked --all
```
successfully locally so I am unable to reproduce this.

**Comment by alexcrichton** [comment]
Ah the msrv policy of this workspace is [here](https://github.com/bytecodealliance/wasm-tools/blob/6325cc4a73eca6b9bd9c71d4a2c4e9b4a7c8b366/Cargo.toml#L90-L109), and it looks like 1.81 is now possible too so I've sent https://github.com/bytecodealliance/wasm-tools/pull/2402 to update the msrv.

> The other job is failing on the `wit-dylib`

I'll look into that, I agree it's probably unrelated to this PR

**Comment by alexcrichton** [comment]
Ok https://github.com/bytecodealliance/wasm-tools/pull/2402 should address the CI issues here. Once that is merged if you can rebase on top of that I think that'll have everything "just work"

File: crates/wasm-wave/src/parser.rs [hunk 5]
```diff
@@ -176,7 +182,7 @@ impl<'source> Parser<'source> {
     }
 
     fn finish_flags(&mut self, start: usize) -> Result<Node, ParserError> {
-        let mut flags = IndexMap::with_capacity(1);
+        let mut flags = BTreeMap::new();
```
lann, line 185: This is a (small) functional change, switching the ordering of serialized flags from definition-ordered to lexicographical. I don't see this behavior documented anywhere so its probably fine. Clearly it wasn't important enough to me to assert it in a test case. 😅

alexcrichton (approved):

PR #2411, opened 2026-01-05
disallow kebab fragment beginning with digit

dicej:
Per the `word` and `acronym` grammar rules in
https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#import-and-export-definitions, a kebab name must not contain a fragment beginning with a digit.  This fixes [this
test](https://github.com/WebAssembly/component-model/blob/b7fc86e31d71796f107aef276f0944ec6df70dba/test/wasm-tools/naming.wast#L25C1-L30C2) in the spec repo.


alexcrichton:
This was changed here in this repo in https://github.com/bytecodealliance/wasm-tools/pull/2355 and provisionally in the component model in https://github.com/WebAssembly/component-model/issues/345, so I think that this might be a case of an upstream test that needs changing

dicej:
Thanks for pointing that out, @alexcrichton .  @lukewagner is going to update the spec when he gets a chance.

PR #2421, opened 2026-01-20
rename fixed size list to fixed-length lists

cpetig:
This came up while designing the patches in wasmtime and wit-bindgen, the feature is not named in accordance with the [spec](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#gated-features) and if we ever would like to rename it, it should be now or never. Because right now the feature is in little use but will be used more widely soon.

I recall talking with @alexcrichton about this somewhere before (couldn't find it) and we decided that it wasn't worth the extra effort. But with this patch available the scale might tip - it feels less weird to have it named fixed-length list everywhere. 

This will require matching patches in wit-bindgen, wac, and wasmtime once these upgrade to this wasm-tools version.


cpetig:
PS: I am not able to add this to https://github.com/orgs/bytecodealliance/projects/19 but it belongs there.

alexcrichton:
CI failure looks unrelated to this PR, I'll investigate soon

alexcrichton (approved):
Thanks! With https://github.com/bytecodealliance/wasm-tools/pull/2422 CI should be fixed now, mind rebasing?

PR #2433, opened 2026-02-02
Update wasi-sdk in CI

alexcrichton:
Keeping it up-to-date and testing it


ricochet:
We also have https://github.com/bytecodealliance/setup-wasi-sdk-action

ricochet:
There's a bug with the install script: https://github.com/bytecodealliance/setup-wasi-sdk-action/issues/12

alexcrichton:
I'm going to skip this update here as `__wasi_init_tp` is a new symbol imported by `libdl.so` which while it's not a bug it's also not necessary, so I don't think it's worth the update.

PR #2453, opened 2026-02-27
Add `(implements "...")` component name support

ricochet:
Add parsing, validation, and uniqueness rules for the new
`(implements "...")` extern name form from the component
model `implements` proposal. An implements name labels an instance
import/export that implements a named interface.

See implements feature:
https://github.com/WebAssembly/component-model/pull/613


crates/wasmparser/src/validator/names.rs
```diff
@@ -962,6 +1052,19 @@ mod tests {
...
+        assert!(parse_kebab_name("[implements=<a:b/c@1.0.0>]name").is_some());
+        assert!(parse_kebab_name("[implements=<ns:pkg/iface>]my-label").is_some());
+        // invalid: not a valid interface name (no colon/slash)
+        assert!(parse_kebab_name("[implements=<not-valid>]name").is_none());
+        // invalid: empty interface name
+        assert!(parse_kebab_name("[implements=<>]name").is_none());
+        // invalid: missing label
+        assert!(parse_kebab_name("[implements=<a:b/c>]").is_none());
+        // invalid: label not kebab
+        assert!(parse_kebab_name("[implements=<a:b/c>]NOT_KEBAB").is_none());
```
alexcrichton: In addition to unit tests here, could you add some unit tests as `*.wast` files in `tests/cli/component-model/*.wast`? Maybe something like `tests/cli/component-model/implements.wast` or similar. It's ok to move these tests outright there where necessary, but by having a `*.wast` test we can help other implementations' tests as well in the future by eventually sharing wast tests

crates/wit-smith/src/generate.rs
alexcrichton, line 1: Thanks for adding wit-smith support! I always forget to do that...

Can you run the fuzzer locally for a few minutes to make sure nothing crops up? You can do that with:

```shell
$ FUZZER=roundtrip_wit cargo +nightly fuzz run -s none run
```

crates/wit-parser/src/lib.rs
```diff
@@ -614,6 +614,18 @@ pub enum WorldItem {
...
+        /// identifies the named interface `I` that this import/export
+        /// implements.
+        #[cfg_attr(
+            feature = "serde",
+            serde(
+                skip_serializing_if = "Option::is_none",
+                serialize_with = "serialize_optional_id"
+            )
+        )]
+        implements: Option<InterfaceId>,
```
alexcrichton: AST-wise I think I might bikeshed this a bit. I find this a bit confusing where there's two `InterfaceId`s listed here, one above with `id` and one for `implements`, and for `import foo: bar;` I'd assume `implements` would point to `bar` but it's not clear to me what `id` above would point to.

Additionally foundationally I'd ideally like to keep things such that `name_world_key` is sufficient for generating the import name of a component/instance import.

Given all that, WDYT about augmenting `WorldKey` instead of augmenting `WorldItem`? What I'm imagining is:

* Add a new `WorldKey::Implements { name: String, interface: InterfaceId }` - this'd corresponds to `import foo: bar` with `name` being `foo` and `interface` pointing to `bar`
* Reuse `WorldItem::Interface` for the `WorldItem`, in this case with `id` also pointing to `bar`

To me that feels like it'll mesh better with the existing generators and such, but want to double-check with how you're thinking about this.

crates/wit-parser/src/decoding.rs
```diff
@@ -744,6 +731,55 @@ impl WitPackageDecoder<'_> {
...
     }
 
+    /// Decodes a component instance import/export name into a
+    /// `(WorldKey, WorldItem)` pair.
+    ///
+    /// Handles three name forms:
+    /// - `[implements=<I>]label` — named import/export implementing interface I
+    /// - `ns:pkg/iface` — qualified interface name, keyed by `InterfaceId`
+    /// - `plain-name` — unqualified name for an inline or local interface
+    fn decode_world_instance<'a>(
```
alexcrichton, line 726: Thanks for doing the refactor here to split this out!

crates/wit-parser/src/decoding.rs
```diff
@@ -18,6 +18,19 @@ use wasmparser::{
     types::Types,
 };
 
+/// Parses an `[implements=<interface_name>]label` name, returning
+/// the interface name and label if the name matches this pattern.
+fn parse_implements_name(name: &str) -> Option<(&str, &str)> {
+    let rest = name.strip_prefix("[implements=<")?;
```
alexcrichton: I'm not sure if it'd make sense, but one option here would be to use `ComponentName` from `wasmparser` to do the parsing here (I think that's what it's called) which could alleviate the string processing from here

alexcrichton (commented):
Thanks again for this! I've left some comments below, the most consequential one being about the AST structure of `wit-parser` itself. I'd like to still look more closely at some details once that's settled, but also wanted to give some initial feedback. I'll need to catch up on the wasmtime/wit-bindgen discussions too

alexcrichton:
Oh, another point -- I forgot this in my first pass so wanted to leave a note here as well -- can you add a new feature to `crates/wasmparser/src/features.rs`? We'll want to gate usage of this new syntax by default (aka require an opt-in in the Wasmtime CLI) while it's being implemented in various places.

alexcrichton:
@ricochet do you mind if I pick this up this week and work on getting this landed?

alexcrichton:
Ok I've performed major updates to this PR. Everything is now up-to-date, the history's a bit messy, but hey it's all there. This PR is coupled with updates to https://github.com/WebAssembly/component-model/pull/613 as well. Major changes are:

* The binary format is completely different. No more `[implements=...]` but instead it's `(implement "...")` as a separate binary form. This is done for future extensibility for `@external-id("...")`.
* The AST representation of everything subsequently changed quite a bit. For example type reflection now yields a `ComponentItem` instead of a `ComponentEntity` enum directly. The item form contains `implements: Option<String>` to carry this annotation.
* The AST for `wit-parser` ended up actually reverting to what it was before this PR. That'll make consumer's lives much easier, hopefully. Notably the way this is encoded is that a `WorldKey::Name`'d thing points at a `WorldItem::Interface` value, and the interface has a name as opposed to not having a name.
* All sorts of plumbing in all sorts of locations have been updated to account for this change. For example decoding/encoding the worlds from wasm works, wit-component support is now implemented for all forms, fuzzing works, etc. This, as always, took a few iterations and some more primitives to figure out the right approach, but I feel the end state is pretty good.
* The `generate_nominal_type_ids` method is becoming more load bearing. It's updated for `(implements "...")` directives to duplicate the interface in question.

alexcrichton:
My plan is to perform a final once-over on review tomorrow, and then I'll ask for another sign-off from someone since so much of this is now self-authored by me. I've verified the fuzzer runs for ~1h locally finding no issues so I'm reasonably confident in the coverage and feature. After a bit of time to bikeshed the binary format details upstream in the spec I plan on landing and then working on the Wasmtime side of things.

alexcrichton:
Ok I've done my own once-over on everything now and I feel this is in a good spot. @dicej would you be willing to look over this as well to confirm you don't find anything awry? I realize this is a pretty big PR, though, so no worries if you're too busy

crates/wit-parser/src/resolve/mod.rs
```diff
@@ -2248,56 +2269,62 @@ impl Resolve {
...
             }
-            // If this is an import and it's already in the `required_imports`
-            // set then we can skip it as we've already visited this interface.
+            let (id, stability) = match &item {
+                WorldItem::Interface { id, stability, .. } => (*id, stability),
+                _ => unreachable!(),
+            };
+            // If this is an import and it's already in the `imports` set then
+            // we can skip it as we've already visited this interface.
+            // if !add_export && imports.contains_key(&key) {
```
dicej: Can this line be removed?

crates/wit-parser/src/resolve/mod.rs
```diff
@@ -2838,92 +2865,293 @@ impl Resolve {
...
+        // types/interfaces/etc within `maps`. We don't want to persist
+        // everything because if an interface is cloned twice then everything
+        // will get overwritten/corrupted within the map. In theory what we want
+        // is for `maps` to track, for any one interface, just the transitive
+        // set of dependencies for that interface and how they've been cloned.
+        // What's implemented here is an approximation of this that should work
+        // for now.
+        //
+        // Notably `to_rewrite` is an ordered list keyed by `InterfaceId`. This
+        // means taht if we walk `to_rewrite` in order we're walking this in
```
dicej:
```suggestion
        // means that if we walk `to_rewrite` in order we're walking this in
```

dicej (approved):

PR #2482, opened 2026-04-07
Structured resolver errors

PhoebeSzmucer:
Part of https://github.com/bytecodealliance/wasm-tools/pull/2460 (part 1 was [here](https://github.com/bytecodealliance/wasm-tools/pull/2465)).

This PR replaces most of `anyhow` usages in the resolver with structured `ParserError` type, which carries structured error information (with spans).


crates/wit-parser/src/ast/error.rs
```diff
@@ -1,11 +1,15 @@
+//! Error types for WIT parsing.
```
PhoebeSzmucer, line 1: Added some missing docs

alexcrichton:
Happy to take a look when you feel this is ready, but I'm interpreting the draft state as "come back later" for now -- just wanted to note this

PhoebeSzmucer:
Thanks @alexcrichton. No review needed just yet - I'm still cleaning up some final things and do one more self review, and I'll mark it as ready soon.

crates/wit-parser/src/resolve/mod.rs
```diff
@@ -230,67 +236,65 @@ impl Resolve {
         Resolve::default()
     }
 
-    /// Parse WIT packages from the input `path`.
```
PhoebeSzmucer, line 233: This doc was very out of date.

crates/wit-smith/src/lib.rs
PhoebeSzmucer, line 37: Both of these are `ResolveErrorKind::Semantic`. We could introduce a dedicated type for them to prevent string matching, either now or in a follow up PR.

PhoebeSzmucer: Although `wit-smith` is just a fuzzer so idk if it's worth it.

crates/wit-parser/src/resolve/mod.rs
```diff
@@ -230,67 +236,65 @@ impl Resolve {
...
-    /// This method returns a `PackageId` and additionally a `PackageSourceMap`.
-    /// The `PackageId` represent the main package that was parsed. For example if a single WIT
-    /// file was specified  this will be the main package found in the file. For a directory this
-    /// will be all the main package in the directory itself. The `PackageId` value is useful
-    /// to pass to [`Resolve::select_world`] to take a user-specified world in a
-    /// conventional fashion and select which to use for bindings generation.
+    /// Merge `main` and `deps` into this [`Resolve`], topologically sorting
+    /// them internally. Returns the [`PackageId`] of `main` and a
+    /// [`PackageSources`] covering all groups.
     fn sort_unresolved_packages(
```
PhoebeSzmucer, line 242: Note - this function was restructured to do upfront merging so that if a `DuplicatePackage` or `PackageCycle` error happens, we already have valid spans in the error

PhoebeSzmucer: I added some tests for this

crates/wit-parser/src/resolve/mod.rs
```diff
@@ -385,24 +379,47 @@ package {name} is defined in two different locations:\n\
...
+    /// [`Resolve`] in a single call, topologically sorting them internally.
+    ///
+    /// This is useful when you have a package and its local dependencies
+    /// available as in-memory [`UnresolvedPackageGroup`]s and want dependency
+    /// ordering and cycle detection handled automatically.
+    ///
+    /// The returned [`PackageId`] corresponds to `main`.
+    ///
+    /// On error, `self` may be partially modified and should not be reused.
+    pub fn push_groups(
```
PhoebeSzmucer, line 401: This function is something I will want to use in the WIT LSP. I was going to add it in a separate PR, but having this function now makes it much easier to test the behavior of `sort_unresolved_packages`.

crates/wit-parser/src/resolve/mod.rs
```diff
@@ -3087,38 +3107,57 @@ fn rename(original_name: &str, include_name: &IncludeName) -> Option<String> {
...
-        let prev = resolve.package_names.insert(unresolved.name.clone(), pkgid);
-        if let Some(prev) = prev {
-            resolve.package_names.insert(unresolved.name.clone(), prev);
-            bail!(
-                "attempting to re-add package `{}` when it's already present in this `Resolve`",
-                unresolved.name,
-            );
-        }
-
+        assert!(
```
PhoebeSzmucer, line 3155: Note we went from `bail!` to `assert!`, but this is a programmer error (pushing the same package twice), not a WIT source error. So I think panic is more appropriate here.

PhoebeSzmucer:
@alexcrichton this is ready for a review now. I left some inline comments on the PR

crates/wit-parser/src/resolve/fs.rs
```diff
@@ -131,7 +131,9 @@ impl Resolve {
             .parse_deps_dir(&deps)
             .with_context(|| format!("failed to parse dependency directory: {}", deps.display()))?;
 
-        let (pkg_id, inner) = self.sort_unresolved_packages(top_pkg, deps)?;
+        let sort_result = self.sort_unresolved_packages(top_pkg, deps);
+        let (pkg_id, inner) =
+            sort_result.map_err(|e| anyhow::anyhow!("{}", e.highlight(&self.source_map)))?;
```
alexcrichton, line 136: I feel like I might have asked this already on the previous PR of yours, so I apologize if so, but here this feels kind of bad to stringify the error and forget about the original error. The main downside here is that the returned value of `anyhow::Result` doesn't have the ability to `downcast` the error to a concrete error type and inspect it.

I forget, though, was this something where the `anyhow::anyhow!("{}", ...)` is temporary? Or is this intended to be more permanent? For example should `ResolveError` have a "from I/O error" variant and this function returns a `ResolveError`?

---

Or, alternatively, one thing I thought of is that it's useful to have a decoupled error from the `SourceMap` where the returned error has all the info it needs to be rendered without pairing to something else, which this call to `highlight` is doing. I was thinking, though, that `highlight` would mutate the error-in-place (e.g. filling in some fields) as oppose to returning just a string

crates/wit-smith/src/lib.rs
alexcrichton: Yeah this is a great use case for dedicated error kinds, but totally fine to do in a follow-up

crates/wit-parser/src/resolve/mod.rs
```diff
@@ -368,14 +362,14 @@ package {name} is defined in two different locations:\n\
...
-    ) -> Result<PackageId> {
+    ) -> ResolveResult<PackageId> {
         unresolved.adjust_spans(span_offset);
         let ret = Remap::default().append(self, unresolved);
         if ret.is_ok() {
             #[cfg(debug_assertions)]
             self.assert_valid();
         }
-        self.source_map.rewrite_error(|| ret)
+        ret
```
alexcrichton, line 372: Question on the removal here, and I feel like again I may have already asked this and/or you already answered  so sorry if so, but this seems like a bit of a regression insofar as callers need to rewrite the error as opposed to that happening automatically internally. That's sort of where above I was thinking that the errors would be rewritten in-place so these calls to `rewrite_error` would stick around.

crates/wit-parser/src/resolve/mod.rs
```diff
@@ -3166,20 +3205,7 @@ impl Remap {
         // referenced along the way.
         for (id, mut iface) in unresolved.interfaces.into_iter().skip(foreign_interfaces) {
             let span = iface.span;
-            if !resolve
-                .include_stability(&iface.stability, &pkgid, span)
-                .with_context(|| {
```
alexcrichton, line 3171: The loss of the context here, and a few below, is a bit unfortunate because it means that the error messages have less contextual information than they did before. Would it be possible to retain a `context`-like method on `ResolveError`  which acts similar-ish to `anyhow::Context`?

crates/wit-parser/src/resolve/fs.rs
PhoebeSzmucer: > I feel like I might have asked this already on the previous PR of yours, so I apologize if so, but here this feels kind of bad to stringify the error and forget about the original error. The main downside here is that the returned value of anyhow::Result doesn't have the ability to downcast the error to a concrete error type and inspect it.

You're right, I was a bit sloppy here. I think I can fix this by using `anyhow::Error::from` instead of `anyhow!`, something like `anyhow::Error::from(e).context(msg)`. I'll play around with it a bit.

> I forget, though, was this something where the anyhow::anyhow!("{}", ...) is temporary? Or is this intended to be more permanent? For example should ResolveError have a "from I/O error" variant and this function returns a ResolveError?

In this place it's not temporary. I personally don't think we should introduce any I/O error variants to `ResolveError` - the resolver's main job is not to interact with the filesystem, but to deal with already loaded files. Adding I/O variants would mix the concerns too much. I/O matters only at the boundary of the resolver.

> Or, alternatively, one thing I thought of is that it's useful to have a decoupled error from the SourceMap where the returned error has all the info it needs to be rendered without pairing to something else, which this call to highlight is doing. I was thinking, though, that highlight would mutate the error-in-place (e.g. filling in some fields) as oppose to returning just a string

I generally prefer the errors to stay very lightweight - ie only contain the span info and the necessary metadata. Adding rendered strings to errors means that we lose a clean separation between what went wrong (spans/metadata), and how to display it (source map rendering). And if we mutate the error, the error now becomes dependent on when and whether you call that mutation (which increases the chance of bugs).

I also plan to revisit the highlight naming and the overall error presentation in a follow-up - now that errors are structured, it should be much easier to reason about where and how rendering happens, and to improve individual error messages.

crates/wit-smith/src/lib.rs
PhoebeSzmucer: I'll just do it now then, it's easy enough

crates/wit-parser/src/resolve/mod.rs
PhoebeSzmucer: I think some of that contet was pretty redundant, and leaked details about the internal workings of the resolver. Example https://github.com/bytecodealliance/wasm-tools/pull/2482/changes#diff-e50c5a3992e0125ef03e2863997fb3e17e33acd1f626f5ed8f5e9a6039880322 - I think the new version is better, but let me know if you think otherwise.

In this case I might have dropped valuable context though - let me generate test cases/snapshots for each one of these cases, and see how they look now vs before. If the errors don't have enough info, I'm more than happy to add extra metadata to the error variants.

I can also add a context method, but I think it's generally better to have typed/structured metadata rather than accumulating strings.

PhoebeSzmucer: I definitely don't want to introduce any regression to how errors are presented (if anything, I want to make them better). Let's make sure we're happy with snapshot changes before this PR merges.

crates/wit-parser/src/resolve/mod.rs
PhoebeSzmucer: There is a broader design question here - whether erros should rewrite themselves, vs rendering being explicit. I'm intentionally trying to keep the errors as pure data (spans + metadata), since I think that's the pattern that most compilers (including rustc) use.

For example rustc has an [Emitter](https://github.com/rust-lang/rust/blob/main/compiler/rustc_errors/src/emitter.rs#L53) which owns the source map, renders the diagnostics, and the [diagnostics](https://github.com/rust-lang/rust/blob/5f36a7f8072e835d73d9e1f7ab8a2134ac7352dc/compiler/rustc_errors/src/diagnostic.rs#L227) themselves are just metadata (without source code).

I think this pattern is the best, because:
- The errors stay light, and tools that don't care about rendered code don't have to pay the cost of it
- We can have emitters for different targets
- The data and presentation layer are clearly separated
- Errors can stay immutable

In practice, rendering only needs to happen at a few boundary points where `ResolveError` crosses into user-facing output. The source map is always available there, so I don't think it will be burdensome enough to warrant prerendering the errors (which has its own downsides).

What do you think?

crates/wit-parser/src/resolve/fs.rs
alexcrichton: Ok this all sounds good, and yeah using `anyhow::Error::from(e).context(msg)` instead of `anyhow::anyhow!("{}", ...)` would indeed resolve my concerns here

crates/wit-parser/src/resolve/mod.rs
alexcrichton: That sounds like solid reasoning to me yeah so overall sounds good. One difference with rustc/others though is that this crate is embedded in a number of different contexts, for example the `wasm-tools` CLI, the `wit-bindgen` CLI, the `wit_bindgen::generate!` proc macro, etc. In all of these contexts the ultimate desire right now is to have a rendered error at some point. I bring this up because ideally the logic to duplicate across these embeddings is relatively minor rather than relatively invasive. Many of them just render an `anyhow::Error` as-is to capture the full chain of context currently.

For keeping errors immutable and such, perhaps the `wit-parser` crate could have a convenience function for rendering an `anyhow::Error`? For example keeping the `highlight` method but also doing something that renders all the layers of the `anyhow::Error` too.

crates/wit-parser/src/resolve/mod.rs
alexcrichton: You've got a good point, and you've also helped me remember that many of these `.context(` methods were related to a lack of span information. Without span information the context errors were the only indicator of what function to look at, which was critical at the time. Nowadays though we've got spans everywhere so the context is no longer critical, and as you point out it can also be redundant.

I'll do a once-over again of all the error message changes in this PR (in the tests) during the final review and double-check it all looks good, but I suspect it is. Thanks!

crates/wit-smith/src/lib.rs
PhoebeSzmucer: Done

crates/wit-parser/src/resolve/fs.rs
PhoebeSzmucer: This should be done now. I also added tests for downcasting in `crates/wit-parser/tests/downcast.rs`, to make sure we don't accidentally regress this.

There is a bit of tension between being able to downcast, and being able to control the order of the snippets/messages/context in the anyhow display. In the end I introduced a `render_anyhow_error` helper that renders the anyhow error in the way that we want, and updated the three places where we display those errors.

We are still mapping `e` to `anoyhow!` + `render_anyhow_error` in a few places, but hopefully that's ok because that's in the binary crate where downcasting shouldn't matter. It might be awkward to push the stringification a bit higher because we don't have the source map there.

Let me know if this sounds good.

crates/wit-parser/src/resolve/mod.rs
PhoebeSzmucer: I added `render_anyhow_error`

crates/wit-parser/tests/ui/parse-fail/bad-gate-record-field.wit.result
```diff
@@ -1,5 +1,5 @@
-failed to update field `field-b`: found a reference to a type which is excluded due to its feature not being activated
-     --> tests/ui/parse-fail/bad-gate-record-field.wit:7:10
+found a reference to a type which is excluded due to its feature not being activated
+     --> tests/ui/parse-fail/bad-gate-record-field.wit:9:5
       |
-    7 |   record r {
-      |          ^
\ No newline at end of file
+    9 |     field-b: gated,
```
PhoebeSzmucer, line 4: The message doesn't contain the field name anymore, but I changed the span to point to the field itself, so no context should be lost.

crates/wit-parser/src/resolve/mod.rs
PhoebeSzmucer: I added some more snapshot tests in https://github.com/bytecodealliance/wasm-tools/pull/2510 - they should hopefully cover all the changes that the current PR introduces.

Most of the snapshot changes in this PR look good IMO - there might be a couple of places where the error message is slightly less specific, but I think combined with the snippet it shouldn't cause any confusion.

Let me know please if any snapshot changes look worse to you, and I can try improving them.

crates/wit-parser/tests/ui/parse-fail/bad-since-on-include.wit.result
```diff
@@ -1,4 +1,4 @@
-failed to process feature gate for included world [a] in package [test:invalid@0.1.0]: feature gate cannot reference unreleased version 0.2.0 of package [test:invalid@0.1.0] (current version 0.1.0)
+feature gate cannot reference unreleased version 0.2.0 of package [test:invalid@0.1.0] (current version 0.1.0)
```
PhoebeSzmucer, line 1: As a follow up I think we should change the snippet to point to the bad version.

PhoebeSzmucer: And this error could become much nicer potentially - I find the current phrasing a bit vague.

alexcrichton (approved):
Everything here looks great to me, thanks again @PhoebeSzmucer! I think it's fine for anything further to be in follow-ups, so let's go ahead and land this.

PR #2485, opened 2026-04-10
Update spec test suite submodule

alexcrichton:
* Add support for `assert_{invalid,malformed}_custom` directives
* Shuffle around how `@custom` is printed to improve round-trip-ability
* Disallow multiple `@metadata.branch_hint` annotations on one instruction.
* Skip a custom-page-sizes test for now until the test upstream is updated.
* Adjust the `name` section emitted during text-to-binary to happen after user-specified `@custom` sections to make round-tripping work.


tests/cli/branch-hinting/branch-hinting-simple.wast
```diff
@@ -70,28 +70,36 @@
...
   (func
-    (@metadata.code.branch_hint "\00")
-    (@metadata.code.branch_hint "\00")
     (@metadata.code.branch_hint "\00")
     i32.const 0
     (@metadata.code.branch_hint "\00")
-    (@metadata.code.branch_hint "\00")
-    (@metadata.code.branch_hint "\00")
     if
     (@metadata.code.branch_hint "\00")
```
keithw, line 101: Re: the `no-test-folded`on this file. The folding printer is rejecting this because of the branch hint on an `end` operator. Branch hints seem to be only allowed on `br_if` and `if` -- we could avoid having to make this test `no-test-folded` if we either:

- remove this particular branch hint (the one just before the `end`) from the test, or
- make it an `assert_malformed_custom`, and enforce the "branch hints are only allowed on br_if or if` rule in the parser

pchickey (approved):
Needs the merge with main to get rid of the typos CI pass

tests/cli/branch-hinting/branch-hinting-simple.wast
alexcrichton: I end up sort of doing this for the `tests/cli/folding/branch-hinting-simple.wast` file where I deleted some of the annotations to get the test to pass, so this test here is asserting that if the hints are present it at least round-trips without folding.

Otherwise though for `assert_malformed_custom`, it's a bit tricky. Technically this needs to be `assert_invalid_custom` because the branch hint section isn't malformed with hints on weird instructions, it's only invalid because it's not allowed on those instructions. Right now wasm-tools doesn't have any validators for custom sections, though, so the `assert_invalid_custom` directive is basically ignored.

PR #2486, opened 2026-04-10
Revert "ci(style): add typos-cli to CI flow (#2468)"

alexcrichton:
This reverts commit d0f7d9890fcf451ed07c5ea7cd550f5023cb2781.

This is currently otherwise blocking CI in https://github.com/bytecodealliance/wasm-tools/pull/2485 for something unrelated. Personally I feel that CI-checking typos is a bit too onerous to place on CI and all contributors insofar as it's a "hard signal" (aka you cannot land this PR) for something that's pretty "soft" (typos rarely make code misunderstandable). I'd personally prefer to stick to the previous state of "occasionally we get typo PRs" and that way there's no need to figure out how to skip things in CI, exempt one-offs, etc.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -364,14 +364,6 @@ jobs:
       - run: rustup component add clippy
       - run: cargo clippy --workspace --all-targets --exclude dl --exclude component
 
-  typos:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v6
-      - uses: cargo-bins/cargo-binstall@v1.17.9
-      - run: cargo binstall typos-cli@1.44.0 --no-confirm
-      - run: typos
-
   verify-publish:
     if: github.repository_owner == 'bytecodealliance'
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


alexcrichton:
cc @mkatychev, you might have thoughts on this too

mkatychev:
@alexcrichton I don't mind at all taking typos-cli out of CI, in my (poor) defence the PR was a request ;). Though looking at https://github.com/bytecodealliance/wasm-tools/pull/2485, one could just exclude the [snapshots as a directory](https://github.com/crate-ci/typos/blob/master/docs/reference.md#configuration-keys):

```toml
# https://github.com/crate-ci/typos/blob/master/docs/reference.md
[workspace.metadata.typos.files.extend-exclude]
extend-exclude = ["tests/snapshots"]
```

Either way is fine. The impetus for it was mainly in the rust code and having the benefit of documenting certain terms & acronyms.

dicej (approved):

mkatychev (approved):

pchickey:
Thanks @mkatychev for contributing this, it seemed like it was worth a try but since we have run into issues I agree with Alex it should be backed out.

alexcrichton:
@mkatychev that makes sense yeah, but personally I'd prefer to avoid the experience of having contributors come to the repo, hit a failure, and have to go search around for the syntax of what file to modify to allow a certain typo. That's what I in theory needed to do here, and while I think it's ok to hold maintainers to a different standard it's still something I'd want to avoid to need for contributors.

PR #2487, opened 2026-04-14
Upgrade to hashbrown v0.17

cuviper:
... plus indexmap v2.14 to depend on the same.

I'm following up on behalf of the Rust compiler, https://github.com/rust-lang/rust/pull/155256#issuecomment-4240494425, trying to eventually sync up the dependencies.


Cargo.toml
```diff
@@ -104,7 +104,7 @@ version = "0.246.2"
 # message to wasm-tools maintainers to discuss. In some cases it's possible to
 # add version detection to build scripts but in other cases this may not be
 # reasonable to expect.
-rust-version = "1.82.0"
+rust-version = "1.85.0"
```
cuviper, line 107: We're not *quite* to N-10 yet, but 1.95 releases this Thursday. I'm in no rush if you want to wait for that.

alexcrichton (approved):
Looks good to me, thank you!

I'm happy to make a release for this now if you need, but if you're ok with "eventual consistency" I'll also let this ride the next on-demand release (probably no more than 1-2 months)

cuviper:
I think it's ok to let this ride -- there are other dependency chains I have to chase down as well.

PR #2491, opened 2026-04-15
wasm-mutate-stats: release worklist mutex before file I/O

### LINKED ISSUE #2490: wasm-mutate-stats writer thread holds queue mutex during file I/O
`wasm-mutate-stats` holds the `to_write` mutex guard across file I/O in the writer thread.

The affected code is in `crates/wasm-mutate-stats/src/bin/wasm-mutate-stats.rs`:

```rust
let encoder = std::thread::spawn(move || -> anyhow::Result<()> {
    let mut counter = 0;

    while !finish_writing_wrap_clone.load(Relaxed) {
        // pop from worklist
        if let Some(wasm) = to_write_clone.lock().unwrap().pop() {  // !! Issue 1 
            let filename = artifact_folder_cp.join(format!("mutated.{counter}.wasm"));
            std::fs::write(filename, &wasm).context("Failed to write mutated wasm")?;
            counter += 1;
        }
    }
    eprintln!("Writing down pending mutated binaries!");
    // Then write pending wasms
    while let Some(wasm) = to_write_clone.lock().unwrap().pop() {  // !! Issue 2
        let filename = artifact_folder_cp.join(format!("mutated.{counter}.wasm"));

        std::fs::write(filename, &wasm).context("Failed to write mutated wasm")?;
        counter += 1;
    }

    Ok(())
});
```

Because the temporary `MutexGuard` created in the `if let` / `while let` scrutinee lives for the whole statement, the lock remains held while std::fs::write executes, and is only released at the closing brace.

## Impact

This code also keeps the queue locked during each file write, which can delay shutdown and reduce throughput.

It is more about a performance and concurrency issue rather than a correctness bug, but it can slow `wasm-mutate-stats`, especially when many mutations are generated or artifact storage is slow.

## Proposed Fix

Pop under the mutex, store the result in a local, and drop the guard before entering the body that writes to disk.

```diff
diff --git a/crates/wasm-mutate-stats/src/bin/wasm-mutate-stats.rs b/crates/wasm-mutate-stats/src/bin/wasm-mutate-stats.rs
@@
-            while !finish_writing_wrap_clone.load(Relaxed) {
-                // pop from worklist
-                if let Some(wasm) = to_write_clone.lock().unwrap().pop() {
+            while !finish_writing_wrap_clone.load(Relaxed) {
+                // Pop under the lock and drop the guard before writing to disk.
+                let wasm = to_write_clone.lock().unwrap().pop();
+                if let Some(wasm) = wasm {
                     let filename = artifact_folder_cp.join(format!("mutated.{counter}.wasm"));
                     std::fs::write(filename, &wasm).context("Failed to write mutated wasm")?;
                     counter += 1;
                 }
             }
             eprintln!("Writing down pending mutated binaries!");
             // Then write pending wasms
-            while let Some(wasm) = to_write_clone.lock().unwrap().pop() {
+            while let Some(wasm) = {
+                let mut pending = to_write_clone.lock().unwrap();
+                pending.pop()
+            } {
                 let filename = artifact_folder_cp.join(format!("mutated.{counter}.wasm"));

                 std::fs::write(filename, &wasm).context("Failed to write mutated wasm")?;
                 counter += 1;
             }
```

## Notes

If needed, I can submit a PR containing this fix.

SAY-5:
Closes #2490.

The writer thread in `crates/wasm-mutate-stats/src/bin/wasm-mutate-stats.rs` popped from the worklist with the lock held for the entire body of the `if let` / `while let`:

```rust
if let Some(wasm) = to_write_clone.lock().unwrap().pop() {
    std::fs::write(filename, &wasm).context("Failed to write mutated wasm")?;
    counter += 1;
}
```

The temporary `MutexGuard` produced by `.lock()` lives for the whole scrutinee, so the queue mutex stayed held across every `std::fs::write` call. Producers and the shutdown signal blocked on each file write, hurting throughput and increasing shutdown latency.

Hoist the `.pop()` into its own statement so the guard is dropped before the I/O, and match on the resulting `Option` in a separate scope. Same fix for both the steady-state `while !finish_writing_wrap_clone.load(...)` loop and the post-shutdown drain loop.

`cargo check -p wasm-mutate-stats` clean against current `main`.

### DIFF
--- a/crates/wasm-mutate-stats/src/bin/wasm-mutate-stats.rs
+++ b/crates/wasm-mutate-stats/src/bin/wasm-mutate-stats.rs
@@ -453,18 +453,24 @@ impl State {
             let mut counter = 0;
 
             while !finish_writing_wrap_clone.load(Relaxed) {
-                // pop from worklist
-                if let Some(wasm) = to_write_clone.lock().unwrap().pop() {
+                // Pop from the worklist while holding the mutex for as short a window
+                // as possible — the previous form `if let Some(_) = to_write_clone.lock()
+                // .unwrap().pop()` keeps the temporary MutexGuard alive for the entire
+                // body of the `if let`, blocking the producers (and shutdown) on every
+                // file write.
+                let wasm = to_write_clone.lock().unwrap().pop();
+                if let Some(wasm) = wasm {
                     let filename = artifact_folder_cp.join(format!("mutated.{counter}.wasm"));
                     std::fs::write(filename, &wasm).context("Failed to write mutated wasm")?;
                     counter += 1;
                 }
             }
             eprintln!("Writing down pending mutated binaries!");
-            // Then write pending wasms
-            while let Some(wasm) = to_write_clone.lock().unwrap().pop() {
+            // Then write pending wasms — same lock-narrowing as above.
+            loop {
+                let wasm = to_write_clone.lock().unwrap().pop();
+                let Some(wasm) = wasm else { break };
                 let filename = artifact_folder_cp.join(format!("mutated.{counter}.wasm"));
-
                 std::fs::write(filename, &wasm).context("Failed to write mutated wasm")?;
                 counter += 1;
             }


alexcrichton (approved):
Thanks!

spidermana:
Cool! It's nice seeing my suggestions (#2490) getting merged.

PR #2492, opened 2026-04-16
Fix merge imports based on semver

fibonacci1729:
#2451 introduced `topologically_sort_interfaces` and correctly calls it at the end of `Resolve::merge_packages` after modifying interface dependencies. However, `merge_world_imports_based_on_semver` was not updated with the same fix.

`merge_world_imports_based_on_semver` calls `update_interface_dep_of_type` to rewire type aliases from older semver-compatible interfaces to their newer replacements. This changes the dependency edges in the interface graph: an interface `A` that previously pointed to `B_old` (positioned before `A` in the arena) may now point to `B_new` (positioned after `A`). This violates the topological ordering invariant that `assert_topologically_sorted` checks.

The `merge` path already handled this by calling `self.topologically_sort_interfaces(...)` after modifying type references, but the semver deduplication path was missed.

### DIFF
--- a/crates/wit-parser/src/resolve/mod.rs
+++ b/crates/wit-parser/src/resolve/mod.rs
@@ -2559,6 +2559,11 @@ package {name} is defined in two different locations:\n\
             assert!(prev.is_none());
         }
 
+        // The type-reference replacements above may have changed interface
+        // dependency edges, so re-sort interfaces topologically before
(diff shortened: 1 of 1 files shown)


dicej:
Looks reasonable, but CI is unhappy.

fibonacci1729:
Closing this to further investigate. #2493

PR #2496, opened 2026-04-17
feat: (CM64) Allow component model `resource`s to be represented by `i64` as well as `i32`. 

michael-weigelt:
This is the first PR in a series to enable verification of 64bit component model modules. The effort is tracked [here](https://github.com/bytecodealliance/wasm-tools/issues/2500).

This PR extends the component model's `resource rep` immediate to allow `i64` and gates this behind a feature flag `CM64`, which corresponds to the elephant emoji 🐘 in the [spec PR](https://github.com/WebAssembly/component-model/pull/624).


crates/wasmparser/src/validator/component.rs
```diff
@@ -588,8 +588,11 @@ impl ComponentState {
...
 
                 // Current MVP restriction of the component model.
-                if rep != ValType::I32 {
-                    bail!(offset, "resources can only be represented by `i32`");
+                if rep != ValType::I32 && rep != ValType::I64 {
+                    bail!(
+                        offset,
+                        "resources can only be represented by `i32` or `i64`"
+                    );
                 }
```
alexcrichton: For `ValType::I64`, could you be sure to [add a new feature here](https://github.com/bytecodealliance/wasm-tools/blob/599a4cddcb26a4294637134d00491d7466acd27b/crates/wasmparser/src/features.rs#L310-L314), probably called `CM64`, which gates all the new validation in this PR?

crates/wasmparser/src/validator/component.rs
```diff
@@ -4366,8 +4370,27 @@ impl ComponentState {
...
+                        memory64: true,
+                        shared: false,
+                        page_size_log2: None,
+                    },
+                    offset,
+                )
+            } else {
+                Err(e)
+            }
+        })
```
alexcrichton: For this I might recommend checking `ty` to see if it's `memory64`-or-not and then performing a subtype with the corresponding type. Either that or creating a `MemoryType` with the `memory64` value mirrored from the origin `ty`.

This'll also want to check for the `cm64` feature introduced in the above comment to ensure that's enabled here.

michael-weigelt: aff4dc27f122643e3e4a04838d02771d19fab9f8

crates/wasmparser/src/validator/component.rs
michael-weigelt: aff4dc27f122643e3e4a04838d02771d19fab9f8

michael-weigelt:
Do we squash merge? I hope so, otherwise I'll be more cautious with my commits.

tests/cli/component-model/resources.wast
```diff
@@ -50,7 +50,7 @@
 
 (assert_invalid
   (component
-    (type $x (resource (rep i64)))
+    (type $x (resource (rep v128)))
```
alexcrichton: Could this stick to using `i64` to ensure that, without the feature, this continues to not validate?

tests/cli/component-model/memory64.wast
```diff
@@ -51,4 +50,3 @@
     (alias core export $A "m" (core memory $m))
     (core func (canon lower (func $x) (memory $m)))
   )
-  "canonical ABI memory is not a 32-bit linear memory")
```
alexcrichton, line 54: Could this test, specifically that it's not accepted without the `cm64` feature, stick around elsewhere in the test suite? (e.g. added to another file or such)

crates/wasmparser/src/validator/component.rs
```diff
@@ -4351,23 +4360,35 @@ impl ComponentState {
...
+                msg = "or 64-bit ";
+                SubtypeCx::memory_type(ty, &valid_memory_type, offset)
+            }
+        }
         .map_err(|mut e| {
-            e.add_context("canonical ABI memory is not a 32-bit linear memory".into());
+            e.add_context(format!(
+                "canonical ABI memory is not a 32-bit {msg}linear memory"
+            ));
             e
```
alexcrichton: Conventionally one thing I shoot for within the validator is to structure things as-if all features are enabled, such that when a feature is eventually removed it'd just deleting some guards. In that sense I try to avoid major branches in control flow when features are enabled and instead just try to gate one or two particular properties based on features. Here for example that might look like this:

```rust
let ty = self.memory_at(idx, offset)?;
let valid_memory_type = MemoryType {
    initial: 0,
    maximum: None,
    memory64: ty.memory64,
    shared: false,
    page_size_log2: None,
};
if ty.memory64 && !self.features.cm64() {
    bail!(
        offset,
        "64-bit memories require the component model 64-bit feature"
    );
}
SubtypeCx::memory_type(ty, &valid_memory_type, offset).map_err(|mut e| {
    e.add_context("canonical ABI memory is valid to use");
    e
})?;
```

that way the `cm64` branch "only barely" diverges from the main path while still gating the same bits. Above it might look like:

```rust
if rep == ValType::I64 && !component.features.cm64() {
    bail!(
        offset,
        "resources with `i64` require the `cm64` feature to be enabled"
    )
}
if rep != ValType::I32 && rep != ValType::I64 {
    bail!(
        offset,
        "resources can only be represented by `i32` or `i64`"
    );
}
```

---

Overall though this is a pretty small stylistic nit, so no need to block this PR on this. The main reason I'm bringing this up is that I'd like to at least retain the tests below about the feature not working when the feature is disabled (e.g. 64-bit memories disallowed), and I figured I'd note this as well.

alexcrichton (approved):
Indeed yeah we do the squash & merge commit strategy so no need to fret too much about commit history

crates/wasmparser/src/validator/component.rs
michael-weigelt: I was going back and forth on how to structure this, so thanks for the input.

tests/cli/component-model/resources.wast
michael-weigelt: Yeah, I forgot to move this behind the feature flag. Same below.

crates/wasmparser/src/validator/component.rs
michael-weigelt: But I am not sure why you propose "canonical ABI memory is valid to use", should we not just not add anything here?

alexcrichton: Oops sorry I typo'd and forgot the word "not" there,  but yeah it might be reasonable to just skip the extra context as it's probably not adding too much any more. No need to fret too much over it though this is a pretty minor concern

tests/cli/component-model/resources.wast
michael-weigelt: 6358e4db

tests/cli/component-model/memory64.wast
michael-weigelt: I think things are in their proper place now: 6358e4db

Issue #2500, opened 2026-04-20
Add validaton for 64bit component model


michael-weigelt:
This is a checklist issue to track the progress of the implementation side of [this 🐘 spec PR](https://github.com/WebAssembly/component-model/pull/624). 

The following is a list of subfeatures that are not or only loosely related and can be worked on independently. Validating a 64bit component model module will (in general) only really work once they are all done, but since the feature is behind a feature flag, it is ok for the feature to be in an incomplete state. 
- [x] Feature flag [#2496](https://github.com/bytecodealliance/wasm-tools/pull/2496)
- [x] resource rep [#2496](https://github.com/bytecodealliance/wasm-tools/pull/2496)
- [x] realloc [#2501](https://github.com/bytecodealliance/wasm-tools/pull/2501)
- [x] context.{get, set} [#2502](https://github.com/bytecodealliance/wasm-tools/pull/2502)
- [x] waitable-set.{wait, poll} [#2514](https://github.com/bytecodealliance/wasm-tools/pull/2514)
- [ ] stream.{read, write}
- [ ] future.{read, write}
- [ ] thread.{new-indirect, spawn-ref, spawn-indirect}
- [ ] error-context.{new, debug-message}

PR #2501, opened 2026-04-23
feat: (CM64) Make `realloc` signature match the memory's address type

michael-weigelt:
This is part of a series of PRs to [enable verification of 64bit component model modules](https://github.com/WebAssembly/component-model/pull/624). The effort is tracked in https://github.com/bytecodealliance/wasm-tools/issues/2500.

This PR validates that the `realloc` canonopt signature matches the `memory` canonopt address type.


tests/cli/component-model/async/task-builtins.wast
```diff
@@ -79,10 +79,11 @@
 (assert_invalid
   (component
     (core module $m
+      (memory (export "m") i32 1)
       (func (export "r") (param i32 i32 i32 i32) (result i32) unreachable)
     )
     (core instance $m (instantiate $m))
-    (core func $task-return (canon task.return (result u32) (realloc (func $m "r"))))
+    (core func $task-return (canon task.return (result u32) (memory $m "m") (realloc (func $m "r"))))
```
michael-weigelt: This shows that the validation of `realloc` depends on `memory`. I think it's inevitable, so this should be ok?

crates/wasmparser/src/validator/component.rs
```diff
@@ -2635,11 +2635,18 @@ impl ComponentState {
                 CanonicalOption::Realloc(idx) => {
                     realloc = match realloc {
                         None => {
+                            let memory_idx = memory.ok_or(BinaryReaderError::new(
+                                "canonical option `realloc` requires the `memory` option",
+                                offset,
+                            ))?;
```
alexcrichton: This'll want to defer validation of the signature until after all canonical options have been processed because the `realloc` option isn't required to show up after the `memory` option (there's no implicit orderings). What this can do, however, is to require, at the end, that if both realloc and memory are present then there's the correct signature.

To preserve the preexisting behavior of specifying `realloc` and no `memory` I think we'll have to continue to validate that it has the `i32`-based signature, but we should probably make it a first-class spec-level error to specify `realloc` with `memory` or basically consider this situation in isolation.

tests/cli/component-model/async/task-builtins.wast
alexcrichton: I commented below, but I think we're going to want to preserve the previous behavior (albeit it's a bit nonsensical with memory64 in the spec), and probably seek out clarification at the spec level of what to do here

crates/wasmparser/src/validator/component.rs
michael-weigelt: Paragraph 1: Will do. 

>To preserve the preexisting behavior of specifying realloc and no memory

Do I understand you correctly that for the present, we must remain backwards compatible, but that you are suggesting to change the spec and move forward at some point? If so, we should gate [this new bullet point](https://github.com/WebAssembly/component-model/blob/main/design/mvp/CanonicalABI.md#canonical-definitions) behind the elephant emoji: 
> - if `realloc` is present then `memory` must be present

Or we remove it altogether for the present.

michael-weigelt: Or, if you can point me in the right direction, I can try to
> make it a first-class spec-level error to specify realloc with[out] memory

tests/cli/component-model/async/task-builtins.wast
michael-weigelt: 168015381943d7fffb0301ba3925c58b11fdf6ee

crates/wasmparser/src/validator/component.rs
alexcrichton: Hm ok yeah good point. That bullet was actually added by me in WebAssembly/component-model#467 as well, but I botched the implementation in wasm-tools. The implementation in wasm-tools is such that if `realloc` is required, then `memory` is also required. The implement needs to be if `realloc` is *specified*, then `memory` is required, however.

Could you file an issue for this and tag this handle-`memory`-is-`None` with a fixme pointing to that issue? This PR preserves the preexisting behavior, which is good, but I'll need to sit down and review to see if this is something we can "just change" or whether it's a problem in practice and needs to change more.

alexcrichton (approved):
In the meantime this all looks good so I'll merge

crates/wasmparser/src/validator/component.rs
michael-weigelt: [Issue](https://github.com/bytecodealliance/wasm-tools/issues/2503): Please edit the text if it's imprecise. 
I also refer to it [here](https://github.com/bytecodealliance/wasm-tools/pull/2504). By tagging do you mean naming or did you want some sort of label?

alexcrichton: Oh no that's perfect, thanks! I was just thinking of `// FIXME(#...)` as you did in https://github.com/bytecodealliance/wasm-tools/pull/2504

PR #2510, opened 2026-04-30
Add more snapshots for feature gate paths and fix a panic on @since include

PhoebeSzmucer:
Adding a couple of snapshots that will be useful for making sure that the error messages don't regress in https://github.com/bytecodealliance/wasm-tools/pull/2482

In the process I discovered a panic in the new `bad-since-on-include.wit` test (coming from `process_world_includes`). We were reading from `resolve.worlds[include.id].name.as_str()` but `include.id` is an index from the unresolved package's arena, not the `Resolve`'s arena. This code path wasn't being hit before in any tests.


alexcrichton (approved):
Thanks!

PR #2513, opened 2026-05-06
update bumpalo

maksimryndin:
Hey guys!

Thanks a lot for your tremendous work over wasm rust tooling!

A small PR to update `bumpalo` as the old version causes version conflicts (https://github.com/bytecodealliance/wasmtime/blob/main/Cargo.toml#L374)

### DIFF
--- a/crates/wast/Cargo.toml
+++ b/crates/wast/Cargo.toml
@@ -20,7 +20,7 @@ all-features = true
 workspace = true
 
 [dependencies]
-bumpalo = "3.14.0"
+bumpalo = "3.20"
 gimli = { workspace = true, optional = true }
 leb128fmt = { workspace = true }
 memchr = "2.4.1"


alexcrichton (approved):
Thanks! Although FWIW this shouldn't be necessary because the version requirement here is "3.x.x, minimum 3.14.0", which is compatible with Wasmtime's requirement of 3.20. Out of curiosity, how did the conflict arise?

Nevertheless this is still fine to merge, so I'll flag it as such here.

maksimryndin:
> Thanks! Although FWIW this shouldn't be necessary because the version requirement here is "3.x.x, minimum 3.14.0", which is compatible with Wasmtime's requirement of 3.20. Out of curiosity, how did the conflict arise?
> 
> Nevertheless this is still fine to merge, so I'll flag it as such here.

Hey @alexcrichton! ✋ I am upgrading dependencies in the project https://github.com/NethermindEth/stellar-private-payments/pull/191 and there was a conflict with the latest wast (https://github.com/NethermindEth/stellar-private-payments/blob/feature-protocol-26/circuits/build.rs#L874). But you're absolutely right - just updating it in Cargo.lock solved the issue.

Thanks a lot for a quick feedback!

PR #2515, opened 2026-05-10
wasmparser: add regression test for relocation_range overflow

### LINKED ISSUE #2499: Add a regression test to cover wrapping addition in `RelocationEntry::relocation_range`
This was addressed in https://github.com/bytecodealliance/wasm-tools/pull/2497, but still needs a test.

eyupcanakman:
Adds the regression test that #2497 was missing, as requested in #2499.

The test calls `relocation_range` on an entry with `offset = 0xFFFF_FFFC` and the same `FunctionIndexLeb` extent that triggered the original report. On 32-bit targets the new `checked_add` returns `Err`; on 64-bit the call succeeds with the expected range.

Closes #2499.

### DIFF
--- a/crates/wasmparser/src/readers/core/reloc.rs
+++ b/crates/wasmparser/src/readers/core/reloc.rs
@@ -302,3 +302,21 @@ impl<'a> FromReader<'a> for RelocationEntry {
         })
     }
 }
+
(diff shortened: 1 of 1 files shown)


alexcrichton (approved):
Thanks!

PR #2521, opened 2026-05-16
addr2line: echo input addresses back as given instead of hex

### LINKED ISSUE #2328: `addr2line` prints addresses in hex if given in decimal
```
$ wasm-tools addr2line --generate-dwarf lines foo.wasm 255 512
0xff: main foo.c:21:9
0x200: main foo.c:25:9
```
It's a bit unclear if this should be considered a bug or not, but I'm leaning towards thinking that if someone provides the addresses in decimal, they probably want to see the results in decimal.

eyupcanakman:
addr2line parsed each address into a `u64` and used that same binding for the printed label, so a decimal input like `255` came back as `0xff`. Keep the original input string for the label and only feed the parsed value into the lookup. Hex input is unchanged. Updated the help-text example and added a cli test.

Closes #2328


alexcrichton (approved):
Thanks!

PR #2524, opened 2026-05-21
Support components with custom page sizes

anlavandier:
This PR partly reverts #2477 by allowing again but now in an explicit manner components whose inner modules have linear memories with custom page sizes.

cc @alexcrichton

See this [Zulip Topic](https://bytecodealliance.zulipchat.com/#narrow/channel/511605-wasm-tools/topic/Making.20component.20from.20modules.20with.20pagesize.200x1) for additional context.

### DIFF
--- a/crates/wasmparser/src/validator/component.rs
+++ b/crates/wasmparser/src/validator/component.rs
@@ -4481,7 +4481,7 @@ impl ComponentState {
             maximum: None,
             memory64: ty.memory64,
             shared: false,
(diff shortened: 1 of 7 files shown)


dicej (approved):
Thanks!  Looks reasonable to me, but I'll let @alexcrichton take a look before merging it since he has more context for this.  He's out this week but should be back next week.

alexcrichton (approved):
Thanks!

PR #2533, opened 2026-05-28
wit-parser: rust docs for ItemName, and instance_name method doing the obvious

pchickey:
ItemName was missing nice rust docs, and the `instance_name` method is something that is expected to be needed in various places.


pchickey:
I'd like to cut a wasm-tools release once this lands.

alexcrichton (approved):

PR #2535, opened 2026-05-30
Converge to unicode-ident and remove unicode-xid dependency

Marcondiro:
Hello,

`unicode-xid` and `unicode-ident` do the exact same thing and they are both in the dependency tree.
`unicode-xid` is imported directly while `unicode-ident` is brought in by `serde`.

Converge over the more performant `unicode-ident`.
Ideally this will slightly reduce binary size and slightly reduce compile time.

This PR does not introduce any functional change.

Thanks!


alexcrichton (approved):
Thanks!

PR #2536, opened 2026-06-01
wasm-smith: Allow limiting the size of const arrays

alexcrichton:
This adds some config knobs to `wasm-smith` to enable limiting the dynamic size of constant expressions in generated modules, notably the array lengths. This additionally tweaks the fuel consumption strategy and moves the compile-time constant to a runtime constant.

cc bytecodealliance/wasmtime#13521


fitzgen (approved):
Nice!

PR #2538, opened 2026-06-11
fix(wit-encoder): only use serde attribute if serde is enabled

MendyBerger:


alexcrichton:
Thanks! Could CI be updated [here](https://github.com/bytecodealliance/wasm-tools/blob/c59fb63e218e8a1b5434038ade23e2629140038f/.github/workflows/main.yml#L252-L334) with the command that failed for you before this PR but passes after?

MendyBerger:
Good point @alexcrichton. Added

alexcrichton (approved):

Issue #1471, opened 2024-03-27
`json-from-wast`: differences to wabt's `wast2json`


ohorn:
I just saw the new `json-from-wast` command. It is great to see a `wast2json`-like tool as part of `wasm-tools`, especially since it supports many more proposed features than `wabt`.

However, I've noticed that `json-from-wast` outputs slightly different JSON than `wabt`. In particular:

* `wabt` writes the module names in their `$`-prefixed form, while `json-from-wast` drops the `$` from the module names.
* `wabt` writes all numeric values in their unsigned representation (even if the corresponding type itself is signed like `i32` and `i64`), while `json-from-wast` writes negative numbers as they are, with a minus sign.

Another difference is how the filenames for the associated modules are derived: `wabt` derives these names from the given output filename. IMHO, this has the advantage that both the JSON file and the associated module files share the same base name.

As noted in the commit, `json-from-wast` also doesn't write the `expected` return types for commands like `assert_trap`. Are there any plans to add this feature in the future?
I also experimented with my own `wast2json` implementation in the past. My approach was to parse and analyze the associated modules after encoding and store their exported function and global types so that they could be queried when needed.

alexcrichton:
Hello and thanks for the report! Yeah I noticed these differences as well when writing it, and while they could be fixed it felt a bit odd given the implementation in this repository to go out of our way to represent that. Are these differences causing issues for consumption?

If you'd like I think it'd be reasonable to change the output in most of the cases here. I had my own reasons for not matching wabt exactly in some of these cases (mostly that it required "going out of the way" to exactly match as opposed to do what this repository does natively), but other cases (like the filename bits) I just didn't realized. I think it'd be totally fine to have a PR to update things, if you're willing.

> also doesn't write the expected return types for commands like assert_trap. Are there any plans to add this feature in the future?

That could be added, we've in theory got all the tooling to support this. It would be a pretty big lift though because we'd have to (a) maintain a mapping of named modules to know which module is being referred to with `assert_{trap,invoke}` and (b) we'd have to parse and validate the module to determine the type signature. I was hoping that the `expected` bits wouldn't really be that necessary to avoid needing this work, but if you'd like to send a PR I think that's reasonable as well.

ohorn:
[...]

PR #2542, opened 2026-06-19
json-from-wast: name emitted files after the output stem

eyupcanakman:
`json-from-wast` named its emitted `*.wasm` and `*.wat` files after the source `.wast` stem, so `-o out.json` still wrote `<source>.0.wasm` rather than `out.0.wasm` the way `wast2json` does. This derives the stem from the `-o` output path when one is given and falls back to the source stem otherwise. Added a cli test for both cases.

Refs #1471.


alexcrichton:
Thanks! The CI failure might be some CRLF thing, but I'm not entirely sure.

Out of curiosity though, could you explain a bit more why you want an exact match? Naively it seems good to strip the leading `$` and it seems better to render "-1" instead of u32::MAX. I'm not opposed to these changes necessarily, but for future-selves I think it'd be good to have written down motivation. (even if it's "just because it's nice to byte-for-byte match" if that's the case)

eyupcanakman:
It was just to match wast2json more closely, but your question made me check who consumes this, and chicory (which runs json-from-wast) parses the i32/i64 values as signed, so the unsigned change would break their `Integer.parseInt` (and probably the `$`-prefix). I'd lean toward trimming this to just the filename fix and dropping the int and name changes, unless you'd rather keep the full match. The Windows fail was a path-separator thing in the test, not CRLF, and I'll fix that.

alexcrichton:
That sounds reasonable to me, yeah 👍

eyupcanakman:
Trimmed this to just the filename change and reverted the int and name bits along with the snapshots. The Windows fail was the cli test runner handing a backslash path to `source_filename`, so I normalized the `%` substitution to forward slashes like `wast` already does.

alexcrichton (approved):
Thanks!

PR #2544, opened 2026-06-23
fix: signature lowering for strings and lists on modules with 64bit memory

christoph-dfinity:
Fixes signature lowering for strings and lists on modules with 64bit memory. 

Before this change the newly added tests fail with a signature mismatch, because the lowering unconditionally produced i32 pointers and lengths. On modules with 64bit memories these need to be i64 instead, based on my reading of the [Canonical ABI document](https://github.com/WebAssembly/component-model/blob/main/design/mvp/CanonicalABI.md#loading)

Before the fix the tests fail with:
```
---- "tests/cli/component-model/memory64/string.wast" ----
failed test "tests/cli/component-model/memory64/string.wast"
...
       Caused by:
           0: failed testing wasm binary produced by `wast`
           1: wasm isn't valid
           2: lowered parameter types `[I32, I32]` do not match parameter types `[I64, I64]` of core function 0 (at offset 0x8f)
```

/cc @michael-weigelt


alexcrichton (approved):
Thanks!

PR #2546, opened 2026-06-26
wasmparser: Rename BinaryReaderError to Error

lann:
This type hasn't been particularly `BinaryReader`-specific for a long time, being returned by e.g. `Validator` which can be used without any direct use of a `BinaryReader`.

- Move `crate::binary_reader::BinaryReaderError` to `crate::error::Error`
- Add `crate::BianryReaderError` type alias for backward compatibility
- Update references in other workspace crates


lann:
I could easily be convinced that this should be e.g. `ParserError` instead. This is just a side quest on the way toward exposing details about `WasmFeatures` validation errors.

alexcrichton (approved):
Definitely a cleanup I've wanted to see in the past myself as well, and I think `Error` is fine given how it's The Error for the crate. Thanks!

PR #2548, opened 2026-06-29
Refactor `KebabStr` to not require unsafe code

alexcrichton:
This requires changing `KebabStr`'s implementation internally and adjusting callers in various locations, but the overall concept remains the same. The main difference now is that `KebabStr` and `KebabString` no longer have a `Borrow` or `ToOwned` relationship, and `KebabString` no longer derefs to `KebabStr`. For making wasmparser 100% safe this seems like a reasonable tradeoff.

This additionally fixes a problem with the previous signature of `KebabStr::new_unchecked` which enabled UB at runtime if called with owned types such as `String` or `Box<str>` or such. While `wasmparser` never actually did this it was still an unsound API to expose to users, so worthwhile to fix.

### DIFF
--- a/crates/wasmparser/src/lib.rs
+++ b/crates/wasmparser/src/lib.rs
@@ -42,7 +42,6 @@ extern crate std;
 /// like `String`. This custom prelude helps bring those types into scope to
 /// avoid having to import each of them manually.
 mod prelude {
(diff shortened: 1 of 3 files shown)


fitzgen (approved):
Thanks!

PR #2549, opened 2026-06-29
wasmparser: Add Error::missing_wasm_feature

lann:
Add generated require_feature::* methods that return Errors with a new ErrorKind::WasmFeature if the named feature is disabled. Replace most feature checks with calls to these methods.
    
Checks in BinaryReader (and ancillary code) are gated on the 'features' feature to preserve existing behavior.

Additional internal changes:

- Made Error fields private by adding a couple of methods
- Renamed ErrorKind variants for clarity

### DIFF
--- a/crates/wasmparser/src/error.rs
+++ b/crates/wasmparser/src/error.rs
@@ -13,10 +13,10 @@ pub struct Error {
 
 #[derive(Debug, Clone)]
 pub(crate) struct ErrorInner {
-    pub(crate) message: String,
-    pub(crate) kind: ErrorKind,
-    pub(crate) offset: usize,
-    pub(crate) needed_hint: Option<usize>,
+    message: String,
+    kind: ErrorKind,
+    offset: usize,
+    needed_hint: Option<usize>,
 }
 
 #[derive(Debug, Clone, Copy)]
@@ -96,5 +96,14 @@ impl Error {
 
     pub(crate) fn set_message(&mut self, message: &str) {
         self.inner.message = message.to_string();
+    }
+
+    pub(crate) fn needed_hint(&self) -> Option<usize> {
+        self.inner.needed_hint
+    }
+
+    pub(crate) fn without_needed_hint(mut self) -> Self {
+        self.inner.needed_hint = None;
+        self
     }
 }
--- a/crates/wasmparser/src/error.rs
+++ b/crates/wasmparser/src/error.rs
@@ -1,5 +1,6 @@
 use core::fmt;
 
+use crate::WasmFeatures;
 use crate::prelude::*;
 
 /// A binary reader for WebAssembly modules.
@@ -23,6 +24,7 @@ pub(crate) struct ErrorInner {
 pub(crate) enum ErrorKind {
     Uncategorized,
     InvalidHeapType,
+    WasmFeature(WasmFeatures),
 }
 
 /// The result for `BinaryReader` operations.
@@ -64,6 +66,15 @@ impl Error {
     }
 
     #[cold]
+    pub(crate) fn wasm_feature(
+        feature: crate::WasmFeatures,
+        msg: impl fmt::Display,
+        offset: usize,
+    ) -> Self {
+        Self::_new(ErrorKind::WasmFeature(feature), msg.to_string(), offset)
+    }
+
+    #[cold]
     pub(crate) fn fmt(args: fmt::Arguments<'_>, offset: usize) -> Self {
         Error::new(args.to_string(), offset)
     }
@@ -75,7 +86,7 @@ impl Error {
         err
     }
 
-    pub(crate) fn kind(&mut self) -> ErrorKind {
+    pub(crate) fn kind(&self) -> ErrorKind {
         self.inner.kind
     }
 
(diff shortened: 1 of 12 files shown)


crates/wasmparser/src/readers/core/imports.rs
```diff
@@ -95,13 +95,13 @@ impl<'a> FromReader<'a> for Imports<'a> {
...
             ("", 0x7F) => {
-                if !reader.compact_imports() {
-                    bail!(
-                        reader.original_position(),
-                        "invalid leading byte 0x7F with compact imports \
-                         proposal disabled"
-                    );
-                }
+                #[cfg(feature = "features")]
+                crate::require_feature::compact_imports(
```
lann, line 99: The existing generated `BinaryReader::<feature>()` methods always return `true` (not `$default`) when `features` is disabled which was already effectively disabling these checks.

crates/wasmparser/src/validator/core/canonical.rs
```diff
@@ -93,11 +93,12 @@ pub(crate) trait InternRecGroup {
...
-        if rec_group.is_explicit_rec_group() && !self.features().gc() {
-            bail!(
+        if rec_group.is_explicit_rec_group() {
+            require_feature::gc(
+                *self.features(),
+                "rec group usage requires `gc` proposal to be enabled",
                 offset,
-                "rec group usage requires `gc` proposal to be enabled"
-            );
+            )?;
```
lann, line 101: This vast majority of the changes in this PR look basically like this. I'll try to annotate the "weirdest" other ones.

crates/wasmparser/src/validator/component.rs
```diff
@@ -1545,13 +1541,18 @@ impl ComponentState {
...
-                "`{operation}` immediate must be zero or one: {immediate}"
-            )
+            )?;
+            if immediate > 1 {
+                bail!(
+                    offset,
+                    "`{operation}` immediate must be zero or one: {immediate}"
+                )
+            }
         }
```
lann, line 1556: It took a distressingly long time to convince myself this was equivalent. 😅

crates/wasmparser/src/error.rs
```diff
@@ -89,12 +91,35 @@ impl Error {
...
+    /// will contain a feature that could be enabled to prevent this error.
+    ///
+    /// **Note:** This is not comprehensive; `None` may be returned in some
+    /// cases where enabling a feature could avoid the error.
+    pub fn missing_wasm_feature(&self) -> Option<WasmFeatures> {
+        match self.inner.kind {
+            ErrorKind::WasmFeature(features) => Some(features),
+            _ => None,
+        }
+    }
```
lann, line 115: It isn't clear to me what this should do when `features` is disabled. Currently it just returns the useless `Some(WasmFeatures{ _priv: () })` which at least signals that _some_ feature is missing. Alternatively it could unconditionally return `None` or the entire method could be gated. 🤷

crates/wasmparser/src/error.rs
```diff
@@ -89,12 +91,35 @@ impl Error {
...
+        self.inner.message = message.into();
+    }
+
+    pub(crate) fn needed_hint(&self) -> Option<usize> {
+        self.inner.needed_hint
+    }
+
+    pub(crate) fn without_needed_hint(mut self) -> Self {
+        self.inner.needed_hint = None;
+        self
```
lann, line 132: This was a tangent that allows the `Error` fields to become private.

crates/wasmparser/src/features.rs
```diff
@@ -105,6 +105,33 @@ macro_rules! define_wasm_features {
             }
         }
         pub(crate) use foreach_wasm_feature;
+
+        #[allow(dead_code)]
```
lann, line 109: The three unused methods are `multi_memory`, `cm_nested_names`, and `call_indirect_overlong`. This does mean that those features are never reflected in `missing_wasm_feature` as noted in its doc comment. That could be fixed but they're all non-trivial (at least to me).

crates/wasmparser/src/error.rs
alexcrichton: Personally I prefer to avoid `#[cfg]` where possible so I think this is fine. There's nothing that can be done with the return value if this is returned in a feature-less build and that'd basically just mean "some off-by-default thing was used".

crates/wasmparser/src/features.rs
```diff
@@ -105,6 +105,33 @@ macro_rules! define_wasm_features {
...
+        pub(crate) mod require_feature {
+            use crate::Error;
+            use super::WasmFeatures;
+
+            $(
+                #[doc = "Returns an error if [`WasmFeatures::"]
+                #[doc = stringify!($const)]
+                #[doc = "`] is not enabled in `features`."]
+                pub fn $field(
```
alexcrichton, line 119: Some mild bikeshedding on this:

* This'll want to use `#[inline]` as otherwise it's not a candidate for inlining.
* The `msg` argument is one where the ideal here is that it's very low-overhead when not used and the processing on the error path ideally happens in a "cold" function somewhere else to avoid cluttering up hot functions. For example `Error::fmt` is itself `#[cold]` so the `to_string` happens out-of-the-hot-function which can help reduce function sizes.

Could there be a new `Error::new_display` function (probably crate-local to start) which is `#[cold]` to avoid the `format!` here? Or perhaps `Error::wasm_feature` which takes `impl Display` and is `#[cold]`?

crates/wasmparser/src/validator/operators.rs
```diff
@@ -2013,7 +2008,11 @@ macro_rules! validate_proposal {
     (validate self $proposal:ident / MemoryCopy) => {};
 
     (validate $self:ident $proposal:ident / $op:ident) => {
-        $self.0.check_enabled($self.0.features.$proposal(), validate_proposal!(desc $proposal))?
+        require_feature::$proposal(
+            $self.0.features,
+            validate_proposal!(desc $proposal),
```
alexcrichton: For the test failures I think this'll want to be `format_args!("{} is not enabled", validate_proposal!(desc $proposal)))`. Or if you want to get fancy I think `concat!` might work here too

alexcrichton (approved):
Thanks for this!

crates/wasmparser/src/features.rs
lann: I had `Error::wasm_feature` and mindlessly inlined it. 😅

Brought that back, added `#[inline]` here.

crates/wasmparser/src/validator/operators.rs
lann: Done.

PR #2550, opened 2026-06-30
Add CustomSectionReader -> BinaryReader conversion

WorldSEnder:
Construction by hand does not preserve features field of the parser.

This also shrinks the parser to the remaining data. Passing the reader as a field of `KnownCustom::Unknown` would break consumers, hence this is a standalone method here. The recommended approach is thus to use it as

```rust
match custom.name() {
  "my-interesting-custom-section" => foo(custom.data_reader()),
  _ => {},
}
```


alexcrichton (approved):
Thanks!

PR #2553, opened 2026-07-01
fix: use ptr-sized type when returning via ptr indirection under memory64

christoph-dfinity:
Follow-up to #2544 (I missed these)


alexcrichton (approved):
Thanks!

PR #2555, opened 2026-07-01
Initial work for `versionsuffix` and `external-id`

### LINKED ISSUE #2551: wast: WASM_PARENS_DEPTH too high on `aarch64-unknown-linux-gnu`
```console
$ rustc -vV
rustc 1.96.0 (ac68faa20 2026-05-25)
binary: rustc
commit-hash: ac68faa20c58cbccd01ee7208bf3b6e93a7d7f96
commit-date: 2026-05-25
host: aarch64-unknown-linux-gnu
release: 1.96.0
LLVM version: 22.1.2

$ cargo test --test cli -- tests/cli/component-model/very-nested.wast
[...]
       thread '<unknown>' (2755691) has overflowed its stack
       fatal runtime error: stack overflow, aborting
[...]

$ RUST_MIN_STACK=3000000 cargo test --test cli -- tests/cli/component-model/very-nested.wast
[...]
test result: ok ...
```

I can also get the test to pass by setting [`MAX_PARENS_DEPTH`](https://github.com/bytecodealliance/wasm-tools/blob/41cf129bfec14e710323271691d3c2c95dee9b05/crates/wast/src/parser.rs#L86) from 100 to...99. 😂

This happens to be a Linux VM on a Mac if that makes any difference.

alexcrichton:
This is intended to at least start laying groundwork and implementation for two features in the component model that this repository does not yet implement:

* `(versionsuffix "...")` for components -- introduced in WebAssembly/component-model#536 this is intended to pave a path forward to canonicalizing versions in component imports/exports in the future. For example where today you'd import `a:b/c@1.0.0` tomorrow you'd import `a:b/c@1`. This is not yet fully integrated into `wit-component`, however, and further work will be necessary for bindings generators as well to take advantage of this. It's all encoded in the same place as `external-id`, though, so I figured I'd at least start to lay the groundwork.

* `(external-id "...")` for components -- being introduced in WebAssembly/component-model#672 this is intended to attach arbitrary metadata to imports/exports in a component. This is surfaced in WIT as an `@external-id("...")` attribute. This should be hooked up throughout this implementation to WIT and such such that the implementation is intended to be complete. There's probably something I forgot, but all the major pieces should be there.

The `versionsuffix` parsing/validation is gated by a new `cm-canon-names` feature, and the `external-id` parsing/validation is gated behind the preexisting `cm-implements` feature to match the specification's classification.

Closes #2551


ricochet:
Heads up that gating `external-id` to instances is different from the spec PR. IMO gating on instances is the right call so we should funnel that back to the spec. AKA right now this PR has `external_id` only on `WorldItem::Interface`, not on Function/TypeDef.

Right now the Explainer's motivating example puts `external-id` on a bare `func` import:

```wat
(component $ESM
  (import "slugify"
    (external-id "https://esm.unpkg.com/slugify@1.6.6")
    (func (param "text" string) (result string)))
)
```

fitzgen (approved):
LGTM

PR #2557, opened 2026-07-06
Make parsing errors downcastable

PhoebeSzmucer:
This is a follow up to my earlier PRs i particular https://github.com/bytecodealliance/wasm-tools/pull/2465 and https://github.com/bytecodealliance/wasm-tools/pull/2482

Originally when the parsing errors were migrated to be structured errors, I didn't make them downcastable (like I did the resolver errors in https://github.com/bytecodealliance/wasm-tools/pull/2482). This PR fixes that - the errors are now downcastable, instead of being eagerly rendered into strings.


crates/wit-parser/src/lib.rs
```diff
@@ -347,32 +347,12 @@ impl UnresolvedPackageGroup {
     /// files.
     #[cfg(feature = "std")]
     pub fn parse_dir(path: impl AsRef<Path>) -> anyhow::Result<UnresolvedPackageGroup> {
```
PhoebeSzmucer, line 349: I kept this function, because it has some external consumers.

alexcrichton (approved):
Thanks!

PR #2559, opened 2026-07-07
Add `SourceMap::resolve_span` for resolving global span locations

PhoebeSzmucer:
This PR adds `SourceMap::resolve_span(Span) -> Option<SpanLocation>`, which resolves a global span to its source file and a file-local byte range. This makes it possible for consumers of wit-parser to get precise error locations.

### DIFF
--- a/crates/wit-parser/src/ast.rs
+++ b/crates/wit-parser/src/ast.rs
@@ -1946,6 +1946,32 @@ impl SourceMap {
         return msg;
     }
 
+    /// Resolves a global `span` to a location within a single source file.
+    ///
+    /// Returns the path the source was registered with and the span's byte
+    /// range (UTF-8) relative to that file's contents. Returns `None` if
+    /// `span` is unknown or does not fall within any source in this map.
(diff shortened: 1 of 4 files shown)


crates/wit-parser/src/ast.rs
```diff
@@ -1946,6 +1946,32 @@ impl SourceMap {
...
 
+    /// Resolves a global `span` to a location within a single source file.
+    ///
+    /// Returns the path the source was registered with and the span's byte
+    /// range (UTF-8) relative to that file's contents. Returns `None` if
+    /// `span` is unknown or does not fall within any source in this map.
+    ///
+    /// Spans pointing at the end of a file resolve to an empty range at the
+    /// file's length.
+    pub fn resolve_span(&self, span: Span) -> Option<SpanLocation<'_>> {
```
PhoebeSzmucer, line 1957: This LSP will call this method for spans that we get from structured errors.

alexcrichton (approved):

PR #2563, opened 2026-07-10
Reject out-of-range NaN payloads and bare type-index value types

vouillon:
Fixes two cases where `wasm-tools` accepts malformed input that the spec grammar and the reference interpreter (`wasm -d`) reject.

- Text parser silently masked over-long NaN payloads (`wast`)

  A NaN literal's payload `n` must satisfy `1 ≤ n < 2^signif_bits` (52 bits for  f64, 23 for f32). The parser masked the payload to the significand width  instead of range-checking it, so an over-long payload was silently truncated and accepted.  The spec testsuite asserts this (`const.wast`,  `float_literals.wast`) but only with the value `nan:0x80_0000` — which is
  exactly `2^23` and masks to `0`, so the old buggy code rejected it by  coincidence.

- Binary decoder accepted a bare type index as a value type (`wasmparser`)

  The short form of a reference type (no `0x63`/`0x64` prefix) is only an abbreviation for `ref null <abstract-heap-type>`. The decoder instead read a  full heap type there, so a bare non-negative type index was accepted as a  value type:  Now concrete/exact heap types reached through the short-form path are rejected  with `malformed reference type` at decode time. This also covers the over-long  LEB variant (`ff 00`).

### DIFF
--- a/crates/wasmparser/src/readers/core/types.rs
+++ b/crates/wasmparser/src/readers/core/types.rs
@@ -1822,17 +1822,25 @@ impl<'a> FromReader<'a> for RefType {
             }
             0x62 => Err(crate::Error::new("unexpected exact type", pos)),
             _ => {
-                // Reclassify errors as invalid reference types here because
-                // that's the "root" of what was being parsed rather than
-                // heap types.
-                let hty = reader.read().map_err(|mut e| {
+                // The short form of a reference type (without a `0x63`/`0x64`
+                // prefix byte) is only an abbreviation for `ref null <ht>` where
+                // `<ht>` is an *abstract* heap type; a bare (non-negative) type
+                // index is not a value type. Parse an abstract heap type
+                // directly, handling the `0x65` prefix for `shared` variants,
+                // rather than parsing a full heap type and then rejecting the
+                // concrete/exact cases that a heap type additionally allows.
+                let shared = reader.peek()? == 0x65;
+                if shared {
+                    reader.read_u8()?;
+                }
+                let ty = reader.read().map_err(|mut e| {
                     if let ErrorKind::InvalidHeapType = e.kind() {
                         e.set_message("malformed reference type");
                     }
                     e
                 })?;
-                RefType::new(true, hty)
-                    .ok_or_else(|| crate::Error::new("type index too large", pos))
+                // `RefType::new` is infallible for abstract heap types.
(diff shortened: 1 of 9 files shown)


crates/wasmparser/src/readers/core/types.rs
```diff
@@ -1831,6 +1831,15 @@ impl<'a> FromReader<'a> for RefType {
...
                 })?;
+                // The short form of a reference type (without a `0x63`/`0x64`
+                // prefix byte) is only an abbreviation for `ref null <ht>` where
+                // `<ht>` is an *abstract* heap type. A bare (non-negative) type
+                // index is not a value type, so reject concrete/exact heap types
+                // reached through this path even though they parse as heap types
+                // in their own right.
+                if let HeapType::Concrete(_) | HeapType::Exact(_) = hty {
+                    return Err(crate::Error::new("malformed reference type", pos));
+                }
```
alexcrichton: Instead of catching the result of parsing here, could this directly parse `AbstractHeapType` above? That way there's no need to reclassify errors or recover from an otherwise-successful parse that should have failed. I think in that case this'll have to handle the 0x65 case for `shared: true` abstract heap types, but that's a pretty small amount of duplication

tests/cli/bad-reftype.wast
```diff
@@ -0,0 +1,26 @@
...
+
+;; A value type is `numtype | vectype | reftype`, and a reference type is either
+;; an abstract heap type shorthand byte or the two-byte `0x64 heaptype` (`ref ht`)
+;; / `0x63 heaptype` (`ref null ht`) forms. There is no bare type-index value
+;; type, so a leading byte of `0x00` (a non-negative sLEB) is malformed.
+
+;; type 0 = (func), type 1 = (func (param <0x00>)) — bare index used as a value type.
+(assert_malformed
+  (module binary
+    "\00asm\01\00\00\00\01\08\02\60\00\00\60\01\00\00")
```
alexcrichton: would you be ok splitting this up like a number of other `module binary` tests throughout the test suite with annotations of what each byte or set of bytes are? For example annotations [like this](https://github.com/bytecodealliance/wasm-tools/blob/main/tests/cli/invalid/empty-br-table.wast)

alexcrichton (commented):
Thanks for this, and good catch! I've got some thoughts below on minor tweaks, but otherwise looks good to me 👍

alexcrichton (approved):

PR #2564, opened 2026-07-13
wasmparser: require result equivalence for `(on $t switch)` handler

vouillon:
The `resume`/`resume_throw` handler clause `(on $tag switch)` was accepted when the switch tag's result types were merely a *subtype* of the resumed continuation's results. This is unsound, and does not follow the spec: the handler judgment has no subsumption rule, so `(on tu switch) : t*` together with the required `hdl : t2*` forces `t* = t2*`. The two result vectors must be equivalent, not just related by subtyping.

V8 and Binaryen already check for equivalence (see https://github.com/v8/v8/commit/cf215c5985d19eb741959558ae0ac06d7810c48f).


alexcrichton (approved):
Thanks!

PR #2566, opened 2026-07-15
Add a parse recurse limit to `wasm-wave`

alexcrichton:
This commit adds a limit to the amount of recursive parsing done to avoid blowing the stack. This is intended, like other limits in wasm-tools, to allow most reasonable code while being tweakable later if necessary.


alexcrichton:
cc @lann

pchickey (approved):

lann (approved):

PR #2573, opened 2026-07-18
Use u64 to track file offsets instead of usize

### LINKED ISSUE #838: wasmparser: `usize` may not be the most appropriate type for offsets
The wasmparser implements an incremental parser for the binary webassembly encoding. In practice this means that tooling can work with files larger the size of the virtual memory. This is not unlike the concept of 32-bit systems still being capable of working with filesystem files larger than 4GiB (i.e. `off_t > uint32_t`).

While `wasmparser` does impose a limit on the number of entities within module contents, these limits are not sufficiently low to limit modules to under-4GiB on 32-bit (or 64-kiB on 16-bit) targets. That is, I believe, quite intentional. However it still uses `usize` to represent the offset while parsing, which seems like a foot gun. We probably should replace the offset with a `u64`.

This issue has not been written with any particular use-case in mind, so it isn’t something that needs to be fixed ASAP, but it still seems like something we probably should look into fixing when the opportunity presents itself.

WorldSEnder:
This replaces usize in a lot of places. Besides being a breaking change, this should not impact functionality for most users. Tracking file offset with `u64` is the more principled choice.

Fixes #838

Solves a TODO left long ago in the entry code of parse. Most of the places used `usize` to pass it along to other functions, sinking into Error code for nice error messages, and are thus not sensitive to the data type used. I found two cases that used `usize::MAX` as a dummy index in a path that called `expect()` on the eventual error.

There is a small abstraction in `offsets.rs` exposed as `InMemData` that allows conversion from `Range<u64>` to byte slices in case the whole data input fits in memory.


WorldSEnder:
A lot of consumers seem to hold the entire wasm input in memory, hence these need some ergonomics to convert their offsets back to `Range<usize>`. Not sure what the best approach is.

WorldSEnder:
I have not removed `MemOffset` outright, I think it's worthwhile the small indirection for `BinaryReader` but I have removed the additional `cfg(debug_assertions)` field. It's now a transparent `usize`.

alexcrichton (commented):
Personally I still feel that `InMemData` and `MemOffset` aren't worth it. This is already a very large change due to how many places it has to touch, and in my mind it's a consequence of the change at all that `as` casts are going to be needed when in-memory data is indexed with the offsets coming out of `wasmparser`. While I've no doubt that certain abstractions could cut down on the number of resulting `as` casts I'm not sure if the benefit of such abstractions outweighs their costs (e.g. new primitives to learn).

WorldSEnder:
> Personally I still feel that `InMemData` and `MemOffset` aren't worth it. This is already a very large change due to how many places it has to touch, and in my mind it's a consequence of the change at all that `as` casts are going to be needed when in-memory data is indexed with the offsets coming out of `wasmparser`. While I've no doubt that certain abstractions could cut down on the number of resulting `as` casts I'm not sure if the benefit of such abstractions outweighs their costs (e.g. new primitives to learn).

I can see that for `MemOffset`, which is internal to the parser only, and serves partly as a namespace to attach the methods to. All the methods are readily inlined and the `u64: Add<MemOffset>` impl could just be written with as `as` cast (losing the debug assertion but oh well). I'm ready to do that at some point today.

`InMemData` on the other hand is repeatedly used in the consuming crates (and would be in my downstream consumer, too). I could doc-comment this a bit more to teach how to use it. The [doc comment change](https://github.com/WorldSEnder/wasm-tools/blob/8c251e028d7246c96b22313de0e945c3def61155/crates/wasm-encoder/src/core/code.rs#L90-L104) in `wasm-encoder` should be a good example already. It wraps a slice to the data, it derefs to that slice so you can pass it instead of the data to `BinaryReader::new` or `parse_all` (e.g. [here](https://github.com/bytecodealliance/wasm-tools/pull/2573/changes#diff-6abaa3c2edecec8fe38805ba402e7400e31ed3e1f3d766eebbedec40ada5e110R58-R59) or [here](https://github.com/bytecodealliance/wasm-tools/pull/2573/changes#diff-17b4facb6adbe870345b29ede8cfe6dfe818260fb6b359e0bc6952981899d993R230-R235) and [here](https://github.com/bytecodealliance/wasm-tools/pull/2573/changes#diff-17b4facb6adbe870345b29ede8cfe6dfe818260fb6b359e0bc6952981899d993R354-R355)). It then lets you index into the data with the ranges and offset the parser emits.

One oddity for example is that since `Range<u64>` does not implement iterator, you can't call `len()` on it as before, which was used to calculate how many bytes. The alternative could be a custom range type that does have this method or writing it out everytime. As is, i think it leaves a clear marker at use site that yeah we are indeed talking about wasm that fits into memory.

In any case, conversion was much easier _with_ `InMemData` than writing the same or similar thing out in eight different crates. Even if it's "just" `as` casts as I reader I would always wonder if we are not silently truncating away bits.

WorldSEnder:
In the course of adding docs to `InMemData` I have found that `RelocationEntry` currently reports the relocation_range as a `usize`. Should this be adjusted to either `u32` (limited by the underlying binary representation?) or `u64` (for consistency) instead?

alexcrichton (commented):
After reading this in full again, personally I'd still  subjectively disagree. I understand taht `InMemData` is used a fair bit,  but it's a pretty nontrivial abstraction. I find it much easier to reason about possibly-slightly-more-verbose code using `try_into` conversions or such rather than seeing an `InMemData` abstraction and immediately having to go look up what it's doing and what it looks like under the hood.

At most I think it'd be ok to have a small set of helpers in `wasmparser::*` to do something with `Range<u64>` and get `usize`-like things out, or maybe indexing, but I'd want that to be constrained to maybe at most 2-3 functions if that were done. Otherwise I feel that `InMemData` is just a bit too foundational in the wrong place and it's not carrying its weight for being a foundational abstraction

WorldSEnder:
> Otherwise I feel that `InMemData` is just a bit too foundational in the wrong place and it's not carrying its weight for being a foundational abstraction

Not sure I follow the "foundational" aspect. To me, the wrapper struct is as much utility as the proposed "2-3 functions". It doesn't cross into parser code, or even other crate boundaries. It's certainly not meant to show up in any public API except as an optional utility to ease the conversion to u64 ranges. A consuming crate can choose to wrap the byte slice it passes to the parsing functions so that it itself can index into that slice with the ranges coming back the parser. The fact that a crate did that wrapping should be a transparent implementation detail, and the wrapper behaves as best it can as that byte slice, except for the additional indexing access.

But alas, I will push the indexing into the consuming crates and duplicate the small bits of code tomorrow.

alexcrichton:
The foundational part I find odd is that every crate in this workspace is migrated to use it, while simultaneously nothing in `wasmparser` requires it. That puts the abstraction in this weird spot where it's supposed to be used but only if you're "in the know". I'm not actually sure if utility functions make sense over just using `as` casts myself, it's mostly just an idea to help reduce duplication if you'd like.

WorldSEnder:
> it's supposed to be used but only if you're "in the know".

I agree on the doc issue and that it will be a bit hard to discover. If you have a better spot where to put it (be it structs or helper methods) feel free to suggest.

The reason for using it in almost every crate in the workspace is simply that they all end up loading the full wasm into memory even if the parser rejects this model, some could reasonably be rewritten into the streaming mode without holding all of the wasm in memory;

- (1) `wasm-metadata` rewrites section by section as far as I can tell and shouldn't need to call `parse_all`. Currently it does and slices to emit some sections byte-by-byte from the input.
- (2) `wit-component` similar to `wasm-metadata` to copy whole sections (and to skip some bytes at the start).
- (3) `demangle` similar to `wasm-metadata` to copy whole sections from the input (and operate on the name custom sections).
- (4) `strip` similar to `wasm-metadata` to copy whole sections that don't get stripped
- (5) `wasm-mutate` reparses the code section to replace opcodes and also copies raw
- (6) `dump` to print bytes literally from the input
- (7) `wasmprinter` uses only the helper `InMemData::range_len` to skip the section in its input
- ~~(8) `wast` similarly uses only `InMemData::translate_offset` in a test case~~
- ~~(9) `validate` to check that the ranges are in bound. This also seems to needlessly be done in `usize` instead of `u64` since the checks boil down to checking `end < len`. Is there even a point where the parser emits a potentially invalid range?~~

At least (1)-(4) serve the same purpose to copy over whole sections. (5) and (6) do some work on the bytes instead of just reencoding them raw. (7) uses the range to avoid parsing the unchecked range of component sections and skip some bytes in the input. ~~(8) and (9) are mostly reuse for reuse sake.~~

---

Perhaps it would be easier to discover and use if the "address translation" into the parsed wasm bytes was returned from the parse function(s) instead of being passed in as a wrapped byte slice - as part of `Chunk::Parsed`? Imo, at least the first two uses deserve _some_ support by the parser library to find the relevant bytes in the input.

WorldSEnder:
Last try to get some utility into the parser. `Chunk` now contains an extra field with a helper struct to convert offsets (using the offset from the start of that parse call). `parse_all` does not return this extra `OffsetConverter` because I expect that most uses are at offset 0 when they call `parse_all`.

The extra field should make the helper discoverable when potentially needed (during incremental parsing). When parsing everything at once, the data is in memory anyway, and the `as` cast is simple enough to reason about.

This leaves two places that use `OffsetConverter::from_start(0)`. One, in a doc comment, which should give it slightly more visibility. Second, [in `wasm-encoder`](https://github.com/bytecodealliance/wasm-tools/pull/2573/changes#diff-4fa630ff71b90900a18ad95294b15a8cf1db2027ffdc3cea7586ebaf3f80e51bR440) which is connected to [these new lines](https://github.com/bytecodealliance/wasm-tools/pull/2573/changes#diff-4fa630ff71b90900a18ad95294b15a8cf1db2027ffdc3cea7586ebaf3f80e51bR80-R81). Since these are `pub` trait functions, but undocumented, I'm not sure if these count as private API and can be just modified to pass through the offset. Anyway, this asserts against a bug when passing a `Parser` that is not positioned at offset `0`, in which case just `as`-casting the offsets would be wrong.

WorldSEnder:
Tried out the extra API but don't think it's worth the hazzle. I still don't like the `as` casts, but better than "low overhead" additional helper structs.

crates/wasmparser/src/binary_reader.rs
```diff
@@ -316,17 +322,13 @@ impl<'a> BinaryReader<'a> {
     /// If `BinaryReader` has no bytes remaining.
     #[inline]
     pub fn read_u8(&mut self) -> Result<u8> {
-        let b = match self.buffer.get(self.position) {
-            Some(b) => *b,
-            None => return Err(self.eof_err()),
-        };
-        self.position += 1;
+        let [b] = self.read_bytes(1)?.try_into().unwrap();
```
alexcrichton: This is a relatively hot method and I'd be somewhat surprised if this performed the same as before. This also looks like it's perhaps derivative of previous changes and reflects changes/refactors that are no longer necessary? I think there's more refactors in `wasm-mutate` for example which are now unrelated to this change. Could the refactors unrelated to `usize` -> `u64` be backed out and saved for a separate PR?

WorldSEnder: For posteriority, the compiler is not smart enough to see a postcondition of `let end = self.ensure_has_bytes(size)?;` namely

`unsafe { std::hint::assert_unchecked(self.position <= end && end <= self.buffer.len()) }`

and inserts extra slice_index_fail calls into the assembly. The unsafe hint would similarly get rid of these error paths in `read_u32`, etc see also [on godbolt](https://godbolt.org/z/d3dqejdsr).

crates/wasm-encoder/src/reencode/component.rs
```diff
@@ -77,6 +77,8 @@ pub trait ReencodeComponent: Reencode {
         parser: wasmparser::Parser,
         data: &[u8],
     ) -> Result<(), Error<Self::Error>> {
+        // so we can slice into the data with the offsets from the parser
+        assert_eq!(parser.offset(), 0, "data must be parsed at offset 0");
```
WorldSEnder, line 81: We could instead pass the parser offset down into `parse_component`. I would like your guidance on this, as it's not clear which part is public API and where we assume "correct" data.

WorldSEnder:
Let me know what to do about the failing clippy warnings. Got a local patch ready, but would rather rebase/merge on some commit from master.

crates/wasmparser/src/offsets.rs
```diff
@@ -0,0 +1,55 @@
...
+//   in some error paths, where saving the position directly would clutter registers.
+
+/// An memory offset into some chunk of bytes occurs at some specified logical
+/// offset in the file. We currently use `usize` to represent memory offsets.
+///
+/// This offset can always be added onto the logical offset without overflow.
+/// Compute the maximum allowable memory offset under both contraints
+// TODO: on platforms where usize::BITS > u64::BITS (currently almost no-where),
+// we could use u64 directly instead of usize to represent memory offsets.
+pub fn max_memory_offset(mut max_logical: u64, max: usize) -> usize {
```
alexcrichton: In the end this didn't end up being used anywhere outside of `BinaryReader`'s constructors, right? I also think it's a bit too surprising for `BinaryReader` to get a sort of truncated view of the buffer silently. If this is indeed private to this crate, then I think it would be best to panic when `max + max_logical` exceeds `u64`'s limit. That'd be a new documented panic condition of `BinaryReader` constructors, but I think that's fine.

alexcrichton (commented):
Looks like the clippy warnings have been resolved and the fuzz-related error looks like just a normal "needs some elbow grease" sort of error. Let me know though if there's some other issue you're running into.

alexcrichton:
Er, actually that's just CI broken on `main`, no need to worry about that -- I'll take care of it.

crates/wasmparser/src/offsets.rs
WorldSEnder: The silent truncation mirrors the truncation in the parser that also sets `eof` to true when it reaches a large enough offset and also silently truncates. Perhaps documenting the behaviour is enough?

alexcrichton: Oh dear that might be inherited by accident from some time ago, personally I think that should either be an error or panic as well.

WorldSEnder: If I understand it correctly, it serves a secondary purpose in the parser. With `component-model` the inner parser stops parsing at the end of its section even when it gets fed bytes past that. You mean that the top-level parser should error instead? That should be simple enough to adjust.

alexcrichton: I forget how exactly this integrates with component parsing, but `BinaryReader` shouldn't silently truncate and to the extent that this is used within `Parser` that also shouldn't silently truncate

WorldSEnder:
The CI error now is caused by a failed fetch in a setup step and unrelated.

alexcrichton (approved):
Looks all good to me, thanks again for working on this!

PR #2576, opened 2026-07-20
Fix generating deeply nested WIT types in `wit-smith`

alexcrichton:
Fixes a fuzz bug fallout from #2565


crates/wit-smith/src/generate.rs
```diff
@@ -1142,6 +1143,13 @@ impl<'a> InterfaceGenerator<'a> {
             ErrorContext,
         }
 
+        // Bail out with a leaf type if a nested type here would nest too
+        // deeply.
+        if depth + 1 >= 50 {
```
dicej: Nit: consider turning the `50` into a named constant.

dicej (approved):

PR #2577, opened 2026-07-23
wasm-tools mutate: validate the input module before mutating

### LINKED ISSUE #2568: wasm-mutate panics when mutating invalid modules
This is something we should probably return an error for or document explicitly, but for example:

```
$ wasm-tools mutate --seed 0 --reduce <<< "(module (global i32 (global.get 5)))" 

thread 'main' (1242810) panicked at crates/wasm-mutate/src/mutators/modify_const_exprs.rs:99:77:
index out of bounds: the len is 1 but the index is 5
zsh: IOT instruction (core dumped)  wasm-tools mutate --seed 0 --reduce <<< "(module (global i32 (global.get 5)))
```

or 

```
$ wasm-tools mutate --seed 1 <<< "(module (func global.get 99))"

thread 'main' (1243187) panicked at crates/wasm-mutate/src/mutators/peephole/eggsy/analysis.rs:142:55:
index out of bounds: the len is 0 but the index is 99
zsh: IOT instruction (core dumped)  wasm-tools mutate --seed 1 <<< "(module (func global.get 99))"
```

(etc)

SAY-5:
The mutators index into type/global/etc. spaces and panic on invalid modules, e.g. `wasm-tools mutate --seed 0 --reduce <<< "(module (global i32 (global.get 5)))"` dies with an index out of bounds in modify_const_exprs. Validate the input up front and report it via the already-documented exit code 2 instead. Closes #2568.


fitzgen (commented):
I would prefer to instead add bounds checks to the mutation that fails, so that we don't need to run the full validator over all the Wasm's opcodes and everything just to e.g. mutate a global.

fitzgen (approved):
Thanks!

PR #2578, opened 2026-07-24
Fix folded branch hints

vouillon:
This fixes the parsing and printing of branch hints, so that the hint is attached to the `if` and not to the `local.get` in the example below (from the [test suite](https://github.com/WebAssembly/branch-hinting/blob/23ce0bb32bbb690ce08877154e3879af060fb1d4/test/custom/metadata.code.branch_hint/branch_hint.wast#L31-L35)).
```wat
        (@metadata.code.branch_hint "\01")
        (if (result i32) (local.get 1)
          (then (call $dummy) (i32.const 9))
          (else (call $dummy) (i32.const 10))
        )
```
The form previously emitted (with the branch hint right before the `then`) is still accepted on parsing.
```wat
        (if (result i32) (local.get 1)
          (@metadata.code.branch_hint "\01")
          (then (call $dummy) (i32.const 9))
          (else (call $dummy) (i32.const 10))
        )
```


fitzgen (approved):
Seems reasonable to me but I think @alexcrichton should probably take a look, since I've really only done minor work on the `wast` crate

keithw:
Thank you for fixing this! Now that branch-hints is merged and there's a normative spec, I would be in favor making the `wast` or `wasmparser` crate stricter to match the reference interpreter, and be able to catch printing errors automatically in the future. This PR is being generous in having `wast` continue to support the previously printed (incorrect) syntax, but I would tend to think that's not necessary. E.g., the reference interpreter rejects all three of these:

```wat
(func
 (@metadata.code.branch_hint "\00")
)
```

```
$ ./wasm -u -ca -cr /tmp/f1.wat 
/tmp/f1.wat:2.2-2.29: custom annotation syntax error: @metadata.code.branch_hint annotation: invalid placement
```

```wat
(func
 (@metadata.code.branch_hint "\00")
 nop
)
```

```
$ ./wasm -ca -cr /tmp/f2.wat
/tmp/f2.wat:2.2-2.35: custom validation error: @metadata.code.branch_hint annotation: invalid target
```

```wat
(func
 (if (i32.const 0) (@metadata.code.branch_hint "\00")
  (then))
)
```

```
$ ./wasm -ca -cr /tmp/f3.wat
/tmp/f3.wat:5.7-5.34: custom annotation syntax error: @metadata.code.branch_hint annotation: invalid placement
```

Currently (pre and post this PR) wasm-tools accepts these as well-formed and valid. I think it probably should reject them too... and this will make it easier to catch future issues when it does the roundtrip tests.

alexcrichton:
Thanks! I agree with @keithw that there's no need to keep parsing the old syntax and it's ok to just break things here. I think there's code in `wast` trying to parse the older syntax and that can all be deleted.

I'll note though that @keithw syntax like:

```wasm
(func
 (@metadata.code.branch_hint "\00")
 nop
)
```

will be relatively difficult to reject I think as that requires running the validator to determine that it's not actually tied to an `if` or `br_if`.

vouillon:
I have updated the PR to reject misplaced annotations.
I'm not sure what to do about branch hints on non-branch instructions. Maybe it make more sense to handle that during validation. The test suite has [a test which asserts that this is a validation error](https://github.com/WebAssembly/branch-hinting/blob/23ce0bb32bbb690ce08877154e3879af060fb1d4/test/custom/metadata.code.branch_hint/branch_hint.wast#L93-L107); not a parsing error..

alexcrichton (approved):

PR #2581, opened 2026-07-28
support cooperative multithreading ABI in `wit_component::linking`

dicej:
- Handle `__init_stack_pointer` global variable imports
- Update `EncodingState::create_export_task_initialization_wrappers` to look for component exports in both the main module _and_ library modules
- Ensure `__wasm_init_{async_}task` is re-exported from the main module if it is exported from a library so that `EncodingState::create_export_task_initialization_wrappers` can find and use it
- Call `__wasm_init_task` if present before calling `__wasm_call_ctors` or `_initialize`

I've tested this using WASI-SDK's main branch, slightly modified to allow cooperative multithreading in shared libraries.


crates/wit-component/tests/components/link-asyncify/component.wat
```diff
@@ -3,27 +3,29 @@
...
-    (export "__asyncify_data" (global 2))
-    (export "__stack_high" (global 3))
-    (export "__stack_low" (global 4))
-    (export "bar:memory_base" (global 5))
-    (export "bar:table_base" (global 6))
-    (export "foo:memory_base" (global 7))
-    (export "foo:table_base" (global 8))
-    (export "__heap_base" (global 9))
-    (export "__heap_end" (global 10))
+    (export "__init_stack_pointer" (global 1))
```
alexcrichton: A bit orthogonal to this PR, but it'd be nice to make the synthesis of globals like this optional and disabled if nothing actually imported it

crates/wit-component/tests/components/link-wasip3-abi/component.wat
```diff
@@ -0,0 +1,335 @@
...
+      i32.add
+      global.set 2
+      global.get 3
+      global.get 4
+      i32.add
+      global.set 5
+      call 0
+      call 2
+      call 1
+      call 3
```
alexcrichton: I'm a bit confused by this -- the first two functions are are `__wasm_apply_data_relocs` and the latter two functiosna re `__wasm_call_ctors`. Shouldn't there be an invocation of `__wasm_init_task` in the middle?

alexcrichton:
Oh, also, there's a lot of hardcoding of various strings here, do you think it would be reasonable to consolidate things in fewer locations?

crates/wit-component/tests/components/link-wasip3-abi/component.wat
dicej: Oof, good catch.  Yes, the logic was buggy; it should have been looking for a `__wasm_init_task` export across _all_ libraries, not just the one exporting `__wasm_call_ctors`.  I just pushed a fix.

crates/wit-component/tests/components/link-asyncify/component.wat
dicej: Done!

dicej:
> Oh, also, there's a lot of hardcoding of various strings here, do you think it would be reasonable to consolidate things in fewer locations?

I just pushed an update.  Let me know if you think more needs to be done.

alexcrichton (commented):
Technically the string literals at least for `_initialize` and `__wasm_init_task` and such are duplicated throughout `encoding.rs` and `validation.rs` which is what I was primarily worried about, but don't feel the need to take those on here.

alexcrichton (approved):

alexcrichton:
(CI failure due to MSRV as well)

PR #2590, opened 2026-07-31
Implement new handling of TLS base/symbols

alexcrichton:
This commit is the wasm-tools half of the implementation of WebAssembly/wasi-libc#857. More information can be found in the writeup of that issue itself, but the general goal here is adjusting linking conventions for wasip3 modules, in particular around dynamic libraries, to ensure that TLS works everywhere.


crates/wit-component/src/linking/metadata.rs
```diff
@@ -424,6 +479,22 @@ impl<'a> Metadata<'a> {
...
                                 }
                             }
+                            (
+                                self::ENV,
+                                name @ (self::GET_STACK_POINTER
+                                | self::SET_STACK_POINTER
+                                | self::GET_TLS_BASE
+                                | self::SET_TLS_BASE),
+                            ) => {
+                                if !matches!(import.ty, TypeRef::Func(_)) {
```
dicej, line 489: Nit: checking the precise types here (e.g. `__wasm_get_stack_pointer` should take no parameters and return a single `i32`) would help catch problems earlier.  Likewise for `[thread-new-indirect-v0]` below.

dicej (approved):
Thanks for doing this!

crates/wit-component/src/linking/metadata.rs
alexcrichton: Makes sense yeah, although there's not much other type-checking in this file from what I can tell for function params/results, only for major items like "is this a global or a func".  I'm going to go ahead and land this for now, but I'll open a follow-up issue for this.

PR #2593, opened 2026-08-10
Enable wide-arithmetic/compact-imports by default

alexcrichton:
These are both now phase 4 proposals as voted last week by the CG and thus meet the threshold for enabled-by-default in this repository.


pchickey (approved):
Congratulations, Alex. Great work on the wide arithmetic proposal

PR #2594, opened 2026-08-11
add `map` support to `wit-dylib`

dicej:
Pushing and popping maps and their elements follows the convention established for lists, except that each item involves two values instead of one, and there's no "fast path" for pushing or popping directly from or to the canonical ABI format.


alexcrichton (approved):

dicej:
Oof, I just noticed all the CI failures.  I'll fix those when I get a chance.

PR #2600, opened 2026-08-12
Consolidate generated modules in `wit-component` into one

alexcrichton:
To fully address some upcoming refactorings for wasip3 intrinsics this commit refactors the output of `wit-component` to ensure that all the little helper modules generated are glommed into one instead of lots of separate ones. The main purpose of this is to address the use case now addressed here where task intrinsics need to hook resource destructors which previously wasn't handled, nor was it easily possible.

This is mostly a lot of refactoring and attempting to move some complexity of `encoding.rs` to a separate file. The end result is that these modules are now all folded into one module:

* The original fixup module which initializes the table of the shim module, for example. This is used when lowerings can't happen until after the main module is instantiated.
* The shim module used to have a `start` function for the `_initialize` hook is now folded into the fixup module.
* The `__init` module used by `wasm-tools component link` is now located in the main fixup module.
* The task initialization hooks module is now part of the fixup module and now handles the use case of resource destructors as well.

This is all intended to be refactoring such that the next commit, adding more task lifetime hooks to get added, will be much easier to do.

> ~~**Note**: this is built on https://github.com/bytecodealliance/wasm-tools/pull/2599~~


alexcrichton:
I'm going to switch this over to @dicej since I'd like to get your eyes in particular on the dynamic library changes here

crates/wit-component/src/encoding/fixup.rs
```diff
@@ -0,0 +1,709 @@
...
+        export: u32,
+        /// The number of parameters to forward to `export`.
+        params: usize,
+        /// The debug-related core name of this function it's hooking.
+        core_name: String,
+    },
+
+    /// Similar to `HookedCoreExport` except that this is specifically for
+    /// resource destructors which don't need to be reexported.
+    HookedResuorceDtor { before: u32, export: u32 },
```
dicej:
```suggestion
    HookedResourceDtor { before: u32, export: u32 },
```

dicej (approved):

PR #2605, opened 2026-08-14
Use `__wasm_task_hook` for `realloc` options

alexcrichton:
This commit updates the `__wasm_task_hook` intrinsic, added in #2603, to additionally get called for the `realloc` function, typically exported as `cabi_realloc` today. The goal of this commit is to uniformly use this hook for wasip3 task initialization/configuration as opposed to the current state of affairs in wasi-libc where `cabi_realloc` is special.

The implementation here is more involved than #2603 `realloc` options are required to create the fixup module and thus come from the original instance raw. When using task hooks, however, the desired state is that the lowerings created use the hooked version of `realloc`. As with all problems in computer science, this is solved with another layer of indirection.

Specifically the shim module may not contain shims for `realloc` options if used for lowerings. These are then hooked up through the fixup module so it's instantiated with both lowered functions and the "real" implementation so everything can get wired up at the same time.

While implementing this I went ahead an implemented a minor optimization where for modules that import linear memory they no longer need an shim-per-import-using-realloc and instead just have a single shim for realloc. This means that when importing memory generated components generally have even fewer shims than before.

This commit then attempts to add a variety of tests for various shapes of this new hook and various other options. This was developed in tandem with changes to wasi-libc and additionally verified against those.


crates/wit-component/src/encoding/world.rs
```diff
@@ -536,17 +536,20 @@ fn options_are_available_before_instantiation(
     options: RequiredOptions,
     info: &ValidatedModule,
 ) -> bool {
-    // Currently realloc always comes from the main module.
-    if options.contains(RequiredOptions::REALLOC) {
-        return false;
+    // If neither memory nor realloc are needed, so this can always be lowered
```
dicej:
```suggestion
    // If neither memory nor realloc are needed, this can always be lowered
```

dicej (approved):

PR #2607, opened 2026-08-15
Add support for package-scope WIT types

yordis:
## Summary

Adds support for declaring `record`, `variant`, `enum`, `flags`, and `type`
aliases directly at package scope in WIT, and encodes/decodes them following the
approved design in WebAssembly/component-model#699.

A package-scope type is encoded like a `use` of an `interface`, just without the
wrapping `instance`: the definition is restated locally and bound with an `eq`
import named by its fully-qualified `namespace:package/name`. Because the type
is structural, the `eq` bound means nothing has to be supplied to satisfy the
import. A referencing `interface` or `world` aliases that import into its
instance/component type; a `world` places the import inside its inner exported
component-type.

This replaces the earlier approach of inlining a package-scope type into every
interface/world instance and re-exporting it, which did not pass Component Model
validation.

## Encoding

Given:

```wit
package local:demo;

record point { x: u32, y: u32 }

interface api {
  move-to: func(p: point);
}
```

the package-scope `point` is exported once at the top level and imported with
`eq` into the wrapping component-type:

```wat
(component
  (type (;0;) (record (field "x" u32) (field "y" u32)))
  (export (;1;) "point" (type 0))
  (type (;2;) (component
    (type (;0;) (record (field "x" u32) (field "y" u32)))
    (import "local:demo/point" (type (;1;) (eq 0)))
    (type (;2;) (instance
      (alias outer 1 1 (type (;0;)))
      (type (;1;) (func (param "p" 0)))
      (export (;0;) "move-to" (func (type 1)))))
    (export (;0;) "local:demo/api" (instance (type 2)))))
  (export (;3;) "api" (type 2)))
```

A package-scope type that depends on one from another package puts the `import`
on the package's own component, since there is no wrapping component-type to
hold it. Decoding recognizes qualified type imports (both nested and at the
package root) as package-scope dependencies and keeps ownership with the
defining package; the previous inline-coalescing pass in the decoder is removed.
The printer emits the corresponding top-level `use` before the definitions that
reference it.

## Types-only packages

Decoding recovers the package name from the fully-qualified export inside an
interface's or world's component type, so a package with only package-scope
types would otherwise have no package name in the binary at all. For that case
the encoding is generalized (as discussed with Luke): each type is additionally
bound with a self-referential `eq` import named by its fully-qualified name,
which is where the package name is recovered from on decode:

```wat
(component
  (type (;0;) (record (field "x" u32) (field "y" u32)))
  (export (;1;) "point" (type 0))
  (import "local:demo/point" (type (;2;) (eq 1)))
  (type (;3;) (list 2))
  (export (;4;) "path" (type 3))
  (import "local:demo/path" (type (;5;) (eq 4))))
```

Later types reference the imported index so every import satisfies the
component model's named-type validation rules, and the import is placed after
the export so V2 format sniffing still sees a label first. Packages that do
have an interface or world encode byte-identically to the approved WIT.md
examples; the self-imports may be worth a small follow-up note in the spec's
Package Format section.

## Tests

- Golden interface fixtures mirroring the approved WIT.md examples
  (`package-scope-spec-interface`, `package-scope-spec-world`), a types-only
  fixture (`package-scope-types-only`), plus the existing package-scope
  fixtures re-blessed to the `eq`-import encoding. All round-trip through
  encode → validate → decode → print → re-encode.
- Combination fixtures covering package-scope types mixed with classic
  `use iface.{...}` projections on the same wrapping component-type:
  `package-scope-with-iface-use`, `package-scope-with-world-use`,
  `package-scope-with-resource-use`, and `package-scope-mixed-use` (foreign
  package-scope `use` plus a local interface projection).
- Unit tests covering: foreign `use` of a package-scope type, a package-scope
  type defined in terms of a foreign one (import on the package component),
  versioned fully-qualified names, the export-before-use ordering for a type
  referring to another, types-only round-trips (with and without a foreign
  dependency), rejection of a `resource` at package scope, and that building a
  component (`component new`) exports the interface directly without inventing
  a types interface (WebAssembly/component-model#694).

## References

- https://github.com/WebAssembly/component-model/issues/694
- https://github.com/WebAssembly/component-model/pull/699


crates/wit-parser/src/decoding.rs
```diff
@@ -278,7 +322,14 @@ impl ComponentInfo {
                 docs: Docs::default(),
                 interfaces,
                 worlds,
+                types,
             }
+        } else if !types.is_empty() {
+            bail!(
```
yordis, line 344: This is odd; so waiting for Luke's direction here

Issue #2338, opened 2025-10-07
Missing tests for some `component new` flags


catamorphism:
I can't find any tests that exercise `component new` with the `--realloc-via-memory-grow` or `--merge-imports-based-on-semver` flags.

Context: I'm trying to write examples in the help text that illustrate various flags. My first step was to look for tests to see how they should be used.

alexcrichton:
I don't think that there are dedicated tests for these flags right now unfortunately. They are used to control the default behavior of how a component is created where, for example, by default `cabi_realloc` is used to allocate memory but `--realloc-via-memory-grow` says "don't do that and instead use `memory.grow`". Similarly for `--merge-imports-based-on-semver` that turns off on-by-default behavior. These were added over time to tweak the default heuristics which is why the aren't super well tested.

The `tests/cli` directory has some files with `--merge-world-imports-based-on-semver` but none with `--merge-world-imports-based-on-semver=false`. The tests at `crates/wit-component/tests/components` test some various behaviors of reallocation but in general don't have a dedicated test for `--realloc-via-memory-grow` I believe.

PR #2611, opened 2026-08-19
Restore the default for `component new --merge-imports-based-on-semver`

Lstarsky0:
I started on #2338 — writing the missing tests for `component new`'s `--realloc-via-memory-grow` and `--merge-imports-based-on-semver` — and the second one turned up a behaviour change rather than just a coverage gap.

`merge_imports_based_on_semver` is the only `Option<bool>` field on `ComponentEncoder`, and it reads as `unwrap_or(true)`. Before #2597 the CLI only called the setter when the flag was present, so leaving it off inherited that `true`. Now it goes through `optional_flag_with_default(..., false)`, which calls the setter with `false` on every run. The flag's own doc comment, and the help output in `tests/cli/help-component-new.wat.stdout`, both still say "This is enabled by default".

#2597 landed after 1.256.0, so 1.257.0 is the first release where `wasm-tools component new` stops merging semver-compatible imports. Two imports of `a:b/c@0.1.0` and `a:b/c@0.1.1` come out as `a:b/c@0.1.0` on `main` today and as `a:b/c@0.1.1` with `--merge-imports-based-on-semver=true`.

I read this as an accident of the refactor rather than an intended change, since the encoder default, the doc comment and the help snapshot all still describe the old behaviour — but you'd know better than me, and if the new default is what you want then the doc comment and the help snapshot are what should move instead.

The one-line change is the default in `ComponentEncoderOpts::configure`. The rest is #2338:

`tests/cli/component-new-merge-imports-based-on-semver.wit` runs the same input three ways. `default` and `merge` produce byte-identical output and `no-merge` differs from both, so the test pins the default rather than just exercising the flag — putting `false` back makes exactly this one test fail out of 825.

`crates/wit-component/tests/components/adapt-realloc-via-memory-grow/` covers the other flag. It needed a new `realloc-via-memory-grow` knob on the components harness, in the same shape as `return-call-ref`. The flag only reaches `gc::run` on the adapter path, which is why I copied `adapt-inject-stack-with-realloc` rather than writing a fresh module — without an adapter the flag has no observable effect at all, which is probably why it never got a test. Against that baseline the encoded component drops the `__main_module__` `cabi_realloc` import and gains a synthesized `$realloc_via_memory_grow` instead.

`cargo fmt --check`, the `components` suite (143) and `tests/cli` (825) are clean with `RUSTFLAGS="-D warnings"`. Clippy is clean too, but I had to add `--exclude artifacts --exclude wit-dylib` locally since I don't have a WASI SDK installed; that part is untested on my machine.


alexcrichton (approved):
Thanks for this, and yes that was an unintentional change of mine, thanks for catching that!

PR #2619, opened 2026-08-24
Remove post-return hooks, adjust sync task hooks

alexcrichton:
This commit updates the invocation of `__wasm_task_hook` in `wit-component` for generated components, first intended to be used with wasip3. Previously hooks were added for all known locations, including post-return, but this turned out to be incompatible with how the stack is managed in `wit-dylib`. The preexisting hooks are also a minor source of overhead where a sync task with post return gets 4 hook points, 2 of which are basically unnecessary. The change in this commit is to avoid post-return-specific hooks entirely and instead move the "sync task finished" hook to the end of the post-return function if one is specified. This makes things slightly more efficient (fewer hook points), and is additionally compatible with what wit-dylib does.

### DIFF
--- a/crates/wit-component/src/encoding/fixup.rs
+++ b/crates/wit-component/src/encoding/fixup.rs
@@ -137,7 +137,7 @@ enum DefinedFunction {
         /// get inserted into `EncodingState`'s export wrapper map.
         export: Option<ImportedInstance>,
         /// The kind of hook to use at the start of this function.
-        start: TaskHook,
+        start: Option<TaskHook>,
         /// The kind of hook to use at the end of this function.
         end: EndTaskHook,
     },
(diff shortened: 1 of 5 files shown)


crates/wit-component/src/encoding/fixup.rs
```diff
@@ -167,17 +167,17 @@ pub enum TaskHook {
     ResourceDtorStart = 8,
     /// A resource destructor is finished.
     ResourceDtorFinish = 9,
-    /// A call to post-return is starting.
-    PostReturnStart = 10,
-    /// A call to post-return is finished.
-    PostReturnFinish = 11,
+    //
+    // 11/12 intentionally missing
```
dicej:
```suggestion
    // 10 and 11 intentionally missing
```

dicej (approved):

PR #2621, opened 2026-08-24
Refactor name resolution of components in the text format

alexcrichton:
This commit is an update to how name resolution happens for components during the text-to-binary transform. The main purpose of this commit is to enable running the upstream component model test suite in this repository as a submodule, and minor issues in the text format are preventing that. No valid component should change as a result of this reimplementation, but invalid components will now have their errors reported at a slightly different time.

Name references in the text format are generally expected to be possible to do in a forwards and backwards way in the text format. For example in core wasm you can do `call $foo` before actually defining function `$foo`. This works because the resolver will first register all names defined and then actually resolve references to those names. Components have historically not worked this way, however, because unlike core modules components cannot refer to names defined later in a component. This influenced the resolver algorithm in the text-to-binary transform in the `wast` crate where it currently resolves a field and then registers the name of the field. This was done originally because the process of resolving a field may inject aliases into a component due to shorthands possible in the text format. This injection requires name resolution information and cannot happen in an earlier expansion phase for outer aliases as well. This meant that if a `$foo` identifier was referring to an item defined later this turned out to be a text-parse-time error.

With a text-parse-time error this clashes with `wasmprinter`'s default printing of components unfortunately. For `wasmprinter` it loads the name section and then prints identifiers for all kinds of references since all names are known up front. This means that for an invalid component `wasmprinter` will produce unparsable text by `wast`. This friction makes working with invalid components difficult and notably also makes it so `assert_invalid` cannot be used for some components that are indeed binary-parsable-but-invalid.

That all leads to the change in this commit. The main refactor here is to how a group of fields are processed. This affects components, nested components, instance types, module types, etc. The goal is to behave more like core wasm in terms of name resolution even if the output component would be invalid (it's not `wast`'s job to validate). Notably the previous resolution is now replaced with a few more phases. Specifically:

* First all names in a group of fields are collected. Indices are assigned here but are known to be invalid because aliases haven't yet been injected.
* All fields are resolved to expand their implicit aliases. This will inject export/outer aliases as necessary into the list of items.
* This temporary resolution state is now discarded now that aliases are fully expanded.
* Next resolution proceeds like core wasm again, but this time indices are "for real". All fields are registered with an index and then resolved internally.

This behavior change means that components now behave more like core wasm where the text format allows for forward-references. This notably reduces complexity when handling `rec` groups, for example, which already needed special-casing since they support forward references.

The only affect on tests here is that some `assert_malformed` tests are, correctly, becoming `assert_invalid`. Otherwise test output isn't changing and everything that parsed yesterday should still parse today. No changes are made to validation of components here, this only affects text parsing.


fitzgen (approved):
LGTM

PR #2622, opened 2026-08-25
Implement canonical interface names in wit-component

chenyan2002:
Follow-up to https://github.com/bytecodealliance/wasm-tools/issues/2556 and replaces https://github.com/bytecodealliance/wasm-tools/pull/2602.

* Add config flag `emit_canonical_names` in `ComponentEncoder` to emit canonical version.
* When both implement and versionsuffix are present, the versionsuffix should refer to the version from implement, instead of the main package name.
* Fixed a bug in `merge_world_imports_based_on_semver` that `implements` version is not updated during the merge.
* Add CLI flag `--emit-canonical-names` to `wasm-tools component new`, `embed` and `wit`.

During the transition period, upstream libraries, e.g., wac, can decode the original binary and re-encode with the `emit_canonical_names` flag enabled, so that we convert all binaries into their canonical encoding as a preprocessing step. Then `wac` can merge the interface via string matching directly.


crates/wit-component/src/encoding/wit.rs
```diff
@@ -197,11 +244,15 @@ impl Encoder<'_> {
...
+            let name = if self.canonical_names {
+                self.resolve.canon_id_of(interface).unwrap()
+            } else {
+                self.resolve.id_of(interface).unwrap()
+            };
             if interface == id {
                 let idx = encoder.encode_instance(interface)?;
                 log::trace!("exporting self as {idx}");
-                encoder.outer.export(name, ComponentTypeRef::Instance(idx));
+                encoder.outer.export(&name, ComponentTypeRef::Instance(idx));
```
alexcrichton: This seems like it might be wrong and/or untested -- if `name` is a canonical name shouldn't this be a `ComponentExternName` to carry the versionsuffix payload?

src/bin/wasm-tools/component.rs
```diff
@@ -208,9 +208,24 @@ struct ComponentEncoderOpts {
...
 
+    /// Merges imports based on canonical version prefixes and emits canonical
+    /// interface names with version suffixes.
+    ///
+    /// When enabled, import/export names use canonical version prefixes (e.g.,
+    /// `wasi:cli/exit@0.2` instead of `wasi:cli/exit@0.2.1`) and the
+    /// `version_suffix` field is populated in the binary. This also forces
+    /// merging of imports that share the same canonical version prefix.
+    #[clap(long, conflicts_with = "merge_imports_based_on_semver")]
+    merge_imports_based_on_canonical_version: bool,
```
alexcrichton: Instead of modeling this as a new `--merge-...` flag, could this be modeled as a `--canonical-names` flag? That feels to me a bit more accurate to what's happening here and while `--canonical-names` would indeed imply `--merge-imports-based-on-semver` it seems to me like it would help plumbing this internally because how something is encoded is somewhat orthogonal. It's mostly just that `--merge-imports-based-on-semver=false --canonical-names` should be an error in theory

src/bin/wasm-tools/component.rs
```diff
@@ -908,10 +928,30 @@ pub struct WitOpts {
...
+        long,
+        conflicts_with = "importize",
+        conflicts_with = "importize_world",
+        conflicts_with = "exportize",
+        conflicts_with = "exportize_world",
+        conflicts_with = "generate_nominal_type_ids",
+        conflicts_with = "merge_world_imports_based_on_semver",
+        value_name = "WORLD"
+    )]
+    merge_world_imports_based_on_canonical_version: Option<String>,
```
alexcrichton: To expand a bit on my comment from above, this is where `--canonical-names` feels more appropriate as a flag as opposed to a conflict with all of these other flags. The flags is basically "if you encode to wasm use canonical names in the wasm intead of the default full name"

crates/wit-parser/src/lib.rs
```diff
@@ -308,6 +308,28 @@ impl PackageName {
...
+            version.major.to_string().len()
+        } else if version.minor > 0 {
+            2 + version.minor.to_string().len()
+        } else {
+            4 + version.patch.to_string().len()
+        };
+        let prefix = s[..split_pos].to_string();
+        let suffix = s[split_pos..].to_string();
+        (prefix, suffix)
+    }
```
alexcrichton: I don't think that this is quite right for a few reasons:

* I believe this is assuming all version number are 1 digit long
* This doesn't take into account pre-release versions where there's no split on those.

crates/wit-parser/src/resolve/mod.rs
```diff
@@ -2575,16 +2651,25 @@ impl Resolve {
...
     /// semver-compatible versions are listed the same way.
-    ///
-    /// The second element in the returned tuple is this interface's package's
-    /// version.
-    fn semver_track(&self, id: InterfaceId) -> Option<((PackageName, String), &Version)> {
+    fn semver_track(
+        &self,
+        id: InterfaceId,
+        use_canonical_version: bool,
+    ) -> Option<((String, String), &Version)> {
```
alexcrichton: I'm a bit surprised that this function needs changing because the preexisting version compat logic is intended to be basically the canon-names logic. If it's different then one should change, but in the end this function shouldn't need an extra `bool` parameter

crates/wit-component/src/encoding.rs
```diff
@@ -993,12 +1014,25 @@ impl<'a> EncodingState<'a> {
...
+                (Some(_), _, WorldItem::Interface { id, .. }) => {
+                    implements = resolve.canon_id_of(*id);
+                    resolve.version_suffix_of(*id)
+                }
+                (None, WorldKey::Interface(id), _) => resolve.version_suffix_of(*id),
+                _ => None,
+            }
+        } else {
+            None
+        };
```
alexcrichton: I feel like I've seen this similar block in a few different locations throughout this diff, all pretty similar but also somewhat different too. Could there be a helper for constructing a `ComponentExternName` which takes all of this into account? I'd find it helpful for example to see all the logic in one place for "what are we gonna call this" as opposed to reviewing it spread out amongst a number of locations.

Ideally as well the inputs to the function wouldn't be a string-y name but rather things like `WorldKey` or `InterfaceId`/`&Function` or similar.

crates/wit-parser/src/lib.rs
chenyan2002: I think this is correct. `else if version.minor > 0` means `version.major == 0`, so it's exactly 2 characters, then we add `version.minor.to_string().len()` to the split_pos. I added some unit tests with double-digit versions to make sure.

> pre-release versions where there's no split on those

Yes, I see that code in the original `version_compat_track`, but I think this is not in the spec.

crates/wit-parser/src/resolve/mod.rs
chenyan2002: The main difference is whether to call `PackageName::canon_version_split` or `PackageName::version_compat_track_string`, depending on the extra bool parameter. `version_compat_track_string` returns the original version if `version.pre` is not empty, which is not in the spec. 

If we update `version_compat_track_string`, it will be a breaking change. This is also the reason we add bunch of functions like `canon_id_of`, `canon_id_of_name`, and `name_canon_world_key`, just to call the `canon_version_split` instead of `version_compat_track` for canonical names.

crates/wit-component/src/encoding/wit.rs
chenyan2002: Good catch! Fixed. I will think about how to write a test for this.

crates/wit-component/src/encoding.rs
chenyan2002: Will refactor this. It's mainly to handle the case where both `implements` and `versionsuffix` are present.

src/bin/wasm-tools/component.rs
chenyan2002: Changed the flag to `emit-canonical-names`

crates/wit-parser/src/lib.rs
alexcrichton: Oh right of course for "0". For the pre-handling, if that's not in the spec then I believe the spec is wrong. I'll go review that and see if changes need to be made.

crates/wit-parser/src/resolve/mod.rs
alexcrichton: After the clarifications now-merged in https://github.com/WebAssembly/component-model/pull/711, I think the `use_canonical_version: bool` parameter here may no longer be necessary?

chenyan2002: Yes, it also simplifies other changes in `wit-parser`. The return type may change to `String` instead of `Version`, but that's probably the only thing we need.

crates/wasmparser/src/validator/component.rs
```diff
@@ -4765,6 +4776,16 @@ impl ComponentNameContext {
...
 
+        // When an `implements` is present, validate the `version_suffix`
+        // against the implements interface name rather than the main name.
+        if let Some(implements) = implements {
+            if let ComponentNameKind::Interface(iface) = implements.kind() {
+                if let Err(e) = iface.version(version_suffix) {
+                    bail!(offset, "invalid interface version: {e}");
+                }
+            }
+        }
```
alexcrichton: Could this validation move up to the above method? I think it'd be a bit clearer to perform this validation when `implements` is specified rather than deferring it to happen later here. That'd probably require validating  the `version_suffix` field before `implements` and reorderin the above `if` statements, but that should be fine to do.

crates/wit-component/src/encoding/wit.rs
```diff
@@ -25,7 +25,16 @@ use wit_parser::*;
...
 pub fn encode(resolve: &Resolve, package: PackageId) -> Result<Vec<u8>> {
-    let mut component = encode_component(resolve, package)?;
+    encode_with_options(resolve, package, false)
+}
+
+/// Same as [`encode`] but with an option to emit canonical interface names.
+pub fn encode_with_options(
+    resolve: &Resolve,
+    package: PackageId,
+    canonical_names: bool,
```
alexcrichton: In retrospect reading over this again, I think let's just add the `bool` parameter to the preexisting function. Given the nature of publishing for this crate, which is that each release is a semver-major release, it's ok to change API signatures. And given that I think it makes more sense rather than to start a `*_with_options` convention because if more options are added in the future it'll just end up breaking this signature anyway. Having just one function helps reduce duplication and cognitive overhead too I think.

crates/wit-component/src/encoding.rs
alexcrichton, line 832: Right now this name logic is now split between the caller and the end of this function, but perhaps this parameter could be changed to `ComponentExternName` directly to avoid this duplication and instead have the logic in just one place?

crates/wit-component/src/encoding/wit.rs
```diff
@@ -125,19 +150,31 @@ fn component_extern_name(
...
     item: &WorldItem,
+    canonical_names: bool,
 ) -> wasm_encoder::ComponentExternName<'static> {
-    ComponentExternName {
-        name: resolve.name_world_key(key).into(),
-        implements: resolve.implements_value(key, item).map(|s| s.into()),
-        external_id: resolve.external_id_value(key, item).map(|s| s.into()),
-        version_suffix: None,
+    let implements = resolve.implements_interface(key, item);
+    if canonical_names {
```
alexcrichton, line 142: In a previous review I was curiuos if it would be possible to deduplicate the number of places that a canonical-names option was taken into account and a `ComponentExternName` were created. I count currently four different locations doing very similar things:

1. here
2. below in this file in `for interface in interfaces`
3. in `encode_interface_import` in `encoding.rs`
4. in `encode_interface_export` in `encoding.rs` (split across two functions)

Were you able to take a look and see if these locations could be unified? Is there perhaps one, or maybe two at most, helpers that could be used to construct these names?

crates/wit-component/tests/components/canonical-names/module.wat
```diff
@@ -0,0 +1,10 @@
+;;! emit-canonical-names = true
+
+(module
+  (import "a:b/c@0.1.1" "x" (func (param i32 i32)))
+  (import "a:b/c@0.1.1" "y" (func))
```
alexcrichton, line 7: In addition to importing these two functions could this test additionally import the 0.1.0 version of `a:b/c` to test that version-merging logic too?

crates/wit-parser/src/resolve/mod.rs
```diff
@@ -1550,21 +1550,49 @@ impl Resolve {
...
+        let interface_id = match key {
+            WorldKey::Interface(id) => *id,
+            WorldKey::Name(_) => {
+                if let WorldItem::Interface { id, .. } = item {
+                    *id
+                } else {
+                    return None;
+                }
+            }
+        };
```
alexcrichton, line 1586: I think this could be simplified by `self.implements_interface`?

crates/wit-parser/src/resolve/mod.rs
```diff
@@ -2515,7 +2543,10 @@ impl Resolve {
         for (key, item) in mem::take(&mut self.worlds[world_id].imports) {
             if let WorldItem::Interface { id, .. } = item {
                 if replacements.contains_key(&id) {
-                    continue;
+                    if let WorldKey::Interface(_) = key {
+                        continue;
+                    }
+                    // Keep labeled imports with `implements` version unchanged
```
alexcrichton, line 2543: I'm a bit surprised that this is necessary since it's expected that there'd be a unique `InterfaceId` for each named import. If this is removed, do tests fail, however?

crates/wasmparser/src/validator/component.rs
chenyan2002: Done. Neither way is perfect. We validate `version_suffix` with `ComponentName` here. Ideally, we would also validate the suffix with implements here. But moving up does make a smaller diff.

crates/wit-component/src/encoding/wit.rs
chenyan2002: Done

crates/wit-parser/src/resolve/mod.rs
chenyan2002: Yes, merge-canon-with-implements.wit test would fail:
```
  import a:b/c@0.1.0;
  import a:b/c@0.1.1;
  import my-thing: a:b/c@0.1.0;
```
`import my-thing: a:b/c@0.1.0` gets deleted, because 0.1.0 have been removed by the above imports.

crates/wit-component/tests/components/canonical-names/module.wat
chenyan2002: Done

crates/wit-component/src/encoding/wit.rs
chenyan2002: Will think about it tomorrow.

crates/wasmparser/src/validator/component.rs
```diff
@@ -4682,22 +4694,21 @@ impl ComponentNameContext {
...
-            )?;
-            match ty {
-                ComponentEntityType::Instance(_) => {}
-                _ => bail!(offset, "only instances can have an `versionsuffix`"),
-            }
-        }
+            Some(implements)
+        } else {
+            None
+        };
```
alexcrichton: I think these changes may no longer be necessary

crates/wit-component/tests/linking.rs
alexcrichton, line 1: FWIW the changes here seem fine, but for future PRs I the test here don't really have much to do with the feature being added here so I think it'd be fine to just hardcode this to either false or true

crates/wit-component/tests/interfaces/wasi-http.wat
```diff
@@ -6,7 +6,7 @@
           (export (;0;) "pollable" (type (sub resource)))
         )
       )
-      (import "wasi:io/poll@0.2.0-rc-2023-11-10" (instance (;0;) (type 0)))
+      (import "wasi:io/poll@0.2.0-rc-2023-11-10" (versionsuffix "") (instance (;0;) (type 0)))
```
alexcrichton, line 9: I think we'll probably want to handle this differently with a few layers:

* In the abstract this is a wast of binary-encoding space so we wouldn't want to emit the `versionsuffix`
* Additionally though the component model should probably add a validation rule that the length of `versionsuffix` is >0 since there's not much point in having a 0-length version suffix

Would you be up for adding the new validation rule, adding tests for that, and then adjusting the encoder to skip the suffix if it's empty?

alexcrichton (approved):
Were there more changes you wanted to make as well? I think the main previously-mentioned-and-possibly-outstanding comment of mine is refactoring the construction fo `ComponentExternName`, but if you'd prefer that can be deferred to a future change too.

Otherwise though in reading over the tests again I'm thinking that the component model should probably disallow an empty `versionsuffix` field, which would then require some changes here too. Other than that though I think this is reasonable to land.

Thanks again for working on this!

crates/wit-component/tests/interfaces/wasi-http.wat
chenyan2002: Eventually yes. During the transition period, is it better to always have a `versionsuffix`, so that tools can easily tell if the binary is using the canonical name or not? Or we can put something in the producer section, like `processed-by wit-component --emit-canonical-names`? Happy to implement this either way.

chenyan2002:
Thanks for the review! There may be some follow up PR to fix the decoder side, so that the decoder doesn't need to reconstruct the full version using the `versionsuffix`. At the moment, this seems fine for my use case.

As for refactoring `ComponentExternName`, there are two places that are kind of ad-hoc: one in `encode_interface`, one in `encode_interface_import`. I add a TODO in the code, but not sure what's the best helper function to cover both cases.

During the transition period, it's helpful to know if the binary is encoded using canonical-names or not. I was wondering the best way to do this, either we add something in the producer section, or we emit versionsuffix everywhere even if it's empty. Which one do you prefer?

alexcrichton:
Sounds good on the follow-ups. For transitioning wouldn't presence of `versionsuffix` at all be suitable for knowing whether canonical names are used? Just looking at the import/export names I'd expect would be enough. The producers section doesn't capture toolchain/compilation flags, just versioning info, and I think the validation should be stricter to disallow empty strings. Validation should also ensure that `versionsuffix` is correct when present (e.g. `0.1.0` is split as `0.1` and `.0`, not `0` and `.1.0`), but that's a separate change to make

chenyan2002:
Sounds good. I added the validation and encoder to exclude empty version suffix. 

The CI error is strange, and I cannot reproduce it locally. `(func (;0;) (type 0) (param i32 i32)` should be equivalent to `(func (;0;) (type 0)`?

alexcrichton:
Ah the test failure should be fixed by rebasing, this was a recent change to `main`

alexcrichton:
Thanks again @chenyan2002!

If you'd like as well, once this is published to crates.io the final piece of integration will be [in wasm-component-ld](https://github.com/bytecodealliance/wasm-component-ld/blob/db16fecf73bcda9cdcfd1b00f38f6073066fe266/src/lib.rs#L258-L347) to add a CLI flag like `--emit-canonical-names` here. Similar to the `--return-call-ref` handling.

PR #2623, opened 2026-08-25
Import component spec tests as a submodule and run them.

alexcrichton:
This was a major motivation for https://github.com/bytecodealliance/wasm-tools/pull/2621 and this PR follow through (building on top of #2621) in importing the component model test suite. One minor change of shifting an error around is additionally required, but otherwise all component spec tests run except for the recently-added `max-value-size.wast` which I'll be tackling next. For now that's expected to fail.

Note that just the last two commits here are part of this PR, and I've separated out the auto-generated changes and the handwritten changes commit-wise.


pchickey (approved):
Obviously I didn't review the entire blessed test results added by this PR, but I spot checked two files and the rest looks good

PR #2626, opened 2026-08-26
Drop hyphens when testing equality of kebab names

alexcrichton:
This commit is an implementation of WebAssembly/component-model#704 for this repository where the main change is to drop the `-` character when comparing kebab-names. More rationale can be found in the upstream PR, and while there's expected to be minor amounts of test-related fallout this is not expected to break anything in practice.


pchickey (approved):

tests/cli/component-model/naming.wast
```diff
@@ -158,3 +158,51 @@
...
+  "conflicts with previous name")
+(assert_invalid
+  (component (import "foo-bar" (func)) (import "FOOBAR" (func)))
+  "conflicts with previous name")
+
+(assert_invalid
+  (component
+    (import "foo" (type $foo (sub resource)))
+    (import "[static]foo.bar-baz" (func))
+    (import "[method]foo.barbaz" (func (param "self" (borrow $foo)))))
```
bvisness, line 186: To maximize terribleness, you could also add a test where you import `foo-bar`, and then import `[method]foo-bar.FO-ob-AR` or something :) (In other words, something that will trigger the `[method]foo.foo` rule.)

You could also add a test that makes `foo-bar` and `[constructor]FOOBAR` conflict, but such a test would either require you to also import a resource type named `FOOBAR` (which would itself conflict as seen above), or it would imply that strong uniqueness is checked before the constructor name itself, which feels wrong to me but I don't think the order is specified today anyhow. (I'd just avoid a test for that at this stage.)

alexcrichton: Added the `[method]...` case, and yeah otherwise the order-of-validation makes it not possible right now to write a test for `foo-bar` and `[constructor]FOOBAR`

PR #2627, opened 2026-08-27
Add compact import generation in wasm-smith

chacailline:
Updates wasm-smith to generate compact import sections.

Allows wasmtime fuzzing of compact import sections for bytecodealliance/wasmtime#14108.

`Module::imports_exports_from_module_shape` and `Module::arbitrary_imports_from_available` are also updated to accept compact imports and to generate them when it's enabled in the config.

I enabled the proposal by default. `config.rs` has a comment saying "Proposals that are not stage4+ are disabled by default." By that logic, we may want to enable wide arithmetic by default too.

Outside of `#[cfg(feature = "component-model")]`, `DuplicateImportsBehavior::Disallowed` was only ever created in an if-test in `Module::arbitrary_imports`, so I had to add a `#[cfg]` gate to pass CI.

The binary format supports empty compact import statements, but it's not clear whether the text format will support them. See WebAssembly/compact-import-section#22. Currently, `wasmparser` doesn't accept them, but `wasmprinter` outputs them if given a binary with empty statements.

For now, I don't generate empty statements. If they're not allowed, we should update `wasmprinter` not to generate them. If they are allowed, we should update `wasmparser` to accept them, and probably generate them in `wasm-smith`.

### DIFF
--- a/crates/fuzz-stats/src/bin/failed-instantiations.rs
+++ b/crates/fuzz-stats/src/bin/failed-instantiations.rs
@@ -106,6 +106,7 @@ impl State {
 
         // Wasmtime doesn't support these proposals yet.
         config.gc_enabled = false;
+        config.compact_imports_enabled = false;
 
         let mut wasm = wasm_smith::Module::new(config, &mut u)?;
         wasm.ensure_termination(10_000).unwrap();
--- a/crates/wasm-smith/src/config.rs
+++ b/crates/wasm-smith/src/config.rs
@@ -103,6 +103,12 @@ macro_rules! define_config {
             /// with exactly the same names and types as those in the provided
             /// WebAssembly module.
             ///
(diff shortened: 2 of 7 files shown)


chacailline:
`Module::arbitrary_imports` gained some complexity because we have to check constraints in the nested loops too. We also can't use `arbitrary_loop` for the outer loop. We could for the inner loops, but I didn't think it would be any cleaner.

`Module::arbitrary_imports_from_available` instead generates all imports to start, then decides whether to group adjacent ones with the same module / type. This is cleaner, but if we did it with random module / type names, I think we would almost never generate any compact imports with more than 1 element.

alexcrichton (approved):
This all looks great to me, thanks! It's expected that this'd generally make things more complicated but everything here looks as I'd expect

PR #2643, opened 2026-09-08
Return an error instead of panicking on fixed-length-list size overflow

### LINKED ISSUE #2637: Integer overflow panic in `wit-parser` `SizeAlign` for nested fixed-length lists
## Summary

`SizeAlign`'s handling of `TypeDefKind::FixedLengthList(ty, size)` computes the total size as `element_size * length` and panics on overflow instead of returning an error:

```rust
// crates/wit-parser/src/sizealign.rs:266-273
TypeDefKind::FixedLengthList(t, size) => {
    let field_align = self.align(t);
    let field_size = self.size(t);
    ElementInfo::new(
        ArchitectureSize::new(
            field_size.bytes.checked_mul(*size as usize).unwrap(),    // <-- panics on overflow
            field_size.pointers.checked_mul(*size as usize).unwrap(),
        ),
        field_align,
    )
}
```

`size` is a `u32` taken directly from the WIT text `list<T, N>` syntax (`crates/wit-parser/src/ast.rs:1454-1463`), parsed with no upper bound — unlike the component *binary* format, which bounds fixed-length list lengths to `MAX_WASM_FIXED_LENGTH_LIST_ELEMENTS = 1 Gi`. Nesting two fixed-length lists whose element type has a non-trivial size (e.g. `u64`) drives `element_size * length` past `usize::MAX`, so `checked_mul` returns `None` and `.unwrap()` panics. This fires in release builds too (it's a real panic, not a `debug_assert`).

`SizeAlign::fill(&Resolve)` iterates all types in the resolve, so the overflow is reachable even when the oversized fixed-length list is never used by any function signature.

## Repro

```wit
package test:types;
interface types {
  type a = list<u64, 4294967295>;
  type b = list<a, 4294967295>;
}
world w { import types; }
```

Parsing this via `wit_parser::Resolve::push_source` and computing sizes via `SizeAlign::fill` (or running `wasm-tools wit-dylib` on it) panics:

```
thread 'main' panicked at .../wit-parser-0.257.1/src/sizealign.rs:271:70:
called `Option::unwrap()` on a `None` value
```

Reproduced on `wit-parser` `0.230.0` (first version shipping this code, panic at `sizealign.rs:268`) and `0.257.1` (latest, panic at `sizealign.rs:271`), and on main HEAD `a7ef4778a98ed6d49c7785487dbd95bf053d788d`.

Controls confirming it's specifically the multiplication overflow, not a parse error:

| Case | Result |
|---|---|
| `type a = list<u64, 2>; type b = list<a, 3>` | exits fine |
| `type a = list<u64, 4294967295>` alone (size 34359738360) | no overflow, exits fine |
| `type a = list<u8, 4294967295>; type b = list<a, 4294967295>` (size 4294967295² = 18446744065119617025, under u64::MAX) | no overflow, exits fine |
| Two nested `list<u64, ...>` | panics |

Happy to open a PR for this if useful.

zikk090:
`SizeAlign::fill` (and its internal `calculate` helper) computed the size of a `FixedLengthList` as `element_size * length` via `checked_mul(..).unwrap()`, which panics when the multiplication overflows `usize`. Since `list<T, N>` places no upper bound on `N` in the WIT text format (unlike the component binary format, which bounds it to `MAX_WASM_FIXED_LENGTH_LIST_ELEMENTS`), a WIT document with two nested fixed-length lists of a non-trivial element type is enough to overflow the multiplication and abort the process, in a release build too.

This changes `SizeAlign::fill`/`calculate` to return `anyhow::Result` and propagates that outward through `wit-dylib`'s `Adapter::encode`, `create`, and `create_with_metadata`, and their two call sites (the `wasm-tools wit-dylib` CLI subcommand and the wit-dylib test-programs harness). A oversized fixed-length list now surfaces as a normal error instead of aborting:

    $ wasm-tools wit-dylib poc.wit
    error: size of fixed-length list of 4294967295 elements overflows
    the target architecture's address space

Adds a regression test covering both the overflow case (now returns `Err`) and a benign nested fixed-length list (still computes correctly).

Fixes #2637

### DIFF
--- a/crates/wit-dylib/src/lib.rs
+++ b/crates/wit-dylib/src/lib.rs
@@ -1491,7 +1491,7 @@ world w {{
             )
             .unwrap();
         let world = resolve.select_world(&[package], None).unwrap();
-        let adapter = super::create(&resolve, world, None);
+        let adapter = super::create(&resolve, world, None).unwrap();
         for payload in Parser::new(0).parse_all(&adapter) {
             match payload.unwrap() {
                 Payload::CodeSectionEntry(body) => {
--- a/crates/wit-dylib/src/lib.rs
+++ b/crates/wit-dylib/src/lib.rs
@@ -51,24 +51,28 @@ pub enum StackPointer {
     TaskContext,
 }
 
-pub fn create(resolve: &Resolve, world_id: WorldId, opts: Option<&mut DylibOpts>) -> Vec<u8> {
-    create_with_metadata(resolve, world_id, opts).0
+pub fn create(
+    resolve: &Resolve,
+    world_id: WorldId,
+    opts: Option<&mut DylibOpts>,
+) -> anyhow::Result<Vec<u8>> {
+    Ok(create_with_metadata(resolve, world_id, opts)?.0)
 }
 
 pub fn create_with_metadata(
     resolve: &Resolve,
     world_id: WorldId,
     mut opts: Option<&mut DylibOpts>,
-) -> (Vec<u8>, Metadata) {
+) -> anyhow::Result<(Vec<u8>, Metadata)> {
     let mut adapter = Adapter::default();
     if let Some(opts) = &mut opts {
         adapter.opts = opts.clone();
     }
-    let result = adapter.encode(resolve, world_id);
+    let result = adapter.encode(resolve, world_id)?;
     if let Some(opts) = &mut opts {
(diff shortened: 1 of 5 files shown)


pchickey (approved):
Thank you! Please fix the error discovered by the clippy CI job. The other failure appears spurrious

Issue #2591, opened 2026-08-10
Consider adding more type-checking to `wasm-tools component link`


alexcrichton:
Inspired by [this comment](https://github.com/bytecodealliance/wasm-tools/pull/2590#discussion_r3730085482) currently the linking phase of `wasm-tools component link` does type-checks at the item level (e.g. if something is a function or global) and type-checks for globals/tables/memories, but fine-grained type-checks of functions is not yet done. For example if functions are imported with the wrong signature that'll show up as a validation error at the end instead of a linking error. Ideally the linking phase itself would perform these sorts of type-checks and provide an error "faster" to avoid delaying it to a confusing time later on.

PR #2646, opened 2026-09-09
fix: validate linker intrinsic function signatures

RanaPriyansh:
Invalid stack, TLS, and thread intrinsic signatures can reach final component validation before the linker reports an error.

Metadata extraction now checks five intrinsic signatures, enforces the `i32` stack ABI, and reports required and actual types.

Verification: 1,845 workspace tests passed. Six existing tests remain ignored. Clippy, formatting, and the component-only feature check passed.

Related to #2591.

Codex assisted with implementation, tests, and review.

### DIFF
--- a/crates/wit-component/src/linking/metadata.rs
+++ b/crates/wit-component/src/linking/metadata.rs
@@ -110,6 +110,32 @@ impl TryFrom<&FuncType> for FunctionType {
     }
 }
 
+fn validate_intrinsic_signature(
+    types: &[FuncType],
+    type_index: u32,
+    module: &str,
+    name: &str,
+    parameters: &[ValueType],
+    results: &[ValueType],
+) -> Result<()> {
+    let type_index = usize::try_from(type_index).context("function type index is too large")?;
+    let ty = types.get(type_index).with_context(|| {
+        format!("invalid function type index {type_index} for `{module}.{name}`")
+    })?;
+    let actual = FunctionType::try_from(ty)
+        .with_context(|| format!("failed to read function type for `{module}.{name}`"))?;
+    if actual.parameters.as_slice() != parameters || actual.results.as_slice() != results {
+        let expected = FunctionType {
+            parameters: parameters.to_vec(),
+            results: results.to_vec(),
+        };
+        bail!(
+            "type mismatch for function `{module}.{name}`: required linker ABI `{expected}` but found `{actual}`"
+        );
+    }
+    Ok(())
+}
+
 /// Represents a core Wasm global variable type
 #[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
 pub struct GlobalType {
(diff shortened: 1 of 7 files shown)


crates/wit-component/tests/linking.rs
```diff
@@ -168,6 +176,202 @@ fn encode(wat: &str, wit: Option<&str>) -> Result<Vec<u8>> {
...
+  (type (func (result i32)))
+  (import "env" "__wasm_get_stack_pointer" (func (type 1)))
+  (func (export "run") (type 0))
+)
+"#,
+        true,
+    )
+    .context("failed to link an intrinsic at a nonzero type index")?;
+    Ok(())
+}
```
alexcrichton, line 232: Where possible I personally prefer to have tests in `tests/components/*` rather than unit tests, and I think tests like this can move there. That being said is this testing something that's not already tested? I would have figured we've already got tests for this.

crates/wit-component/tests/linking.rs
```diff
@@ -168,6 +176,202 @@ fn encode(wat: &str, wit: Option<&str>) -> Result<Vec<u8>> {
...
+        let error = format!("{error:#}");
+        assert!(error.contains("failed to extract linking metadata from app.wasm"));
+        assert!(error.contains(&format!("function `{namespace}.{name}`")));
+        assert!(
+            error.contains(&format!(
+                "required linker ABI `{expected}` but found `{actual}`"
+            )),
+            "{error}"
+        );
+    }
```
alexcrichton, line 304: Could you move these tests to `tests/components/*`?

crates/wit-component/tests/linking.rs
```diff
@@ -168,6 +176,202 @@ fn encode(wat: &str, wit: Option<&str>) -> Result<Vec<u8>> {
...
+    let error = link_intrinsic(
+        "other",
+        "__wasm_get_stack_pointer",
+        "(param i64) (result i32)",
+        false,
+    )
+    .unwrap_err();
+    let error = format!("{error:#}");
+    assert!(error.contains("module requires an import interface named `other`"));
+    assert!(!error.contains("required linker ABI"));
```
alexcrichton, line 372: It's ok to be not quite so exhaustive in testing, these are pretty noisy tests and there's not necessarily a need to test every single possible observable behavior. Can these either be dropped or moved to `tests/components/*`?

tests/cli/component-model/link-intrinsic-signature.wat
alexcrichton, line 1: This is already tested in `wit-component`, so it's ok to skip the test here as well

crates/wit-component/src/linking/metadata.rs
```diff
@@ -110,6 +110,32 @@ impl TryFrom<&FuncType> for FunctionType {
     }
 }
 
+fn validate_intrinsic_signature(
+    types: &[FuncType],
+    type_index: u32,
+    module: &str,
+    name: &str,
```
alexcrichton, line 117: One possible refactoring to simplify this slightly would be to take the `wasmparser::Import` here directly, and internally this would validate the type is a function to get the type index. That way not quite as many arguments need to be passed.

alexcrichton (approved):
Thanks! Some minor comments on tests, but otherwise seems reasonable to me.

alexcrichton (approved):

PR #2652, opened 2026-09-11
Support multiple supertypes in everything but validation

alexcrichton:
Upstream a new spec test has been added that a type with multiple supetypes is considered invalid. This is distinct from malformed meaning that the binary doesn't even parse, but by being invalid that means that the binary parses but is dynamically considered invalid. This test requires handling multiple supertypes in all locations throughout the tooling here, for example `wast`, `wasm-encoder`, and `wasmparser`. Notably `wasmparser` now has a `Vec<u32>` for supertype indices, and lengths > 1 are rejected during validation.

While here this is changing all of the `dump` test outputs anyway so this switches to printing types in a more human-readable format.

### DIFF
--- a/crates/wasm-encoder/src/core/types.rs
+++ b/crates/wasm-encoder/src/core/types.rs
@@ -9,7 +9,7 @@ pub struct SubType {
     pub is_final: bool,
     /// The list of supertype indexes. As of GC MVP, there can be at most one
     /// supertype.
-    pub supertype_idx: Option<u32>,
+    pub supertype_idxs: Vec<u32>,
     /// The composite type of the subtype.
     pub composite_type: CompositeType,
 }
@@ -666,7 +666,7 @@ impl<'a> CoreTypeEncoder<'a> {
         // We only need to emit a prefix byte before the actual composite type
         // when either the `sub` type is not final or it has a declared super
         // type (see notes on `push_prefix_if_component_core_type`).
-        if ty.supertype_idx.is_some() || !ty.is_final {
+        if !ty.supertype_idxs.is_empty() || !ty.is_final {
             if ty.is_final {
                 self.bytes.push(0x4f);
             } else {
@@ -675,7 +675,7 @@ impl<'a> CoreTypeEncoder<'a> {
                 }
                 self.bytes.push(0x50);
             }
-            ty.supertype_idx.encode(self.bytes);
+            ty.supertype_idxs.encode(self.bytes);
         }
         if ty.composite_type.shared {
             self.bytes.push(0x65);
--- a/crates/wasm-encoder/src/core/types.rs
+++ b/crates/wasm-encoder/src/core/types.rs
@@ -730,7 +730,7 @@ mod tests {
         let mut types = TypeSection::new();
         types.ty().subtype(&SubType {
             is_final: true,
-            supertype_idx: None,
+            supertype_idxs: Vec::new(),
             composite_type: CompositeType {
                 inner: CompositeInnerType::Func(FuncType::new([], [])),
                 shared: false,
--- a/crates/wasmparser/src/limits.rs
+++ b/crates/wasmparser/src/limits.rs
@@ -20,7 +20,7 @@
 //
 // See https://webassembly.github.io/spec/js-api/#limits for details.
 pub const MAX_WASM_TYPES: usize = 1_000_000;
-pub const MAX_WASM_SUPERTYPES: usize = 1;
+pub const MAX_WASM_SUPERTYPES: usize = 5;
 pub const MAX_WASM_FUNCTIONS: usize = 1_000_000;
 pub const MAX_WASM_IMPORTS: usize = 1_000_000;
 pub const MAX_WASM_EXPORTS: usize = 1_000_000;
--- a/crates/wasmparser/src/readers/core/types.rs
+++ b/crates/wasmparser/src/readers/core/types.rs
@@ -450,21 +450,21 @@ pub struct SubType {
     /// Is the subtype final.
     pub is_final: bool,
     /// The list of supertype indexes. As of GC MVP, there can be at most one supertype.
-    pub supertype_idx: Option<PackedIndex>,
+    pub supertype_idxs: Vec<PackedIndex>,
     /// The composite type of the subtype.
     pub composite_type: CompositeType,
 }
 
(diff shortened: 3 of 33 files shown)


crates/wasm-encoder/src/core/types.rs
```diff
@@ -9,7 +9,7 @@ pub struct SubType {
     pub is_final: bool,
     /// The list of supertype indexes. As of GC MVP, there can be at most one
     /// supertype.
-    pub supertype_idx: Option<u32>,
+    pub supertype_idxs: Vec<u32>,
```
fitzgen, line 12: `SmallVec` to avoid the heap allocation in practice for valid modules?

crates/wasmparser/src/readers/core/types.rs
```diff
@@ -450,21 +450,21 @@ pub struct SubType {
     /// Is the subtype final.
     pub is_final: bool,
     /// The list of supertype indexes. As of GC MVP, there can be at most one supertype.
-    pub supertype_idx: Option<PackedIndex>,
+    pub supertype_idxs: Vec<PackedIndex>,
```
fitzgen, line 453: Ditto

crates/wasmparser/src/limits.rs
```diff
@@ -20,7 +20,7 @@
 //
 // See https://webassembly.github.io/spec/js-api/#limits for details.
 pub const MAX_WASM_TYPES: usize = 1_000_000;
-pub const MAX_WASM_SUPERTYPES: usize = 1;
+pub const MAX_WASM_SUPERTYPES: usize = 5;
```
fitzgen, line 23: This seems wrong.

crates/wast/src/core/types.rs
```diff
@@ -957,7 +957,7 @@ pub struct TypeDef<'a> {
     /// Whether the type is shared or not.
     pub shared: bool,
     /// The declared parent type of this definition.
-    pub parent: Option<Index<'a>>,
+    pub parents: Vec<Index<'a>>,
```
fitzgen, line 960: Ditto re `SmallVec`

fitzgen (approved):
LGTM with below addressed

crates/wasmparser/src/limits.rs
alexcrichton: This is required insofar as without this the test case doesn't get past the parsing stage -- an error is generated because 2 supertypes exceeds the maximum of 1. I needed to change this to something, and so I just sort of arbitrarily bumped this up a little bit to get tests passing but not enough that we'd have to worry about it.

crates/wasm-encoder/src/core/types.rs
alexcrichton: How strongly do you feel about that? This has come up a few times in the past about using `smallvec` in `wasmparser` and other crates here, so this isn't new, but so far we haven't added the dependency. This would add a new dep to all of these crates, and the only real perf-critical one is `wasmparser` which is already, before this PR, collecting the results into the heap and then taking it back out. 

If `smallvec` or similar were in libstd I'd use it without hesitation, but as an extra dep it's something extra for all users to pull in and depend on, and `wasmparser` in particular shows up in a good number of dependency trees. Keeping our deps slim for minimal feature builds (e.g. without the component-model) feels relatively important at least.

fitzgen: It feels like a very light dep to me, but if we're concerned about that, then I think we can do some [small shenanigans](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=c35be1180da3a15fdc6cc9e95d31da5b) to make it optional while making the uses inside the crates not too bad. Luckily that is pretty simple and concise, so not too annoying to duplicate across each crate.

alexcrichton: That unfortunately wouldn't work though b/c toggling a Cargo feature would be able to cause compile-time breakages -- basically it's public knowledge that `SmallVec<...> == Vec<T>` so when the crate dep is pulled in it'd break builds that relied on such equivalence. We've historically solved this in wasmparser with [collection wrappers](https://github.com/bytecodealliance/wasm-tools/tree/main/crates/wasmparser/src/collections) but that requires duplicating the API surface area of collections which I think crosses the threshold of "probably shouldn't copy across crates".

I suppose another way to say what I'm thinking is that I don't think that the crate dependency is worth it unless we've got benchmarks showing such. At this time I don't think we do have such benchmarks, and at least for Wasmtime's use cases I don't think validation/parsing really factors into any of them, so I don't think we actually have a benchmark-driven use case for taking a new dependency at this time

fitzgen: Eh, I guess. I suspect you'll be able to measure the slowdown on Sightglass's `kotlin-richards`, but I'm not gonna push on this any harder. For the record, I agree it is marginal for the non-`wasmparser` crates, so if we ever do revisit, it is probably good enough to just do something for `wasmparser` only.

alexcrichton: Benchmarking that with `crates/wasmparser/benchmarks/benchmark.rs` in this repo I get:

**this PR compared to `main`**

```
validate/kotlin-richards
                        time:   [337.07 µs 337.20 µs 337.38 µs]
                        change: [+1.4892% +1.7770% +1.9513%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (1.00%) low severe
  2 (2.00%) low mild
  3 (3.00%) high mild
  3 (3.00%) high severe

parse/kotlin-richards   time:   [86.062 µs 86.242 µs 86.568 µs]
                        change: [+5.0978% +5.3059% +5.5673%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 5 outliers among 100 measurements (5.00%)
  5 (5.00%) high severe
```

**this PR + smallvec compared to `main`**


```
validate/kotlin-richards
                        time:   [335.91 µs 336.08 µs 336.27 µs]
                        change: [+1.1715% +1.4555% +1.6291%] (p = 0.00 < 0.05)
                        Performance has regressed.

parse/kotlin-richards   time:   [85.705 µs 85.734 µs 85.771 µs]
                        change: [+4.7187% +4.8553% +4.9887%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

as a reference point at least

PR #2654, opened 2026-09-11
Print params/results of imports

alexcrichton:
This commit is a small change to the `wasmprinter` logic to print `(param ...)` and `(result ...)` for function imports instead of only printing `(type N)`. This helps improve debuggability-at-a-glance since there's no need to cross-reference types, and this is additionally done in preparation for a spec test suite update where this'll be necessary to pass all tests in the extended-name-section proposal.

This has a very large fallout in golden outputs for this repository, hence the very large diff. The only source code change is `crates/wasmprinter/src/lib.rs`, however.


pchickey (approved):
Nice improvement to readability, thanks

PR #2658, opened 2026-09-14
Update extended-name-section proposal implementation

alexcrichton:
Notably implement the parameter and tag-parameter subsections. These are then integrated in all the relevant locations throughout the tools.


fitzgen (approved):

bvisness:
Very glad to see this done so quickly! Saves me the trouble of making my own patches :smile:

Does this mean the extended names implementation is complete and passing all the tests that were recently upstreamed in https://github.com/WebAssembly/testsuite/pull/167?

alexcrichton:
Sort of and sort of not -- there's three main gotchas to the test coverage:

* One is that I haven't actually updated the submodule here yet so there's some other misc changes needed. I'll have that update ready after https://github.com/bytecodealliance/wasm-tools/pull/2652 is figured out, however.
* Another gotcha is that wasm-tools [doesn't actually validate custom sections at all](https://github.com/bytecodealliance/wasm-tools/blob/3b6380472e7797661632e58665248b54259e4584/src/bin/wasm-tools/wast.rs#L324-L326) so `assert_invalid_custom` is basically an unimplemented directive. IIRC it was just branch hinting that previously used this and that was pretty minor. Otherwise though wasmparser's architecture isn't very amenable to validation the `name` custom section right now API-wise, so I haven't figured out how best to do that. Notably validating the `name` section requires knowledge of all parsed functions to validate labels and locals. Anyway, just wanted to point this out b/c I wouldn't consider wasm-tools a conforming implementation until that directive is implemented.
* The final gotcha is that wasm-tools uses `*.wast` tests to go a bit further than the spec tests have on the surface, specifically w.r.t. round-tripping the module. All preexisting tests assert that if you parse text to a binary, then print a binary, then parse that back to binary again, that the two binaries produced are the exact same. This works out because the text section can't represent things like empty function sections, but the [tests outlined in my comment here](https://github.com/WebAssembly/extended-name-section/pull/11#pullrequestreview-5201655673) don't round-trip because the `name` section doesn't pop out the other end (the text format can't represent a name section, or at least not too easily in the current impelmentation). That means that `wasm-tools wast` fails on those tests, but only for the more strict assertions that wasm-tools has internally in a sense. That's where I was thinking a separate `assert_valid_custom` directive would be a good indicator for "don't roundtrip test this" or something like that.

For phase-advancement purposes it may be there though? None of these issues have anything to do with the extended-name-section proposal itself, all of what I've described here is a property of the `name` section already as-specified.

If you're curious, https://github.com/alexcrichton/wasm-tools/commit/088508b2cdcaf9920a59b27fb1359d4daf31a8d7 is the commit I'm working on to update the submodule to have the extended-name-section tests. The current failures look like [this](https://github.com/bytecodealliance/wasm-tools/blob/088508b2cdcaf9920a59b27fb1359d4daf31a8d7/tests/cli/spec/proposals/extended-name-section/custom/name.wast.stderr)

PR #2660, opened 2026-09-14
Add getter and setter support

bvisness:
Adds support for getters and setters to the component model / WIT tooling per WebAssembly/component-model#701.

Disclosure: While the wasmparser code and tests are all handwritten, the WIT stuff and surrounding crates were significantly AI-assisted, as were the tests, and while I tried to very carefully review all of it over the last couple days, it's all still pretty unfamiliar to me. I will not be offended you deem it trash and I need to spend more time on it.


crates/wasmparser/src/validator/names.rs
```diff
@@ -515,6 +477,202 @@ impl Hash for ResourceFunc<'_> {
...
+        // (Really we have to add back those annotations because there's
+        // nothing in `raw` to strip.)
+        if resource_func_kind == Some(ResourceFuncKind::Constructor) {
+            format!("{CONSTRUCTOR}{lower}")
+        } else if accessor_kind == Some(AccessorKind::Set) {
+            format!("{SET}{lower}")
+        } else {
+            lower
+        }
+    }
```
alexcrichton, line 566: For this method the step 1 here at least is duplicating the preexisting `KebabStr::canonical_chars` method above and it's finnicky enough that I'd like to keep them deduplicated.

Could this method here perhaps skip the lowercase/`-`-filter and instead return a `KebabString`? That would break the match-the-steps-to-the-spec here but would help keep things deduplicated

crates/wasmparser/src/validator/names.rs
```diff
@@ -515,6 +477,202 @@ impl Hash for ResourceFunc<'_> {
     }
 }
 
+/// A plain name, possibly with resource func annotations like `[constructor]`
+/// or `[method]` and with accessor annotations like `[get]` or `[set]`.
+#[derive(Debug, Clone)]
+pub struct PlainName<'a> {
+    /// Stores only the part after the annotations, if any.
+    raw: &'a str,
```
alexcrichton, line 485: Could this be renamed to maybe `unannotated` or something similar? To me "raw" evokes "ok this is what was literally in the component" which isn't the case here. I'd also be ok with a bland `name` with the documentation sufficient to explain what it is

crates/wasmparser/src/validator/component.rs
```diff
@@ -4784,81 +4787,145 @@ impl ComponentNameContext {
...
+                        }
+
+                        // Setters take one value and return either nothing or `(result (error $E)?)`.
+                        AccessorKind::Set => {
+                            if num_own_params != 1 {
+                                bail!(
+                                    offset,
+                                    "setter function should have exactly one parameter{besides_self}"
+                                );
+                            }
```
alexcrichton, line 4912: Right now there's no validation of the parameters themselves, which means that this component is valid:

```wasm
(component
  (import "[get]a" (func (result u32)))
  (import "[set]a" (func (param "v" u64)))
)
```

is that intentional, or something to also add some validation/tests for? (if it's intentional that's totally fine, I haven't kept up-to-speed on the latest spec discussions myself)

crates/wit-parser/src/ast/resolve.rs
```diff
@@ -1759,6 +1811,38 @@ impl<'a> Resolver<'a> {
...
+                    {
+                        Ok(Some(resolved))
+                    } else {
+                        Err(ParseError::new_syntax(
+                            ty.span(),
+                            "if a setter return type is declared it must be a `result` with no `ok` type".to_owned(),
+                        ))
+                    }
+                }
+            },
```
alexcrichton, line 1844: FWIW so far wit-parser has generally been pretty lax about validation of incoming WITs. This has the benefit of deferring "real validation" to `wasmparser` and when something hits components, but has the downside of now errors, when they happen, are far removed from the source. Implementing validation in wit-parser flips those two, where the benefit is more relevant errors more quickly, and the downside is it's parallel checks to maintain.

I don't have a strong opinion myself but wanted to point this out. I expect that in the limit of time everything will get a parallel implementation in `wit-parser` since better errors are indeed better. With LLMs and such in theory keeping the two in sync isn't so onerous either.

(I also realize that the preexisting validation is a bit mish-mash, some stuff is validated, some stuff isn't, it's not super consistent other than "everything is for sure validated in `wasmparser`")

crates/wit-parser/src/ast/resolve.rs
```diff
@@ -738,9 +746,26 @@ impl<'a> Resolver<'a> {
...
+                };
+                names.define(name, kind.span())?;
+            }
         }
         self.type_lookup.clear();
 
+        // Getter/setter pairing is checked in `process_world_includes`, rather
+        // than here with all the other uniqueness etc. conditions, because
+        // includes need to be resolved first.
+
```
alexcrichton, line 768: It's ok to drop this comment

crates/wit-parser/src/decoding.rs
```diff
@@ -1197,28 +1199,61 @@ impl WitPackageDecoder<'_> {
...
+                                FunctionKind::Freestanding
+                            }
+                        }
+                        (None, Some(ComponentAccessorKind::Get)) => FunctionKind::Getter,
+                        (None, Some(ComponentAccessorKind::Set)) => FunctionKind::Setter,
+
+                        (Some(ResourceFuncKind::Constructor), None) => {
+                            FunctionKind::Constructor(resource.unwrap())
+                        }
+                        (Some(ResourceFuncKind::Constructor), _) => {
```
alexcrichton, line 1224: To the extent of scoping `_` in patterns as specific-as-possible, I think `_` here could be replaced with `Some(_)`

crates/wit-parser/src/lib.rs
```diff
@@ -1057,38 +1124,140 @@ impl FunctionKind {
...
+    /// A `set` function, encoded with a `[set]` name.
+    Setter,
+}
+
+/// Reorders `items` so that every setter immediately follows its getter. (WIT
+/// allows functions to be declared in any order, but the component binary
+/// format requires getters to precede setters.)
+pub(crate) fn place_setters_after_getters<K, V>(
+    items: &mut IndexMap<K, V>,
+    func: impl for<'a> Fn(&'a V) -> Option<&'a Function>,
```
alexcrichton, line 1214: You should be able to drop the `for<'a>` here and the explicit `'a` annotations: `impl Fn(&V) -> Option<&Function>`

alexcrichton (commented):
Thanks! I've got a mixture of high-level and rust-style-ish comments below, but overall everything looks pretty reasonable to me.

crates/wasmparser/src/validator/component.rs
bvisness: Hm, I think this is probably just an oversight in my spec work.

crates/wit-parser/src/ast/resolve.rs
bvisness: Would you prefer to drop the WIT-level validation of getter/setter rules for simplicity, or keep them so that we can try and improve error messages? (I'm more likely to be able to wrangle this PR if I just drop WIT-level validation...and with other component changes coming down the pipe, maybe it's better to keep the WIT system simpler for now and make it friendlier later.)

alexcrichton: Either way's fine by me, I mostly figured that this was likely LLM-generated and wanted to note it for possibly-future-PRs in case you wanted to trim things down. Can consider this a bit of an experiment to see how it holds up!

crates/wasmparser/src/validator/component.rs
bvisness: This has been addressed in WebAssembly/component-model#722. I'll get this PR updated to reflect that.

bvisness:
@alexcrichton I think I've addressed your feedback. I chose to keep the validation in WIT since it's already done, but I'm not sure I will in the future.

alexcrichton (approved):
Thanks!

Some minor failing tests on CI, but otherwise lgtm

bvisness:
Ok, finally fixed those.

PR #2667, opened 2026-09-16
Update spec test suite submodule

alexcrichton:
This commit updates both the core wasm and the component model spec test suites. This drops a number of FAIL directives for component-model tests now that the upstream tests have updated. This then additionally fills out some various corners/holes for core wasm such as

* The `extended-name-section` proposal's tests are new and now run here. The main test added, though, is marked as expected to fail for now. Most of the test passes but there's a few minor ones that don't round-trip correctly just yet.

* Parsing of `name` sections, subsections, and maps now all have extra validation applied while parsing. Namely subsections must be ordered and maps-of-names must all be ordered correctly.

* Validation for the `custom-descriptors` proposal has been adjusted to handle new tests that are added.

* The `assert_malformed_custom` directive now actually parses all custom sections to look for an error. Previously it just encoded into bytes but this was an insufficient check.

* Roundtripping modules with invalid custom sections is disabled because the `name` section, when invalid, doesn't roundtrip.


fitzgen (approved):
Seems reasonable to me

PR #2670, opened 2026-09-22
wit-parser: distinguish doc comments from non-doc comments

cestercian:
## Summary

Split `Token::Comment` so `wit-parser` only puts rustdoc-style doc comments (`///`, `/** */`) into `docs`, and treats `//` / `/* */` (plus `////`, `/**/`, `/*** */`) as non-doc.

Motivation came up on [componentize-py#251](https://github.com/bytecodealliance/componentize-py/pull/251) / [#177](https://github.com/bytecodealliance/componentize-py/issues/177): bindings tools were absorbing `//` lines into generated docs because the parser folded every comment into `Docs`. Fixing this in `wit-parser` helps wit-bindgen, componentize-go, ComponentizeJS, etc. without each shipping a preprocessor.

## Changes

- `Token::DocComment` + classification in the tokenizer
- `parse_docs` only aggregates doc comments; non-doc comments stay transparent (like whitespace)
- Regression UI test + re-blessed in-tree fixtures (`wit-component`, CLI snapshots)

## Test plan

- [x] `cargo test -p wit-parser`
- [x] `cargo test -p wit-component -p wit-encoder`
- [x] `BLESS=1 cargo test --test cli` (fixtures re-blessed)


tschneidereit:
We deliberately aren't doing this right now, because we want to do it more properly before trying it at all. Specifically, doc comments should have a fully specified syntax that allows code generators to also generate full doc comments in the target syntax, including proper references to symbols, etc.

As such, I think we should first have a spec change as a PR to the [Component Model spec](https://github.com/WebAssembly/component-model/) that defines all this, and then tackle the implementation.

cestercian:
closing for now — agree we should wait on a Component Model spec change for doc-comment syntax before landing this in wit-parser. thanks @tschneidereit

